PR #33, opened 2024-03-19
Remove unnecessary NuGet references to S.R.CS.Unsafe

PaulusParssinen:
Removed unnecessary NuGet package references to the OOB `System.Runtime.CompilerServices.Unsafe`. All the project TFMs have `Unsafe` class in their BCL.

Apologies for VS autoformatting..


PaulusParssinen:

badrishc:
Thanks for your thoughtful contributions! Just a question - we support .NET 6, so wouldn't we need System.Runtime.CompilerServices.Unsafe NuGet package for that runtime?

PaulusParssinen:
> Just a question - we support .NET 6, so wouldn't we need System.Runtime.CompilerServices.Unsafe NuGet package for that runtime?

CI is green and everything build so it seems like we're fine 😄. I believe Visual Studio would flag us if we used any APIs from `System.Runtime.CompilerServices.Unsafe` that were not supported by project target frameworks or explicitly `#ifdef`'d out. Also `Unsafe` has had minimal updates since all the goodies we got around .NET Core 2.1 and 3.1

badrishc (approved):

PR #35, opened 2024-03-19
Use BCL provided BitOperations class to accelerate bit-twiddling hacks

PaulusParssinen:
Use BCL provided `BitOperations` class to accelerate bit-twiddling hacks. These use hardware provided intrinsics if they're available.

This PR also removes some of the many duplicated and unused methods in the Utility classes. These classes will evaporate in its entirety eventually.


vazois (approved):
Looks good, thanks!

Issue #38, opened 2024-03-19
GC Hole for SpanByte and other ones


hamarb123:
https://github.com/microsoft/garnet/blob/1e5c7ad3d80a57a58e233e76267e5a21f9d2113f/libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs#L61

this member is called here

https://github.com/microsoft/garnet/blob/1e5c7ad3d80a57a58e233e76267e5a21f9d2113f/libs/client/ClientSession/GarnetClientSessionMigrationExtensions.cs#L403

which this API is called from the public API `TryWriteKeyValueSpanByte` here

https://github.com/microsoft/garnet/blob/1e5c7ad3d80a57a58e233e76267e5a21f9d2113f/libs/client/ClientSession/GarnetClientSessionMigrationExtensions.cs#L296

which takes `ref SpanByte`.

It is a GC hole since it's never pinned or otherwise fixed.

This is also a GC hole, since the memory is not fixed (you can also just use `Unsafe.SizeOf` here):
https://github.com/microsoft/garnet/blob/1e5c7ad3d80a57a58e233e76267e5a21f9d2113f/libs/client/Utility.cs#L28

This is also seems to be a GC hole, since the memory is never fixed on this public API:
https://github.com/microsoft/garnet/blob/1e5c7ad3d80a57a58e233e76267e5a21f9d2113f/libs/storage/Tsavorite/cs/src/core/VarLen/SpanByteComparer.cs#L25

(note this list is not exhaustive)

badrishc:
None of these seem like real issues. We know `SpanByte` structs are simply a view over pre-pinned memory buffers or pinned pre-pinned network buffers. Garnet is being used as a server. At best, it might be a case of things being marked public when they shouldn't have?

EgorBo:
> None of these seem like real issues.

[...]

PR #39, opened 2024-03-19
Use Unsafe.SizeOf to calculate object sizes

PaulusParssinen:
Fixes potential GC hole by using BCL's `Unsafe.SizeOf<T>()` to calculate sizes for given type parameter more _safely_ 😆.

Contributes to #38


PaulusParssinen:
No idea why build is failing in the CI..

libs/storage/Tsavorite/cs/src/core/Allocator/BlittableAllocator.cs
```diff
@@ -17,10 +17,10 @@ internal sealed unsafe class BlittableAllocator<Key, Value> : AllocatorBase<Key,
         private readonly long* nativePointers;
 
         // Record sizes
-        private static readonly int recordSize = Utility.GetSize(default(Record<Key, Value>));
-        private static readonly int recordInfoSize = Utility.GetSize(default(RecordInfo));
-        private static readonly int keySize = Utility.GetSize(default(Key));
-        private static readonly int valueSize = Utility.GetSize(default(Value));
+        private static readonly int recordSize = Unsafe.SizeOf<Record<Key, Value>>();
```
jkotas, line 20: It is unnecessary to cache these values in statics. `Unsafe.SizeOf` evaluates to constant, so it is cheaper to call it inline or to wrap it in a property.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Revivification/RevivificationManager.cs
```diff
@@ -13,20 +13,17 @@ internal struct RevivificationManager<Key, Value>
         internal RevivificationStats stats = new();
 
         internal readonly bool IsEnabled = false;
-        internal readonly int FixedValueLength;
+        internal readonly int FixedValueLength => Unsafe.SizeOf<Value>();
         internal bool restoreDeletedRecordsIfBinIsFull;
         internal bool useFreeRecordPoolForCTT;
 
-        internal readonly bool IsFixedLength => FixedValueLength != 0;
+        internal readonly bool IsFixedLength { get; }
```
jkotas, line 20: Is this change needed? If original implementation would be more efficient.

badrishc:
> ~No idea why build is failing in the CI..~
> 
> nvm. There apparently was a Tsavorite.test project that is _not_ visible in the normal project solution view. Fixing..

Yes, there are too many tests in Tsavorite.test and that degrades the dev experience for Garnet, so those tests are not included in the Garnet solution. They are limited to Tsavorite. Tsavorite CI kicks off if any changes to Tsavorite folder are detected.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Revivification/RevivificationManager.cs
PaulusParssinen: The revivifaction logic was broken due to how isFixedLen was flowed in to this ctor. This was quickest fix I could come up to restore previous logic. Shouldn't JIT be able to consider this bool as constant in next tier as it is `readonly`?

PaulusParssinen: I wrecked my brain and couldn't come up how one of these variations would result less branching or worse CQ.. What am I missing?

jkotas: I agree. I think what you have is fine after taking a closer look.

libs/storage/Tsavorite/cs/src/core/Allocator/BlittableAllocator.cs
badrishc: is this true for .NET 6 as well?

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Revivification/RevivificationManager.cs
badrishc: @TedHartMS - pls check this thread as well.

libs/storage/Tsavorite/cs/src/core/Allocator/BlittableAllocator.cs
jkotas: Yes, it is true for all .NET versions.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Revivification/RevivificationManager.cs
TedHartMS: This change fixes 9bdb889 that set FixedValueLength always, rather than setting it only if isFixedLen is true. While the fix in ef5a01a was sufficient to fix the GC hole, the change in 9bdb889 made it consistent with the pattern of calling Unsafe.SizeOf only in properties. This requires a separate bool value for IsFixedLength. Unfortunately, as you've discovered, this bool can't be considered a constant even though it is readonly.

TedHartMS (approved):

PR #40, opened 2024-03-19
Use more UTF-8 string literals

PaulusParssinen:
This PR
* replaced many of the `Encoding.ASCII.GetBytes([constant string})` calls with the newish [UTF-8 string literals syntax ](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals) where applicable. 
* used [RVA statics](https://github.com/dotnet/roslyn/pull/24621) to store some of the cosntant data directly in `.data` -section where possible but there's still a lot of opportunities for that.
* Updated `RespWriteUtils.WriteBulkString` and `RespWriteUtils.WriteDirect` to take `ReadOnlySpan<byte>`s instead of arrays
  * This allowed me to remove unnecessary `Memory<byte>` overloads which should pretty much never be parameters in sync paths. We can just pass the `Memory<byte>.Span` to the write methods instead.


libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -430,7 +430,7 @@ private void IncrementIntegerOrFloat(HashOperation op, byte* input, int length,
                     }
                     else
                     {
-                        var errorMessage = Encoding.ASCII.GetBytes("-ERR field value is not a number\r\n");
+                        ReadOnlySpan<byte> errorMessage = "-ERR field value is not a number\r\n"u8;
```
neon-sunset, line 433: u8 literals aren't collection expressions - they don't require target-typing as their natural type is already ROS<byte> :)

neon-sunset (approved):

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -445,11 +445,11 @@ public unsafe GarnetStatus SortedSetRange<TObjectContext>(ArgSlice key, ArgSlice
             {
                 case SortedSetOrderOperation.ByScore:
                     sortedOperation = SortedSetOperation.ZRANGEBYSCORE;
-                    operation = Encoding.ASCII.GetBytes("BYSCORE");
+                    operation = "BYSCORE"u8.ToArray();
```
sebastienros, line 448: Shouldn't these arrays be made `static readonly` as class members instead?

PaulusParssinen: That'd be improvement yes, I kept changes minimal but this one is easy enough, thanks! 👍

libs/client/ClientSession/GarnetClientSessionMigrationExtensions.cs
```diff
@@ -18,11 +18,11 @@ namespace Garnet.client
...
-        static readonly Memory<byte> AUTH = Encoding.ASCII.GetBytes("AUTH");
-        static readonly Memory<byte> SETSLOTSRANGE = Encoding.ASCII.GetBytes("SETSLOTSRANGE");
-        static readonly Memory<byte> MIGRATE = Encoding.ASCII.GetBytes("MIGRATE");
-        static readonly Memory<byte> DELKEY = Encoding.ASCII.GetBytes("DELKEY");
-        static readonly Memory<byte> GETKVPAIRINSLOT = Encoding.ASCII.GetBytes("GETKVPAIRINSLOT");
+        static ReadOnlySpan<byte> AUTH => "AUTH"u8;
+        static ReadOnlySpan<byte> SETSLOTSRANGE => "SETSLOTSRANGE"u8;
+        static ReadOnlySpan<byte> MIGRATE => "MIGRATE"u8;
+        static ReadOnlySpan<byte> DELKEY => "DELKEY"u8;
+        static ReadOnlySpan<byte> GETKVPAIRINSLOT => "GETKVPAIRINSLOT"u8;
```
jaredpar, line 25: Why not leave these as `static readonly`?

jaredpar: oh duh, it's property

PaulusParssinen: The RVA pattern relies on the constant values to wrapped in properties, which can't be marked with `readonly` modifier because it's lowered to initializer expression.

vazois (approved):

PR #51, opened 2024-03-20
Update type in Cluster Overview docs

jeremylcarter:
The command line arguments for clustering incorrectly had the 2nd and 3rd servers using the same checkpoint directory.


darrenge:
Not sure what happened here, but the CIs did not automatically start and aren't listed in GH Actions either ... almost like those CIs were either never started or were deleted.

darrenge:
There was a recent ci.yml file changed that ignores all .md files and won't run CI when only .md files are changed. Therefore, the CIs didn't run (by design) but the PR seems to be looking for that CI to be ran.

jeremylcarter:
A good test for .md PRs then!

It's a pretty trivial change, just noticed it and thought it would be easy to fix.

darrenge:
Yeah, early stages of a release hiccups. @Vazois can still merge it after reviewing it.

vazois (approved):

PR #59, opened 2024-03-20
Added expiration to garnet objects at time of serialization deserialization

pradeepyadavmsft:
This fix addresses a bug where expiration is not being added to serialization, potentially causing some objects to become live again after recovery from a checkpoint.

If you are using the object store, please clean up older checkpoints as it will not be able to deserialize them due to the breaking nature of this change.

### DIFF
--- a/libs/server/Objects/Hash/HashObject.cs
+++ b/libs/server/Objects/Hash/HashObject.cs
@@ -43,26 +43,22 @@ namespace Garnet.server
         readonly Dictionary<byte[], byte[]> hash;
 
         /// <summary>
-        ///  HashObject Constructor
+        ///  Constructor
         /// </summary>
-        public HashObject(long expiration = 0) : base(expiration, MemoryUtils.DictionaryOverhead)
+        public HashObject(long expiration = 0)
+            : base(expiration, MemoryUtils.DictionaryOverhead)
         {
             hash = new Dictionary<byte[], byte[]>(new ByteArrayComparer());
         }
 
         /// <summary>
-        /// Copy constructor
-        /// </summary>
-        public HashObject(Dictionary<byte[], byte[]> hash, long expiration, long size) : base(expiration, size)
-        {
-            this.hash = hash;
-        }
-
-        /// <summary>
         /// Construct from binary serialized form
         /// </summary>
-        public HashObject(BinaryReader reader, long expiration) : this(expiration)
+        public HashObject(BinaryReader reader)
+            : base(reader, MemoryUtils.DictionaryOverhead)
         {
+            hash = new Dictionary<byte[], byte[]>(new ByteArrayComparer());
+
             int count = reader.ReadInt32();
             for (int i = 0; i < count; i++)
             {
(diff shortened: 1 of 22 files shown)


badrishc (approved):

libs/server/Objects/Hash/HashObject.cs
```diff
@@ -45,25 +45,23 @@ public unsafe partial class HashObject : GarnetObjectBase
...
         {
             this.hash = hash;
-            this.Size = size;
         }
 
         /// <summary>
         /// Construct from binary serialized form
         /// </summary>
-        public HashObject(BinaryReader reader) : this()
+        public HashObject(BinaryReader reader, long expiration) : this(expiration)
```
yrajas, line 64: Why not pass in the reader to the base class so that it can handle reading expiration direclty? That way, the base class fields info can stay within the base class and not have to be exposed to the derived class.

yrajas (approved):

libs/server/Objects/Hash/HashObject.cs
badrishc: great idea - i have made this change to the PR.

libs/server/Objects/Types/GarnetObjectBase.cs
```diff
@@ -62,6 +79,7 @@ public void CopyUpdate(ref IGarnetObject newValue)
                     using (var ms = new MemoryStream())
                     {
                         using var writer = new BinaryWriter(ms, new UTF8Encoding(), true);
+                        writer.Write(Expiration);
```
pradeepyadavmsft, line 82: We are writing expiration in base.DoSerialize so this should be removed.

libs/server/Objects/Types/GarnetObject.cs
```diff
@@ -83,14 +83,14 @@ public static IGarnetObject Create(byte[] data)
         {
             using var ms = new MemoryStream(data);
             using var reader = new BinaryReader(ms);
+            var expiration = reader.ReadInt64();
```
pradeepyadavmsft, line 86: Expiration will come after type.

libs/server/Objects/Types/GarnetObjectSerializer.cs
```diff
@@ -23,21 +23,22 @@ public GarnetObjectSerializer(CustomCommandManager customCommandManager)
         /// <inheritdoc />
         public override void Deserialize(out IGarnetObject obj)
         {
+            var expiration = reader.ReadInt64();
```
pradeepyadavmsft: Here also

badrishc:
The proposed serialization format is:
[base class fields][derived class fields]
where [base class fields] = [type][expiration]

test/Garnet.test/GarnetObjectTests.cs
```diff
@@ -184,6 +184,10 @@ public override void Deserialize(out IGarnetObject obj)
         }
 
         /// <inheritdoc />
-        public override void Serialize(ref IGarnetObject obj) => obj.Serialize(writer);
+        public override void Serialize(ref IGarnetObject obj)
+        {
+            writer.Write(obj.Type);
```
pradeepyadavmsft, line 189: Isn't obj.Serialize will add type also?

libs/server/StoreWrapper.cs
```diff
@@ -165,6 +166,26 @@ public sealed class StoreWrapper
...
+        {
+            using var ms = new MemoryStream();
+            var serializer = new GarnetObjectSerializer(customCommandManager);
+            serializer.BeginSerialize(ms);
+            serializer.Serialize(ref garnetObject);
+            serializer.EndSerialize();
+            return ms.ToArray();
+        }
+
+        public IGarnetObject CreateGarnetObject(byte[] data)
```
badrishc, line 179: Call this Deserialize to be symmetrical.

libs/server/StoreWrapper.cs
```diff
@@ -165,6 +166,26 @@ public sealed class StoreWrapper
...
+        {
+            using var ms = new MemoryStream(data);
+            var serializer = new GarnetObjectSerializer(customCommandManager);
```
badrishc, line 182: I think object serializer needs to be created per session in respserversession as it is not thread safe.

yrajas: The GarnetObjectSerializer is instantiated per call to tackle thread safety.

PR #82, opened 2024-03-22
Remove ifdefs from ClusterStress to fix build

PaulusParssinen:
`dotnet build` kept failing for me. Seems like either wrapping the `using System.Diagnostics` into `#ifdef` or removing them from logic should fix it. `Debug.Assert` is still only a guard for the Debug builds unless the intent was to keep cost of the `NumUtils.GetHashSlot` from Release configuration stresstest?


playground/ClusterStress/OnlineReqGen.cs
```diff
@@ -75,10 +75,10 @@ public byte[] GenerateKeyBytes(out int slot)
             int key = randomGen ? (zipf ? zipfg.Next() : keyRandomGen.Next(DbSize)) : (keyIndex++ % DbSize);
             slot = Garnet.common.NumUtils.HashSlot(Encoding.ASCII.GetBytes(key.ToString()));
             byte[] keyBytes = GetClusterKeyBytes(key);
-#if DEBUG
```
vazois, line 78: It might be best to use the full namespace for Debug instead of removing the guard. #ifdef is there to avoid the cost of NumUtils.HashSlot call.

Sooraz2 (approved):

vazois (commented):
Looks good!

PR #84, opened 2024-03-22
Enhance Code Readability

nimanikoo:
Hello, I hope you are well
Thank you for helping to open source and build this project for the .NET community
Out of curiosity, I browsed through the code in this repository and found some hints that might be good.
One of the performance improvements was where you used lists.
I tried to use the new features of .NET 8 and C# 12 and refactored where needed
In the commits, I stated the reason for the changes and on what basis the changes were made.
In some files, it was necessary to sync namespace , and I think namespace scope can be used for all files, according to the rest of the new .NET projects such as Aspire.
The next thing is that you can use the primary constructor  according to the new .NET 8 feature
In some cases, we can make the code more concise and readable and use expressions.
If this model of changes is approved by you, let me know so that I can apply it to the whole project and move forward step by step.
I hope these things are useful and I will be happy to contribute to the development of this project
Please check and let me know the result


nimanikoo:

libs/client/LightEpoch.cs
```diff
@@ -15,22 +15,22 @@ namespace Garnet.client
     public sealed unsafe class LightEpoch
     {
         /// Size of cache line in bytes
-        public const int kCacheLineBytes = 64;
+        private const int KCacheLineBytes = 64;
```
PaulusParssinen, line 18: I believe the lowercase ``k`` came from C++ convention for constants. I suggest this project to use normal PascalCase as per the [.NET Framework Design Guidelines guides:](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)
```suggestion
        private const int CacheLineBytes = 64;
```

PaulusParssinen (approved):
Not a maintainer but otherwise I agree with all other changes 👍

libs/client/LightEpoch.cs
nimanikoo: Hi @PaulusParssinen , how are you?
I think that the access modifiers should be modified in many places, and what I saw in the code was not given much attention.
In changing the name of the variable, I tried not to change the original name and only change the naming convention
But if you agree, it can be corrected

PaulusParssinen: I'm good, thanks for asking 😄 

I agree with access modifiers being sort of all over the place in the current state of project. In many places the explicit access modifier is missing completely which makes it even worse, but state of this can be improved in later PRs by introducing good convention that are enforced with `.editorconfig` rules. 

As for this change, other files in the project use proper the naming convention for constant fields as I suggested above.

DUWENINK:
I vote for this

libs/client/LightEpoch.cs
PaulusParssinen: Please remove the `K` prefix from the other constants below too 😄

libs/client/LightEpoch.cs
```diff
@@ -15,22 +15,22 @@ namespace Garnet.client
...
     {
         /// Size of cache line in bytes
-        public const int kCacheLineBytes = 64;
+        private const int CacheLineBytes = 64;
 
         /// <summary>
         /// Default invalid index entry.
         /// </summary>
-        const int kInvalidIndex = 0;
+        const int KInvalidIndex = 0;
```
PaulusParssinen, line 23:
```suggestion
        private const int InvalidIndex = 0;
```

libs/client/LightEpoch.cs
```diff
@@ -15,22 +15,22 @@ namespace Garnet.client
...
 
         /// <summary>
         /// Default number of entries in the entries table
         /// </summary>
-        static readonly ushort kTableSize = Math.Max((ushort)128, (ushort)(Environment.ProcessorCount * 2));
+        static readonly ushort KTableSize = Math.Max((ushort)128, (ushort)(Environment.ProcessorCount * 2));
```
PaulusParssinen, line 28:
```suggestion
        private static readonly ushort TableSize = Math.Max((ushort)128, (ushort)(Environment.ProcessorCount * 2));
```

libs/client/LightEpoch.cs
```diff
@@ -15,22 +15,22 @@ namespace Garnet.client
...
 
         /// <summary>
         /// Default drainlist size
         /// </summary>
-        const int kDrainListSize = 16;
+        const int KDrainListSize = 16;
```
PaulusParssinen, line 33:
```suggestion
        private const int DrainListSize = 16;
```

libs/client/LightEpoch.cs
```diff
@@ -53,7 +53,7 @@ public sealed unsafe class LightEpoch
         /// ensure latest value is seen by the last suspended thread.
         /// </summary>
         volatile int drainCount = 0;
-        readonly EpochActionPair[] drainList = new EpochActionPair[kDrainListSize];
+        readonly EpochActionPair[] drainList = new EpochActionPair[KDrainListSize];
```
PaulusParssinen, line 56:
```suggestion
        private readonly EpochActionPair[] drainList = new EpochActionPair[DrainListSize];
```

libs/client/LightEpoch.cs
```diff
@@ -100,15 +100,15 @@ static LightEpoch()
 
             // Over-allocate to do cache-line alignment
 #if NET5_0_OR_GREATER
-            threadIndex = GC.AllocateArray<Entry>(kTableSize + 2, true);
+            threadIndex = GC.AllocateArray<Entry>(KTableSize + 2, true);
             p = (long)Unsafe.AsPointer(ref threadIndex[0]);
 #else
             threadIndex = new Entry[kTableSize + 2];
```
PaulusParssinen, line 106: This and all `#if !NET5_0_OR_GREATER` is dead code, which is why this wasn't automatically renamed. You can either just rename this or remove all `#if !NET5_0_OR_GREATER` if defs.

PaulusParssinen (commented):
Didn't suggest all rename changes so please go through the constants in `LightEpoch` again to make them consistent with my previous suggestion, otherwise LGTM :shipit:

Sooraz2 (approved):

ritish78 (approved):

libs/client/LightEpoch.cs
```diff
@@ -306,13 +306,13 @@ public void Mark(int markerIdx, long version)
...
             // check if all threads have reported complete
-            for (int index = 1; index <= kTableSize; ++index)
+            for (int index = 1; index <= TableSize; ++index)
             {
-                int entry_epoch = (*(tableAligned + index)).localCurrentEpoch;
-                int fc_version = (*(tableAligned + index)).markers[markerIdx];
-                if (0 != entry_epoch)
+                int entryEpoch = (*(tableAligned + index)).localCurrentEpoch;
+                int fcVersion = (*(tableAligned + index)).markers[markerIdx];
+                if (0 != entryEpoch)
```
JeremyWu917, line 313:
```suggestion
                if (entryEpoch != 0)
```

JeremyWu917: Uniform style ?

libs/client/LightEpoch.cs
```diff
@@ -344,14 +344,14 @@ int ComputeNewSafeToReclaimEpoch(int currentEpoch)
         {
             int oldestOngoingCall = currentEpoch;
 
-            for (int index = 1; index <= kTableSize; ++index)
+            for (int index = 1; index <= TableSize; ++index)
             {
-                int entry_epoch = (*(tableAligned + index)).localCurrentEpoch;
-                if (0 != entry_epoch)
+                int entryEpoch = (*(tableAligned + index)).localCurrentEpoch;
+                if (0 != entryEpoch)
```
JeremyWu917, line 350:
```suggestion
                if (entryEpoch != 0)
```

libs/client/LightEpoch.cs
```diff
@@ -373,10 +373,10 @@ void SuspendDrain()
                 // Barrier ensures we see the latest epoch table entries. Ensures
                 // that the last suspended thread drains all pending actions.
                 Thread.MemoryBarrier();
-                for (int index = 1; index <= kTableSize; ++index)
+                for (int index = 1; index <= TableSize; ++index)
                 {
-                    int entry_epoch = (*(tableAligned + index)).localCurrentEpoch;
-                    if (0 != entry_epoch)
+                    int entryEpoch = (*(tableAligned + index)).localCurrentEpoch;
+                    if (0 != entryEpoch)
```
JeremyWu917, line 379:
```suggestion
                    if (entryEpoch != 0)
```

JeremyWu917 (changes requested):
I don't know if it makes sense, but it looks coooool 😎

libs/client/LightEpoch.cs
nimanikoo: Thank you for the good tips,
I solved it

libs/client/LightEpoch.cs
douglasg14b: Definitely the variable to the left IMHO

douglasg14b:
What is the appetite for automatic styling? Many of these things can be linted for and even auto-fixed to some arbitrary level of consistency without requiring human intervention.

nimanikoo:
I appreciate your point, but I don't think we can ignore the impact of human creativity on coding.
You're right that automation can make many tasks easier and faster.
However, When it comes to code refactoring and using new .NET features, I believe that humans are more productive than automation.
Let's agree to disagree 😃

@douglasg14b

JeremyWu917 (approved):
LGTM

douglasg14b (approved):
Like the cleanup :)

TalZaccai:
@nimanikoo thank you so much for your contribution to Garnet, I can see that there was quite a bit of work that went into this and that people are overall happy with it. We would like to address most of these via our .editorconfig rules so that we're able to enforce them in the future and that we're all on the same page with regards to styling choices. We're planning on a code-style overhaul soon, and I will definitely use your PR for reference for desired choices. I'm therefore closing it for now... Thanks again!

PR #95, opened 2024-03-22
Fix build error when configuration is Release

fallenwood:
The error log looks like
```
→ dotnet build -c Release
MSBuild version 17.9.4+90725d08d for .NET
  Determining projects to restore...
...
/home/vbox/workspace/build/garnet/playground/ClusterStress/OnlineReqGen.cs(6,1): error IDE0005: Using directive is unnecessary. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0005) [/home/vbox/workspace/build/garnet/playground/ClusterStress/ClusterStress.csproj::TargetFramework=net7.0]
/home/vbox/workspace/build/garnet/playground/ClusterStress/OnlineReqGen.cs(6,1): error IDE0005: Using directive is unnecessary. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0005) [/home/vbox/workspace/build/garnet/playground/ClusterStress/ClusterStress.csproj::TargetFramework=net6.0]
/home/vbox/workspace/build/garnet/playground/ClusterStress/OnlineReqGen.cs(6,1): error IDE0005: Using directive is unnecessary. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0005) [/home/vbox/workspace/build/garnet/playground/ClusterStress/ClusterStress.csproj::TargetFramework=net8.0]
```

PR #96, opened 2024-03-22
fix(chore): build error in Dockerfile

yzy613:
Now it builds normally. :v:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,11 @@ WORKDIR /source
 
 # Copy files
 COPY . .
+WORKDIR /source/main/GarnetServer
 RUN dotnet restore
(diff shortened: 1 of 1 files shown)


badrishc (approved):

badrishc:
Can you make the same change to the two other Dockerfiles in the root directory?

irinasp:
Recently release build encountered an issue. It appears that the problem (build error in Dockerfile) is not related to Dockerfile itself but the side effect of the build break (fixed by PR #95). The changes to the Dockerfile are not necessary because it is back working fine since the build break issue is fixed.

yzy613:
Well, it looks like you changed it yourself.

PR #99, opened 2024-03-22
Uses the `new` expression in RespCommandsInfo.cs

TStand90:
This uses the targeted-type `new` expressions in RespCommandsInfo cs, which follows the advice laid out in [Style rule IDE0090](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0090)


TalZaccai:
Thank you for your contribution to Garnet! We will address issues such as this in a wider code-style "overhaul" using .editorconfig rules to enforce, so closing this for now.... Thanks again!

PR #106, opened 2024-03-23
Removing rule severity for a category of analyzer rules + switching some rules from dotnet_diagnostic syntax

TalZaccai:
So it turns out that severity of specific "dotnet_diagnostic" style rules take precedence over severity of category of analyzer rules BUT severity of category of analyzer rules takes precedence over dotnet_style_ / csharp_style_ rules. 
More about this [here](https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#configure-severity-levels).
So in essence, any specified severity in dotnet_style_ / csharp_style_ rules is effectively ignored and marked as "error".
I think then the "appropriate" way to do this is to remove the category severity of error, and run "dotnet format --verify-no-changes" in the CI to make sure that current formatting follows the rules.
LMK if you have any comments.


Rsdgsc:
111

badrishc (approved):

Issue #76, opened 2024-03-22
API Coverage - Implement HSTRLEN


TalZaccai:
Here's your opportunity to implement a RESP command in Garnet! 

**Syntax:** 
`HSTRLEN key field`

**Command description:**
Returns the string length of the value associated with `field` in the hash stored at `key`. If the `key` or the `field` do not exist, 0 is returned.

**Response:**
_Integer Reply:_ the string length of the value associated with the `field`, or zero when the `field` isn't present in the hash or the `key` doesn't exist at all.

---

**How to approach this task:**
1. Add the new command info to the hashCommandsInfoMap in the RespCommandsInfo class _(Garnet.server/Resp/RespCommandsInfo.cs)_
2. Add the new command string to the returned HashSet in the RespInfo.GetCommands method _(Garnet.server/Resp/RespInfo.cs)_
3. Add the appropriate fast parsing logic for the new command in the RespCommand.FastParseArrayCommand method (use other commands as reference) _(Garnet.server/Resp/RespCommand.cs)_
4. Implement the wrapper method to the storage layer in StorageSession _(Garnet.server/Storage/Session/ObjectStore/HashOps.cs)_
5. Define a new method in IGarnetAPI and implement it in GarnetApiObjectCommands, calling the method you have implemented in step #4 _(Garnet.server/API/IGarnetAPI.cs, Garnet.server/API/GarnetApiObjectCommands.cs)_
6. Add a new method to the RespServerSession class which will call the storage API and return the appropriate response _(Garnet.server/Resp/Objects/HashCommands.cs)_
8. Add a new method to HashObjImpl which will perform the required operation on the hash object _(Garnet.server/Objects/Hash/HashObjectImpl.cs)_
9. Add a new enum value to HashOperation and add the appropriate case to the switch in HashObject.Operate, in which you will call the method defined in step #8 _(Garnet.server/Objects/Hash/HashObject.cs)_
10. Add an appropriate case to the switch in TransactionManager.HashObjectKeys (Garnet.server/Transaction/TxnKeyManager.cs)
11. Add tests for the new command in the RespHashTests class _(Garnet.test/RespHashTests.cs)_

**Tip:** First add a simple test that calls the new command and use it to debug as you develop, it will make your life much easier!

If you have any questions, the Garnet team is here to help!

TalZaccai:
> I can help, can you assign this issue to me please? Thanks😊

Done! Thanks for your help! Let me know if you have any questions :)

PR #115, opened 2024-03-24
Support `HSTRLEN` command.

argsno:
Support `HSTRLEN` command. 

Returns the string length of the value associated with field in the hash stored at key. If the key or the field do not exist, 0 is returned.

Interested in being able to deepen my understanding by implementing some small features. thx.

refer: https://github.com/microsoft/garnet/issues/76


libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -148,6 +148,69 @@ private void HashLength(byte* output)
             ((ObjectOutputHeader*)output)->countDone = hash.Count;
         }
 
+        private void HashStrLength(byte* input, int length, ref SpanByteAndMemory output)
+        {
+            var _input = (ObjectInputHeader*)input;
+            int count = _input->count;
```
TalZaccai, line 154: Won't count always be 1 here? HSTRLEN doesn't accept more than 1 key per command

test/Garnet.test/RespHashTests.cs
```diff
@@ -160,6 +160,19 @@ public void CanDoHExists()
...
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+            db.HashSet("user.user1", new HashEntry[] { new HashEntry("Title", "Tsavorite") });
+            long r = db.HashStringLength("user.user1", "Title");
+            Assert.AreEqual(9, r, 0);
+            r = db.HashStringLength("user.user1", "NoExist");
+            Assert.AreEqual(0, r, 0);
+            r = db.HashStringLength("user.user2", "Title");
+            Assert.AreEqual(0, r, 0);
+        }
```
TalZaccai, line 175: Would be good to add a wrong # of tokens test here as well

TalZaccai (changes requested):
Nice work! Added a few comments :)
Thanks for your contribution!

badrishc:
Make sure you use the error handling logic introduced in this recent PR: https://github.com/microsoft/garnet/pull/103

test/Garnet.test/RespHashTests.cs
argsno: Sorry, what do you mean by "wrong # of tokens"? Is it the wrong number of arguments? This is the interface offer by StackExchange.Redis and have constraint two arguments already.

libs/server/Objects/Hash/HashObjectImpl.cs
argsno: You are right. I have rewritten the code to simplify the overhead needed to process.

argsno:
> Make sure you use the error handling logic introduced in this recent PR: #103

Yes, I have modify the code to use the same handling logic

test/Garnet.test/RespHashTests.cs
TalZaccai: Yeah same as the last test you wrote for LightClient, you can execute it with db.Execute(...) instead of using the predefined API methods... (e.g. db.Execute("HSTRLEN hashset key1 key2"))

argsno: Yeah, I have tried it and it shows that StackExchange.Redis has already checked the params number and threw an ArgumentOutOfRangeException in https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/CommandBytes.cs#L111.

So I add another test with LightClient.

yrajas:
Please make sure to update documentation to indicate HSTRLEN is supported now.
Under website/docs/commands/api-compatibility.md

argsno:
> Please make sure to update documentation to indicate HSTRLEN is supported now. Under website/docs/commands/api-compatibility.md

thanks, The api-compatibility.md  and data-structures.md file have been updated to reflect HSTRLEN support.

TalZaccai (approved):

TalZaccai:
I'm encountering an issue that seems to happen subsequently to when an error is returned. Please take a look...
image

argsno:
> I'm encountering an issue that seems to happen subsequently to when an error is returned. Please take a look... Note how after "wrong number of arguments" we're getting "unknown command" for the same input image

After parsing the incorrect command, a return is left. I apologize for the bug I introduced.

TalZaccai (approved):
Looks good! Thanks again for your work!

PR #118, opened 2024-03-24
Add ZMSCORE command support

### LINKED ISSUE #77: API Coverage - Implement ZMSCORE
Here's your opportunity to implement a RESP command in Garnet! 

**Syntax:** 
`ZMSCORE key member [member ...]`

**Command description:**
Returns the scores associated with the specified `members` in the sorted set stored at `key`.

For every `member` that does not exist in the sorted set, a nil value is returned.

**Response:**
One of the following:

_Nil reply:_ if the member does not exist in the sorted set.
_Array reply:_ a list of string `member` scores as double-precision floating point numbers.

---

**How to approach this task:**
1. Add the new command info to the sortedSetCommandsInfoMap in the RespCommandsInfo class _(Garnet.server/Resp/RespCommandsInfo.cs)_
2. Add the new command string to the returned HashSet in the RespInfo.GetCommands method _(Garnet.server/Resp/RespInfo.cs)_
3. Add the appropriate fast parsing logic for the new command in the RespCommand.FastParseArrayCommand method (use other commands as reference) _(Garnet.server/Resp/RespCommand.cs)_
4. Implement the wrapper method to the storage layer in StorageSession _(Garnet.server/Storage/Session/ObjectStore/SortedSetOps.cs)_
5. Define a new method in IGarnetAPI and implement it in GarnetApiObjectCommands, calling the method you have implemented in step #4 _(Garnet.server/API/IGarnetAPI.cs, Garnet.server/API/GarnetApiObjectCommands.cs)_
6. Add a new method to the RespServerSession class which will call the storage API and return the appropriate response _(Garnet.server/Resp/Objects/SortedSetCommands.cs)_
8. Add a new method to SortedSetObjImpl which will perform the required operation on the hash object _(Garnet.server/Objects/SortedSet/SortedSetObjectImpl.cs)_
9. Add a new enum value to SortedSetOperation and add the appropriate case to the switch in SortedSetObject.Operate, in which you will call the method defined in step #8 _(Garnet.server/Objects/SortedSet/SortedSetObject.cs)_
10. Add an appropriate case to the switch in TransactionManager.SortedSetObjectKeys (Garnet.server/Transaction/TxnKeyManager.cs)
11. Add tests for the new command in the RespSortedSetTests class _(Garnet.test/RespSortedSetTests.cs)_

**Tip:** First add a simple test that calls the new command and use it to debug as you develop, it will make your life much easier!

If you have any questions, the Garnet team is here to help!

ProTip:
I have tried my hand at adding the `ZMSCORE` command using the(very helpful) direction in the issue and surrounding code.

The score/scores fetching implementations could likely be merged.

resolves #77


ProTip:

badrishc:
Make sure you use the error handling logic introduced in this recent PR: https://github.com/microsoft/garnet/pull/103

ProTip:
> Make sure you use the error handling logic introduced in this recent PR: #103

I believe I am using the newly introduced command, however I have improved the count check and removed the arithmetic to be inline with the changes in the referenced PR.

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -450,7 +450,83 @@ private unsafe bool SortedSetScore<TGarnetApi>(int count, byte* ptr, ref TGarnet
...
+                *inputPtr = save;
+
+                switch (status)
+                {
+                    case GarnetStatus.OK:
+                        //process output
+                        var objOutputHeader = ProcessOutputWithHeader(outputFooter.spanByteAndMemory);
+                        ptr += objOutputHeader.bytesDone;
+                        break;
+                    case GarnetStatus.NOTFOUND:
```
TalZaccai, line 516: If key is not found, this is the error that should be displayed: "WRONGTYPE Operation against a key holding the wrong kind of value" (You might need to add it to CmdStrings...)

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -341,6 +341,29 @@ public void AddScore()
             Assert.AreEqual(expectedResponse, actualValue);
         }
 
+        [Test]
+        public void CanGetMemberScores()
```
TalZaccai, line 345: Please check the following scenario as I see an issue here:
(note that key2 does not exist in the set)

127.0.0.1:3278> zmscore key2 "one"
(nil)
127.0.0.1:3278> zmscore key2 "one"
(error) ERR unknown command

TalZaccai (changes requested):
Great work, thank you for your contribution to Garnet!
Added a few comments here :)

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -450,7 +450,83 @@ private unsafe bool SortedSetScore<TGarnetApi>(int count, byte* ptr, ref TGarnet
...
+                // Prepare header in input buffer
+                inputPtr->header.type = GarnetObjectType.SortedSet;
+                inputPtr->header.SortedSetOp = SortedSetOperation.ZMSCORE;
+                inputPtr->count = inputCount;
+                inputPtr->done = 0;
+
+                // Prepare GarnetObjectStore output
+                var outputFooter = new GarnetObjectStoreOutput { spanByteAndMemory = new SpanByteAndMemory(dcurr, (int)(dend - dcurr)) };
+
+                var status = storageApi.SortedSetScore(key, new ArgSlice((byte*)inputPtr, inputLength), ref outputFooter);
```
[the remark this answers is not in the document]
ProTip: Indeed.. Interesting that it gets routed to the correct location later.

libs/server/Resp/RespCommand.cs
```diff
@@ -515,6 +515,10 @@ private RespCommand FastParseCommand(byte* ptr)
                     if (*(long*)ptr == 4702694082085729572L && *(ushort*)(ptr + 8) == 3406 && *(ptr + 10) == 10)
                         return (RespCommand.SortedSet, (byte)SortedSetOperation.ZSCAN);
 
+                    //[$7|ZMSCORE|] = 13 bytes = 8 (long) + 2 (ushort) + 3 bytes
+                    if (*(long*)ptr == 4851306272719189796L && *(ushort*)(ptr + 8) == 21071 && *(ptr + 10) == 69)
```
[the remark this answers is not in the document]
ProTip: Ahh, that last pointer is a single int comparison. Cheers, it's been a long time since I've worked with pointers 😄

libs/server/Resp/Objects/SortedSetCommands.cs
ProTip: For reference I've checked the latest Redis docker image and for missing keys it will return `nil` on `ZSCORE` and a list of `nil` on `ZMSCORE` if the key does not exist.

I'm not sure if response parity is a goal here so I'm just asking about that(the documentation doesn't appear to be explicit about the missing key case).

I've also noticed that Garnet returns `nil` for `ZSCORE` currently even though it properly catches `NOTFOUND` and is supposedly writing the error response...

I can make both calls consistent and address that issue if it's desirable.

ProTip: Ah, so `RESP_ERRNOTFOUND` seems to be an alias for return `nil`. In any case I can update both to be consistent with Redis, the new error message, or etc.

Please confirm.

test/Garnet.test/RespSortedSetTests.cs
ProTip: This was a tricky one due to my lack of familiarity with the project. I was not "draining" the input in the case of the key not being found. This was resulting in the `ptr` not advancing and thus the `readHead` not advancing correctly..

I have a couple thoughts on this:

**Testing calling twice now seems moot**

I've resolved it here but are there tests patterns in place that would catch this issue on any command, or more generally?

```
127.0.0.1:3278> ZMSCORE key2 one
(error) WRONGTYPE Operation against a key holding the wrong kind of value.
127.0.0.1:3278> ZMSCORE key2 one
(error) WRONGTYPE Operation against a key holding the wrong kind of value.
127.0.0.1:3278> ZMSCORE key2 one
(error) WRONGTYPE Operation against a key holding the wrong kind of value.
```

**The existing patterns for constructing obj methods seems error prone**

Each method needing to individually implement the logic to advance the `readHead` properly is very tricky. Particularly as there are variations between methods on how it gets advanced, how `ptr` gets advanced, and etc.

I'm wondering in what circumstances it shouldn't be advanced to the end of the input command after object commands complete? Or rather, if there is some way to handle this reliably in one place.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -164,6 +164,57 @@ private void SortedSetScore(byte* input, ref SpanByteAndMemory output)
...
+            var _input = (ObjectInputHeader*)input;
+            ObjectOutputHeader _output = default;
+
+            int count = _input->count;
+
+            bool isMemory = false;
+            MemoryHandle ptrHandle = default;
+
+            byte* outPtr = output.SpanByte.ToPointer();
+            var outCurr = outPtr;
```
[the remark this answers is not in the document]
ProTip: I've updated the in/out ptr naming to be consistent with each other, within this method. I'm not sure if you meant to use a different method as a reference; it seems to be variable.

LMK if this isn't what you had in mind.

ProTip:
Cheers, that was a very good review😅 I believe I have addressed all the issues less some potential clarification in the unresolved review convos.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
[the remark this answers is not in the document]
TalZaccai: Oh sorry if I wasn't clear, I meant it's better to be consistent between methods in this class. Look at SortedSetAdd, for example.

libs/server/Resp/Objects/SortedSetCommands.cs
TalZaccai: Yes, we do strive to have parity with the existing RESP protocol responses, makes life easier for everyone (and especially when using 3rd party libraries to test Garnet against). 

And it looks like you're right, you should be returning Nil even when the key doesn't exist. Not sure where I got that error message before...

test/Garnet.test/RespSortedSetTests.cs
TalZaccai: Good catch! 

**Testing calling twice now seems moot** - you're correct, let me bring this up with the team and see what they think...

**The existing patterns for constructing obj methods seems error prone** - you're VERY correct, in fact, one of our engineers is working on an overhaul for this part of the code, so should get better pretty soon! :)

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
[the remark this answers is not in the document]
ProTip: Looking at `SortedSetAdd` it does not maintain a set of output pointers so `startptr`, `ptr`, and `end` are a bit ambiguous for this method. `SortedSetIncrement` uses `input_startptr` and etc for input, and `ptr`/`curr`/`end` for the output pointers 😐 

I could base the naming off `SortedSetIncrement`? I'm not sure that would be consistent with any other method or an improvement though..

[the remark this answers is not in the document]
TalZaccai: I agree, it is ambiguous, but at least it's consistent :D Sure, you can use any other method as naming reference...

libs/server/Resp/Objects/SortedSetCommands.cs
ProTip: I have updated the output to match the observed behavior from Redis; an array of nil matching the input member count.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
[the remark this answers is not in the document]
ProTip: The pointer names have been updated based on `SortedSetIncrement`.

TalZaccai (approved):
Looks good! Thanks again for your contribution! :)

PR #119, opened 2024-03-25
Improve FastParseCommand readability

aromaa:
Modified the code to use switch expression with pattern matching.

### DIFF
--- a/libs/server/Resp/RespCommand.cs
+++ b/libs/server/Resp/RespCommand.cs
@@ -204,11 +204,11 @@ namespace Garnet.server
                 count = ptr[1] - '1';
                 Debug.Assert(count is >= 0 and < 9);
 
-                var oldReadHead = readHead;
-
                 // Extract length of the first string header
                 var length = ptr[5] - '0';
                 Debug.Assert(length is > 0 and <= 9);
+
+                var oldReadHead = readHead;
 
                 // Ensure that the complete command string is contained in the package. Otherwise exit early.
                 // Include 10 bytes to account for array and command string headers, and terminator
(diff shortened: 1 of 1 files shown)


aromaa:

PaulusParssinen (approved):

aromaa:
Failures seem to be unrelated, other PRs are failing with same tests related to the TLS handshake.

aromaa:
Was actually able to find small workaround for the small regression on .NET 6. This is now a perf win all across the board.

libs/server/Resp/RespCommand.cs
```diff
@@ -125,201 +127,430 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
...
+                            switch ((*(1 + (int*)ptr) >> 8) & 0xFF)
+                            {
+                                case '6':
+                                    if (bytesRead - readHead >= 16) // TODO: make sure this condition is appropriate per command
+                                    {
+                                        if (*(1 + (long*)ptr) == Read<long>("RUNTXP\r\n"u8))
+                                            return RespCommand.RUNTXP;
+                                    }
+
+                                    break;
```
PaulusParssinen, line 157: Some variation of following should be possible to reduce indentation. Roslyn should lower to same conditions
```suggestion
                                case '6' when bytesRead - readHead >= 16:
                                    if (*(1 + (long*)ptr) == Read<long>("RUNTXP\r\n"u8))
                                        return RespCommand.RUNTXP;
```

aromaa: Thanks! Realized I could take this to the extreme actually x)

lmaas:
Great job! Thank you for your diligent work on this. There's definitely a lot of room for performance optimization in the parsing loop, and your approach seems spot-on.

However, I am holding off on merging this pull request for now. We're currently in the process of refactoring the parsing logic, and we have made changes to the semantics of the `FastParseCommand()` function (see PR #164). The new version you've proposed aligns well with the new interface, and I appreciate the improved readability.

If you have the time, I’d appreciate it if you could review the PR and provide any comments. Once #164 is merged, we can proceed with integrating this one.

lmaas:
PS: Thanks also for the benchmark! This is super useful. Once we have command parsing and execution separated, we should think about adding a proper parsing benchmark that follows your example.

aromaa:
I see, glad someone was already working on larger clean up on this. Looks like we had similar ideas already. Seems to be a very nice clean up :)

aromaa:
Updated the PR on top of main as #164 is merged now. I didn't touch any of the FastParseArrayCommand as the mileage varies on how much benefit you get from there with this pattern.

libs/server/Resp/RespCommand.cs
```diff
@@ -226,235 +226,67 @@ private RespCommand FastParseCommand(out int count)
...
+                        4 when lastWord == MemoryMarshal.Read<ulong>("\r\nPING\r\n"u8) => RespCommand.PING,
+                        4 when lastWord == MemoryMarshal.Read<ulong>("\r\nEXEC\r\n"u8) => RespCommand.EXEC,
+                        5 when lastWord == MemoryMarshal.Read<ulong>("\nMULTI\r\n"u8) => RespCommand.MULTI,
+                        6 when lastWord == MemoryMarshal.Read<ulong>("ASKING\r\n"u8) => RespCommand.ASKING,
+                        7 when lastWord == MemoryMarshal.Read<ulong>("ISCARD\r\n"u8) && ptr[8] == 'D' => RespCommand.DISCARD,
+                        7 when lastWord == MemoryMarshal.Read<ulong>("NWATCH\r\n"u8) && ptr[8] == 'U' => RespCommand.UNWATCH,
+                        8 when lastWord == MemoryMarshal.Read<ulong>("ADONLY\r\n"u8) && *(ushort*)(ptr + 8) == MemoryMarshal.Read<ushort>("RE"u8) => RespCommand.READONLY,
+                        9 when lastWord == MemoryMarshal.Read<ulong>("DWRITE\r\n"u8) && *(uint*)(ptr + 8) == MemoryMarshal.Read<uint>("READ"u8) => RespCommand.READWRITE,
+
+                        // Commands with fixed amount of arguments
```
lmaas, line 242: NIT: should be "number of arguments"

lmaas (changes requested):
Great work! The code looks solid overall and the performance looks good as well. I noticed two minor typos in the comments. Once those are fixed, I'll be happy to merge this PR. Thank you for your contribution! 🙌

libs/server/Resp/RespCommand.cs
aromaa: Thanks, fixed 👍

lmaas (approved):

PR #127, opened 2024-03-25
feat: add SISMEMBER command

Zzhiter:
I was going to finish this earlier: https://github.com/microsoft/garnet/issues/76. 
But before I raised the PR, I realized someone had already submitted it: https://github.com/microsoft/garnet/pull/115
So I went back and added the API that is not in garnet but is in Redis: SISMEMBER
Usage: SISMEMBER key member
which returns if member is a member of the set stored at key.
I'm happy to contribute code, looking forward to your review!


Zzhiter:

PR #137, opened 2024-03-26
bugfix: variable length parameter judge error

Zzhiter:
**The bug is not completely fixed in this PR**: https://github.com/microsoft/garnet/pull/103
There is also a lack of judgment about commands that can receive multiple parameters.
Examples：
image
image
But in Redis:
image
image


lmaas (approved):
Good catch! Code looks good to me. Thank you for your contribution! 🙌

PR #138, opened 2024-03-26
Add Arity Checks for Command Parameters in Garnet

argsno:
It has been noticed that the current implementation of checking parameters in each command is independent. 
Furthermore, some commands were previously implemented incorrectly 
and did not drain commands as they were supposed to, but this has been fixed as mentioned in https://github.com/microsoft/garnet/pull/103/files. 

However, it is still necessary to check the parameters for each command 
that may be constrained by arity. 

Therefore, this pull request adds an arity check for each command.


TalZaccai:
Hi! We've reviewed your PR and found that while the problem of arity values not being checked is correctly identified, this solution is unfortunately insufficient. We have added this work item to our list and will get to resolving it soon. Thank you for your contribution!

PR #139, opened 2024-03-26
Add format step to CI + continue on error

TalZaccai:
- Added dotnet format step to build jobs
- Set continue-on-error to true


darrenge:
My concern with continue on fail is that if it fails, it will show in the UI as a fail but will it pass the "check" for PR \ Merges.

TalZaccai:
> My concern with continue on fail is that if it fails, it will show in the UI as a fail but will it pass the "check" for PR \ Merges.

Addressed by switching "continue-on-error: true" with "fail-fast: false"

PR #144, opened 2024-03-26
Remove the subscriber when unsubscribe

hez2010:
Fix the issue where unsubscribe doesn't remove the subscriber. 
This can cause `ObjectDisposedException` later if the publisher publishes a new message after a subscriber unsubscribed and disconnected from the server.


hez2010:
To reproduce the bug:
1. Run RedisInsight, goto the Sub/Pub page
2. Subscribe and publish some messages
3. Unsubscribe and then continue to publish some messages
4. See exception from the server

badrishc (approved):

badrishc:
LGTM, thanks!

PR #145, opened 2024-03-26
Spanify TsavoriteLogRecoveryInfo deserialization

PaulusParssinen:
This PR uses `BinaryPrimitives.Read*` overloads to deserialize `TsavoriteLogRecoveryInfo` from a span instead of a stream that was always either ``MemoryStream`` or ``UnmanagedMemoryStream`` created from `byte[]` or a pointer.

I started to spanify the serialization too but got annoyed by potential cost of multiple UTF8 encoding calls, which cost is much harder to reason about mentally, so I'll leave that for later.


TedHartMS (approved):
Nice cleanup, thanks

PR #148, opened 2024-03-26
Minor follow-up optimizations

PaulusParssinen:
By making the fields properties or by inlining the statements, we allow JIT to consider the values such as record sizes and blittability to be constants. 

I missed these in https://github.com/microsoft/garnet/pull/39

edit: came across ~~one~~ two log2 acceleration oppurtunities I missed in #35 so included it in this PR if that's ok, i think that should be them all.. sry, for expanding the scope of PR


libs/storage/Tsavorite/cs/src/core/Allocator/MallocFixedPageSize.cs
```diff
@@ -76,21 +73,19 @@ public class MallocFixedPageSize<T> : IDisposable
...
         private volatile int writeCacheLevel;
 
         private volatile int count;
 
-        internal bool IsPinned => isPinned;
-        private readonly bool isPinned;
+        internal static int RecordSize => Unsafe.SizeOf<T>();
+        internal static bool IsPinned => Utility.IsBlittable<T>();
 
         private const bool ReturnPhysicalAddress = false;
```
PaulusParssinen, line 83: Could we remove this knob too? I checked through the FASTER git history and it seems to be unused.

TedHartMS: Yes, go ahead and remove it

TedHartMS (approved):

badrishc (approved):

libs/storage/Tsavorite/cs/src/core/Allocator/MallocFixedPageSize.cs
badrishc: Yes, let's remove these. Haven't needed in all these years. Can always resurrect if necessary later.

PaulusParssinen: We could simplify the class even further by removing support for non-blittable types, which are only used by a test where T is `IHeapContainer<Value>`. Might be better to leave that as potential follow-up, as there is more proper constraining opportunities in the allocator design elsewhere.

PR #154, opened 2024-03-27
Fix SpanByte ToString() to prevent Index out of range exception

KimJuHeok:
Fix SpanByte's ToString() that can cause index out of range exception when the length of the span is smaller than 8


badrishc (approved):

badrishc:
LGTM, thanks!

TedHartMS (approved):

PR #160, opened 2024-03-27
Seal more internal/private classes

PaulusParssinen:
This PR enables per [CA1852](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852) and [CA1813](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1813) as suggestions. We can also alternatively enable entire performance category analyzers. I went through some of the internal and private types and sealed them, unsealing them later is easy as is it is not a breaking change.

You can read more about potential benefits by sealing internal/private classes by default here: https://github.com/dotnet/runtime/issues/49944

note: skimming through usage of the classes I sealed, the benefits sealing are negligible at best as very few is stored in arrays (we would avoid covariance checks) or having virtual call dispatches.. However the concrete attributes should benefit from sealing and I think it's good pratice to seal any private or internal class with no-derived types by default as unsealing it later is free.


TedHartMS (approved):

badrishc:
Amazing stuff! You are on a roll :).

badrishc (approved):

PR #164, opened 2024-03-27
Refactor: Separate RESP Command Parsing and Execution

lmaas:
_This PR is a first step towards a more maintainable RESP parser design._

## Overview
This PR tries to separate command parsing and execution by making the following changes:

- All parsing has been unified in a single `ParseCommand()` function with clear semantics. If parsing is successful, the function returns the `RespCommand` ID, optionally a subcommand ID (if parsed) and the number of remaining elements in the RESP array (i.e., the number of command arguments)
- Every command has been assigned an ID in the `RespCommand` enum.

Additionally, the parsing code is now enforcing the following semantics on the parsing function and all operators:

1. **Parsing advances the read head** to the first string after the parsed command and (optional) subcommand
2. **All operators read from the read head** to remove inconsistency between reading from pointers and read head.
3. **count** represents the number of arguments *remaining* in the current RESP package.

Separating parsing and command execution will allow easier integration with components such as ACL and easier replacement of the parsing logic. In addition, it allows us to properly identify any potential bottlenecks in the current parser.

## Known Caveats
- Currently all parsing functions return `(RespCommand, byte)`, but most commands still do their own subcommand parsing inside the command executors. This will change in the future with new parser updates.
- There are inconsistencies in the interfaces of different command executors. Eventually we will need to converge to a standardized interface.


aromaa:
When working on #119 I ran to the problem that the heavy usage of `MemoryMarshal.Read` would hit the inlining budget where the JIT refuses to inline anymore methods no matter what (It starts to even ignore AggressiveInlining) at the end of the method. I worked around this by implementing simpler (and more unsafe) version of the method to reduce the IL size. The limits are increased on .NET 8 compared to on .NET 6. Did you happen to check check the assembly that the JIT actually inlines all these calls?

badrishc:
Hi @aromaa, we really appreciate your expertise with JIT and inlining to help make this work more optimized, thanks!

lmaas:
> When working on #119 I ran to the problem that the heavy usage of `MemoryMarshal.Read` would hit the inlining budget where the JIT refuses to inline anymore methods no matter what (It starts to even ignore AggressiveInlining) at the end of the method. I worked around this by implementing simpler (and more unsafe) version of the method to reduce the IL size and folding the JIT needs to do. These limits are increased on .NET 8 compared to .NET 6. Did you happen to check the assembly that the JIT actually inlines all these calls?

This is great insight! We have not checked yet if the inlining is happening correctly. This should definitely go on our to-do list.

I think we'll need a few optimization passes on this - especially on the core loop. So far, we have not tuned the parsing loop much and I am sure there is a lot of room for improvement. We'd definitely appreciate your help with this!

badrishc:
> When working on #119 I ran to the problem that the heavy usage of `MemoryMarshal.Read` would hit the inlining budget where the JIT refuses to inline anymore methods no matter what (It starts to even ignore AggressiveInlining) at the end of the method. I worked around this by implementing simpler (and more unsafe) version of the method to reduce the IL size and folding the JIT needs to do. These limits are increased on .NET 8 compared to .NET 6. Did you happen to check the assembly that the JIT actually inlines all these calls?

How do you check the assembly for this, any pointers?

PaulusParssinen:
> How do you check the assembly for this, any pointers?

I recommend using excellent VS extension by developed by one the JIT team members, called [Disasmo](https://github.com/EgorBo/disasmo). <sub><sup>warning: it's pretty addicting</sup></sub>

I believe you need to clone and build runtime for using the `print-inliness` to see the JIT inlining heuristics in action. You can also toggle JitDump -option to see every stage of JIT compilation.

https://github.com/microsoft/garnet/assets/26825431/a39f8c43-23bc-4743-9ff9-3a5296405247

If VS is not available for you (I doubt it 😆) you can follow [.NET Runtime guide](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/viewing-jit-dumps.md) to getting the (almost all) same information manually using ``dotnet run`` with ``DOTNET_JitDisasm=MethodName`` environment variables.

badrishc (approved):

vazois (approved):

libs/server/Resp/RespCommand.cs
```diff
@@ -120,602 +146,1200 @@ enum RespCommandOption : byte
...
-                            (*(long*)(ptr + 8) == 7929164607238311248L && *(int*)(ptr + 16) == 168653935))
-                            return (RespCommand.BITOP, (byte)BitmapOperation.NOT);
+                        // TODO: Error Handling should be done outside of parsing function.
+                        if (bytesRead - readHead < 2)
+                        {
+                            success = false;
+                            return (RespCommand.INVALID, subCmd);
+                        }
+
+                        if (*(ushort*)(recvBufferPtr + bytesRead - 2) != 2573)
```
TalZaccai, line 1309: nit: Might want to add a comment explaining the error conditions we're checking here

libs/server/Resp/RespServerSession.cs
```diff
@@ -351,61 +383,46 @@ private bool ProcessBasicCommands<TGarnetApi>(byte* ptr, RespCommand cmd, ref TG
...
-            var cmd = FastParseArrayCommand(count, recvBufferPtr + readHead);
-            var success = cmd switch
+            var success = (cmd, subcmd) switch
             {
                 (RespCommand.MGET, 0) => NetworkMGET(count, ptr, ref storageApi),
                 (RespCommand.MSET, 0) => NetworkMSET(count, ptr, ref storageApi),
                 (RespCommand.MSETNX, 0) => NetworkMSETNX(count, ptr, ref storageApi),
-                (RespCommand.DEL, 0) => NetworkDEL(count, ptr, false, ref storageApi),
-                (RespCommand.UNLINK, 0) => NetworkDEL(count, ptr, true, ref storageApi),
+                (RespCommand.UNLINK, 0) => NetworkDEL(count, ptr, ref storageApi),
```
TalZaccai, line 420: Where did RespCommand.DEL & RespCommand.EXISTS go? Was that on purpose?

libs/server/Resp/AdminCommands.cs
```diff
@@ -363,24 +340,26 @@ private bool ProcessAdminCommands<TGarnetApi>(ReadOnlySpan<byte> command, ReadOn
                         SendAndReset();
                 }
             }
-            else if (command.SequenceEqual(CmdStrings.RESET))
-            { }
-            else if (command.SequenceEqual(CmdStrings.QUIT))
+            else if (command == RespCommand.RESET)
+            {
+                // XXX: Not Implemented?
```
TalZaccai, line 345: Shouldn't we return something here?

libs/server/Resp/AdminCommands.cs
```diff
@@ -17,42 +17,19 @@ namespace Garnet.server
...
-            else if (command.SequenceEqual(CmdStrings.CONFIG)) return (RespCommand.CONFIG, 0);
-            else if (command.SequenceEqual(CmdStrings.CLIENT)) return (RespCommand.CLIENT, 0);
-            else if (command.SequenceEqual(CmdStrings.MEMORY)) return (RespCommand.MEMORY, 0);
-            else if (command.SequenceEqual(CmdStrings.MONITOR)) return (RespCommand.MONITOR, 0);
-
-            return (cmd, subcmd);
-        }
-
-        private bool ProcessAdminCommands<TGarnetApi>(ReadOnlySpan<byte> command, ReadOnlySpan<byte> bufSpan, int count, ref TGarnetApi storageApi)
+        private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<byte> bufSpan, int count, ref TGarnetApi storageApi)
```
TalZaccai, line 20: nit: Can we rename "count" to something more explanatory (like "tokenCount")?

libs/server/Resp/RespServerSession.cs
```diff
@@ -89,6 +88,24 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
         /// </summary>
         public readonly IClusterSession clusterSession;
 
+        /// <summary>
+        /// Current custom transaction to be executed in the session.
+        /// NOTE: This is temporary until the parser contains a proper parsing context.
```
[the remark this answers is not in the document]
lmaas: Those comments have been removed in the cleanup.

libs/server/Resp/RespServerSession.cs
```diff
@@ -330,12 +362,12 @@ private bool ProcessBasicCommands<TGarnetApi>(byte* ptr, RespCommand cmd, ref TG
...
                 RespCommand.PSETEX => NetworkSETEX(ptr, true, ref storageApi),
-                RespCommand.SETEXNX => NetworkSETEXNX(ptr, ref storageApi),
-                RespCommand.DEL => NetworkDEL(ptr, ref storageApi),
+                RespCommand.SETEXNX => NetworkSETEXNX(ptr, count - 2, ref storageApi),
+                RespCommand.DEL => NetworkDEL(count, ptr, ref storageApi),
                 RespCommand.RENAME => NetworkRENAME(ptr, ref storageApi),
-                RespCommand.EXISTS => NetworkEXISTS(ptr, ref storageApi),
-                RespCommand.EXPIRE => NetworkEXPIRE(ptr, RespCommand.EXPIRE, ref storageApi),
-                RespCommand.PEXPIRE => NetworkEXPIRE(ptr, RespCommand.PEXPIRE, ref storageApi),
+                RespCommand.EXISTS => NetworkEXISTS(count, ptr, ref storageApi),
```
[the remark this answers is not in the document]
lmaas: I fully agree! In the future, we should try to encapsule the whole parsing state (command/subcommand, current read head, and remaining arguments) and only pass those arguments together. For now, I have addressed this to unify the ordering to `(Command, count, ptr, storageAPI)`, as this required fewer changes than ensuring `ptr` is the first argument.

libs/server/Resp/RespServerSession.cs
lmaas: Those previously had to be checked twice, because they could be parsed in two different places (for single key and array commands). Now that parsing is happening before execution we are only keeping the invocations in `ProcessBasicCommands()`.

libs/server/Resp/AdminCommands.cs
lmaas: I have fully removed the command for now. When we decide to implement the command, we can add it back to the parsing loop.

libs/server/Resp/AdminCommands.cs
lmaas: I would say we keep that for the next iteration of parsing refactor. The argument handling for command executors probably has to change completely.

libs/server/Resp/RespCommand.cs
```diff
@@ -120,602 +146,1200 @@ enum RespCommandOption : byte
...
+                                }
+                                break;
+                        }
+
+                        // Reset optimistically changed state, if no matching command was found
+                        count += 1;
+                        readHead = oldReadHead;
+                    }
+                }
+                else if ((*(uint*)ptr & 0xFF0000FF) == MemoryMarshal.Read<uint>("$\0\0\r"u8))
```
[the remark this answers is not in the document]
lmaas: This has been addressed in the cleanup.

libs/server/Resp/RespCommand.cs
lmaas: This has been addressed by simplifying this function.

TalZaccai (approved):

PR #172, opened 2024-03-28
Fix HashOperation.HMSET Cannot Set On Existing Data

KimJuHeok:
When calling HMSET, the first initialization works as expectation, but second HMSET to initialized key just ignores the input.

And finally i found HashObject.SetOrNX() only targeted on HashOperation.HSET, so it ignores HMSET to existing data.
I think it needs to check both of HashOperation.


libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -364,7 +364,7 @@ private void SetOrSetNX(HashOperation op, byte* input, int length, byte* output)
                     this.UpdateSize(key, value);
                     _output->opsDone++;
                 }
-                else if (_input->header.HashOp == HashOperation.HSET && _value != default && !_value.SequenceEqual(value))
+                else if ((_input->header.HashOp == HashOperation.HSET || _input->header.HashOp == HashOperation.HMSET) && _value != default && !_value.SequenceEqual(value))
```
yrajas, line 367: Could you please add a test to fix the gap?

KimJuHeok:
```
        [Test]
        public void CanSetAndGetOnePair()
        {

            //First Attempt
            db.HashSet("user:user1", new HashEntry[] { new HashEntry("Title", "Tsavorite") });
            string r = db.HashGet("user:user1", "Title");
            Assert.AreEqual("Tsavorite", r);

            //Second Attempt <- It fails
            db.HashSet("user:user1", new HashEntry[] { new HashEntry("Title", "Tsavorite2") });
            string r2 = db.HashGet("user:user1", "Title");
            Assert.AreEqual("Tsavorite2", r2);
        }

```
I added the test code. This will fail when above fix is not applied.

yrajas: Thanks.
May be it wasn't clear. 
Could you please add it to the tests file like RespHashTests.cs

KimJuHeok: I added the test in RespHashTests.cs

KimJuHeok:

KimJuHeok:

yrajas (approved):

PR #185, opened 2024-03-29
Fixing additional "wrong number of arguments" issues in object store commands

TalZaccai:
Additional "wrong number of arguments" fixes in object store commands
See previous related pr #103 

Fixed commands in this PR: ZADD, HKEYS, LPUSH / RPUSH, LPOP, SPOP


libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -344,6 +344,12 @@ private unsafe bool SetPop<TGarnetApi>(int count, byte* ptr, ref TGarnetApi stor
         {
             ptr += 10;
 
+            if (count < 2)
```
lmaas, line 347: I think we need to check the upper bound here as well (count > 3).

lmaas (changes requested):

lmaas (approved):

Issue #187, opened 2024-03-29
COMMAND COUNT not supported


tisilent:
![commandcount](https://github.com/microsoft/garnet/assets/48614781/b6b8a380-7e1c-4536-b11e-04fe8705e7d7)

PR #188, opened 2024-03-29
add COMMAND COUNT

tisilent:
#187


TalZaccai:
Thank you for your contribution! Unfortunately this does not currently return the correct response and can be confusing as there are commands missing from RespInfo that will not be counted (e.g. CLUSTER commands, custom commands etc.). 
We will address this missing COUNT command when we implement the actual COMMAND command itself first (in the works). 
Closing this for now, but we might use some of it when we do implement COMMAND COUNT...
Thanks again & much appreciated!

PR #201, opened 2024-03-30
Reduce allocations around the use of ArgSlice

PaulusParssinen:
This PR includes various little refactoring and changes that can be considered ["peanut-butter"](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#peanut-butter) as Toub terms them.

<sup>I have shared my concerns about this type in general on Garnet discord already so I'm not gonna repeat them here 😄</sup> 

### Commits
* [Make `ArgSlice.Bytes` a method and rename it to ToArray](https://github.com/microsoft/garnet/commit/05e0f135ee14de007c5eecff194bbfe8ae13041c)
  * This makes it clearer to the API consumers that this action is not free. This ``To``-prefix pattern may seem familiar from the BCL.
*  [Mark all properties on ArgSlice as readonly](https://github.com/microsoft/garnet/commit/eae66f37dd487d7f6b8f9299306e3afc273b1d2b)
* [Remove unnecessary allocations for ArgSlice length checks](https://github.com/microsoft/garnet/commit/642365e2a41f5313fb738bb5cd81146c306fb10e)
  * It indeed was not clear whether accessing `Bytes` is cheap. Use `Length` property directly instead and avoid the unnecessary allocation.
* [Remove misc. allocations relating to use of ArgSlice](https://github.com/microsoft/garnet/commit/a7d416eaf180ba34f12a74119f8e5d4be02a16b3)
* [Switch some ArgSlice initializations to use constructor instead of initializer expression](https://github.com/microsoft/garnet/commit/989b387598512113ad5321e03bb8736d720b396e)
* [Use the public ``ArgSlice.Length`` property instead of the internal field where applicable](https://github.com/microsoft/garnet/commit/aac454a85f13f19d8d3348aec8c62ec4d88c3561)


test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -526,7 +526,7 @@ public void CreateTestLibrary(string[] namespaces, string[] referenceFiles, stri
                 var source = File.ReadAllText(filePath);
                 var stringText = SourceText.From(source, Encoding.UTF8);
                 return SyntaxFactory.ParseSyntaxTree(stringText,
-                    CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.CSharp10), string.Empty);
+                    CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.Latest), string.Empty);
```
PaulusParssinen, line 529: The test failure was caused by https://github.com/microsoft/garnet/pull/201/commits/a7d416eaf180ba34f12a74119f8e5d4be02a16b3 where I updated ``ReadWriteTxn`` to use new C# 12 features. This caused `RespCustomCommandTests.CreateTestLibraries` to fail as it pulls the ``ReadWriteTxn.cs`` in for compilation:
https://github.com/microsoft/garnet/blob/4971e20e884da234cfcbcec6e59b939b54a0fa06/test/Garnet.test/RespCustomCommandTests.cs#L610-L614

libs/server/ArgSlice/ArgSlice.cs
```diff
@@ -4,14 +4,16 @@
...
 using Tsavorite.core;
 
 namespace Garnet.server
 {
     /// <summary>
-    /// Slice of Key
+    /// Represents contiguous region of arbitrary _pinned_ memory.
     /// </summary>
+    /// <remarks>
+    /// This type is used to represent arguments that are assumed to point to pinned memory.
```
PaulusParssinen, line 15:
```suggestion
    /// SAFETY: This type is used to represent arguments that are assumed to point to pinned memory.
```

TedHartMS (approved):
Good changes--nice catch on "Bytes" usage

badrishc:
@PaulusParssinen - could you resolve the conflicts on this PR? Thanks!

PR #203, opened 2024-03-31
Fix typos in SortedSetOps.cs

### LINKED ISSUE #202: Copy paste err from line above?
https://github.com/microsoft/garnet/blob/05643c36e636dd6914653e408aa27a9c73d3ddf4/libs/server/Storage/Session/ObjectStore/SortedSetOps.cs#L178

PaulusParssinen:
Closes #202 

Where should I add regression test for this? It seems the testing coverage for this layer of the API is minimal (I only see couple overloads tested in `TestProcedureSortedSets`


libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -175,7 +175,7 @@ public unsafe GarnetStatus SortedSetRemoveRangeByLex<TObjectContext>(ArgSlice ke
                 return GarnetStatus.OK;
 
             var minBytes = Encoding.ASCII.GetBytes(min);
-            var maxBytes = Encoding.ASCII.GetBytes(min);
+            var maxBytes = Encoding.ASCII.GetBytes(max);
```
PaulusParssinen, line 178: Why these APIs that take scores as strings instead of just `double`? Am I missing something 😅

gmartinezsan: TLDR;

ZREMRANGEBYSCORE myzset -inf (2
ZREMRANGEBYSCORE myzset -inf +inf
ZREMRANGEBYSCORE myzset -inf (2
ZREMRANGEBYLEX myzset [alpha [omega

the zrange* api commands are about selecting a range of elements in the sorted set based on the value or member or score of the each pair. In the case of the ZREMRANGEBYLEX, this command works in the scenario where the sorted set has many elements with the same score, so the values of min and max parameters are string, here are the docs from redis: 

I'm sharing also the ZRANGEBYLEX because it has the full explanation of how the min and max parameters work, and also the variations both use. 
SE Redis has something very similar, but they are just using their own RedisValue type which is a string.
 
ZREMRANGEBYLEX can use also specific characters to indicate close or open range, or the +inf, -inf etc. Same case for ZREMRANGEBYSCORE, but this is selecting based on the score,  also it can have the characters [, (, ), [, or the +inf, -inf as ZREMRANGEBYLEX to indicate the way the range will work.
  
ZRANGEBYLEX
https://redis.io/commands/zrangebylex/
 
ZREMRANGEBYLEX
https://redis.io/commands/zremrangebylex/

PaulusParssinen: I see, thank you for the rationale behind it! I should've probably checked the spec 😅

TalZaccai (approved):

TalZaccai:
> Closes #202
> 
> Where should I add regression test for this? It seems the testing coverage for this layer of the API is minimal (I only see couple overloads tested in `TestProcedureSortedSets`

SortedSet tests are located here: Garnet.test/RespSortedSetTests.cs

PR #209, opened 2024-04-01
docs: provide helpful example on how to enable password auth when using docker compose

wmundev:
- provide helpful example on how to enable password auth when using docker compose so that whoever uses docker compose can easily add password auth if they wanted to

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
     # To avoid docker NAT, consider `host` mode.
     # https://docs.docker.com/compose/compose-file/compose-file-v3/#network_mode
     # network_mode: "host"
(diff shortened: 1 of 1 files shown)


badrishc:
Instead of editing docker-compose.yml, we should add this documentation here: 

https://microsoft.github.io/garnet/docs/welcome/releases

Can do this by a PR to this file: https://github.com/microsoft/garnet/blob/main/website/docs/welcome/releases.md

wmundev:
thanks badrishc, 
i've updated the PR, something like this? https://github.com/microsoft/garnet/pull/209/files#diff-0348affd91795660ce75ed4eead47960881f2511e6b5f63b224680c4f044d161R41

Issue #210, opened 2024-04-01
lrange performance problem


hbprotoss:
CPU: 16C
MEM: 16G

test server process inside docker, both garnet and redis use one single process node

gartner built with release conf

use redis-benchmark for testing

```
garnet
PING_INLINE: 37230.08 requests per second, p50=0.743 msec                   
PING_MBULK: 36576.45 requests per second, p50=0.759 msec                   
SET: 36376.86 requests per second, p50=0.759 msec                   
GET: 35676.06 requests per second, p50=0.767 msec                   
INCR: 35855.14 requests per second, p50=0.759 msec                   
LPUSH: 36258.16 requests per second, p50=0.767 msec                   
RPUSH: 36697.25 requests per second, p50=0.759 msec                   
LPOP: 36616.62 requests per second, p50=0.759 msec                   
RPOP: 37091.99 requests per second, p50=0.759 msec                   
SADD: 36536.36 requests per second, p50=0.759 msec                   
HSET: 35625.22 requests per second, p50=0.783 msec                   
SPOP: 36443.15 requests per second, p50=0.767 msec                   
ZADD: 36805.30 requests per second, p50=0.751 msec                   
ZPOPMIN: 35893.75 requests per second, p50=0.759 msec                   
LPUSH (needed to benchmark LRANGE): 36536.36 requests per second, p50=0.759 msec                   
LRANGE_100 (first 100 elements): 14766.69 requests per second, p50=2.783 msec                   
LRANGE_300 (first 300 elements): 2128.43 requests per second, p50=18.831 msec                   
LRANGE_500 (first 500 elements): 801.81 requests per second, p50=47.871 msec                  
LRANGE_600 (first 600 elements): 587.45 requests per second, p50=64.447 msec                   
MSET (10 keys): 30413.62 requests per second, p50=0.919 msec
```

CPU 400%
lrange use full cpu, memory 25%

```
redis
PING_INLINE: 34855.35 requests per second, p50=0.815 msec                   
PING_MBULK: 36995.93 requests per second, p50=0.783 msec                   
SET: 36536.36 requests per second, p50=0.799 msec                   
GET: 35625.22 requests per second, p50=0.807 msec                   
INCR: 35298.27 requests per second, p50=0.807 msec                   
LPUSH: 36075.04 requests per second, p50=0.807 msec                   
RPUSH: 35536.61 requests per second, p50=0.799 msec                   
LPOP: 36995.93 requests per second, p50=0.799 msec                   
[...]

PR #220, opened 2024-04-02
fix lrange perf issue

hbprotoss:
#210


libs/server/Objects/List/ListObjectImpl.cs
```diff
@@ -187,9 +187,15 @@ private void ListRange(byte* input, ref SpanByteAndMemory output)
                         while (!RespWriteUtils.WriteArrayLength(count, ref curr, end))
                             ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
 
-                        for (int i = start; i <= stop; i++)
+                        var i = -1;
+                        foreach (var bytes in list)
                         {
-                            while (!RespWriteUtils.WriteBulkString(list.ElementAt(i), ref curr, end))
+                            i++;
```
badrishc, line 193: These offsets can also be negative numbers indicating offsets starting at the end of the list. They can also be larger than the number of elements in the list.

For example, -1 is the last element of the list, -2 the penultimate, and so on. See https://redis.io/commands/lrange/.

So if a start or end offset is -X, you should convert it to `Max(list.Count - X, 0)`
If a start of end offset is non-negative X, you should convert it to `Min(X, list.Count - 1)`

Test case would be: `LRANGE mylist -100 100`

badrishc: If start is larger than the end of the list, an empty list is returned. If stop is larger than the actual end of the list, Redis will treat it like the last element of the list.

hbprotoss: > These offsets can also be negative numbers indicating offsets starting at the end of the list. They can also be larger than the number of elements in the list.
> 
> For example, -1 is the last element of the list, -2 the penultimate, and so on. See https://redis.io/commands/lrange/.
> 
> So if a start or end offset is -X, you should convert it to `Max(list.Count - X, 0)` If a start of end offset is non-negative X, you should convert it to `Min(X, list.Count - 1)`
> 
> Test case would be: `LRANGE mylist -100 100`



`start` and `stop` have been fixed to a positive number in previous code, and test is ok

```
127.0.0.1:3278> lpush list 1 2 3 4 5
(integer) 5
127.0.0.1:3278> lrange list 0 99
1) "5"
2) "4"
3) "3"
4) "2"
5) "1"
127.0.0.1:3278> lrange list -100 100
1) "5"
2) "4"
3) "3"
4) "2"
5) "1"
127.0.0.1:3278> lrange list -3 -1
1) "3"
2) "2"
3) "1"
127.0.0.1:3278> lrange list 2 4
1) "3"
2) "2"
3) "1"
```

badrishc (approved):

PR #221, opened 2024-04-02
Conditionally cleanup Garnet files when disposing GarnetServer object.

### LINKED ISSUE #135: All the file in the working directory will be deleted automatically when the exception was throw in Start() method in GarnetServer.cs
It was found that when a port is occupied by another program, it can cause an exception to be thrown from Start() method in GarnetServer.cs. Then the dispose() method wil be executed (when the above exception was catched outside), and all the files in the working directory will be deleted automatically without any warnings or reminders by "ckptdir.Delete()". Is this a bug?

```csharp
public void Dispose()
   {
       InternalDispose();

       logFactory?.Delete(new FileDescriptor { directoryName = "" });
       if (opts.CheckpointDir != opts.LogDir && !string.IsNullOrEmpty(opts.CheckpointDir))
       {
           var ckptdir = opts.DeviceFactoryCreator();
           ckptdir.Initialize(opts.CheckpointDir);
           ckptdir.Delete(new FileDescriptor { directoryName = "" });
       }
   }

```

in this case opts.LogDir is the default value (null ).

yrajas:
Fixes #135 

Default behavior is to not cleanup data directories.
However, it is possible to create GarnetServer object with an option to cleanup folders on dispose.

### DIFF
--- a/libs/host/GarnetServer.cs
+++ b/libs/host/GarnetServer.cs
@@ -70,7 +70,7 @@ namespace Garnet
         /// </summary>
         /// <param name="commandLineArgs">Command line arguments</param>
         /// <param name="loggerFactory">Logger factory</param>
-        public GarnetServer(string[] commandLineArgs, ILoggerFactory loggerFactory = null)
+        public GarnetServer(string[] commandLineArgs, ILoggerFactory loggerFactory = null, bool cleanupDir = false)
         {
             Trace.Listeners.Add(new ConsoleTraceListener());
 
@@ -119,6 +119,7 @@ namespace Garnet
 
             // Assign values to GarnetServerOptions
             this.opts = serverSettings.GetServerOptions(this.loggerFactory.CreateLogger("Options"));
+            this.cleanupDir = cleanupDir;
             this.InitializeServer();
         }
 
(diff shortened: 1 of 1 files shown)


luojin765:
Hi,  in my case, this cleanup occurred with all option using default . I debugged the process, and traced behavior to "LocalStorageNamedDeviceFactory.cs". In "Delete" method, fileInfo.directoryName was found to be "",  and then the basePath ("dir") would be cleaned.
When the start of garnet failed, all the files in the directory are missing, including the config file and others. 

```csharp
//LocalStorageNamedDeviceFactory.cs

public void Delete(FileDescriptor fileInfo)
{
    long startTime = DateTimeOffset.UtcNow.Ticks;
    if (fileInfo.fileName != null)
    {
        //codes not executed in my case...
    }
    else
    {
        var dir = new DirectoryInfo(Path.Combine(baseName, fileInfo.directoryName)); //fileInfo.directoryName=="", dir==baseName
        while (true)
        {
            try
            {
                if (dir.Exists) dir.Delete(true);
                break;
            }
            catch { }
            Thread.Yield();
            // Retry until timeout
            if (DateTimeOffset.UtcNow.Ticks - startTime > TimeSpan.FromSeconds(5).Ticks) break;
        }
    }
}
```

badrishc (approved):

PR #224, opened 2024-04-02
Improve dockerfiles

akoken:
This PR contains the following changes.

- Updated Dockerfiles to respect Docker layer caching. The current Dockerfiles copy all the files to the build context, which means every time you make a change in the codebase, the `dotnet restore` command runs even though you haven't changed anything in the csproj file.
- Set linux distro as Ubuntu Jammy in [Dockerfile.ubuntu](https://github.com/microsoft/garnet/blob/main/Dockerfile.ubuntu). The default Linux distro for . NET container images is Debian.
- Added Alpine based Dockerfile.
- Added non-root user option to Dockerfiles.


badrishc:
@Niek - could you review this PR as well?

Dockerfile
```diff
@@ -1,24 +1,40 @@
...
+RUN dotnet publish -a $TARGETARCH -c Release -o /app --no-restore --self-contained false -f net8.0
 
 # Final stage/image
-FROM mcr.microsoft.com/dotnet/runtime:8.0
+FROM mcr.microsoft.com/dotnet/runtime:8.0 AS runtime
 WORKDIR /app
 COPY --from=build /app .
 
+# Run container as a non-root user
+USER $APP_UID
```
Niek, line 34: Where is `APP_UID` coming from? It's not a standard build environment variable AFAIK, and it's not imported with `ARG`>

Niek (changes requested):
What about changes to the https://github.com/microsoft/garnet/blob/main/.github/workflows/docker.yml file? Caching can be enabled there, as well as adding the other docker files as separate tags.

Dockerfile
akoken: It's coming from [base](https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/8.0/bookworm-slim/amd64/Dockerfile) .NET image as default.

akoken:
> What about changes to the https://github.com/microsoft/garnet/blob/main/.github/workflows/docker.yml file? Caching can be enabled there, as well as adding the other docker files as separate tags.

Setting up CI pipeline for docker layer caching is not part of this PR. I'm not very familiar with Github Actions, but my next PR could be about it. This PR only contains Dockerfile best practices.

Niek (approved):

akoken:
> Fair enough re GitHub workflows.
> 
> One more nitpick: the dockerignore doesn't add much value it seems, since there is no wildcard copy any longer in the dockerfiles.

In addition to excluding Dockerfiles, we've also omitted future possible docker-compose and README files from the dockerignore. While the few KB saved might not seem significant for GitHub Actions, the value is considerably higher in the context of the development environment. For instance, when I build the solution and then create a Docker image in my local environment, excluding the bin and obj folders from the build context, as highlighted in [#20](https://github.com/akoken/garnet/blob/main/Dockerfile#L20), proves to be quite beneficial. 

As I mentioned before, it's a best practice and having a `.dockerignore` file is better than not having one at all IMHO.

badrishc:
Good comments and discussions, thanks to both of you!

Dockerfile.nanoserver
irinasp, line 1: Dockerfile.nanoserver is throwing the error while building the image. The build command 
docker build -t nano -f dockerfile.nanoserver .
The error is 
C:\Program Files\dotnet\sdk\8.0.203\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(154,5): error NETSDK1083: The specified RuntimeIdentifier 'win-
$TARGETARCH' is not recognized.
See
https://aka.ms/netsdk1083
for more information. [C:\src\main\GarnetServer\GarnetServer.csproj::TargetFramework=net8.0]
...

irinasp:
@akoken, thank you for contributing to Garnet. I built and run all Linux dokerfiles successfully but Dockerfile.nanoserver is failed to build. I've added the comment for this file.  Could you please take a look.

akoken:
Hi @irinasp 

It should be fixed now.

Dockerfile.nanoserver
akoken: It should be fixed now.

Dockerfile.nanoserver
```diff
@@ -1,17 +1,29 @@
...
+COPY Garnet.snk Garnet.snk
+COPY libs/ libs/
+COPY main/ main/
+COPY metrics/ metrics/
+COPY test/testcerts test/testcerts
 
-# Copy and publish app and libraries
-RUN dotnet publish -c Release -o /app -r win-x64 --self-contained false -f net8.0
+WORKDIR /src/main/GarnetServer
+RUN dotnet publish -c Release -o /app -r win-x64 --no-restore --self-contained false -f net8.0
```
irinasp, line 26: The build is still failing for Dockerfile.nanoserver but in a different place now. 
The build command is: docker build -t nanosvr -f dockerfile.nanoserver .
MSBuild version 17.9.6+a4ecab324 for .NET
C:\Program Files\dotnet\sdk\8.0.203\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1047: Assets file 'C:\src\main\GarnetServer\obj\project.assets.json' doesn't have a target for 'net8.0/win-x64'. Ensure that restore has run and that you have included 'net8.0' in the TargetFrameworks for your project. You may also need to include 'win-x64' in your project's RuntimeIdentifiers. [C:\src\main\GarnetServer\garnetserver.csproj::TargetFramework=net8.0]
The command 'cmd /S /C dotnet publish -c Release -o /app -r win-x64 --no-restore --self-contained false -f net8.0' returned a non-zero code: 1

irinasp:
Hi @akoken, I built and run the fixed version of Dockerfile.nanoserver but it still failing to build. I've added the comment. Could you please take a look.

Dockerfile.nanoserver
akoken: Looks like there's a runtime identifier problem between restore and publish operations. I've pushed a fix so that both operations use the current system runtime.

akoken:
Hi @irinasp 

I'm sorry for inconveniencing you. Since my development environment isn't on Windows, I have been unable to replicate the issue. Would you kindly consider testing the latest change I made? If the issue persists, I believe I may need to create a virtual machine to conduct further testing.

irinasp:
@akoken, there are a lot of warnings while building the image from the dockerfile with your new fix. And there is the error while running the container: 
docker: Error response from daemon: container e3e0ce56eb2163da9a1507ed388a21140066806acc218e8831e2944a6110b3ca encountered an error during hcs::System::CreateProcess: ./GarnetServer.exe -i 128m --port 6379: failure in a Windows system call: The system cannot find the file specified. (0x2)
[Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
[Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
[Event Detail: onecore\vm\compute\management\orchestration\vmhostedcontainer\processmanagement.cpp(173)\vmcomputeagent.exe!00007FF79388FDFD: (caller: 00007FF793834BB7) Exception(2) tid(3cc) 80070002 The system cannot find the file specified.
    CallContext:[\Bridge_ProcessMessage\VmHostedContainer_ExecuteProcess]
 Provider: 00000000-0000-0000-0000-000000000000].

Would you consider removing the update for Dockerfile.nanoserver from this PR and do separate PR for windows dockerfile after tunning the changes?

akoken:
> Would you consider removing the update for Dockerfile.nanoserver from this PR and do separate PR for windows dockerfile after tunning the changes?

I was thinking the same way @irinasp. I reverted the change for nanoserver. Thanks for your help!

irinasp (approved):
The changes for all Linux dockerfiles are tested. Dockerfile.nanoserver is reverted to original version.

badrishc (approved):

PR #227, opened 2024-04-02
fix: Replication checkpoint recovery fix 

### LINKED ISSUE #226: CLUSTERDOWN for read  - Replica stuck in recovery
Replicas are marked as recovering after they receive `CLUSTER REPLICATE` and before they attach to primary to snapshot. However, if primary is down and sync is requested, replica will reset it's state but not recovery flag. If it's assigned as primary later, it will accept writes but not reads and report CLUSTERDOWN as per https://github.com/microsoft/garnet/blob/main/libs/cluster/Session/ClusterSlotVerify.cs#L33

``` txt
Repro:
1. Start N1 primary and replica N2. 
2. Issue `CLUSTER REPLICATE`<node id of N1> to N2. (N2 transitions to secondary state)
3. Kill N1 (there might be a race and this needs to happen quickly)
4. N2 will revert to unassigned state if N1 fails initial checkpoint request but is stuck in recovery state.
5. Assign N2 as primary with some slots
6.  Send read request to N2
7. CLUSTERDOWN reported
```

Though a node can still be stuck in recovery after this stage - it will atleast transition to secondary if this step completes. As a result, we can take corrective action  - ideally good idea would be to expose metric like Redis's  `master_sync_last_io_seconds_ago` so admin can take corrective action.

msft-paddy14:
This PR resolves #226 
There is already support to reset state to unassigned when initiating checkpointing fails, we will also reset recovery state with this as this clears the node for usage as primary ("recovery" on primary doesn't make sense). 
Also, adding support to expose a `master_sync_last_io_seconds_ago` metric similar to redis which when combined with `master_sync_in_progress` can give information on checkpoint status and whether completion can be expected based on reasonable timeouts for cases where we initiate checkpoint and move to secondary state.


libs/cluster/Server/Replication/ReplicationManager.cs
```diff
@@ -24,9 +24,16 @@ internal sealed partial class ReplicationManager : IDisposable
 
         readonly ILogger logger;
         bool _disposed;
+
+        private DateTime primary_sync_last_time;
```
vazois, line 28: Use timestamp instead of DateTime

libs/cluster/Server/Replication/ReplicationManager.cs
```diff
@@ -126,6 +133,10 @@ void CheckpointVersionShift(bool isMainStore, long oldVersion, long newVersion)
             storeWrapper.EnqueueCommit(isMainStore, newVersion);
         }
 
+        public void Reset()
+        {
+            recovering = false;
+        }
```
vazois, line 139: nit; add newline here

vazois (changes requested):

vazois (approved):

PR #229, opened 2024-04-03
Increment type check

### LINKED ISSUE #228: incr a non-number key-value should return error
__> set key1 'asd' 
OK
__> get key1
"asd"

__> incr key1
(integer) 5623

expect:  error
actual:  (integer) 5623

__> get key1
"5623"

expect: "asd"
actual: "5623"

tisilent:
fix #228 
todo: increment or decrement would overflow  -- done


libs/common/NumUtils.cs
```diff
@@ -513,5 +513,51 @@ public static unsafe bool TryBytesToInt(byte* source, int len, out int result)
...
+                        return false;
+                    }
+                }
+                else
+                {
+                    if (length > 19)
+                    {
+                        return false;
+                    }
+                }
```
yrajas, line 544: Length check could be skipped.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -829,8 +829,12 @@ public GarnetStatus Increment<TContext>(ArgSlice key, ArgSlice input, ref ArgSli
             var status = context.RMW(ref _key, ref _input, ref _output);
             if (status.IsPending)
                 CompletePendingForSession(ref status, ref _output, ref context);
-            Debug.Assert(_output.IsSpanByte);
```
PaulusParssinen, line 832: Why removed?

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -829,8 +829,13 @@ public GarnetStatus Increment<TContext>(ArgSlice key, ArgSlice input, ref ArgSli
             var status = context.RMW(ref _key, ref _input, ref _output);
             if (status.IsPending)
                 CompletePendingForSession(ref status, ref _output, ref context);
+
             Debug.Assert(_output.IsSpanByte);
             output.length = _output.Length;
+            if (_output.SpanByte.AsReadOnlySpan().SequenceEqual(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER))
+            {
+                return GarnetStatus.ERROR;
+            }
```
tisilent, line 838: I'm not sure if there are any other ways, as it may affect performance. 😕

vazois:
Great work identifying the issue. I think a better approach would be to increment the output buffer by a single byte and use this to implicitly pass the error back to the caller.
I have created a PR for it covering the invalid number case (https://github.com/microsoft/garnet/pull/262)

PR #233, opened 2024-04-04
Make MemorySize a multiple of page size and not necessarily a power of two

yrajas:
MemorySize is currently enforced to be a power of two. If memory is configured to be not a power of two, then the size is adjusted dynamically using EmptyPageCount. Also, the EmptyPageCount is not enforced during recovery.

This change removes the power of two constraint for MemorySize. With this, the memory size is constrained to be a multiple of PageSize. This makes MemorySize to be constant during recovery as well as regular operation.

The EmptyPageCount would still be used to control Object Store GC heap memory size.


libs/storage/Tsavorite/cs/src/core/TsavoriteLog/TsavoriteLogSettings.cs
```diff
@@ -204,7 +204,7 @@ internal LogSettings GetLogSettings()
                 LogDevice = LogDevice,
                 PageSizeBits = Utility.NumBitsPreviousPowerOf2(PageSize),
                 SegmentSizeBits = Utility.NumBitsPreviousPowerOf2(SegmentSize),
-                MemorySizeBits = ReadOnlyMode ? 0 : Utility.NumBitsPreviousPowerOf2(MemorySize),
+                MemorySizePages = ReadOnlyMode ? 0 : (int)(MemorySize / (1L << Utility.NumBitsPreviousPowerOf2(PageSize))),
                 ReadCopyOptions = ReadCopyOptions.None,
```
TedHartMS, line 208: Use PageSizeBits instead of repeating Utility.NumBitsPreviousPowerOf2

Issue #235, opened 2024-04-04
SPOP with count bug


TalZaccai:
Another bug relating to SPOP - when calling SPOP with `count` greater than the set size, subsequent calls to SADD return an empty array.

Example: 

![image](https://github.com/microsoft/garnet/assets/18443527/e85310e3-f890-4637-9725-3539689570f7)

TalZaccai:
@Zzhiter FYI - this is not related to your recent PR #136, but a new bug that I've discovered due to your fix.

Zzhiter:
![image](https://github.com/microsoft/garnet/assets/61727602/80bdc2ea-3e44-400a-a84d-ca32d1072db2)
I repeated what you did and realized that when I entered the SADD command, it actually still judged it as a SPOP, which is why the response "(empty list or set)" looks like a bug introduced by this PR: https://github.com/ microsoft/garnet/commit/4cba6ff89ac19f9d5aae67b4b47c664abe15ca36

Zzhiter:
> @Zzhiter FYI - this is not related to your recent PR #136, but a new bug that I've discovered due to your fix.

Actually this fix is ​​not mine😂

Zzhiter:
![image](https://github.com/microsoft/garnet/assets/61727602/fc66cd96-2d9a-49df-8ee0-19f45d3a1cc7)
I seem to have found the cause.
When calling SPOP with count greater than the set size，it will return false.
Fix: https://github.com/microsoft/garnet/pull/239

TalZaccai:
> > @Zzhiter FYI - this is not related to your recent PR #136, but a new bug that I've discovered due to your fix.
> 
> Actually this fix is ​​not mine😂

Oops sorry, meant #191 :)

TalZaccai:
[...]

PR #239, opened 2024-04-05
fix: SPOP with count bug

Zzhiter:
https://github.com/microsoft/garnet/issues/235


TalZaccai (approved):
Looks good! Thanks for looking into this and fixing the issue!

Issue #231, opened 2024-04-03
API Coverage - Implement SRANDMEMBER


TalZaccai:
Here's your opportunity to implement a couple of RESP commands in Garnet! 

## SRANDMEMBER

**Syntax:** 
`SRANDMEMBER key [count]`

**Command description:**

When called with just the `key` argument, return a random element from the set value stored at `key`.

If the provided `count` argument is positive, return an array of **distinct elements**. The array's length is either `count` or the set's cardinality (`SCARD`), whichever is lower.

If called with a negative `count`, the behavior changes and the command is allowed to return the **same element multiple times**. In this case, the number of returned elements is the absolute value of the specified `count`.

**Response:**

One of the following:

_Bulk string reply_: without the additional `count` argument, the command returns a randomly selected member, or a _Nil reply_ when `key` doesn't exist.
_Array reply_: when the optional `count` argument is passed, the command returns an array of members, or an empty array when `key` doesn't exist.

---

**How to approach this task:**
1. Add the new command info to the setCommandsInfoMap in the RespCommandsInfo class _(Garnet.server/Resp/RespCommandsInfo.cs)_
2. Add the new command string to the returned HashSet in the RespInfo.GetCommands method _(Garnet.server/Resp/RespInfo.cs)_
3. Add the appropriate fast parsing logic for the new command in the RespCommand.FastParseArrayCommand method (use other commands as reference) _(Garnet.server/Resp/RespCommand.cs)_
4. Implement the wrapper method to the storage layer in StorageSession _(Garnet.server/Storage/Session/ObjectStore/SetOps.cs)_
5. Define a new method in IGarnetAPI and implement it in GarnetApiObjectCommands, calling the method you have implemented in step #4 _(Garnet.server/API/IGarnetAPI.cs, Garnet.server/API/GarnetApiObjectCommands.cs)_
6. Add a new method to the RespServerSession class which will call the storage API and return the appropriate response _(Garnet.server/Resp/Objects/SetCommands.cs)_
8. Add a new method to SetObjImpl which will perform the required operation on the hash object _(Garnet.server/Objects/Set/SetObjectImpl.cs)_
9. Add a new enum value to SetOperation and add the appropriate case to the switch in SetObject.Operate, in which you will call the method defined in step #8 _(Garnet.server/Objects/Set/SetObject.cs)_
10. Add an appropriate case to the switch in TransactionManager.SetObjectKeys (Garnet.server/Transaction/TxnKeyManager.cs)
[...]

PR #240, opened 2024-04-05
feat: add SRANDMEMBER command

Zzhiter:
https://github.com/microsoft/garnet/issues/231

### DIFF
--- a/libs/server/Objects/Set/SetObjectImpl.cs
+++ b/libs/server/Objects/Set/SetObjectImpl.cs
@@ -267,7 +267,7 @@ namespace Garnet.server
                     countDone += count - countParameter;
                     
                 }
-                else if (count == int.MinValue) // // no count parameter is present
+                else if (count == int.MinValue) // no count parameter is present
                 {
                     // Return a single random element from the set
                     if (set.Count > 0)
@@ -289,7 +289,7 @@ namespace Garnet.server
                 {
                     // Return an array with potentially duplicate elements
                     int countParameter = Math.Abs(count);
-                    
+
                     // Write the size of the array reply
                     while (!RespWriteUtils.WriteArrayLength(countParameter, ref curr, end))
                         ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
--- a/libs/server/Objects/Set/SetObjectImpl.cs
+++ b/libs/server/Objects/Set/SetObjectImpl.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
 using System.Buffers;
+using System.Collections.Generic;
 using System.Linq;
 using System.Security.Cryptography;
 using Garnet.common;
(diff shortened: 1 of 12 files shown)


libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,107 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+                    var countParameter = count > set.Count ? set.Count : count;
+                    List<byte[]> randomElements = new List<byte[]>();
+
+                    // Write the size of the array reply
+                    while (!RespWriteUtils.WriteArrayLength(countParameter, ref curr, end))
+                        ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
+
+                    while (randomElements.Count < countParameter)
+                    {
+                        int index = RandomNumberGenerator.GetInt32(0, set.Count);
```
TalZaccai, line 294: This is a bit inefficient, once you get a random index, you need to get it from the set (O(M)) and compare it to all the other elements (O(n*m)), and you do that n times. Collectively you get O(M*m*n^2), where M is the set size, m is the max length in the set and n is the count, so worse case would be O(m*M^3) if n == m.

Consider the following solution - generate a set of n non-repeating random indexes (O(n)), sort it (O(nlogn)) and then iterate over the target set object, check if the index is in the index set and if it is, set the next item in a randomElements array (O(M)). Overall you get O(n) + O(nlogn) + O(M), which is worst case O(MlogM).

Sorry for the long-winded comment, hope this makes sense to you :)

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,107 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+                    int countParameter = Math.Abs(count);
+
+                    // Write the size of the array reply
+                    while (!RespWriteUtils.WriteArrayLength(countParameter, ref curr, end))
+                        ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
+                    for (int i = 0; i < countParameter; i++)
+                    {
+                        if (set.Count > 0)
+                        {
+                            int index = RandomNumberGenerator.GetInt32(0, set.Count);
```
TalZaccai, line 336: Again, a bit inefficient, since you're iterating over the set n times. If you generate the indexes first, you can do it with one pass over the set :)

libs/server/Resp/RespCommandsInfo.cs
```diff
@@ -216,6 +216,7 @@ public static RespCommandsInfo findCommand(RespCommand cmd, byte subCmd = 0)
             {(byte)SetOperation.SMEMBERS,   new RespCommandsInfo("SMEMBERS", RespCommand.Set,    1, null, (byte)SetOperation.SMEMBERS)},
             {(byte)SetOperation.SREM,       new RespCommandsInfo("SREM",     RespCommand.Set,   -2, null, (byte)SetOperation.SREM)},
             {(byte)SetOperation.SCARD,      new RespCommandsInfo("SCARD",    RespCommand.Set,    1, null, (byte)SetOperation.SCARD)},
+            {(byte)SetOperation.SRANDMEMBER,new RespCommandsInfo("SRANDMEMBER", RespCommand.Set, -1, null, (byte)SetOperation.SRANDMEMBER)},
```
TalZaccai, line 219: Arity should be -2

TalZaccai (changes requested):
Great work! Added some comments mainly concerning runtime optimizations.

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,115 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+
+            try
+            {
+                int[] indexes = default;
+
+                if (count > 0)
+                {
+                    // Return an array of distinct elements
+                    var countParameter = count > set.Count ? set.Count : count;
+                    List<byte[]> randomElements = new List<byte[]>();
```
TalZaccai, line 288: This list is not used...

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,115 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+
+                    // The order of fields in the reply is not truly random
+                    indexes = new HashSet<int>(Enumerable.Range(0, set.Count).OrderBy(x => Guid.NewGuid()).Take(countParameter)).ToArray();
```
TalZaccai, line 291: No need for the HashSet, right? values are already unique

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,115 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+                            ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
+                    }
+                    countDone++;
+                }
+                else // count < 0
+                {
+                    // Return an array with potentially duplicate elements
+                    int countParameter = Math.Abs(count);
+
+                    Random random = new Random();
```
TalZaccai, line 329: random variable is not used

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,115 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+                    List<int> randomIndexes = new List<int>();
```
TalZaccai, line 330: Instead of adding to a list and then converting to an array, you could add the values directly to an array, since you already know the array size in advance

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,115 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+                    for (int i = 0; i < countParameter; i++)
+                    {
+                        randomIndexes.Add(RandomNumberGenerator.GetInt32(0, set.Count));
+                    }
+                    indexes = randomIndexes.ToArray();
+
+                    // Write the size of the array reply
+                    while (!RespWriteUtils.WriteArrayLength(countParameter, ref curr, end))
```
TalZaccai, line 338: This probably needs to move inside the if (set.Count > 0), right?

TalZaccai (changes requested):
Few more small comments :)

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -256,5 +258,112 @@ private void SetPop(byte* input, int length, ref SpanByteAndMemory output)
...
+            {
+                int[] indexes = default;
+
+                if (count > 0)
+                {
+                    // Return an array of distinct elements
+                    var countParameter = count > set.Count ? set.Count : count;
+
+                    // The order of fields in the reply is not truly random
+                    indexes = new List<int>(Enumerable.Range(0, set.Count).OrderBy(x => Guid.NewGuid()).Take(countParameter)).ToArray();
```
TalZaccai, line 290: nit: You don't need a list here either:

Enumerable.Range(0, set.Count).OrderBy(x => Guid.NewGuid()).Take(countParameter).ToArray()

That way you're only enumerating the collection once and only allocating memory once instead of twice

TalZaccai (approved):

PR #242, opened 2024-04-05
Remove redundant Clear() calls

### LINKED ISSUE #241: Redundant "Clear" in many buffer allocations
In a number of places in the storage layer, `GC.AllocateArray` is used to allocate an array on the pinned heap, followed shortly by an `Array.Clear`. This is redundant, since `GC.AllocateArray` pre-initializes the data to zero. I'm guessing this might be a hangover from a previous unmanaged allocator, or from migration from a C/C++ implementation.

If the intent is to use unitialized data in the oversized portion: `GC.AllocateUnitializedArray` exists (but only works for blittable types); otherwise and probably simpler to grok (and easier to validate correctness): just remove the `Array.Clear` calls, leaving the default/inbuilt zero initialize (i.e. do not switch to `GC.AllocateUninitializedArray`).

impact: performance

examples:

- `BlittableFrame.Allocate`
- `MallocFixedPageSize.InternalAllocate` (2 of)
- `SpanByteAllocator.AllocatePage`

similar with simple array creation (no GC alloc):

- `GenericAllocator<,>.AllocatePage`
- `GenericFrame<,>.Allocate`

mgravell:
fix #241

1. removes redundant `Clear()` calls from new arrays that are already zero-initialized by the runtime
2. removes a branch in paging math for allocations

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/Allocator/BlittableAllocator.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Allocator/BlittableAllocator.cs
@@ -151,7 +151,6 @@ namespace Tsavorite.core
 
             byte[] tmp = GC.AllocateArray<byte>(adjustedSize, true);
             long p = (long)Unsafe.AsPointer(ref tmp[0]);
-            Array.Clear(tmp, 0, adjustedSize);
             pointers[index] = (p + (sectorSize - 1)) & ~((long)sectorSize - 1);
             values[index] = tmp;
         }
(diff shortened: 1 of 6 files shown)


badrishc:
This makes sense, we probably should switch to the uninitialized version in a future PR after we can prove that there aren't cases where we use without writing first. I think this is true already but it requires a more careful code examination. cc @TedHartMS

badrishc (approved):

TedHartMS (approved):

PR #243, opened 2024-04-05
Enable [SkipLocalsInit]

mgravell:
Same logic etc as https://github.com/microsoft/FASTER/pull/906

Punting on "server" for now - much more `stackalloc` usage, requires manual inspection to prove not assuming zero-init

In passing, also fixes some allocs in `CreateHexId`/`DefaultHexId` discovered while scanning for `stackalloc` usage:

- in `CreateHexId` the original string was upper-case, so we end up allocating 2 `string` each time, not just one
- in `DefaultHexId` allocated a new `string` every time; most uses (all, at current) always pass the default; optimize for that

impact: performance (paper-cuts, no single big win)


badrishc:
This is a solid idea. You are right that server is a big user of stackalloc. Out of curiosity, did you try adding it and notice tests failing out of the box?

libs/common/Generator.cs
```diff
@@ -16,11 +16,22 @@ public static class Generator
         /// </summary>
         /// <param name="size">The length of the hex identifier string</param>
         /// <returns></returns>
-        public static string CreateHexId(int size = 40)
+        public static unsafe string CreateHexId(int size = 40)
         {
             Span<byte> nodeIdBuffer = stackalloc byte[size / 2];
             RandomNumberGenerator.Fill(nodeIdBuffer);
```
Tornhoof, line 22: Maybe add at least a debug assert or another arg validation if size is above 256 or so?

libs/common/Generator.cs
```diff
@@ -16,11 +16,22 @@ public static class Generator
...
-            return Convert.ToHexString(nodeIdBuffer).ToLowerInvariant();
+
+            char* charBuffer = stackalloc char[nodeIdBuffer.Length * 2]; // not the same as size (if size is odd)
```
Tornhoof, line 24: use string.Create if possible? Then you have one stackalloc place least to care about.

libs/common/Generator.cs
```diff
@@ -29,10 +40,9 @@ public static string CreateHexId(int size = 40)
...
         /// <returns></returns>
         public static string DefaultHexId(int size = 40)
-        {
-            Span<byte> nodeIdBuffer = stackalloc byte[size / 2];
-            nodeIdBuffer.Clear();
-            return Convert.ToHexString(nodeIdBuffer).ToLowerInvariant();
-        }
+            => size == 40 ? DefaultHexId40 : new string('0', size & ~1); // note: trimmed to multiples of 2
+
+        const string DefaultHexId40 = "0000000000000000000000000000000000000000"; // '0' x40
```
Tornhoof, line 45: Change x40 to something like 40 times?
I literally read the comment as 0x40 and counted the zeros and asked myself it's not 64 chars.

mjgaux:
> This is a solid idea. You are right that server is a big user of stackalloc. Out of curiosity, did you try adding it and notice tests failing out of the box?

No; I simply didn't have the time to look at all the uses and evaluate what the code does.

libs/common/Generator.cs
mjgaux: Honestly, I wonder if it might be a better idea to just nuke the parameter :)

TedHartMS (approved):
Looks good. It omits libs\server and Tsavorite.test which are heavy users of stackalloc and would require close review. It does include Cluster which has one use of stackalloc in MIgrateSessionKeys but that writes the full allocation with data. I can't think of a place we do something like *&i

libs/common/Generator.cs
badrishc: yes, this would be great to add in this PR, thanks.

libs/common/Generator.cs
badrishc: it seems safe to nuke the parameter and always return a string of size 40. @vazois?

vazois: Yes, it should be fine. We use the method only to generate node-ids which are of fixed size 40 characters.

mgravell: This is on the public API, so it would be bad form to remove it; I've instead switched to "stackalloc for small, lease for large" strategy that should work for everyone

mgravell: @Tornhoof re `string.Create`: that's not ideal here, because you can't use a `Span<T>` as a generic type argument (`TState`); sure, we could switch to `string.Create` passing in a `(byte[], int)` tuple, but that seems to make things even messier; I've left it as 2 length-constrained `stackalloc` spans (with array-pool fallback) using `new string(ReadOnlySpan<char>)`

libs/common/Generator.cs
mgravell: is on public API, so can't make too many assumptions; now has suitable strategies to handle any reasonable input

libs/common/Generator.cs
PaulusParssinen: Well, there's _a way_ to do that but it's not worth it here 😆 

https://github.com/dotnet/runtime/blob/07648642d345087d344bc8b7b9b3ea7c8793bbea/src/libraries/Common/src/System/HexConverter.cs#L196-L197

badrishc:
As discussed offline, closing PR for now as we are not seeing end-to-end perf advantages for cache workloads. We can revisit this setting in the future. Thanks!

PR #244, opened 2024-04-05
Refactor the SpanByte usage

PaulusParssinen:
This PR brings in the good changes from https://github.com/microsoft/FASTER/pull/903 and renames some terminology around "fixed" buffers to "pinned". The latter term is more commonly used in .NET APIs and avoids the problem with possibly getting mixed with [fixed sized](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/inline-arrays) buffers.

Also marked some of the `SpanByte(AndMemory)` members readonly and removed some unnecessary methods. 

~~Also where is `Reinterpret` overloads supposed to be used?~~ Nvm, just saw the command parsing logic using it 👍


libs/storage/Tsavorite/cs/src/core/VarLen/SpanByteExtensions.cs
```diff
@@ -24,7 +24,7 @@ public static unsafe class SpanByteExtensions
         {
             fixed (byte* k = key)
             fixed (byte* v = desiredValue)
-                return clientSession.Upsert(SpanByte.FromFixedSpan(key), SpanByte.FromFixedSpan(desiredValue), userContext, serialNo);
+                return clientSession.Upsert(SpanByte.FromPinnedSpan(key), SpanByte.FromPinnedSpan(desiredValue), userContext, serialNo);
```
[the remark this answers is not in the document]
PaulusParssinen: This seems to be actually sorta ok. The builtin `LibraryImport` generator relies on same behavior so I guess it's stable. We can still just use directly use `SpanByte` ctor.

libs/cluster/Server/Migration/MigrateSessionKeys.cs
```diff
@@ -75,7 +75,7 @@ private bool MigrateKeysFromMainStore(ref List<(long, long)> keysWithSize, out L
                     if (!o.IsSpanByte)
                     {
                         memoryHandle = o.Memory.Memory.Pin();
-                        value = SpanByte.FromPinnedMemory(o.Memory.Memory);
+                        value = SpanByte.FromPinnedSpan(o.Memory.Memory.Span);
```
Tornhoof, line 78: I think this makes the Code more complex to understand, before it was clear what SpanByte  method to use. ``FromPinnedMemory`` -> you need to pin the memory with ``Pin()``. Now this relation is missing. There is no ``Pin()`` for span

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByteExtensions.cs
```diff
@@ -13,85 +13,49 @@ public static unsafe class SpanByteExtensions
...
-        /// <returns></returns>
         public static Status Upsert<Input, Output, Context, Functions>(this ClientSession<SpanByte, SpanByte, Input, Output, Context, Functions> clientSession, ReadOnlySpan<byte> key, ReadOnlySpan<byte> desiredValue, Context userContext = default, long serialNo = 0)
             where Functions : IFunctions<SpanByte, SpanByte, Input, Output, Context>
         {
-            fixed (byte* k = key)
-            fixed (byte* v = desiredValue)
-                return clientSession.Upsert(SpanByte.FromFixedSpan(key), SpanByte.FromFixedSpan(desiredValue), userContext, serialNo);
+            fixed (byte* keyPtr = key)
+            fixed (byte* desiredValuePtr = desiredValue)
+                return clientSession.Upsert(SpanByte.FromPointer(keyPtr, key.Length), SpanByte.FromPointer(desiredValuePtr, desiredValue.Length), userContext, serialNo);
```
Tornhoof, line 21: Similar as my previous comment, this loses the ``Fixed`` hint and it is very important the ptr is fixed.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs
```diff
@@ -264,78 +257,42 @@ public SpanByte Deserialize()
...
+        /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static SpanByte FromFixedSpan(ReadOnlySpan<byte> span)
+        public static SpanByte FromPinnedSpan(ReadOnlySpan<byte> span)
         {
-            fixed (byte* ptr = span)
-            {
-                return new SpanByte(span.Length, (IntPtr)ptr);
-            }
+            return new SpanByte(span.Length, (nint)Unsafe.AsPointer(ref MemoryMarshal.GetReference(span)));
```
Tornhoof, line 289: I'm not sure it's wise to add another use of ``Unsafe.AsPointer``, which is just another place we'd need to think about GC holes

PaulusParssinen: In ideal world, I'd love to rip out of `SpanByte`, `SpanByteAndMemory` and `ArgSlice` out of existence and rewrite Tsavorite & Faster.. However, these changes are just imported from https://github.com/microsoft/FASTER/pull/903

libs/cluster/Server/Migration/MigrateSessionKeys.cs
PaulusParssinen: Sure. Reverting.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByteExtensions.cs
PaulusParssinen: I'll add "Pinned" word to this overload to make it clearer, good catch.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs
PaulusParssinen: Yeah, I was on the fence whether to split it. I'll wait for maintainer opinion if that's okay for you? This should still be reviewable commit by commit, I think.

badrishc:
Folks, everything you see in the code was there for a reason. Either evolution over many years, workarounds that were necessary at some point but probably rendered unnecessary as .NET evolved, or careful micro benchmarking that necessitated the non traditional styles. It was not easy to get 100 million plus ops/sec with low latency variance that beats the simple C code of systems like Redis even on a single thread. Can it be revisited? Sure, and we welcome it, including all the contributions to leverage the latest language speedups. Should we have delayed open sourcing by 3-6 months just for syntactic happiness from purists? I think not.

PaulusParssinen:
> Folks, everything you see in the code was there for a reason. Either evolution over many years, workarounds that were necessary at some point but probably rendered unnecessary as .NET evolved, or careful micro benchmarking that necessitated the non traditional styles. It was not easy to get 100 million plus ops/sec with low latency variance that beats the simple C code of systems like Redis even on a single thread. Can it be revisited? Sure, and we welcome it, including all the contributions to leverage the latest language speedups. Should we have delayed open sourcing by 3-6 months just for syntactic happiness from purists? I think not.

I apologize for my loaded wording in the comments. It's not directed at you or anyone in the team in particular but more towards the general problem of extermely complex nature of managing low-level .NET code. There's certain APIs in .NET that can cause pitfalls for even most experienced .NET developers and `Unsafe.AsPointer` is one of them. There are instances of .NET team members misusing it and it signals that the API is truly to be avoided. It also seems there has not been enough communication of all the dangers of that API from documentation perspective.. 

Once again, sorry for the inappropiate comment. It's easy for me criticize from hindsight. I have huge respect for the thing you guys have done here as the code has lived over the years and .NET has evolved at the same time. I want to help make this thing safer while maintaining the performance with modern .NET.

badrishc:
Thanks for your perspective! I'm excited for the next few months as we explore scaling back some of the unsafe code without compromising performance. Your stellar PRs are definitely much appreciated, and folks on our team are actively learning and looking at ways to improve and microbenchmark as well.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs
badrishc: This looks good. We used fixed here because `MemoryMarshal.GetReference(span)` wasn't available at the time.

badrishc (approved):

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs
```diff
@@ -37,7 +39,7 @@ public unsafe struct SpanByte
         [FieldOffset(4)]
         private IntPtr payload;
 
-        internal IntPtr Pointer => payload;
+        internal readonly IntPtr Pointer => payload;
```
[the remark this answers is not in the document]
PaulusParssinen: No changes here or in the only place where this is used; `StaticGetHashCode64`.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByteExtensions.cs
```diff
@@ -13,85 +13,49 @@ public static unsafe class SpanByteExtensions
...
-        /// <param name="userContext"></param>
-        /// <param name="serialNo"></param>
-        /// <returns></returns>
         public static Status Upsert<Input, Output, Context, Functions>(this ClientSession<SpanByte, SpanByte, Input, Output, Context, Functions> clientSession, ReadOnlySpan<byte> key, ReadOnlySpan<byte> desiredValue, Context userContext = default, long serialNo = 0)
             where Functions : IFunctions<SpanByte, SpanByte, Input, Output, Context>
         {
-            fixed (byte* k = key)
-            fixed (byte* v = desiredValue)
-                return clientSession.Upsert(SpanByte.FromFixedSpan(key), SpanByte.FromFixedSpan(desiredValue), userContext, serialNo);
+            fixed (byte* keyPtr = key)
```
TedHartMS, line 19: I don't see these Extensions used anywhere outside Tsavorite.Test, and only 10 calls in there total for Upsert and Read. This could be moved there, but probably better to change the callsites to make the proper pattern more visible

PaulusParssinen: Move the extension over to the tests or just delete it entirely try not hide the pinning?

TedHartMS: I think the latter is best as it makes the pinning visible and doesn't introduce a test-only layer

TedHartMS (approved):

PR #249, opened 2024-04-06
feat: Add SMOVE Command

graknow:
Added the SMOVE command.


graknow:
> @microsoft-github-policy-service agree

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -361,6 +363,50 @@ public unsafe GarnetStatus SetScan<TObjectContext>(ArgSlice key, long cursor, st
...
+                return GarnetStatus.OK;
+            }
+
+            SetLength(sourceKey, out var initCount, ref objectStoreContext);
+            SetRemove(sourceKey, member, out _, ref objectStoreContext);
+            SetLength(sourceKey, out var endCount, ref objectStoreContext);
+
+            SetLength(destinationKey, out var destinationCount, ref objectStoreContext);
+
+            if (initCount == endCount || destinationCount == 0)
```
TalZaccai, line 398: If destination doesn't exist we should add a new set object and insert the element to it

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -361,6 +363,50 @@ public unsafe GarnetStatus SetScan<TObjectContext>(ArgSlice key, long cursor, st
...
+            // If the keys are the same, no operation is performed.
+            var sameKey = sourceKey.ReadOnlySpan.SequenceEqual(destinationKey.ReadOnlySpan);
+            if (sameKey)
+            {
+                return GarnetStatus.OK;
+            }
+
+            SetLength(sourceKey, out var initCount, ref objectStoreContext);
+            SetRemove(sourceKey, member, out _, ref objectStoreContext);
+            SetLength(sourceKey, out var endCount, ref objectStoreContext);
```
TalZaccai, line 394: I think you can look at the 2nd out parameter in SetRemove to get the # of operations performed, that may save the need for calling SetLength

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -361,6 +363,50 @@ public unsafe GarnetStatus SetScan<TObjectContext>(ArgSlice key, long cursor, st
...
+            SetLength(destinationKey, out var destinationCount, ref objectStoreContext);
+
+            if (initCount == endCount || destinationCount == 0)
+            {
+                return GarnetStatus.OK;
+            }
+
+            SetAdd(destinationKey, member, out _, ref objectStoreContext);
+
+            smoveResult = 1;
```
TalZaccai, line 405: I'd probably set the result to the out parameter given by SetAdd (saddCount), just for good measure

test/Garnet.test/RespSetTest.cs
```diff
@@ -496,6 +497,88 @@ public void CanDoSPOPWithMoreCountThanSetSizeCommandLC()
...
+            lightClientRequest.SendCommand("SADD \"myDestinationSet\" \"twoS\"");
+            lightClientRequest.SendCommand("SADD \"myDestinationSet\" \"threeS\"");
+            lightClientRequest.SendCommand("SADD \"myDestinationSet\" \"fourS\"");
+            lightClientRequest.SendCommand("SADD \"myDestinationSet\" \"common\"");
+
+            var expectedSuccessfulResponse = ":1\r\n";
+            var expectedFailureResponse = ":0\r\n";
+
+            // Successful move
+            var response = lightClientRequest.SendCommand("SMOVE \"mySourceSet\" \"myDestinationSet\" \"oneS\"");
```
TalZaccai, line 523: Probably good to also check that the correct value moved (i.e. verify contents of both sets)

test/Garnet.test/RespSetTest.cs
```diff
@@ -496,6 +497,88 @@ public void CanDoSPOPWithMoreCountThanSetSizeCommandLC()
...
+            await db.ExecuteForStringResultAsync("SADD", new string[] { "destinationSet", "destinationValue", "commonValue" });
+
+            //Same key.
+            response = await db.ExecuteForLongResultAsync("SMOVE", new string[] { "sourceSet", "sourceSet", "sourceValue" });
+            Assert.AreEqual(response, 0);
+
+            //Move non-common member.
+            response = await db.ExecuteForLongResultAsync("SMOVE", new string[] { "sourceSet", "destinationSet", "sourceValue" });
+            Assert.AreEqual(response, 1);
+            Assert.AreEqual(await db.ExecuteForLongResultAsync("SCARD", new string[] { "sourceSet" }), 1);
```
TalZaccai, line 572: Again, would be good to check the contents as well as the cardinality

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -3,7 +3,9 @@
 
 using System;
 using System.Text;
+using System.Xml.Linq;
```
TalZaccai, line 6: Looks like these 2 added usings are not needed... Please verify

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -505,5 +506,68 @@ private unsafe bool SetPop<TGarnetApi>(int count, byte* ptr, ref TGarnetApi stor
...
+                if (NetworkKeyArraySlotVerify(ref keys, false))
+                {
+                    var bufSpan = new ReadOnlySpan<byte>(recvBufferPtr, bytesRead);
+                    if (!DrainCommands(bufSpan, count)) return false;
+                    return true;
+                }
+
+                var status = storageApi.SetMove(sourceKey, destinationKey, sourceMember, out var output);
+
+                if (status == GarnetStatus.NOTFOUND)
```
TalZaccai, line 554: So as it is, we're never returning GarnetStatus.NOTFOUND. What you can do is in case the sourcekey doesn't exist, return  NOTFOUND and write 0 to the output as that's what's expected (CmdStrings.RESP_RETURN_VAL_0).

TalZaccai (changes requested):
Great work! Few small comments...

TalZaccai (changes requested):
Couple of small comments :)

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Text;
+using System.Xml.Linq;
```
badrishc, line 6: where is this used? we try to avoid Linq if possible.

libs/server/Objects/Set/SetObject.cs
```diff
@@ -192,5 +193,7 @@ public override unsafe void Scan(long start, out List<byte[]> items, out long cu
             if (cursor == set.Count)
                 cursor = 0;
         }
+
+        public HashSet<byte[]> Set => set;
```
badrishc, line 197: what is this property needed for?

libs/server/Resp/Objects/SetCommands.cs
graknow: It isn't being used anywhere, it was added when I was messing with the set object (Which isn't being used anywhere either).

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -363,6 +363,49 @@ public unsafe GarnetStatus SetScan<TObjectContext>(ArgSlice key, long cursor, st
...
+                return GarnetStatus.OK;
+
+            // If the keys are the same, no operation is performed.
+            var sameKey = sourceKey.ReadOnlySpan.SequenceEqual(destinationKey.ReadOnlySpan);
+            if (sameKey)
+            {
+                return GarnetStatus.OK;
+            }
+
+            var sremStatus = SetRemove(sourceKey, member, out var sremOps, ref objectStoreContext);
```
TalZaccai, line 392: Sorry for not catching this sooner - since you are interacting with more than one object, you need to acquire the appropriate locks. See ListMove in ListOps.cs for reference.

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -363,6 +363,49 @@ public unsafe GarnetStatus SetScan<TObjectContext>(ArgSlice key, long cursor, st
...
+        /// <param name="destinationKey"></param>
+        /// <param name="member"></param>
+        /// <param name="smoveResult"></param>
+        /// <param name="objectStoreContext"></param>
+        internal unsafe GarnetStatus SetMove<TObjectContext>(ArgSlice sourceKey, ArgSlice destinationKey, ArgSlice member, out int smoveResult, ref TObjectContext objectStoreContext)
+            where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, SpanByte, GarnetObjectStoreOutput, long>
+        {
+            smoveResult = 0;
+
+            if (sourceKey.Length == 0 || destinationKey.Length == 0 || member.Length == 0)
```
TalZaccai, line 382: No need for member.Length == 0, you can have a string of length 0 as a set member...

TalZaccai (changes requested):
Couple more comments :)

libs/server/Storage/Session/ObjectStore/SetOps.cs
TalZaccai: Note that you'll also have to add a case for SMOVE in the SetObjectKeys method in TxnKeyManager.cs

TalZaccai (approved):

PR #252, opened 2024-04-06
Unify RESP error serialization

PaulusParssinen:
In #223 @vazois [wished](https://github.com/microsoft/garnet/pull/223#discussion_r1552596681) to remove the need from callers to format error messages.

~~In RESP spec, it is said that the serialized errors of the form ```-ERR [errorMessage]\r\n``` are "_generic errors_". I used this distinction to add `WriteGenericError` to `RespWriteUtils`.~~

The original plan was to have `WriteGenericError`, which would only take in the error message for the generic errors which are most common. However that specialization would complicate and confuse I abandoned it and instead went for the `WriteError`

Before this PR all error response serialization happened either by copying the static ASCII encoded error constants or by manually formatting them in the middle of parsing code. The parsing logic often returned a `ROS<byte>` that could either be a simple error or a `CmdString.RESP_OK` and used this as the indicator whether an operation was successful or not. Often this returned `ROS<byte>` was directly copied to the send buffer. This approach was very error prone, as the caller had to correctly format the error string to start with ``-`` and end with a terminator ``\r\n``. This was hard to remember and while doing this PR I came across more where this was forgotten or lost during refactors.

This PR flows all simple error serialization string to `RespWriteUtils.WriteError`. This requires making the error serialization a separate operation from serializing "+OK\r\n" and such required some adjustments in many method overloads. A common pattern introduced in this PR and that already existed in the project is familiar and intuitive ``bool TryX(.., out Y value)`` pattern from BCL. In majority of the places where we previously returned a `ROS<byte>` of either error or OK, we now return boolean indicating whether the there was error or not. If there was error (it returned false), we can grab the error string from ``out`` parameter.


libs/cluster/Session/ClusterCommands.cs
```diff
@@ -58,10 +58,14 @@ private List<byte[]> GetKeysInSlot(int slot, int keyCount)
             return keys;
         }
 
-        private bool ParseSlots(int count, ref byte* ptr, out HashSet<int> slots, out ReadOnlySpan<byte> resp, bool range)
+        /// <summary>
+        /// Try to parse slots
+        /// </summary>
+        /// <param name="errorMessage">The ASCII encoded error message if the method return <c>false</c>; otherwise <c>default</c></param>
+        private bool TryParseSlots(int count, ref byte* ptr, out HashSet<int> slots, out ReadOnlySpan<byte> errorMessage, bool range)
```
PaulusParssinen, line 65: The state this method holds is extremely complex.. Am I right thinking that it is quietly swallowing errors and is that intended behavior or not?

PaulusParssinen: After staring at this for a while, it seems the return value boolean that flows from cluster command processing will trigger the unknown subcommand error. However we should/can exit the loop as soon as we know that there is for example repeat slots.. Got tests failures from that which is nice 👍

PaulusParssinen: I think the way I can return original behavior here is by restoring the behavior where:
* If ``TryParseSlots`` returns `false`:
  * There was wrong number of arguments, no errors thrown here but a "wrong number of arguments" will be thrown at higher layer.
* If ``TryParseSlots`` returns `true`:
  * If the out parameter ``errorMessage == default`` evals to ``true``, we have error to write, however don't short-circuit to return false that would also incorrectly write "wrong num of parameters" fallback error. (I had this behavior here in the PR for while)
  * If the out parameter ``errorMessage == default`` evals to ``false``, everything is good, continue.

vazois: I just saw your messages. When in doubt you can check the Redis protocol for the corresponding command. Of course, some functionality maybe missing but we try our hardest to much the original protocol specification.
I think you nailed down the overall workflow, just summarizing the keys points of ParseSlots below.

- If slot provided is in range else return error.
- I slot provided is specified multiple times else return error. The error here is weird because we are using a set so if a slot is specified multiple times we can still continue without any issues but it a restriction of the Redis protocol.
- We are trying to do everything in place, meaning that we keep reading the slots and gathering possible errors in order to drain the receive buffer. It is a left over from before any attempts to refactor the parser which are still ongoing. Ideally, we should fail immediately, and make the caller is responsible for draining.

I think the last point is what making the ParseSlots method so complicated. If it gets in the way of having a common WriteError method, should be fine to leave it as is until we do a more comprehensive refactoring of the parser.

vazois: An edge condition that we are not tested is what happens when the input contains a non-numeric value for a slot. But I guess, this is an issue we need to revisit because it is unrelated to the current PR.

PaulusParssinen: > * I slot provided is specified multiple times continue processing else capture error. This condition is weird and possibly does not make sense to have since we consolidate the slots into a set. We could omit it but I am worried that because it is part of the specification, somebody will create an issue.

I believe the spec says:
> The command fails if the same slot is specified multiple times.

For both [DELSLOTS](https://redis.io/commands/cluster-delslots/) and [ADDSLOTS](https://redis.io/commands/cluster-addslots/)..  So presumable we can just bailout instantly with ``return false``, and let the command draining handle rest of args?

However the other paths which hit this are the ranges and I'm not clear where to read what we should do for overlapping ranges for example? The slot range variants don't specify failure for overlapping if I'm reading [this](https://redis.io/commands/cluster-addslotsrange/) right?

> * Last condition is that we are processing inputs in place. Therefore, we need to capture the errors and continue processing to drain the recv buffer. Ideally, we should fail immediately, and make the caller is responsible for draining.

I now read this and yea, it'd be ideal if we didn't have to care about draining the arguments here 🤔

PaulusParssinen: I'll restore the proper argument draining behavior here.

vazois: > I'll restore the argument draining behavior here.
Yes, do that. I was going to suggest, that we should probably not change the behavior of ParseSlots since this PR has a different purpose.

vazois: > > * I slot provided is specified multiple times continue processing else capture error. This condition is weird and possibly does not make sense to have since we consolidate the slots into a set. We could omit it but I am worried that because it is part of the specification, somebody will create an issue.
> 
> I believe the spec says:
> 
> > The command fails if the same slot is specified multiple times.
> 
> For both [DELSLOTS](https://redis.io/commands/cluster-delslots/) and [ADDSLOTS](https://redis.io/commands/cluster-addslots/).. So presumable we can just bailout instantly with `return false`, and let the command draining handle rest of args?
> 
> However the other paths which hit this are the ranges and I'm not clear where to read what we should do for overlapping ranges for example? The slot range variants don't specify failure for overlapping if I'm reading [this](https://redis.io/commands/cluster-addslotsrange/) right?
> 
> > * Last condition is that we are processing inputs in place. Therefore, we need to capture the errors and continue processing to drain the recv buffer. Ideally, we should fail immediately, and make the caller is responsible for draining.
> 

Just to note, this is not a redis requirement more of our way to do things fast in-place without extra allocations.

> I now read this and yea, it'd be ideal if we didn't have to care about draining the arguments here 🤔

We have DrainCommand that we can use, though to do it correctly we need to keep track how many of the args we read correctly before the error, so we can drain the rest.

Maybe a more appropriate solution will be a method that reads all parameters for ADDSLOTS/DELSLOTS in a array/list (ensuring that everything is parsed/drained) and then validate the list input.
It might be fine for perf reasons since CLUSTER commands are in the slow path.

PaulusParssinen:
Starting to think special casing the generic errors would be a mistake and everything should flow into `WriteError` instead..

libs/cluster/Session/ClusterCommands.cs
```diff
@@ -967,39 +1052,43 @@ public bool ProcessSlotManageCommands(ReadOnlySpan<byte> bufSpan, ReadOnlySpan<b
...
-                                    SendAndReset();
+                                errorMessage = Encoding.ASCII.GetBytes($"ERR Slot state {subcommand} not supported.");
                                 break;
                         }
 
-                        if (resp.SequenceEqual(CmdStrings.RESP_OK)) UnsafeWaitForConfigTransition();
-
-                        while (!RespWriteUtils.WriteDirect(resp, ref dcurr, dend))
-                            SendAndReset();
+                        if (errorMessage == default)
```
vazois, line 1077: nit; use returned boolean from above  to check for errors.

libs/cluster/Session/ClusterCommands.cs
```diff
@@ -1055,38 +1144,46 @@ public bool ProcessSlotManageCommands(ReadOnlySpan<byte> bufSpan, ReadOnlySpan<b
...
-                                    SendAndReset();
+                                errorMessage = Encoding.ASCII.GetBytes($"ERR Slot state {subcommand} not supported.");
                                 break;
                         }
                     }
 
-                    if (resp.SequenceEqual(CmdStrings.RESP_OK)) UnsafeWaitForConfigTransition();
-                    while (!RespWriteUtils.WriteDirect(resp, ref dcurr, dend))
-                        SendAndReset();
+                    if (errorMessage == default)
```
vazois, line 1175: nit; use returned boolean from above to check for errors.

libs/cluster/Server/ClusterManagerWorkerState.cs
```diff
@@ -131,53 +141,52 @@ public ReadOnlySpan<byte> TryReset(bool soft, int expirySeconds = 60)
...
             while (true)
             {
                 var current = CurrentConfig;
-                //(error) ERR Can't replicate myself
                 if (current.GetLocalNodeId().Equals(nodeid))
                 {
-                    logger?.LogError("{msg}", Encoding.ASCII.GetString(CmdStrings.RESP_CANNOT_REPLICATE_SELF_ERROR));
-                    resp = CmdStrings.RESP_CANNOT_REPLICATE_SELF_ERROR;
+                    errorMessage = CmdStrings.RESP_ERR_GENERIC_CANNOT_REPLICATE_SELF;
+                    logger?.LogError("-{msg}", Encoding.ASCII.GetString(errorMessage));
```
vazois, line 160: nit; we don't need - in front of logged message

vazois (changes requested):

libs/cluster/CmdStrings.cs
```diff
@@ -92,43 +92,44 @@ static class CmdStrings
...
-        public static ReadOnlySpan<byte> RESP_SLOTSTATE_TRANSITION_ERROR => "-ERR Slot already in that state\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_CANNOT_REPLICATE_SELF_ERROR => "-ERR Can't replicate myself\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_CANNOT_MAKE_REPLICA_WITH_ASSIGNED_SLOTS_ERROR => "-ERR Primary has been assigned slots and cannot be a replica\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_CANNOT_FORGET_MYSELF_ERROR => "-ERR I tried hard but I can't forget myself\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_CANNOT_FORGET_MY_PRIMARY_ERROR => "-ERR Can't forget my primary\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_CANNOT_FAILOVER_FROM_NON_MASTER => "-ERR Cannot failover a non-master node\r\n"u8;
-        public static ReadOnlySpan<byte> RESP_UNKNOWN_ENDPOINT_ERROR => "-ERR Unknown endpoint\r\n"u8;
+        /// Generic error respone strings, i.e. these are sent in the form "-ERR responseString\r\n"
+        /// </summary>
+        public static ReadOnlySpan<byte> RESP_ERR_GENERIC_CLUSTER => "ERR This instance has cluster support disabled"u8;
```
badrishc, line 107: why do we not write the string `ERR` in the calling code, similar to `-`.

vazois: The WriteError function takes care of this. It adds  prefix "-" and suffix "\r\n". I think we need to use WriteError more so as to avoid forgetting to resp format the string.

badrishc: My question was, why not let `WriteError` write the `ERR` as well.

libs/cluster/Server/ClusterManagerSlotState.cs
```diff
@@ -419,14 +419,12 @@ public bool TryPrepareSlotsForOwnershipChange(HashSet<int> slots, string nodeid,
...
 
         /// <summary>
-        /// Reset slot state to stable
+        /// Try to reset slot state to <see cref="SlotState.STABLE"/>
         /// </summary>
-        /// <param name="slot">Slot id to reset state</param>       
-        /// <param name="errorMessage"></param>
+        /// <param name="slot">Slot id to reset state</param>
         /// <returns>True on success, false otherwise</returns>
-        public bool TryResetSlotState(int slot, out ReadOnlySpan<byte> errorMessage)
```
PaulusParssinen: Should this method have error messages/failures 😅 Should we check `currentConfig.GetWorkerIdFromSlot` inside the loop?

libs/cluster/CmdStrings.cs
vazois: Yes, exactly. The protocol requires "-" prefix including "\r\n" suffix. Not all error messages start with ERR e.g. -WRONGPASS.

libs/cluster/Session/ClusterCommands.cs
PaulusParssinen: [Addressed](https://github.com/microsoft/garnet/pull/252#discussion_r1556396125)

libs/cluster/Server/ClusterManagerSlotState.cs
vazois: Check it for what? In general, callers to this method have to do validity checking.
This method tries to reset the slot state to stable if it was either in MIGRATING or IMPORTING state. The former means that the local node was the owner of the slot before it was set to MIGRATING, so we reset to 1.
In the latter situation, we still point to the old owner in the slotmap so we just need to get that info and set the node to STABLE.

PS: I have been trying to write cluster dev guide for a while now. Maybe this week, if PR reviews settle down 😄.
I am sure I am going to find couple of issues😅

badrishc:
Fantastic cleanup!!

libs/cluster/Server/ClusterManagerSlotState.cs
PaulusParssinen: Alright, I don't have good mental model for all the cluster state transitions so I kinda hoped this could fail somehow so we could keep the `Try*` pattern and the nice switch expression 😄 ResetSlotState it is then?

vazois (approved):

PaulusParssinen:
Gonna do one more commit where I move more of these error strings to `CmdStrings`

vazois:
> Gonna do one more commit where I move more of these constant error strings to `CmdStrings`

Sounds good, though try to avoid big PRs since they hard to review and could result in a lot of conflicts at merge time.

PaulusParssinen:
> > Gonna do one more commit where I move more of these constant error strings to `CmdStrings`
> 
> Sounds good, though try to avoid big PRs since they hard to review and could result in a lot of conflicts at merge time.

Yeah.. these tend to get a bit out of hand too easily😬I'll leave the constant as follow-up for anyone to do instead..

PR #255, opened 2024-04-07
Command argument parsing improvements

PaulusParssinen:
### Culture invariant parsing
* Avoid re-encoding ASCII and allocating temporary strings, instead feed bytes into directly to `Utf8Parser.TryParse`
  * In https://github.com/microsoft/garnet/pull/255/commits/02cca169e280f00348e20c9ac149429125ef9dd2, this change indirectly fixes the issue where cultures with decimal seperator being comma instead of dot caused some of Geo RESP tests fail.
  * **UPDATE**: Fixed all remaining floating point formatting too by passing `InvariantCulture` to `Parse` and `ToString`.
* Avoid some `CurrentCulture` lookups by using `ToUpperInvariant` instead of `ToUpper`
* Use `StringComparison.OrdinalIgnoreCase` where we otherwise allocated uppercase string for string comparison. This equality comparison is also much faster as it doesn't care about cultures.
  * And use `StringComparison.Ordinal` where applicable for perf (it gets unrolled for constant inputs à la `SequenceEquals`).`StringComparison.OrdinalIgnoreCase` also gets unrolled for constants in .NET 8 but has to deal with some bit flipping for the case insensitivity. 
### Enums
* Use `Enum.TryParse` instead of catching exceptions for `Enum.Parse`, much more cheaper.
* Use generic enum parsing overloads, allows JIT to see the type and optimize it better.
### Misc.
* Removed unused methods from `RespReadUtils`
* Added internal `ArgSlice.FromPinnedSpan` 
  * Necessary evil that will allow further simplifications in future contributions.
  * In this PR, it allow simplifying the HashObject's `IncrementIntegerOrFloat` logic.


libs/cluster/Session/ClusterCommands.cs
```diff
@@ -1025,20 +1021,18 @@ public bool ProcessSlotManageCommands(ReadOnlySpan<byte> bufSpan, ReadOnlySpan<b
...
 
-                    // Parse slot state
-                    try { slotState = (SlotState)Enum.Parse(typeof(SlotState), subcommand); }
-                    catch (Exception ex)
+                    // Try parse slot state
+                    if (!Enum.TryParse(subcommand, out SlotState slotState))
                     {
                         // Log error for invalid slot state option
-                        logger?.LogError(ex, "");
+                        logger?.LogError($"The given {nameof(SlotState)} is invalid.");
```
PaulusParssinen, line 1035: Let's give a little bit better error for the logs..
```suggestion
                        logger?.LogError("The given '{input}' is not a valid slot state option.", subcommand);
```

vazois (changes requested):

libs/cluster/Session/ClusterCommands.cs
```diff
@@ -1058,9 +1055,8 @@ public bool ProcessSlotManageCommands(ReadOnlySpan<byte> bufSpan, ReadOnlySpan<b
                     if (!RespReadUtils.ReadStringWithLengthHeader(out var subcommand, ref ptr, recvBufferPtr + bytesRead))
                         return false;
 
-                    var slotState = SlotState.STABLE;
-                    try { slotState = (SlotState)Enum.Parse(typeof(SlotState), subcommand); }
-                    catch { }
+                    if (!Enum.TryParse(subcommand, out SlotState slotState))
```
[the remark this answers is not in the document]
PaulusParssinen:
```suggestion
                    if (!Enum.TryParse(subcommand, ignoreCase: true, out SlotState slotState))
```

libs/cluster/Session/FailoverCommand.cs
```diff
@@ -24,15 +24,8 @@ private bool TryFAILOVER(int count, byte* ptr)
...
                     return false;
 
-                FailoverOption failoverOption;
-                try
-                {
-                    failoverOption = (FailoverOption)Enum.Parse(typeof(FailoverOption), option);
-                }
-                catch
-                {
+                if (!Enum.TryParse(option, out FailoverOption failoverOption))
```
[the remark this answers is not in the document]
PaulusParssinen:
```suggestion
                if (!Enum.TryParse(option, ignoreCase: true, out FailoverOption failoverOption))
```

badrishc:
@vazois @lmaas - any additional comments on this PR?

vazois (approved):

libs/server/Metrics/Info/InfoCommand.cs
```diff
@@ -41,7 +41,7 @@ private bool ProcessInfoCommand(int count)
                         default:
                             try
                             {
-                                var sectionType = (InfoMetricsType)Enum.Parse(typeof(InfoMetricsType), section);
+                                var sectionType = Enum.Parse<InfoMetricsType>(section);
```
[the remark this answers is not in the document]
PaulusParssinen: Absolutely a place for `TryParse`, thanks! No idea how I missed this..

libs/server/ACL/ACLParser.cs
```diff
@@ -179,11 +179,11 @@ public static void ApplyACLOpToUser(ref User user, string op)
                     user.AddCategory(category);
                 }
             }
-            else if ((op == "~*") || (op == "allkeys"))
+            else if (op.Equals("~*", StringComparison.OrdinalIgnoreCase) || op.Equals("allkeys", StringComparison.OrdinalIgnoreCase))
```
PaulusParssinen, line 182: No case insestivity needed.
```suggestion
            else if (op.Equals("~*", StringComparison.Ordinal) || op.Equals("allkeys", StringComparison.OrdinalIgnoreCase))
```

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -551,20 +552,20 @@ private bool StringBitField<TGarnetApi>(int count, byte* ptr, ref TGarnetApi sto
...
-                    byte bitCount = (byte)Int32.Parse(encodingArg.Substring(1));
-                    //At most 64 bits can fit into enconding info
+                    byte bitCount = (byte)int.Parse(encodingArg.AsSpan(1));
+                    //At most 64 bits can fit into encoding info
                     encodingInfo = (byte)(sign | bitCount);
 
                     //Calculate number offset from bitCount if offsetArg starts with #
-                    bool offsetType = offsetArg.StartsWith("#");
-                    offset = offsetType ? Int64.Parse(offsetArg.Substring(1)) : Int64.Parse(offsetArg);
+                    bool offsetType = offsetArg.StartsWith("#", StringComparison.OrdinalIgnoreCase);
```
PaulusParssinen, line 567: ditto
```suggestion
                    bool offsetType = offsetArg.StartsWith("#", StringComparison.Ordinal);
```

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -740,14 +741,14 @@ private bool StringBitFieldReadOnly<TGarnetApi>(int count, byte* ptr, ref TGarne
...
+                    byte sign = encoding.StartsWith("i", StringComparison.OrdinalIgnoreCase) ? (byte)BitFieldSign.SIGNED : (byte)BitFieldSign.UNSIGNED;
                     //Number of bits in signed number
-                    byte bitCount = (byte)Int32.Parse(encoding.Substring(1));
+                    byte bitCount = (byte)int.Parse(encoding.AsSpan(1));
                     encodingInfo = (byte)(sign | bitCount);
 
                     //Calculate number offset from bitCount if offsetArg starts with #
-                    bool offsetType = offsetArg.StartsWith("#");
-                    offset = offsetType ? Int64.Parse(offsetArg.Substring(1)) : Int64.Parse(offsetArg);
+                    bool offsetType = offsetArg.StartsWith("#", StringComparison.OrdinalIgnoreCase);
```
PaulusParssinen, line 750:
```suggestion
                    bool offsetType = offsetArg.StartsWith("#", StringComparison.Ordinal);
```

libs/server/ACL/ACLParser.cs
```diff
@@ -36,10 +38,8 @@ class ACLParser
         /// <exception cref="ACLUnknownOperationException">Thrown if the given operation does not exist.</exception>
         public static User ParseACLRule(string input, AccessControlList acl = null)
         {
-
             // Tokenize input string 
-            Regex regex = new Regex("\\s+");
-            string[] tokens = regex.Split(input.Trim());
+            string[] tokens = input.Trim().Split(WhitespaceChars);
```
lmaas, line 42: I don't think this is a valid substitution for that regex. `Split()` will add empty entries to token array, if it finds multiple whitespace characters in a row (e.g., `"   "`).

libs/server/ACL/ACLParser.cs
```diff
@@ -97,28 +97,28 @@ public static void ApplyACLOpToUser(ref User user, string op)
                 return;
             }
 
-            if (op == "on")
+            if (op.Equals("on", StringComparison.Ordinal))
```
lmaas, line 100: I think we should use `OrdinalIgnoreCase` for all ACL op keywords to improve robustness (`on`, `off`, `nopass`, `reset`, `resetpass`, `allkeys `and `resetkeys`).

libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -257,7 +260,7 @@ private void HashRandomField(byte* input, int length, ref SpanByteAndMemory outp
                     countDone = count;
 
                     // Prepare response
-                    if (!Int32.TryParse(Encoding.ASCII.GetString(countParameterByteArray), out var countParameter))
+                    if (!Utf8Parser.TryParse(countParameterByteArray, out int countParameter, out _, default))
```
lmaas, line 263: I like the use of `Utf8Parser` for this. However, we should check the `bytesConsumed` as well if we are consuming any unsanitized user input. `Utf8Parser.TryParse` will match substrings - so this could have unintended side-effects for invalid strings such as `123abs`.

This applies to all the `Utf8Parser`-related changes in this PR.

libs/server/Objects/ObjectUtils.cs
```diff
@@ -100,7 +100,7 @@ public static unsafe bool ReadScanInput(byte* input, int length, ref SpanByteAnd
                     if (!RespReadUtils.ReadByteArrayWithLengthHeader(out var countParameterValue, ref input_currptr, input + length))
                         return false;
 
-                    int.TryParse(Encoding.ASCII.GetString(countParameterValue), out countInInput);
+                    _ = Utf8Parser.TryParse(countParameterValue, out countInInput, out _, default);
```
lmaas, line 103: **NOTE:** We should make sure we handle parsing errors gracefully, as this can have unintended consequences. I don't think we can handle this in this PR, so this only serves as a note for the future.

libs/server/ACL/ACLParser.cs
```diff
@@ -97,28 +97,28 @@ public static void ApplyACLOpToUser(ref User user, string op)
...
             {
                 // Enable user
                 user.IsEnabled = true;
             }
-            else if (op == "off")
+            else if (op.Equals("off", StringComparison.Ordinal))
```
lmaas, line 105: While we are at it and we are turning these comparisons into case insensitive comparisons, we can also make the constant string operands upper-case to make the style consistent the rest of the code.

lmaas (changes requested):
Excellent work! There are a lot of good changes in this PR. Thank you for working on this.

I have added a few comments with suggestions for improvement. Let’s address those points before merging.  Once we’ve resolved the questions, I’ll be happy to perform another review.

libs/server/ACL/ACLParser.cs
PaulusParssinen: Let's trim and drop them then :+1: Good catch, thanks!
```suggestion
            string[] tokens = input.Trim().Split(WhitespaceChars, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```

libs/server/Objects/ObjectUtils.cs
PaulusParssinen: Yeah, it hurts to leave them like this but I really can't address all the places where we swallow errors in this PR.

libs/server/Objects/Hash/HashObjectImpl.cs
PaulusParssinen: Ah very good catch! Will restore original behavior by checking we consumed the entire buffer which is fed into U8.TryParse :+1:

lmaas (changes requested):

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -257,10 +257,10 @@ public List<(string, int)> GetLocalNodePrimaryEndpoints(bool includeMyPrimaryFir
             List<(string, int)> primaries = new();
             for (ushort i = 2; i < workers.Length; i++)
             {
-                if (workers[i].role == NodeRole.PRIMARY && workers[i].nodeid != myPrimaryId)
+                if (workers[i].role == NodeRole.PRIMARY && !workers[i].nodeid.Equals(myPrimaryId, StringComparison.OrdinalIgnoreCase))
                     primaries.Add((workers[i].address, workers[i].port));
 
-                if (workers[i].nodeid == myPrimaryId)
+                if (workers[i].nodeid.Equals(myPrimaryId, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
PaulusParssinen: Done.

lmaas (approved):
Great job on the code! Your thorough work is much appreciated. I believe we’re ready to merge this PR. Thank you for your valuable contribution! 🙌

Issue #256, opened 2024-04-07
Secret comparisons should be constant time


SteveSyfuhs:
Any time a symmetric secret whether cleartext, hashed, or encrypted, is compared to another value it needs to use a constant time comparison function. If this isn't done, the secret material is leaked whenever a negative comparison occurs.

```csharp
var a = "foo";
var b = "bar";
var c = "foobar";

a == a // foo == foo 3/3
a == b // foo == bar 0/3
a == c // foo == foobar 3/6
```

First comparison leaks nothing. Second comparison tells us it's three characters. Third tells us it's at least three characters and the first three are "foo".

badrishc:
Good find, thanks!

PR #257, opened 2024-04-07
Constant time secret check

SteveSyfuhs:
Switch all secret comparisons to be constant time to reduce the risk of creating an oracle through side channel timing.

Fixes https://github.com/microsoft/garnet/issues/256

Unit tests pass.


libs/server/ACL/ACLPassword.cs
```diff
@@ -92,20 +92,7 @@ public override string ToString()
...
-            {
-                if (PasswordHash[i] != password.PasswordHash[i])
-                {
-                    return false;
-                }
-            }
-
-            return true;
+            return SecretsUtility.ConstantEquals(password.PasswordHash.AsSpan(), PasswordHash.AsSpan());
         }
```
filipnavara, line 96: Is there a reason not to use `CryptographicOperations.FixedTimeEquals`?

SteveSyfuhs: I couldn't remember the name of the method when I searched for it. 😅 I'll update that.

badrishc:
Looks like lint is failing on this PR, please check - https://github.com/microsoft/garnet/actions/runs/8604728571/job/23579529200?pr=257

SteveSyfuhs:
> Looks like lint is failing CI runs (style check) on this PR, please check - https://github.com/microsoft/garnet/actions/runs/8604728571/job/23579529200?pr=257

Fixed, I think. Verifying.

libs/server/ACL/SecretsUtility.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static bool ConstantEquals(ReadOnlySpan<byte> a, ReadOnlySpan<byte> b) => CryptographicOperations.FixedTimeEquals(a, b);
+
+        public static bool ConstantEquals(string a, string b) => ConstantEquals(a.AsSpan(), b.AsSpan());
+
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public static bool ConstantEquals(ReadOnlySpan<char> a, ReadOnlySpan<char> b)
+        {
+            // all secrets are ASCII or UTF-8
+
+            return ConstantEquals(MemoryMarshal.Cast<char, byte>(a), MemoryMarshal.Cast<char, byte>(b));
```
filipnavara, line 22: I am assuming we don't care about Unicode normalization, right? (just checking)

filipnavara (approved):

badrishc (approved):

libs/server/ACL/SecretsUtility.cs
SteveSyfuhs: I scanned through all the different places encoding changes occur and it looks like they're all ASCII or UTF-8. At the moment this method isn't actually called anywhere and it's added as matter of completeness so I could

1. Do nothing
2. Remove it
3. Add a UTF-16 detection+conversion before cast
4. Re-add the original implementation that compares `char` directly
5. Throw if UTF-16

Not keen on (3) but everything else is fair game.

Thoughts?

filipnavara: I don’t have any specific concern, so it was merely a question for curiosity/completeness. I guess option (1) or (2) would be fine for now.

Lot of authentication specifications impose explicit requirements for string preparation (notably, the StringPrep framework and SASLprep RFCs specifically come to mind), so I figured I would ask if any of that was a concern here if a secret is represented as a string.

lmaas: Since all plain-text secrets are currently encoded in ASCII (because passwords need to be passed through the RESP-protocol), I would suggest removing the unneeded functions for now. If we ever need constant time comparisons for UTF-16 strings, we can add back the original implementation.

lmaas (changes requested):

libs/server/ACL/SecretsUtility.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+
+namespace Garnet.server.ACL
+{
+    internal static class SecretsUtility
+    {
+        public static bool ConstantEquals(ReadOnlySpan<byte> a, ReadOnlySpan<byte> b) => CryptographicOperations.FixedTimeEquals(a, b);
```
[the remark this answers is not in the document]
SteveSyfuhs: ACK, adding

libs/server/ACL/SecretsUtility.cs
SteveSyfuhs: > Since all plain-text secrets are currently encoded in ASCII (because passwords need to be passed through the RESP-protocol), I would suggest removing the unneeded functions for now. If we ever need constant time comparisons for UTF-16 strings, we can add back the original implementation.

Fair enough. Let me it and I'll add some comments to indicate potential options for string conversions.

badrishc:
Sorry closed by mistake, reopened now.

lmaas (approved):
Code looks good to me. Thank you for your valuable contribution! 🙌 Let's merge this in.

PR #267, opened 2024-04-11
A work flow to run secondary configurations every night - currently only MacOS

darrenge:
These secondary configurations could be not only MacOS (which is set up for now) but could also include other flavors of Linux and some perf tests. This is meant to run on a schedule (nightly) and doesn't matter if files changed or not. The tests that are ran are Garnet, Garnet Cluster and Tsvaorite.

NOTE: This work flow has been ran in a private fork and there are test failures in all three areas, so it is recommended that this will be "disabled" until we get tests fixed. However, I feel it is a good idea to get this merged and in place so when devs start fixing the test failures, they will have the workflow available to verify their fixes.


darrenge:
The code on the fork was a bit outdated, so I would like to get this on current main and see where it stands.

TalZaccai (approved):

PR #271, opened 2024-04-12
Website update

TalZaccai:
- Fixed SETNX status in API Compatibility page
- Added SRANDMEMBER command
- Added ACL commands


website/docs/commands/api-compatibility.md
```diff
@@ -297,10 +310,10 @@ Note that this list is subject to change as we continue to expand our API comman
 |  | [SET](raw-string.md#set) | ➕ |  |
 |  | [SET ... NX](raw-string.md#set) | ➕ |  |
 |  | [SETEX](raw-string.md#setex) | ➕ |  |
-|  | [SETNX](raw-string.md#setnx) | ➕ |  |
+|  | SETNX | ➖ | _Deprecated, use `SET` with `NX` argument instead_ |
```
yeurch, line 313: `SETEX` is similarly deprecated (and not implemented in Garnet). It should also be marked as such on the line above.  Instead, users should use `SET` with `EX` argument.

TalZaccai: Yeah that's a good point, I can fix the wording a little bit. `SETEX` is indeed implemented, by the way.

vazois (approved):

Issue #277, opened 2024-04-13
[BUG]: LMOVE operation direction parse error


Zzhiter:
In Redis, when LEFT or RIGHT is missed, it will return syntax error.
![image](https://github.com/microsoft/garnet/assets/61727602/a9e373bb-a55c-47b1-8c46-c23173d26358)
However, We didn't handle this situation
![image](https://github.com/microsoft/garnet/assets/61727602/bd372def-034b-473a-8b86-c198262165cb)

PR #278, opened 2024-04-13
fix: operation direction parse error

Zzhiter:
fix: https://github.com/microsoft/garnet/issues/277

### DIFF
--- a/libs/server/Resp/Objects/ListCommands.cs
+++ b/libs/server/Resp/Objects/ListCommands.cs
@@ -679,33 +679,8 @@ namespace Garnet.server
                 if (!RespReadUtils.ReadPtrWithLengthHeader(ref param2.ptr, ref param2.length, ref ptr, recvBufferPtr + bytesRead))
                     return false;
 
-                OperationDirection sourceDirection, destinationDirection;
-                if (Ascii.EqualsIgnoreCase("RIGHT", param1.ReadOnlySpan))
-                {
-                    sourceDirection = OperationDirection.Right;
-                }
-                else if (Ascii.EqualsIgnoreCase("LEFT", param1.ReadOnlySpan))
-                {
-                    sourceDirection = OperationDirection.Left;
-                }
-                else
-                {
-                    sourceDirection = OperationDirection.Unknown;
-                }
-
-                if (Ascii.EqualsIgnoreCase("RIGHT", param2.ReadOnlySpan))
-                {
-                    destinationDirection = OperationDirection.Right;
-                }
-                else if (Ascii.EqualsIgnoreCase("LEFT", param2.ReadOnlySpan))
-                {
-                    destinationDirection = OperationDirection.Left;
-                }
-                else
-                {
-                    destinationDirection = OperationDirection.Unknown;
-                }
-
+                OperationDirection sourceDirection = GetOperationDirection(param1.ReadOnlySpan);
+                OperationDirection destinationDirection = GetOperationDirection(param2.ReadOnlySpan);
                 if (sourceDirection == OperationDirection.Unknown || destinationDirection == OperationDirection.Unknown)
                 {
                     return AbortWithErrorMessage(count, CmdStrings.RESP_ERR_GENERIC_SYNTAX_ERROR);
--- a/libs/server/Resp/Objects/ListCommands.cs
+++ b/libs/server/Resp/Objects/ListCommands.cs
@@ -684,7 +684,7 @@ namespace Garnet.server
                 if (sourceDirection == OperationDirection.Unknown || destinationDirection == OperationDirection.Unknown)
                 {
                     return AbortWithErrorMessage(count, CmdStrings.RESP_ERR_GENERIC_SYNTAX_ERROR);
-                }                
+                }
 
                 result = ListMove(count, sourceKey, destinationKey, sourceDirection, destinationDirection, out var node, ref storageApi);
                 if (node != null)
--- a/libs/server/Resp/Objects/ListCommands.cs
+++ b/libs/server/Resp/Objects/ListCommands.cs
@@ -692,7 +692,7 @@ namespace Garnet.server
                 {
                     sourceDirection = OperationDirection.Unknown;
                 }
-                
+
                 if (Ascii.EqualsIgnoreCase("RIGHT", param2.ReadOnlySpan))
                 {
                     destinationDirection = OperationDirection.Right;
--- a/libs/server/Resp/Objects/ListCommands.cs
+++ b/libs/server/Resp/Objects/ListCommands.cs
@@ -679,8 +679,12 @@ namespace Garnet.server
                 if (!RespReadUtils.ReadPtrWithLengthHeader(ref param2.ptr, ref param2.length, ref ptr, recvBufferPtr + bytesRead))
                     return false;
 
-                var sourceDirection = param1.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : OperationDirection.Left;
-                var destinationDirection = param2.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : OperationDirection.Left;
+                OperationDirection sourceDirection = GetOperationDirection(param1.ReadOnlySpan);
+                OperationDirection destinationDirection = GetOperationDirection(param2.ReadOnlySpan);
+                if (sourceDirection == OperationDirection.Unknown || destinationDirection == OperationDirection.Unknown)
+                {
+                    return AbortWithErrorMessage(count, CmdStrings.RESP_ERR_GENERIC_SYNTAX_ERROR);
+                }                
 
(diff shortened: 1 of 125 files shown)


libs/server/Resp/Objects/ObjectStoreUtils.cs
```diff
@@ -65,5 +65,28 @@ private bool AbortWithErrorMessage(int count, ReadOnlySpan<byte> errorMessage)
...
+
+        /// <summary>
+        /// Tries to parse the input as "LEFT" or "RIGHT" and returns the corresponding OperationDirection.
+        /// If parsing fails, returns OperationDirection.Unknown.
+        /// </summary>
+        /// <param name="input">The input to parse.</param>
+        /// <returns>The parsed OperationDirection, or OperationDirection.Unknown if parsing fails.</returns>
+        public OperationDirection GetOperationDirection(ReadOnlySpan<byte> input)
+        {
+            string inputString = Encoding.UTF8.GetString(input).ToUpperInvariant();
```
badrishc, line 77: Do not create strings please. Strings are expensive. Do this:

param1.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : (param1.ReadOnlySpan.SequenceEqual("LEFT"u8) ? OperationDirection.Left : OperationDirection.Unknown)

Also, no need to pay the cost of a method call, do this directly in the caller.

PaulusParssinen: In case we need case insensitive comparison, you can use ``Ascii.EqualsIgnoreCase`` for .NET 8 (and fallback to string.Equals(.., StringComparison.OrdinalIgnoreCase for .NET 7 and below, or do case masking if the string allocation matters)

Zzhiter: I understand what you mean. Could you please help me know how to distinguish the .NET version in the program?

PaulusParssinen: Well, I guess we should just allow capitalized for now. We can later make it case-insensitive if the spec says so. 
However if you go with case insensitive path, you can conditionally compile lines against specific .NET versions with preprocessor directives (``if #NET8_OR_GREATER`` etc.).

Zzhiter: Thank you! I will try it. 
And Redis supports ignoring case. Look at the last line
image

Zzhiter: > Do not create strings please. Strings are expensive. Do this:
> 
> param1.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : (param1.ReadOnlySpan.SequenceEqual("LEFT"u8) ? OperationDirection.Left : OperationDirection.Unknown)
> 
> Also, no need to pay the cost of a method call, do this directly in the caller.

I think it would be better to encapsulate it into a function? Since there are other places where this function will be used. It seems a bit long if I expand the function directly at the call point?

PaulusParssinen: You could make static helper with `MethodImpl.AggressiveInlining` attribute, if the static call is that much of a concern.

libs/server/Resp/Objects/ObjectStoreUtils.cs
```diff
@@ -65,5 +66,40 @@ private bool AbortWithErrorMessage(int count, ReadOnlySpan<byte> errorMessage)
...
         }
+
+        /// <summary>
+        /// Tries to parse the input as "LEFT" or "RIGHT" and returns the corresponding OperationDirection.
+        /// If parsing fails, returns OperationDirection.Unknown.
+        /// </summary>
+        /// <param name="input">The input to parse.</param>
+        /// <returns>The parsed OperationDirection, or OperationDirection.Unknown if parsing fails.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public OperationDirection GetOperationDirection(ReadOnlySpan<byte> input)
```
badrishc, line 77: Use this version to optimize for the common case and avoid strings for pre-NET8"

```cs
        public OperationDirection GetOperationDirection(ArgSlice input)
        {
            // Optimize for the common case
            if (input.ReadOnlySpan.SequenceEqual("LEFT"u8))
                return OperationDirection.Left;
            if (input.ReadOnlySpan.SequenceEqual("RIGHT"u8))
                return OperationDirection.Right;
            // Rare case: try making upper case and retry
            MakeUpperCase(input.ptr);
            if (input.ReadOnlySpan.SequenceEqual("LEFT"u8))
                return OperationDirection.Left;
            if (input.ReadOnlySpan.SequenceEqual("RIGHT"u8))
                return OperationDirection.Right;
            return OperationDirection.Unknown;
        }
```

libs/server/Resp/Objects/ObjectStoreUtils.cs
```diff
@@ -65,5 +66,40 @@ private bool AbortWithErrorMessage(int count, ReadOnlySpan<byte> errorMessage)
...
             return true;
         }
+
+        /// <summary>
+        /// Tries to parse the input as "LEFT" or "RIGHT" and returns the corresponding OperationDirection.
+        /// If parsing fails, returns OperationDirection.Unknown.
+        /// </summary>
+        /// <param name="input">The input to parse.</param>
+        /// <returns>The parsed OperationDirection, or OperationDirection.Unknown if parsing fails.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
badrishc, line 76: No need for this aggressive inline - let JIT decide if it is worth it.

libs/server/Resp/Objects/ObjectStoreUtils.cs
Zzhiter: Great, you've taken into account the majority of cases and the minority of cases!

libs/server/Resp/Objects/ObjectStoreUtils.cs
```diff
@@ -65,5 +65,38 @@ private bool AbortWithErrorMessage(int count, ReadOnlySpan<byte> errorMessage)
...
+
+        /// <summary>
+        /// Tries to parse the input as "LEFT" or "RIGHT" and returns the corresponding OperationDirection.
+        /// If parsing fails, returns OperationDirection.Unknown.
+        /// </summary>
+        /// <param name="input">The input to parse.</param>
+        /// <returns>The parsed OperationDirection, or OperationDirection.Unknown if parsing fails.</returns>
+        public OperationDirection GetOperationDirection(ArgSlice input)
+        {
+#if NET8_0_OR_GREATER
```
badrishc, line 77: there is no need for the conditional now. just use the else part for all .NET versions.

Zzhiter: Good suggestion, I have fixed it, thank you for your review again!

libs/server/Resp/Objects/ListCommands.cs
```diff
@@ -679,8 +679,12 @@ private unsafe bool ListMove<TGarnetApi>(int count, byte* ptr, ref TGarnetApi st
                 if (!RespReadUtils.ReadPtrWithLengthHeader(ref param2.ptr, ref param2.length, ref ptr, recvBufferPtr + bytesRead))
                     return false;
 
-                var sourceDirection = param1.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : OperationDirection.Left;
-                var destinationDirection = param2.ReadOnlySpan.SequenceEqual("RIGHT"u8) ? OperationDirection.Right : OperationDirection.Left;
+                OperationDirection sourceDirection = GetOperationDirection(param1);
+                OperationDirection destinationDirection = GetOperationDirection(param2);
+                if (sourceDirection == OperationDirection.Unknown || destinationDirection == OperationDirection.Unknown)
+                {
+                    return AbortWithErrorMessage(count, CmdStrings.RESP_ERR_GENERIC_SYNTAX_ERROR);
```
TalZaccai, line 686: Thanks for adding this fix! If you have a couple of minutes, would you please add a couple of test cases to check that the correct error is raised? Probably best to add in the CanUseLMoveGC method in RespListTest.cs
Appreciate it!

Zzhiter: Of course. I'm currently participating in a competition outside, I'll add it after the competition.

TalZaccai: Bonus points if you can also do a similar thing in ListInsert for BEFORE | AFTER string comparisons :)

Zzhiter: > Bonus points if you can also do a similar thing in ListInsert for BEFORE | AFTER string comparisons :) (No pressure, we can always add that later)

I have added test case. BEFORE | AFTER string comparisons I will finish it later :)

TalZaccai: Sounds good, appreciate it!

PR #279, opened 2024-04-13
Update NOTICE.md, README.md and SECURITY.md

skanda890:

PR #281, opened 2024-04-14
Update NOTICE.md, README.md and SECURITY.md

skanda890:


MartinJohns:
Duplicate of #279. Just re-creating the PR when it was earlier rejected is very very rude.

skanda890:
I didn't know that

skanda890:
But can you tell why that got rejected

badrishc:
Most of these files are boilerplate provided by github/microsoft org and we cannot change them. Cache-store is what we are calling this class of systems, not "cache store". Very minor edits like commas do not merit a pull request either. Thanks.

PR #283, opened 2024-04-14
Publish additional docker images to the registry

akoken:
Currently, the repository contains Dockerfiles for various Linux distributions and nanoserver, but only the default (debian) image is being published to the registry. This pull request addresses this limitation by extending the publication process to include all Docker images for supported Linux distributions and nanoserver.

Due to the lack of support for Windows in [buildx](https://github.com/moby/buildkit/issues/616), I've split the pipeline into separate ones for Linux and Windows, ensuring that all images are properly built and published. The registry should look like [this](https://github.com/akoken/garnet/packages) after changes have been merged.

### DIFF
--- a/.github/workflows/docker-linux.yml
+++ b/.github/workflows/docker-linux.yml
@@ -1,4 +1,4 @@
-name: Garnet Docker images
+name: Garnet Docker images for Linux
 
 on:
   workflow_dispatch: # allow manual run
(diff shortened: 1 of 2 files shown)


badrishc:
Great work! Thanks for the contribution.

irinasp:
Thank you for your contribution, it looks good!

irinasp (approved):

akoken:
Hi @irinasp

I think there is a problem with the registry. Some of the container images received a `403 Forbidden` error, which is quite interesting. Could you please remove the existing images from the registry and trigger the pipeline again? Before doing so, you can check the permissions from [here](https://github.com/orgs/microsoft/packages/container/garnet/settings). Based on what I see in this [issue](https://github.com/docker/build-push-action/issues/463#issuecomment-924995909), this situation is very common.

Thanks.

PR #286, opened 2024-04-17
Fix typo in GarnetServerMonitor

Mathos1432:
Fixing a typo in the GarnetServerMonitor.


TalZaccai (approved):

badrishc:
Even though `GetAllLocksets` is unused, it is a very useful tool/method used to debug issues with locking. Is there a good reason to get rid of it? If not, it would be better to retain it.

PR #287, opened 2024-04-17
Supported RESP Commands Info

TalZaccai:
1. Restructuring RespCommandInfo to read supported RESP commands info from JSON file (Garnet.server/Resp/RespCommandsInfo.json)
2. Adding support for specifying command info when registering custom commands (example implemented in GarnetServer, parsing custom command info from GarnetServer/CustomRespCommandsInfo.json)
4. Implementing COMMAND, COMMAND INFO & COMMAND COUNT RESP commands, matching information returned with Redis
5. Implementing serialization & deserialization of command info to/from JSON and to/from RESP
6. Adding CommandInfoUpdater tool for easily adding / removing supported commands & sub-commands info, generating an updated JSON file to replace RespCommandsInfo.json
7. Adding tests for new commands specified in #4


libs/server/Resp/AdminCommands.cs
```diff
@@ -250,33 +250,102 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+                            foreach (var cmd in respCommandsInfo)
+                            {
+                                cmdCount++;
+                                resultSb.Append(cmd.RespFormat);
+                            }
+                        }
+
+                        while (!RespWriteUtils.WriteArrayLength(cmdCount, ref dcurr, dend))
+                            SendAndReset();
+                        while (!RespWriteUtils.WriteDirect(Encoding.ASCII.GetBytes(resultSb.ToString()), ref dcurr, dend))
```
PaulusParssinen, line 301: nit
```suggestion
                        while (!RespWriteUtils.WriteAsciiDirect(resultSb.ToString(), ref dcurr, dend))
```

libs/server/Resp/AdminCommands.cs
```diff
@@ -250,33 +250,102 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+
+                        for (var i = 0; i < count - 1; i++)
+                        {
+                            if (!RespReadUtils.ReadStringWithLengthHeader(out var cmdName, ref ptr, recvBufferPtr + bytesRead))
+                                return false;
+
+                            if (RespCommandsInfo.TryGetRespCommandInfo(cmdName, logger, out var cmdInfo) ||
+                                storeWrapper.customCommandManager.TryGetCustomCommandInfo(cmdName, out cmdInfo))
+                            {
+                                while (!RespWriteUtils.WriteDirect(Encoding.ASCII.GetBytes(cmdInfo.RespFormat), ref dcurr, dend))
```
PaulusParssinen, line 319:
```suggestion
                                while (!RespWriteUtils.WriteAsciiDirect(cmdInfo.RespFormat, ref dcurr, dend))
```

libs/server/Resp/RespCommandInfoParser.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// <param name="supportedCommands">Mapping between command name and Garnet RespCommand and ArrayCommand values</param>
+        /// <param name="commandInfo">Parsed RespCommandsInfo object</param>
+        /// <param name="parentCommand">Name of parent command, null if none</param>
+        /// <returns>True if parsing successful</returns>
+        public static unsafe bool TryReadFromResp(ref byte* ptr, byte* end, IReadOnlyDictionary<string, (RespCommand, byte?)> supportedCommands, out RespCommandsInfo commandInfo, string parentCommand = null)
+        {
+            commandInfo = default;
+
+            // Command info is null
+            if (Encoding.ASCII.GetString(ptr, 5) == "$-1\r\n") return true;
```
PaulusParssinen, line 29: We can avoid allocation by
```suggestion
            if (new ReadOnlySpan<byte>(ptr, 5).SequenceEquals("$-1\r\n"u8)) return true;
```
(Probably requires adding `using System;`, <sub>`<ImplicitUsings>` would be nice</sub>)

libs/server/Resp/RespCommandInfoParser.cs
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        private static unsafe bool TryReadKeySpecHeader(ref byte* ptr, byte* end, out string keySpecType)
+        {
+            keySpecType = default;
+
+            RespReadUtils.ReadArrayLength(out var ksTypeElemCount, ref ptr, end);
+            if (ksTypeElemCount != 4) return false;
+            RespReadUtils.ReadStringWithLengthHeader(out var ksTypeStr, ref ptr, end);
+            if (ksTypeStr != "type") return false;
```
PaulusParssinen, line 223: nit: A bit faster & "precise" comparisons with `StringComparison.Ordinal` which wont allow any funny business by culture normalizations etc.
```suggestion
            if (!ksTypeStr.Equals("type", StringComparison.Ordinal)) return false;
```

ditto for all string comparisongs in RESP parsing :+1:

PaulusParssinen (commented):
Few small nits for RESP writing/parsing parts, can be left as follow ups too.

TalZaccai:
> Few small nits for RESP writing/parsing parts, can be left as follow ups too.

Much appreciated @PaulusParssinen!

libs/server/Resp/AdminCommands.cs
```diff
@@ -250,33 +250,102 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+                        respCommandCount = 0;
+                    }
+
+                    var commandCount = storeWrapper.customCommandManager.CustomCommandsInfoCount + respCommandCount;
+
+                    while (!RespWriteUtils.WriteInteger(commandCount, ref dcurr, dend))
+                        SendAndReset();
+                }
+                // Handle COMMAND and COMMAND INFO
+                else if (count == 0 || subCommand.SequenceEqual(CmdStrings.INFO) || subCommand.SequenceEqual(CmdStrings.info))
```
lmaas, line 276: I know this is not specific to this PR, but I think `COMMAND INFO` shouldn't require admin permissions.

libs/server/Resp/RespCommandInfoParser.cs
```diff
@@ -0,0 +1,457 @@
...
+                else return false;
+            }
+            else return false;
+
+            if (!parser.TryReadFromResp(ref ptr, end, out keySpecMethod)) return false;
+
+            return true;
+        }
+
+        private static unsafe bool TryReadKeySpecHeader(ref byte* ptr, byte* end, out string keySpecType)
```
lmaas, line 234: NIT: Missing documentation comment

libs/server/Resp/RespCommandsInfoProvider.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Garnet.common;
+using Microsoft.Extensions.Logging;
+using JsonSerializer = System.Text.Json.JsonSerializer;
+
+namespace Garnet.server
+{
+    /// <summary>
+    /// Interface for importing / exporting resp commands info from different file types
```
lmaas, line 19: NIT: We should probably be consistent in the spelling of RESP (all uppercase).

lmaas (commented):
Great work on this PR! I'm happy to see the command info becoming much more maintainable. From what I can see there are no major changes needed besides some additional cleanup (see below).

test/Garnet.test/RespCommandTests.cs
```diff
@@ -0,0 +1,181 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using Garnet.server;
+using NUnit.Framework;
+using StackExchange.Redis;
+using SetOperation = Garnet.server.SetOperation;
+
+namespace Garnet.test
```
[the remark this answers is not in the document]
TalZaccai: I think the only RESP test classes that are currently under the Garnet.test.Resp namespace are the ACL ones. Let's leave it as-is in this PR for consistency and then consider moving all the RESP test classes under Resp/ in a separate PR.

playground/CommandInfoUpdater/SupportedCommand.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// Garnet RespCommand
+        /// </summary>
+        public RespCommand RespCommand { get; set; }
+
+        /// <summary>
+        /// Garnet ArrayCommand
+        /// </summary>
+        public byte? ArrayCommand { get; set; }
+
+        public SupportedCommand()
```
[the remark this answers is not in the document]
TalZaccai: I did :) Since this class is being serialized, it needs to have an explicit parameterless constructor (since it also has a non-parameterless constructor)

test/Garnet.test/RespScanCommandsTests.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using System.Text;
 using Garnet.server;
+using Microsoft.Extensions.Logging.Abstractions;
```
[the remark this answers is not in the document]
TalZaccai: Nope, thanks for noticing!

playground/CommandInfoUpdater/CommandInfoUpdater.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        /// <summary>
+        /// Update the mapping of commands info
+        /// </summary>
+        /// <param name="existingCommandsInfo">Existing command info mapping</param>
+        /// <param name="commandsToAdd">Commands to add</param>
+        /// <param name="commandsToRemove">Commands to remove</param>
+        /// <param name="queriedCommandsInfo">Queried commands info</param>
+        /// <returns></returns>
+        private static IReadOnlyDictionary<string, RespCommandsInfo> GetUpdatedCommandsInfo(IReadOnlyDictionary<string, RespCommandsInfo> existingCommandsInfo, IDictionary<SupportedCommand, bool> commandsToAdd,
```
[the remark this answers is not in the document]
TalZaccai: Thanks, I work with an ultra-wide screen so I sometimes miss those 😆

libs/server/Resp/RespCommandInfoParser.cs
```diff
@@ -0,0 +1,457 @@
...
+            /// <param name="keySpecMethod">Parsed KeySpecMethodBase object</param>
+            /// <returns>True if parsing successful</returns>
+            unsafe bool TryReadFromResp(ref byte* ptr, byte* end, out KeySpecMethodBase keySpecMethod);
+        }
+
+        internal sealed class BeginSearchIndexParser : IKeySpecParser
+        {
+            private static BeginSearchIndexParser instance;
+
+            private BeginSearchIndexParser()
```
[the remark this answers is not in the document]
TalZaccai: Just a private constructor, necessary to keep the singleton design... Not sure if worth commenting :)

libs/server/Resp/AdminCommands.cs
TalZaccai: Good catch!

libs/server/Resp/RespCommandInfoParser.cs
```diff
@@ -0,0 +1,488 @@
...
+            {
+            }
+
+            public static BeginSearchIndexParser Instance
+            {
+                get { return instance ??= new BeginSearchIndexParser(); }
+            }
```
lmaas, line 286: I would personally comment this as follows:

```suggestion
            /// <summary>
            /// Disallow default constructor (singleton)
            // </summary>
            private BeginSearchIndexParser() { }

            /// <summary>
            /// Returns the singleton instance of <see cref="BeginSearchIndexParser">.
            /// </summary>
            public static BeginSearchIndexParser Instance
            {
                get { return instance ??= new BeginSearchIndexParser(); }
            }
```

And obviously the same for the remaining singletons.

playground/CommandInfoUpdater/SupportedCommand.cs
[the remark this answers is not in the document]
lmaas: That makes sense. I would personally add that explicitly as a comment. Something like this:

```suggestion
        /// <summary>
        /// Default constructor provided for JSON serialization
        /// </summary>
        public SupportedCommand()
```

test/Garnet.test/RespCommandTests.cs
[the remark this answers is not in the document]
lmaas: Ok, let's make those changes in a separate PR. Generally, I think it's a good idea to start moving new tests into the correct place in the namespace hierarchy (like we did with the `RespReadUtils` tests).

lmaas (commented):
A few more small suggestions from my side, but I think the code looks good and can be merged afterwards.

badrishc (approved):

libs/server/Resp/RespCommandInfoParser.cs
PaulusParssinen: Looking back at this, I was confused.. The `string.Equals` and the equality operators already do Ordinal comparison. But it's still [good practice](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings) to explicitly specify the `CultureInfo` as this is very inconsistent across all the methods..

image

lmaas (approved):
Thank you for addressing all the comments above. The code looks good to me now. I think we can go ahead and merge this into main. Thanks for your hard work! 🚀

PR #292, opened 2024-04-19
fix INCR DECR operation zero

### LINKED ISSUE #290: [BUG]: INCR DECR operation zero - "ERR value is not an integer or out of range."
### Describe the bug

If a number is set to 0 using INCR or DECR, the INCR and DECR operations can no longer be used.

* case 1
127.0.0.1:6379> INCR key100
(integer) 1
127.0.0.1:6379> INCR key100
(integer) 2
127.0.0.1:6379> DECR key100
(integer) 1
127.0.0.1:6379> DECR key100
(integer) 0
127.0.0.1:6379> DECR key100
(error) ERR value is not an integer or out of range.
127.0.0.1:6379> INCR key100
(error) ERR value is not an integer or out of range.
127.0.0.1:6379> 

* case 2
127.0.0.1:6379> SET key101 0
OK
127.0.0.1:6379> INCR key101
(error) ERR value is not an integer or out of range.

### Expected behavior

_No response_

### Screenshots

_No response_

### Release version

127.0.0.1:6379> INFO SERVER
# Server
garnet_version:1.0.5
garnet_mode:standalone
os:Unix 6.6.16.0
processor_count:8
arch_bits:64
uptime_in_seconds:45
uptime_in_days:0
monitor_task:disabled
monitor_freq:0
latency_monitor:disabled
run_id:2004b00cc75ed9f0b0b6496173deb77ceb17d657
redis_version:6.2.11

### IDE

redis-cli

### OS version

MAC M1 - docker - ghcr.io/microsoft/garnet:latest

### Additional context

_No response_

tisilent:
fix #290

### DIFF
--- a/libs/common/NumUtils.cs
+++ b/libs/common/NumUtils.cs
@@ -113,14 +113,20 @@ namespace Garnet.common
         /// <returns>True if sequence contains only numeric digits, otherwise false</returns>
         public static bool TryBytesToLong(int length, byte* source, out long result)
         {
+            result = 0;
+
+            // Check empty value
+            if (length == 0)
+                return false;
+
             var fNeg = *source == '-';
             var beg = fNeg ? source + 1 : source;
             var end = source + length;
             var digit = *beg - '0';
-            result = 0;
 
             // Check first digit which needs to be non-zero
-            if ((digit is <= 0 or > 9 && end - beg > 1) || length == 0)
+            // Skip zero value
+            if (digit is <= 0 or > 9 && end - beg > 1)
                 return false;
 
             while (beg < end)
--- a/libs/common/NumUtils.cs
+++ b/libs/common/NumUtils.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Buffers.Text;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -112,7 +113,7 @@ namespace Garnet.common
         /// <param name="result">Long value extracted from sequence</param>
         /// <returns>True if sequence contains only numeric digits, otherwise false</returns>
         public static bool TryBytesToLong(int length, byte* source, out long result)
-                {
+        {
             var fNeg = *source == '-';
             var beg = fNeg ? source + 1 : source;
             var len = fNeg ? length - 1 : length;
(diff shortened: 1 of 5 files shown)


libs/server/Resp/BasicCommands.cs
```diff
@@ -753,16 +753,6 @@ private bool NetworkIncrement<TGarnetApi>(byte* ptr, RespCommand cmd, ref TGarne
                 *(valPtr + RespInputHeader.Size) = (byte)'1';
                 input = new ArgSlice(valPtr, vsize);
             }
-            else if (cmd == RespCommand.DECR)
```
vazois, line 756: We better leave this as it was so in the case where the key does not exist we do a direct upsert of "-1".

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -142,7 +142,7 @@ public bool InitialUpdater(ref SpanByte key, ref SpanByte input, ref SpanByte va
                         return false;
                     CopyUpdateNumber(incrBy, ref value, ref output);
                     break;
-
+                case RespCommand.DECR:
```
vazois, line 145: This is not needed if the input provided is "-1".

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
```diff
@@ -74,7 +74,7 @@ public int GetRMWInitialValueLength(ref SpanByte input)
                     var ndigits = NumUtils.NumDigitsInLong(next, ref fNeg);
 
                     return sizeof(int) + ndigits + (fNeg ? 1 : 0);
-
+                case RespCommand.DECR:
```
vazois, line 77: This is not needed if the input provided is "-1".

test/Garnet.test/RespTests.cs
```diff
@@ -1753,5 +1767,43 @@ public void AppendTest()
             Assert.IsTrue(time.Value.TotalSeconds > 0);
         }
 
+        #region Regression tests
+
+        #region #290 INCR DECR operation zero
+        [Test]
+        public void IncrAndDecrTest_290()
```
vazois, line 1774: Isn't @The fix for the #290 covered by [TestCase("key1", 0)] SingleDecr?
If so the test case is not needed.

libs/common/NumUtils.cs
```diff
@@ -113,14 +113,20 @@ public static long BytesToLong(int length, byte* source)
         /// <returns>True if sequence contains only numeric digits, otherwise false</returns>
         public static bool TryBytesToLong(int length, byte* source, out long result)
         {
```
vazois, line 115: This method needs to be replaced with Utf8Parser

```suggestion
                {
            var fNeg = *source == '-';
            var beg = fNeg ? source + 1 : source;
            var len = fNeg ? length - 1 : length;
            result = 0;

            // Do not allow leading zeros
            if (len > 1 && *beg == '0')
                return false;

            // Parse number and check consumed bytes to avoid alphanumeric strings
            if (!Utf8Parser.TryParse(new Span<byte>(beg, len), out result, out var bytesConsumed) || bytesConsumed != len)
                return false;

            // Negate if parsed value has a leading negative sign
            result = fNeg ? -result : result;
            return true;
        }
```

vazois (changes requested):

libs/common/NumUtils.cs
PaulusParssinen: If we can go Utf8Parser, we could just entirely use Utf8Parser for everything in NumUtils?

vazois: There is no official guideline yet. We are currently working on making iterative changes to the NumUtils methods and evaluate their performance impact.

libs/server/Resp/BasicCommands.cs
tisilent: emm....
https://github.com/microsoft/garnet/blob/03611e19292bd27c0a2c185a33279b3c429f7010/libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs#L130-L143
Because of this part, I modified the command.

PR #293, opened 2024-04-19
Fix FAILOVER replica attach workflow

vazois:
This PR refactors the failover workflow logic to deal with the following issues:

1. Attaching to many replicas some of which are unreachable. In main this would have caused the attach loop to fail and result in some of the reachable replicas never receiving the attach message.
2. When FAILOVER with default option is called the old primary will receive a request to become a replica as per the Redis spec.
3. Added flag to useGossipConnections for the failover but did not currently expose it until requested.


badrishc (approved):
LGTM, please resolve conflicts and merge, thanks!

PR #296, opened 2024-04-20
Optimize glob matching logic

PaulusParssinen:
Make the `GlobUtils.Match` do direct ASCII comparisons instead of more expensive culture-sensitive ones.

Sample diff of one of the conditions before and after:
image


libs/server/GlobUtils.cs
```diff
@@ -9,11 +9,13 @@ namespace Garnet.server
...
     {
         /// <summary>
-        /// Glob-style pattern matching
+        /// Glob-style ASCII pattern matching
         /// </summary>
         /// <returns>Whether match was found</returns>
-        public static unsafe bool Match(byte* pattern, int patternLen, byte* key, int stringLen, bool nocase = false)
+        public static unsafe bool Match(byte* pattern, int patternLen, byte* key, int stringLen, bool ignoreCase = false)
         {
+            static byte ToLowerAscii(byte value) => (byte)(value | 0x20);
```
lmaas, line 17: This code considers @=`, and ^=~ (all other relevant special characters are already handled below). IMO, this behavior is absolutely fine for our use case, but it might be worth adding this limitation to the doc comment for Match().

lmaas (approved):

badrishc (approved):

libs/server/GlobUtils.cs
PaulusParssinen: Yeah, I thought about whether to add checks around that but decided it's okay in here, i.e. it is only used for both operands to equality/or in the range pattern.

I'll add the check for upper-case ASCII just in case 👍

libs/server/GlobUtils.cs
```diff
@@ -9,11 +9,18 @@ namespace Garnet.server
...
-        /// Glob-style pattern matching
+        /// Glob-style ASCII pattern matching
         /// </summary>
         /// <returns>Whether match was found</returns>
-        public static unsafe bool Match(byte* pattern, int patternLen, byte* key, int stringLen, bool nocase = false)
+        public static unsafe bool Match(byte* pattern, int patternLen, byte* key, int stringLen, bool ignoreCase = false)
         {
+            static byte ToLowerAscii(byte value)
+            {
+                if ((value - 'A') <= ('Z' - 'A')) // Is in [A-Z]
```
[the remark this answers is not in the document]
PaulusParssinen:
```suggestion
                if ((uint)(value - 'A') <= (uint)('Z' - 'A')) // Is in [A-Z]
```

lmaas (approved):
Code looks good to me. Let's merge this in. Thanks for your contribution! 🙌

PR #297, opened 2024-04-21
Small Tsavorite project file clean-up

PaulusParssinen:
Remove some .NET FX style remains from the Tsavorite project files.


libs/storage/Tsavorite/cs/test/BasicTests.cs
```diff
@@ -799,16 +800,17 @@ public static void KVBasicsSampleEndToEndInDocs()
             Assert.AreEqual(10, output);
         }
 
-#if WINDOWS
         [Test]
         [Category("TsavoriteKV")]
         public static void LogPathtooLong()
         {
+            if (!OperatingSystem.IsWindows())
+                Assert.Inconclusive();
```
TedHartMS, line 808: .Inconclusive shows up in VS as "Not Run", so let's make it "Skipped".

libs/storage/Tsavorite/cs/test/TestUtils.cs
```diff
@@ -134,14 +129,14 @@ internal static IDevice CreateTestDevice(DeviceType testDeviceType, string filen
 
             switch (testDeviceType)
             {
-#if WINDOWS
-                case DeviceType.LSD:
+                case DeviceType.LSD when !OperatingSystem.IsWindows():
+                    Assert.Inconclusive();
```
TedHartMS, line 133: same comment as above. This might be better as an Assert.Fail, since we shouldn't be asking for LSD specifically on non-Windows.

libs/storage/Tsavorite/cs/test/BasicTests.cs
PaulusParssinen: I think this one should do that then?
```suggestion
                Assert.Ignore("Skipped");
```

libs/storage/Tsavorite/cs/test/TestUtils.cs
PaulusParssinen: I removed the preprocessor which previously were on the DeviceType enum. Is following better or do I revert last commit entirely?
```suggestion
                    Assert.Ignore($"Skipping {nameof(DeviceType.LSD)} on non-windows");
```

TedHartMS: Better to make this Assert,Fail; the test needs attention if it's requesting LSD on non-Windows. It returned  null device before in that case, so there'll be no new break.

PaulusParssinen: Ok 👍 I'll check all `DeviceType` enums to not include LSD

PaulusParssinen: The change above this method removed the #ifdef from the LSD enum definition, so I fear we can't make this fail 🤔

TedHartMS: Oh, right--[Values] runs them all. Ignore() it is, then.

TedHartMS (approved):

PR #302, opened 2024-04-22
Enable and apply CA1822: Mark members as static

PaulusParssinen:
This PR enables [`CA1822: Mark members as static`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822) analyzer from the Performance-category. The rule was applied using the VS automation.

Static calls are cheaper.


libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Revivification/FreeRecordPool.cs
```diff
@@ -575,7 +575,7 @@ internal void ScanForEmpty(CancellationToken cancellationToken)
         public void Dispose()
         {
             foreach (var bin in bins)
-                bin.Dispose();
+                FreeRecordBin.Dispose();
```
TedHartMS, line 578: This loop becomes a no-op with this change. Better to remove FreeRecordBin.Dispose entirely as there's currently no need for it

TedHartMS (commented):
Tsavorite changes look good, with one comment

badrishc:
For `StorageSession`, it has become rather confusing now with the API being a mix of static and non-static methods. It would be nice for the storage API to be discoverable via the instance `storageSession` instead of being split between static and non-static. Particularly as we might want to expose this API to users in future.

PaulusParssinen:
> For `StorageSession`, it has become rather confusing now with the API being a mix of static and non-static methods. It would be nice for the storage API to be discoverable via the instance `storageSession` instead of being split between static and non-static. Particularly as we might want to expose this API to users in future.

I'll revert those files where the automation made things more confusing 👍

badrishc (approved):
LGTM; please resolve the conflicts.

PR #306, opened 2024-04-22
Refactor Tsavorite checkpoint path handling to be OS-aware

PaulusParssinen:


TedHartMS (approved):

PaulusParssinen:
Something funny happening with the paths used in the cloud tests?

PaulusParssinen:
The problem is the fact that Tsavorite previously assumed all path-seperators to be ``/`` and let the windows NT APIs to deal with this instead. This PR makes this OS-aware and uses ``\`` for Windows which made the Azure storage tests explode as they use the same naming scheme. I propose to introducing a seperate `ICheckpointNamingScheme` in the AzureStorageDevice for this.. I'll test how that'd work out.

PaulusParssinen:
Making rest of the Tsavorite tests use `System.IO.Path` APIs for consistency.

PaulusParssinen:
Actually, I'll have that in separate PR.

TedHartMS (approved):
Looks good

PR #262, opened 2024-04-10
Increment/Decrement Valid Number Check

vazois:
Added validity check to catch increment/decrement of alphanumeric strings.
This is a proposed solution that minimizes the perf overhead of passing the error from RMWMethods.
However, the overhead of TryBytesToLong will not be avoided.
Maybe it is possible to include a type at upsert and avoid TryBytesToLong validation every time INCR/DECR is executed.

Second iteration updated the PR to catch overflow exceptions and to validate operands on copy update and at InitialUpdate.

Third iteration combined bdn.benchmark with embedded bench to evaluate performance of new validity checks.

**Main without validity checks**

| Method      | Job    | EnvironmentVariables | Runtime  | input    | Mean     | Error   | StdDev  |
|------------ |------- |--------------------- |--------- |--------- |---------:|--------:|--------:|
| Increment   | .NET 6 | Empty                | .NET 6.0 | ?        | 297.4 ns | 0.51 ns | 0.47 ns |
| Decrement   | .NET 6 | Empty                | .NET 6.0 | ?        | 303.4 ns | 0.49 ns | 0.38 ns |
| Increment   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | ?        | 206.5 ns | 0.28 ns | 0.25 ns |
| Decrement   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | ?        | 214.2 ns | 0.55 ns | 0.49 ns |
| IncrementBy | .NET 6 | Empty                | .NET 6.0 | Byte[30] | 210.0 ns | 0.34 ns | 0.32 ns |
| DecrementBy | .NET 6 | Empty                | .NET 6.0 | Byte[30] | 214.8 ns | 0.30 ns | 0.28 ns |
| IncrementBy | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | Byte[30] | 132.7 ns | 0.30 ns | 0.29 ns |
| DecrementBy | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | Byte[30] | 131.2 ns | 0.36 ns | 0.34 ns |
| IncrementBy | .NET 6 | Empty                | .NET 6.0 | Byte[31] | 227.9 ns | 1.09 ns | 1.02 ns |
| DecrementBy | .NET 6 | Empty                | .NET 6.0 | Byte[31] | 236.1 ns | 0.86 ns | 0.81 ns |
[...]

PR #307, opened 2024-04-22
Fix INCR Regression

vazois:
Fix increment regression as identified by issue #290.
Also changed TryBytesLong to utilize Utf8Parser.
BDN benchmark did show any perf regression.
Additionally fixed, also issue with DECR when value is zero and in-place-update fails.


badrishc (approved):

libs/common/NumUtils.cs
```diff
@@ -115,21 +116,18 @@ public static bool TryBytesToLong(int length, byte* source, out long result)
...
 
-            while (beg < end)
-            {
-                digit = *beg++ - '0';
-                if (digit is < 0 or > 9)
-                    return false;
-                checked { result = (result * 10) + digit; }
-            }
+            // Parse number and check consumed bytes to avoid alphanumeric strings
+            if (!Utf8Parser.TryParse(new Span<byte>(beg, len), out result, out var bytesConsumed) || bytesConsumed != len)
```
lmaas, line 127: Won't `Utf8Parser.TryParse(ReadOnlySpan<Byte>, Int64, Int32, Char)` take care of the leading sign? If we allow leading zeros we can get rid of the previous checks as well and just fully map to `Utf8Parser.TryParse()`.

badrishc: Is `Utf8Parser.TryParse` better or more efficient than `TryBytesToLong`? Historically, these APIs have way too many checks that might not be relevant to our needs.

lmaas: I benchmarked this earlier, and we can do better than `Utf8Parser.TryParse` by relaxing the requirements a bit. Maybe keep the existing code and make `digit `explicitly unsigned to simplify the bounds check (`digit <= 9`).

vazois: > Won't `Utf8Parser.TryParse(ReadOnlySpan<Byte>, Int64, Int32, Char)` take care of the leading sign? If we allow leading zeros we can get rid of the previous checks as well and just fully map to `Utf8Parser.TryParse()`.

In both cases, the problem is the leading zeros. 
If the input is something like -010, bytesConsumed will be 4 so we cannot identify that this is not a valid number string. Same applies if it not a negative number (e.g. 010 will return 3 bytesConsumed).
So I broke it down into two distinct checks.
Redis does not allow leading zeros.

image

Performance wise it does not make much of a difference. check PR #262 for perf comparison.
Generally, I noticed there is +/-50ns difference.
Also checked the TryParseCustom vs Utf8Parser and it seems the former is better for numbers with few digits and the latter is better for numbers with many digits.
I think that having a custom parser could be better for reading headers of commands since there we expect few digits most of the time.
Here it should not make much difference, because it does not also touch the critical path.

lmaas (approved):

PR #308, opened 2024-04-22
Make Tsavorite.test path handling OS-aware

PaulusParssinen:
Use `System.IO.Path.Join` to construct paths in Tsavorite tests.

Split off from #306


TedHartMS (approved):
Nice cleanup

PR #327, opened 2024-04-25
Release epoch between lock retries

pradeepyadavmsft:
It avoids the system going into a deadlock. Also, it gives other threads a fair chance to proceed. 

It will retry 1000 times before releasing the epoch. We have kept this number high so that it doesn't release epochs unnecessarily and increase latency. But if it's unable to acquire a lock it will release it and retry so that other threads can proceed.

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/ClientSession/LockableContext.cs
+++ b/libs/storage/Tsavorite/cs/src/core/ClientSession/LockableContext.cs
@@ -21,6 +21,8 @@ namespace Tsavorite.core
         /// <summary>Indicates whether this struct has been initialized</summary>
         public bool IsNull => clientSession is null;
 
+        const int KeyLockMaxRetryAttempts = 1000;
+
         internal LockableContext(ClientSession<Key, Value, Input, Output, Context, Functions> clientSession)
         {
             this.clientSession = clientSession;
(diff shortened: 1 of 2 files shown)


libs/storage/Tsavorite/cs/src/core/ClientSession/LockableContext.cs
```diff
@@ -21,6 +21,8 @@ namespace Tsavorite.core
         /// <summary>Indicates whether this struct has been initialized</summary>
         public bool IsNull => clientSession is null;
 
+        static readonly int KeyLockMaxRetryAttempts = 1000;
```
badrishc, line 24: if we never change this, can we make this a const?

badrishc (approved):

libs/storage/Tsavorite/cs/src/core/ClientSession/LockableContext.cs
```diff
@@ -65,9 +67,14 @@ internal LockableContext(ClientSession<Key, Value, Input, Output, Context, Funct
             // This is the same as DoInternalTryLock but without timeout; it will keep trying until it acquires all locks.
             var end = start + count - 1;
 
+            int retryCount = 0;
         Retry:
-            long prevBucketIndex = -1;
+            if (retryCount >= KeyLockMaxRetryAttempts)
+            {
+                return false;
+            }
```
TedHartMS, line 75: let's move this down to line 95, where you update retryCount. That way we don't pay the "if" cost on the first iteration

TedHartMS (approved):

TedHartMS (approved):

PR #328, opened 2024-04-25
Add CODEOWNERS as part of security best practices by GitHub

darrenge:
GitHub best practices recommends CODEOWNERS for WorkFlows and will require it in the next few months (already contacted FASTER).  It is best to just get this taken care of so I added me as the default for WorkFlows files.


.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Recommended by GitHub best practices to have a Code Owner defined especially for .GitHub folder
+# NOTE: Order is important; the last matching pattern takes the most precedence.
+
+# For Workflows used by GitHub Actions
+/github/workflows/ @darrenge
```
badrishc, line 5: should be `.github` and not `github`?

PR #332, opened 2024-04-26
Add explicit protocol validation when reading RESP messages

lmaas:
# Overview
This PR adds explicit syntax checks to ensure RESP packages follow the correct RESP syntax, safeguarding against vulnerabilities that result from malicious RESP packages.

## Why is this important?
Not validating the complete RESP package syntax can expose vulnerabilities that allow attackers to divert Garnet's control flow and execute malicious actions. It also makes debugging of unintentionally malformed RESP messages much harder.

Take the following invalid RESP message as an example:

`*3\r\n$3\r\nSET\r\n$5\r\nmykey\r\n$\r\n1\r\nPING\r\n`

Without validation, this message will cause Tsavorite to crash and terminates the client connection without an explicit error message. With the changes in this PR Garnet will return a proper error message back to the client:

`-ERR Protocol Error: Unexpected character '\x0d'`

Please note that a parsing error will still close the affected RESP session, because it is impossible to safely recover from a malformed package without discarding the remaining read buffer.

## Implementation Details
The PR makes the following changes:

1. Adds a new `RespParsingException` exception type that is used to indicate parsing errors and terminate the active RESP session. Note that `RespParsingExceptions` should only be used when the error is expected to be forwarded to the client.
2. Adds explicit syntax checks in the RESP parsing logic for "non-essential" RESP tokens such as `'$'`,`'*'`,`':'` and `'\r\n'`
3. Adds digit and overflow checks for integer parsing.
4. Adds length checks for length headers (that could be used to offset into the package).

Known limitations:
- RESP error messages do not contain a direct indication where the error occurred
- The RESP NULL value (`$-1\r\n`) requires specific opt-in by the caller of `ReadLengthHeader()` to reduce the risk of unhandled null values.

## Performance Impact
Considering the requirement for additional checks in the RESP parsing logic — specifically bounds checking for digits and overflow validation for integers — we observed a noticeable impact on parsing performance. In certain scenarios, this impact extends to overall end-to-end performance as well. However, I believe this trade-off is justifiable given the significant improvements for security and stability.

Below I am listing the performance degradation I have measured for the different RESP parsing functions:

Method | Runtime | Testcase | Before (Mean) | PR (Mean)
-- | -- | -- | -- | --
ReadLengthHeader | .NET   6.0 | $-1\r\n | 1.926   ns | 2.462   ns
ReadLengthHeader | .NET   8.0 | $-1\r\n | 1.888   ns | 2.755   ns
ReadLengthHeader | .NET   6.0 | $-2147483648\r\n | 9.317   ns | NA
ReadLengthHeader | .NET   8.0 | $-2147483648\r\n | 9.067   ns | NA
ReadLengthHeader | .NET   6.0 | $0\r\n | 1.665   ns | 4.321   ns
ReadLengthHeader | .NET   8.0 | $0\r\n | 1.944   ns | 4.608   ns
ReadIntWithLengthHeader | .NET   6.0 | $1\r\n0\r\n | 9.153   ns | 9.072   ns
ReadLongWithLengthHeader | .NET   6.0 | $1\r\n0\r\n | 7.778   ns | 10.861   ns
ReadULongWithLengthHeader | .NET   6.0 | $1\r\n0\r\n | 6.918   ns | 8.177   ns
ReadIntWithLengthHeader | .NET   8.0 | $1\r\n0\r\n | 8.359   ns | 10.863   ns
ReadLongWithLengthHeader | .NET   8.0 | $1\r\n0\r\n | 6.878   ns | 9.337   ns
ReadULongWithLengthHeader | .NET   8.0 | $1\r\n0\r\n | 6.347   ns | 7.955   ns
ReadIntWithLengthHeader | .NET   6.0 | $10\r\n2147483647\r\n | 12.330   ns | 14.044   ns
ReadULongWithLengthHeader | .NET   6.0 | $10\r\n2147483647\r\n | 12.203   ns | 13.767   ns
ReadIntWithLengthHeader | .NET   8.0 | $10\r\n2147483647\r\n | 14.085   ns | 14.555   ns
ReadULongWithLengthHeader | .NET   8.0 | $10\r\n2147483647\r\n | 12.070   ns | 13.565   ns
ReadIntWithLengthHeader | .NET   6.0 | $11\r\n-2147483648\r\n | 12.685   ns | 14.707   ns
ReadIntWithLengthHeader | .NET   8.0 | $11\r\n-2147483648\r\n | 14.082   ns | 14.326   ns
ReadLongWithLengthHeader | .NET   6.0 | $19\r\n(...)807\r\n   [26] | 18.076   ns | 19.628   ns
ReadLongWithLengthHeader | .NET   8.0 | $19\r\n(...)807\r\n   [26] | 19.797   ns | 19.582   ns
ReadIntWithLengthHeader | .NET   6.0 | $2\r\n-1\r\n | 7.495   ns | 9.146   ns
ReadLongWithLengthHeader | .NET   6.0 | $2\r\n-1\r\n | 8.030   ns | 8.997   ns
ReadIntWithLengthHeader | .NET   8.0 | $2\r\n-1\r\n | 6.892   ns | 9.010   ns
ReadLongWithLengthHeader | .NET   8.0 | $2\r\n-1\r\n | 6.861   ns | 9.094   ns
ReadLongWithLengthHeader | .NET   6.0 | $20\r\n(...)808\r\n   [27] | 18.307   ns | 19.433   ns
ReadLongWithLengthHeader | .NET   8.0 | $20\r\n(...)808\r\n   [27] | 19.846   ns | 19.310   ns
ReadULongWithLengthHeader | .NET   6.0 | $20\r\n(...)615\r\n   [27] | 18.118   ns | 20.328   ns
ReadULongWithLengthHeader | .NET   8.0 | $20\r\n(...)615\r\n   [27] | 17.530   ns | 19.560   ns
ReadLengthHeader | .NET   6.0 | $2147483647\r\n | 8.712   ns | 9.340   ns
ReadLengthHeader | .NET   8.0 | $2147483647\r\n | 8.245   ns | 9.361   ns
ReadInt64 | .NET   6.0 | :-1\r\n | 4.028   ns | 5.553   ns
ReadInt64 | .NET   8.0 | :-1\r\n | 3.848   ns | 5.782   ns
ReadInt64 | .NET   6.0 | :-922(...)808\r\n   [23] | 37.710   ns | 15.601   ns
ReadInt64 | .NET   8.0 | :-922(...)808\r\n   [23] | 39.468   ns | 17.011   ns
ReadInt64 | .NET   6.0 | :0\r\n | 3.305   ns | 6.120   ns
ReadInt64 | .NET   8.0 | :0\r\n | 2.636   ns | 6.544   ns
ReadInt64 | .NET   6.0 | :9223(...)807\r\n   [22] | 36.654   ns | 16.752   ns
ReadInt64 | .NET   8.0 | :9223(...)807\r\n   [22] | 38.434   ns | 16.985   ns

_Please note that NA indicates a parsing error being thrown (i.e., length headers must not be negative)._

<details>
<summary>Complete performance before PR</summary><p>

| Method                    | Job    | EnvironmentVariables | Runtime  | testCase                 | Mean      | Error     | StdDev    | Median    |
|-------------------------- |------- |--------------------- |--------- |------------------------- |----------:|----------:|----------:|----------:|
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $-1\r\n                  |  1.926 ns | 0.0027 ns | 0.0024 ns |  1.925 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $-1\r\n                  |  1.888 ns | 0.0017 ns | 0.0016 ns |  1.888 ns |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $-2147483648\r\n         |  9.317 ns | 0.1888 ns | 0.1766 ns |  9.457 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $-2147483648\r\n         |  9.067 ns | 0.1523 ns | 0.1424 ns |  9.139 ns |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $0\r\n                   |  1.665 ns | 0.0007 ns | 0.0006 ns |  1.665 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $0\r\n                   |  1.944 ns | 0.0011 ns | 0.0009 ns |  1.944 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              |  9.153 ns | 0.0058 ns | 0.0055 ns |  9.154 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              |  7.778 ns | 0.0188 ns | 0.0176 ns |  7.776 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              |  6.918 ns | 0.0024 ns | 0.0023 ns |  6.918 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              |  8.359 ns | 0.0129 ns | 0.0121 ns |  8.353 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              |  6.878 ns | 0.0120 ns | 0.0100 ns |  6.877 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              |  6.347 ns | 0.0068 ns | 0.0057 ns |  6.349 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $10\r\n2147483647\r\n    | 12.330 ns | 0.0299 ns | 0.0279 ns | 12.324 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $10\r\n2147483647\r\n    | 12.203 ns | 0.0231 ns | 0.0216 ns | 12.199 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $10\r\n2147483647\r\n    | 14.085 ns | 0.0601 ns | 0.0562 ns | 14.090 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $10\r\n2147483647\r\n    | 12.070 ns | 0.0122 ns | 0.0114 ns | 12.070 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $11\r\n-2147483648\r\n   | 12.685 ns | 0.0183 ns | 0.0153 ns | 12.684 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $11\r\n-2147483648\r\n   | 14.082 ns | 0.0636 ns | 0.0595 ns | 14.081 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $19\r\n(...)807\r\n [26] | 18.076 ns | 0.0182 ns | 0.0170 ns | 18.086 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $19\r\n(...)807\r\n [26] | 19.797 ns | 0.0231 ns | 0.0216 ns | 19.796 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $2\r\n-1\r\n             |  7.495 ns | 0.0027 ns | 0.0025 ns |  7.495 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $2\r\n-1\r\n             |  8.030 ns | 0.0088 ns | 0.0083 ns |  8.031 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2\r\n-1\r\n             |  6.892 ns | 0.0128 ns | 0.0113 ns |  6.892 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2\r\n-1\r\n             |  6.861 ns | 0.0143 ns | 0.0134 ns |  6.857 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $20\r\n(...)808\r\n [27] | 18.307 ns | 0.0043 ns | 0.0038 ns | 18.307 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $20\r\n(...)808\r\n [27] | 19.846 ns | 0.0180 ns | 0.0159 ns | 19.843 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $20\r\n(...)615\r\n [27] | 18.118 ns | 0.0130 ns | 0.0121 ns | 18.118 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $20\r\n(...)615\r\n [27] | 17.530 ns | 0.0356 ns | 0.0333 ns | 17.518 ns |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $2147483647\r\n          |  8.712 ns | 0.1950 ns | 0.1523 ns |  8.712 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2147483647\r\n          |  8.245 ns | 0.1511 ns | 0.1413 ns |  8.297 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :-1\r\n                  |  4.028 ns | 0.0222 ns | 0.0208 ns |  4.030 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :-1\r\n                  |  3.848 ns | 0.0137 ns | 0.0128 ns |  3.847 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :-922(...)808\r\n [23]   | 37.710 ns | 0.0223 ns | 0.0174 ns | 37.705 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :-922(...)808\r\n [23]   | 39.468 ns | 0.0391 ns | 0.0365 ns | 39.474 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :0\r\n                   |  3.305 ns | 0.0044 ns | 0.0039 ns |  3.306 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :0\r\n                   |  2.636 ns | 0.0094 ns | 0.0083 ns |  2.635 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :9223(...)807\r\n [22]   | 36.654 ns | 0.0333 ns | 0.0312 ns | 36.654 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :9223(...)807\r\n [22]   | 38.434 ns | 0.0451 ns | 0.0400 ns | 38.442 ns |

</p>
</details>

<details>
<summary>Complete performance after PR</summary><p>

| Method                    | Job    | EnvironmentVariables | Runtime  | testCase                 | Mean      | Error     | StdDev    |
|-------------------------- |------- |--------------------- |--------- |------------------------- |----------:|----------:|----------:|
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $-1\r\n                  |  2.462 ns | 0.0072 ns | 0.0068 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $-1\r\n                  |  2.755 ns | 0.0048 ns | 0.0045 ns |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $-2147483648\r\n         |        NA |        NA |        NA |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $-2147483648\r\n         |        NA |        NA |        NA |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $0\r\n                   |  4.321 ns | 0.0008 ns | 0.0008 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $0\r\n                   |  4.608 ns | 0.0075 ns | 0.0070 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              |  9.072 ns | 0.0220 ns | 0.0195 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              | 10.861 ns | 0.0052 ns | 0.0046 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n              |  8.177 ns | 0.0085 ns | 0.0076 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              | 10.863 ns | 0.0114 ns | 0.0095 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              |  9.337 ns | 0.0010 ns | 0.0009 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n              |  7.955 ns | 0.0092 ns | 0.0081 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $10\r\n2147483647\r\n    | 14.044 ns | 0.0438 ns | 0.0410 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $10\r\n2147483647\r\n    | 13.767 ns | 0.0366 ns | 0.0324 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $10\r\n2147483647\r\n    | 14.555 ns | 0.0181 ns | 0.0169 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $10\r\n2147483647\r\n    | 13.565 ns | 0.1113 ns | 0.1041 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $11\r\n-2147483648\r\n   | 14.707 ns | 0.3044 ns | 0.2848 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $11\r\n-2147483648\r\n   | 14.326 ns | 0.0291 ns | 0.0272 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $19\r\n(...)807\r\n [26] | 19.628 ns | 0.0130 ns | 0.0116 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $19\r\n(...)807\r\n [26] | 19.582 ns | 0.0124 ns | 0.0110 ns |
| ReadIntWithLengthHeader   | .NET 6 | Empty                | .NET 6.0 | $2\r\n-1\r\n             |  9.146 ns | 0.0199 ns | 0.0186 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $2\r\n-1\r\n             |  8.997 ns | 0.0077 ns | 0.0072 ns |
| ReadIntWithLengthHeader   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2\r\n-1\r\n             |  9.010 ns | 0.0022 ns | 0.0018 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2\r\n-1\r\n             |  9.094 ns | 0.0020 ns | 0.0019 ns |
| ReadLongWithLengthHeader  | .NET 6 | Empty                | .NET 6.0 | $20\r\n(...)808\r\n [27] | 19.433 ns | 0.0188 ns | 0.0167 ns |
| ReadLongWithLengthHeader  | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $20\r\n(...)808\r\n [27] | 19.310 ns | 0.0077 ns | 0.0069 ns |
| ReadULongWithLengthHeader | .NET 6 | Empty                | .NET 6.0 | $20\r\n(...)615\r\n [27] | 20.328 ns | 0.1163 ns | 0.1031 ns |
| ReadULongWithLengthHeader | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $20\r\n(...)615\r\n [27] | 19.560 ns | 0.0235 ns | 0.0219 ns |
| ReadLengthHeader          | .NET 6 | Empty                | .NET 6.0 | $2147483647\r\n          |  9.340 ns | 0.0372 ns | 0.0330 ns |
| ReadLengthHeader          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $2147483647\r\n          |  9.361 ns | 0.1513 ns | 0.1415 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :-1\r\n                  |  5.553 ns | 0.0029 ns | 0.0025 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :-1\r\n                  |  5.782 ns | 0.0045 ns | 0.0043 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :-922(...)808\r\n [23]   | 15.601 ns | 0.1549 ns | 0.1209 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :-922(...)808\r\n [23]   | 17.011 ns | 0.1305 ns | 0.1221 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :0\r\n                   |  6.120 ns | 0.0109 ns | 0.0102 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :0\r\n                   |  6.544 ns | 0.0035 ns | 0.0033 ns |
| ReadInt64                 | .NET 6 | Empty                | .NET 6.0 | :9223(...)807\r\n [22]   | 16.752 ns | 0.0662 ns | 0.0620 ns |
| ReadInt64                 | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | :9223(...)807\r\n [22]   | 16.985 ns | 0.0037 ns | 0.0033 ns |

</p>
</details>


libs/common/Parsing/RespParsingException.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        /// <summary>
+        /// Throw an "Unexcepted Token" exception.
+        /// </summary>
+        /// <param name="c">The character that was unexpected.</param>
+        [DoesNotReturn]
+        public static void ThrowUnexpectedToken(byte c)
+        {
+            Throw($"Unexpected byte ({c}) in RESP command package.");
```
mgravell, line 30: since tokens are ASCII, I wonder if it would be preferable to use `{(char)c}` here, or maybe `{(char)c} ({c})` if we're worried about random corruption

libs/common/Parsing/RespParsingException.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        /// <summary>
+        /// Throw NaN (not a number) exception.
+        /// </summary>
+        /// <param name="buffer">The input buffer that could not be converted into a number.</param>
+        [DoesNotReturn]
+        public static void ThrowNotANumber(ReadOnlySpan<byte> buffer)
+        {
+            var ascii = new System.Text.ASCIIEncoding();
```
mgravell, line 60: prefer shared `Encoding.ASCII`

libs/common/Parsing/RespParsingException.cs
```diff
@@ -0,0 +1,82 @@
...
+            Throw($"Unable to parse number: {ascii.GetString(buffer)}");
+        }
+
+        /// <summary>
+        /// Throw a exception indicating that an integer overflow has occurred.
+        /// </summary>
+        [DoesNotReturn]
+        public static void ThrowIntegerOverflow()
+        {
+            var ascii = new System.Text.ASCIIEncoding();
```
mgravell, line 70: unused

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,363 @@
...
+                    return false;
+                }
+
+                var nextDigit = (uint)(*readHead - '0');
+                if (nextDigit > 9)
+                {
+                    goto Done;
+                }
+
+                if ((value == 1844674407370955161UL && ((int)nextDigit > 5)) || (value > 1844674407370955161UL))
```
mgravell, line 85: untested, but I wonder if `value = checked((10 * value) + nextDigit);` might be clearer and more consistent here (using CPU overflow flag, and well-known `OverflowException`)

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,363 @@
...
+            }
+
+            // Ensure terminator has been received
+            ptr = readHead + 2;
             if (ptr > end)
+            {
                 return false;
+            }
+
+            if (*(ushort*)readHead != MemoryMarshal.Read<ushort>("\r\n"u8))
```
mgravell, line 282: I wonder if this can be JIT-optimized to avoid repeated parse; perhaps

``` c#
internal static readonly ushort CrLf = BitConverter.IsLittleEndian ? (ushort)0x0A0D : (ushort)0x0D0A; // see: ASCII
```

which *should* be treated like `const` by modern JITs

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,363 @@
...
         /// <summary>
-        /// Read the length of an array of bulk strings
+        /// Read signed 64 bit number
         /// </summary>
-        /// <param name="number"></param>
-        /// <param name="ptr"></param>
-        /// <param name="end"></param>
-        /// <returns></returns>
-        public static bool ReadArrayLength(out int number, ref byte* ptr, byte* end)
+        public static bool Read64Int(out long number, ref byte* ptr, byte* end)
```
mgravell, line 293: IIRC, current RESP *at least with Redis* defines all client-to-server comms to be limited to arrays (at the root) and strings (inside the root); see https://redis.io/docs/latest/develop/reference/protocol-spec/#sending-commands-to-a-redis-server

as such, is `:123` etc actually used?

libs/common/RespReadUtils.cs
```diff
@@ -241,19 +473,32 @@ public static bool ReadByteArrayWithLengthHeader(out byte[] result, ref byte* pt
         /// </summary>
         public static bool ReadBoolWithLengthHeader(out bool result, ref byte* ptr, byte* end)
         {
-            //$1\r\n1\r\n
-            //$1\r\n0\r\n
             result = false;
-            if (ptr + 7 >= end)
+
```
mgravell, line 477: suggestion: can this be fast-pathed in the "we have 8+ bytes" case using pre-defined `ulong`, masking off one byte?

mgravell (commented):
only got about half-way through so far, but looks promising and much safer; some notes included

libs/common/RespReadUtils.cs
PaulusParssinen: ``MemoryMarshal.Read<ushort>("\r\n"u8)`` is folded from as constant from the RVA static too. Either will work.

mgravell: > edit: actually static readonly fields can't be considered const.

Yes they can - see https://github.com/dotnet/runtime/issues/4287 - this is why reflection against static readonly fields was hard blocked at (some version, when the JIT change was added)

PaulusParssinen: Interesting.. Here's `FullOpts` for:
```CSharp

internal unsafe static class Resp
{
    internal static readonly ushort CrLf = BitConverter.IsLittleEndian ? (ushort)0x0A0D : (ushort)0x0D0A; 

    public static bool EqNewline_StaticGet(ref byte* ptr) => *(ushort*)ptr != CrLf;
}
```

```asm
; Assembly listing for method DisasmoPlayground.Resp:EqNewline_StaticGet(byref):ubyte (FullOpts)
; Emitting BLENDED_CODE for X64 with AVX - Windows
; FullOpts code
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; Final local variable assignments
;
;  V00 arg0         [V00,T00] (  3,  3   )   byref  ->  rcx         single-def
;  V01 OutArgs      [V01    ] (  1,  1   )  struct (32) [rsp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;  V02 rat0         [V02,T01] (  2,  4   )     int  ->  rbx         "Spilling to split statement for tree"
;
; Lcl frame size = 32

G_M48716_IG01:  ;; offset=0x0000
       push     rbx
       sub      rsp, 32
						;; size=5 bbWeight=1 PerfScore 1.25
G_M48716_IG02:  ;; offset=0x0005
       mov      rax, qword ptr [rcx]
       movzx    rbx, word  ptr [rax]
       test     byte  ptr [(reloc 0x7ffd7b16899a)], 1      ; global ptr
       je       SHORT G_M48716_IG05
						;; size=15 bbWeight=1 PerfScore 8.00
G_M48716_IG03:  ;; offset=0x0014
       movzx    rax, word  ptr [(reloc 0x7ffd7b1689a0)]      ; static handle
       cmp      ebx, eax
       setne    al
       movzx    rax, al
						;; size=15 bbWeight=1 PerfScore 3.50
G_M48716_IG04:  ;; offset=0x0023
       add      rsp, 32
       pop      rbx
       ret      
						;; size=6 bbWeight=1 PerfScore 1.75
G_M48716_IG05:  ;; offset=0x0029
       mov      rcx, 0x7FFD7B168968
       mov      edx, 2
       call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
       jmp      SHORT G_M48716_IG03
						;; size=22 bbWeight=0 PerfScore 0.00

; Total bytes of code 63, prolog size 5, PerfScore 14.50, instruction count 17, allocated bytes for code 63 (MethodHash=ce0a41b3) for method DisasmoPlayground.Resp:EqNewline_StaticGet(byref):ubyte (FullOpts)
; ============================================================
```

libs/server/Resp/RespServerSession.cs
```diff
@@ -210,13 +211,33 @@ public override int TryConsumeMessages(byte* reqBuffer, int bytesReceived)
...
+                    ProcessMessages();
+                }
+                catch (RespParsingException ex)
+                {
+                    logger?.LogCritical($"Aborting open session due to RESP parsing error: {ex.Message}");
+                    logger?.LogDebug(ex, "RespParsingException in ProcessMessages:");
+
+                    // Forward parsing error as RESP error
+                    var resp = new ReadOnlySpan<byte>(Encoding.ASCII.GetBytes($"ERR Protocol Error: {ex.Message}"));
+                    while (!RespWriteUtils.WriteError(resp, ref dcurr, dend))
```
PaulusParssinen, line 226: nit: does the UTF16 -> ASCII encoding directly on the buffer
```suggestion
                    while (!RespWriteUtils.WriteError($"ERR Protocol Error: {ex.Message}", ref dcurr, dend))
```

libs/common/RespReadUtils.cs
PaulusParssinen: This might actually be just be blamed on Disasmo. There might depend on timing of the static ctor..

PaulusParssinen: > on net8, all three approaches are basically identical (within measurability); the only delta is on down-level TFMs without modern JIT - on net48, for example, static readonly fields are the clear winner, and memorymarshal read is the clear loser: https://gist.github.com/mgravell/0f9306681f3ae6bbff2ee4891783f303

Ah, thanks for benchmarks! Seems like the rule to always measure is right in different way this time 😄  _Hopefully_ Garnet will never ship parts of it to legacy .NET.

badrishc: We have recently been converting large swaths of Garnet protocol code to use `MemoryMarshal.Read` which improves code readability compared to the random number comparisons. We are only supporting .NET 6 and 8 for now, but this result means that if we ever support older .NET then we would need to revisit this choice.

libs/common/RespReadUtils.cs
lmaas: Good catch! You are right, client messages should never include integers using this syntax and integer reads are only used internally by some of the bitmap operations. Those reads probably don't need to be explicitly checked for overflows and RESP syntax errors, but given the performance impact seems acceptable I think it's good to keep the checks for internal reads as well.

libs/common/Parsing/RespParsingException.cs
lmaas: I think this would make the error messages more readable. We just need to be careful to escape control characters.

libs/common/RespReadUtils.cs
lmaas: I agree that it would be nice to use `checked` for this. However, the problem that I saw with using `checked` is that we need to convert the exception into a `RespParsingException`, so that we know to forward the error to the client. But catching and rethrowing is prohibitively expensive. The check above should be relatively close in performance to using `checked` for valid 20-digit ulong values.

libs/common/RespReadUtils.cs
lmaas: That's a good idea! We are doing something similar for fast-parsing array headers of common commands. At a minimum we can bypass the checks for the length header by just checking for '$1\r\n'.

libs/common/Parsing/RespParsingException.cs
lmaas: This has been addressed with the last commits.

libs/common/RespReadUtils.cs
lmaas: I have added a fast path for the common case '**$1\r\n**'. Performance improvements are definitely visible.

<details>
<summary>Performance impact</summary><p>

| Method         | Job    | EnvironmentVariables | Runtime  | testCase    | Mean     | Error     | StdDev    |
|--------------- |------- |--------------------- |--------- |------------ |---------:|----------:|----------:|
| ReadBoolHeader (old) | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n | 7.982 ns | 0.0343 ns | 0.0321 ns |
| ReadBoolHeader (old) | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n | 6.795 ns | 0.0017 ns | 0.0013 ns |
| ReadBoolHeader (old) | .NET 6 | Empty                | .NET 6.0 | $1\r\n1\r\n | 7.954 ns | 0.0279 ns | 0.0261 ns |
| ReadBoolHeader (old) | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n1\r\n | 6.797 ns | 0.0034 ns | 0.0032 ns |
| ReadBoolHeader (new) | .NET 6 | Empty                | .NET 6.0 | $1\r\n0\r\n | 7.085 ns | 0.0088 ns | 0.0073 ns |
| ReadBoolHeader (new)| .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n0\r\n | 4.156 ns | 0.0018 ns | 0.0017 ns |
| ReadBoolHeader (new)| .NET 6 | Empty                | .NET 6.0 | $1\r\n1\r\n | 7.082 ns | 0.0046 ns | 0.0041 ns |
| ReadBoolHeader (new)| .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | $1\r\n1\r\n | 4.163 ns | 0.0059 ns | 0.0050 ns |

</p>
</details>

libs/server/Resp/RespServerSession.cs
lmaas: Resolved with the latest cleanup commit. Good catch as always! :)

libs/common/Parsing/RespParsingException.cs
lmaas: Good catch! This has been addressed in the latest cleanup commit.

libs/common/Parsing/RespParsingException.cs
lmaas: This has been addressed in the latest cleanup commit.

libs/common/RespReadUtils.cs
lmaas: This is a super interesting discussion. Using `MemoryMarshal.Read` definitely results in much more readable and maintainable code, so I would suggest we keep using it for RESP parsing in the main parser loop.

We could still hard-code common checks such as `CrLf` or wrap those checks in separate inline functions, so that we could easily change the implementation in the future, if we believe we'll ever need to support older .net versions.

libs/common/RespReadUtils.cs
badrishc: Why convert to `RespParsingException`, if we can instead catch that specific exception in the outer block?

lmaas: The connection will be closed for any exception regardless of its type. However, the type of exception determines how we are handling the response to the client. In the parsing case we would like to output as a protocol error to the user - but if an overflow happens anywhere outside of parsing this should be logged as a critical error on the server side only.

Of course, we could technically write logic to match the `OverflowException` back to the parsing functions - but that's a bad design for many many reasons. We would also lose important contextual information to output meaningful error messages. E.g., this is what this PR will throw:

`-ERR Protocol Error: Unable to parse integer. The given number is larger than allowed: 3000000000000000000`

lmaas: As for readability, maybe it would be useful to add a comment to remind the reader that UINT64 Max is 1844674407370955161**5**. Any opinions on this?

libs/common/RespReadUtils.cs
PaulusParssinen: I'm still with @mgravell here and really think comparing against a field called ``CrLf`` or ``Newline`` is more readable than MM.Read + RVA magic that is more expensive for JIT and potentially eats away from its tight budget to do other optimizations. But the difference is so miniscule so either is fine.

libs/common/RespReadUtils.cs
badrishc: We don't use 'checked' anywhere else in the codebase, and can keep it that way. This we could catch overflow exception and kill the connection. All assuming that is indeed more efficient ofc.

badrishc: I would think overflow anywhere in the call path should kill the connection anyway. I can't think of a place where overflow could occur and we would want to continue processing.

libs/common/RespReadUtils.cs
badrishc: Agreed on this. Shouldn't waste the JIT budget unless we get good returns.

libs/common/RespReadUtils.cs
lmaas: `OverflowException` can be thrown in other contexts as well. So, even if we don't use `checked` anywhere else, that does not guarantee an `OverflowException` caught in `TryConsumeMessages()` is a parsing exception that should be forwarded to the client.

Personally, I am not sure it's worth giving up clean error handling for the gains we are getting from using checked (keep in mind this is also just in isolation).  See below for a quick benchmark run.

<details>
<summary>TryReadLong: checked vs RespParsingException</summary><p>

| Method              | Job    | EnvironmentVariables | Runtime  | testCase               | Mean      | Error     | StdDev    | Median    |
|-------------------- |------- |--------------------- |--------- |----------------------- |----------:|----------:|----------:|----------:|
| TryReadUlong        | .NET 6 | Empty                | .NET 6.0 | 0\r\n                  |  2.101 ns | 0.0026 ns | 0.0024 ns |  2.101 ns |
| TryReadUlongChecked | .NET 6 | Empty                | .NET 6.0 | 0\r\n                  |  1.656 ns | 0.0034 ns | 0.0030 ns |  1.655 ns |
| TryReadUlong        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 0\r\n                  |  2.124 ns | 0.0016 ns | 0.0014 ns |  2.123 ns |
| TryReadUlongChecked | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 0\r\n                  |  1.955 ns | 0.0030 ns | 0.0024 ns |  1.954 ns |
| TryReadUlong        | .NET 6 | Empty                | .NET 6.0 | 18446(...)615\r\n [22] | 13.097 ns | 0.0076 ns | 0.0067 ns | 13.096 ns |
| TryReadUlongChecked | .NET 6 | Empty                | .NET 6.0 | 18446(...)615\r\n [22] | 12.247 ns | 0.0059 ns | 0.0052 ns | 12.247 ns |
| TryReadUlong        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 18446(...)615\r\n [22] | 12.823 ns | 0.0064 ns | 0.0060 ns | 12.825 ns |
| TryReadUlongChecked | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 18446(...)615\r\n [22] | 12.145 ns | 0.0096 ns | 0.0090 ns | 12.144 ns |
| TryReadUlong        | .NET 6 | Empty                | .NET 6.0 | 2147483647\r\n         |  7.475 ns | 0.0448 ns | 0.0419 ns |  7.473 ns |
| TryReadUlongChecked | .NET 6 | Empty                | .NET 6.0 | 2147483647\r\n         |  7.444 ns | 0.0703 ns | 0.0658 ns |  7.427 ns |
| TryReadUlong        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 2147483647\r\n         |  7.223 ns | 0.0499 ns | 0.0442 ns |  7.249 ns |
| TryReadUlongChecked | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 2147483647\r\n         |  6.320 ns | 0.1547 ns | 0.3089 ns |  6.167 ns |
</p>
</details>

mgravell:
On the magic numbers vs readability topic vs performance topic; the
benchmark shows that on net8+ the inline read looks fine for perf, but if
you wanted to generalize without compromising, a
suggestion:

static readonly uint MeaningfulNameHere = Utils.CpuEndianUInt32("$1\r\n"u8);

(etc for different widths)

where the method would assert the length, do a coerced read, and reverse
endianness if needed, without using random numbers in the code.

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+        /// <param name="ptr">Pointer to the beginning of the ASCII encoded input string.</param>
+        /// <param name="end">The end of the string to parse.</param>
+        /// <param name="value">If parsing was successful, contains the parsed ulong value.</param>
+        /// <param name="bytesRead">If parsing was successful, contains the number of bytes that were parsed.</param>
+        /// <returns>
+        /// True if a ulong was successfully parsed, false if the input string did not start with
+        /// a valid integer or the end of the string was reached before finishing parsing.
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool TryReadUlong(ref byte* ptr, byte* end, out ulong value, out ulong bytesRead)
```
PaulusParssinen, line 44: Big fan of the``TryRead`` prefix pattern 👍 

Can we do it for the other overloads too that return ``bool`` and have a ``byref`` result, that is ``out T value``?

PaulusParssinen (approved):
LGTM :shipit: A lot of general goodness done here and I'm glad that people are contributing their BDN benchmarks to the repo!

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+            }
+
+            // Special case: '$-1' (NULL value)
+            if (negative)
+            {
+                if (readHead + 4 > end)
+                {
                     return false;
+                }
+                if (*(uint*)readHead == MemoryMarshal.Read<uint>("-1\r\n"u8))
```
vazois, line 243: Are we expecting a command to have a null length header?
Will the following input result in a crash?
*2\r\n$3\r\nGET\r\n$-1\r\nfoo\r\n.

test/Garnet.test/RespTests.cs
```diff
@@ -655,7 +655,7 @@ public void SimpleIncrementInvalidValue(RespCommand cmd, bool initialize)
         {
             using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
             var db = redis.GetDatabase(0);
-            string[] values = ["", "7 3", "02+(34", "笑い男", "01", "-01", "7ab"];
+            string[] values = ["7 3", "02+(34", "笑い男", "01", "-01", "7ab"];
```
vazois, line 658: Don't we allow empty strings as values?

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+        /// <param name="negative">Whether the sign is '-'.</param>
+        /// <returns>True if either '+' or '-' was found, false otherwise.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool TryReadSign(byte* ptr, out bool negative)
         {
-            len = -1;
-            if (ptr + 3 >= end)
-                return false;
+            negative = (*ptr == '-');
+            return negative || (*ptr == '+');
```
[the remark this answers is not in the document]
lmaas: `+` is an optional prefix for RESP2 integers (see [here](https://redis.io/docs/latest/develop/reference/protocol-spec/#integers)). While integers are currently rarely read inside Garnet and we could probably get away without parsing for `+`, I recommend including the `+` prefix for robustness.

libs/common/RespReadUtils.cs
lmaas: I assume you are asking if the NULL value should only be parsed for commands that opt in, right? I think that's a very good question. `$-1` should not be allowed for GET, but currently will not result in an error message. If we make NULL opt-in, we could return an `InvalidStringLength` exception. Any thoughts on this?

test/Garnet.test/RespTests.cs
lmaas: While the array is called `values`, these values are also used as keys and empty strings are not valid keys. However, because this PR forces the length check for key reading to happen during parsing this is not a specific test case for the increment function anymore.

libs/common/RespReadUtils.cs
badrishc: Which other context? It only gets thrown in `checked` and we don't use that anywhere else.

libs/common/RespReadUtils.cs
lmaas: I fully agree! I think we should have a separate cleanup PR to unify the naming of all parsing related functions. As for BCL naming, I personally find BCL naming clearer for integers, but we should be consistent across the whole codebase.

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+                readHead++;
             }
 
-            ptr += 2;
-            if (ptr > end)
-                return false;
-            Debug.Assert(*(ptr - 2) == '\r');
-            Debug.Assert(*(ptr - 1) == '\n');
-            len = neg ? -ksize : ksize;
+            // Parse remaining digits, while checking for overflows.
```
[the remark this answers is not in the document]
lmaas: Exactly! The current code allows leading zeros and therefore inputs can be more than twenty digits. We just loop on the digits remaining digits, but in the normal case we will just exit this loop right away (either because we don't have any remaining digits, or because the string did not have leading zeros, and this is the last digit).

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool TryReadUlong(ref byte* ptr, byte* end, out ulong value, out ulong bytesRead)
+        {
+            bytesRead = 0;
+            value = 0;
+            var readHead = ptr;
+
+            // Fast path for the first 19 digits.
+            // NOTE: UINT64 overflows can only happen on digit 20 or later (if integer contains leading zeros).
```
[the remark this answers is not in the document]
lmaas: I don't think it'll simplify the code significantly unless there is a way to check for the overflow after it has occurred. If you have any ideas for this, please let me know.

The reason I split the loop is to reduce the number of checks for integer overflows (or checks for whether we need to check). But if we have a 20-digit number we will always have to check for an overflow on the last digit (NOTE: We do not know how long the number is until we are done parsing - so we cannot use that for an initial shortcut).

libs/common/RespReadUtils.cs
lmaas: What I meant is that it can be thrown by functions that use `checked` internally. For example, `Convert.ToInt64()` can throw an `OverflowException` as well.

libs/common/RespReadUtils.cs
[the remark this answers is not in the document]
vazois: I get your point, but I was thinking that since we expect to read a ulong, we should eliminate the second loop because we can have at most 20 digits/bytes.
First read up to 19 digits/bytes (as it is done now), then check 20th digit/byte for overflow. If 21st byte is stil a digit then fail.
I might be missing something here, then semantics of this method are slightly confusing.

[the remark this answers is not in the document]
lmaas: See [below](https://github.com/microsoft/garnet/pull/332#discussion_r1583923524). Using a loop for the last digit allows us to support leading zeros for robustness and doesn't come at a visible performance cost.

libs/common/RespReadUtils.cs
badrishc: sounds good, resolving this conversation.

badrishc (approved):

libs/common/RespReadUtils.cs
```diff
@@ -4,156 +4,372 @@
...
+            var fastPathEnd = ptr + 19;
+            while (readHead < fastPathEnd)
             {
-                Debug.Assert(*ptr >= '0' && *ptr <= '9');
-                ksize = ksize * 10 + *ptr++ - '0';
-                if (ptr >= end)
+                if (readHead > end)
```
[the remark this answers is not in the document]
lmaas: I was considering combining the check with `fastPathEnd` above. However, this will hit short numbers at the expense of faster parsing for long numbers. In the end, I don't think the trade-off is worth it, given most numbers probably only have 2 or 3 digits. We can revisit this later, if it becomes a problem.

test/Garnet.test/RespTests.cs
lmaas: It seems Redis allows the empty string as a key. I have restored this behavior, which means `""` is now a valid test case.

libs/common/RespReadUtils.cs
lmaas: I am resolving this conversation for now. We should consider this issue in the next cleanup for the parsing logic. In particular, for readability it would be nice to just have a function of the form `Expect(NewLine)` (or similar) that does the check using a static field and handles any potential errors.

PR #333, opened 2024-04-26
Cleanup of the release pipeline:  removed a task that didn't work and a switch in Nuget Pack that wasn't used 

darrenge:
This is what we found when testing csproj PR.  The "Copy Files for Nuget package ..." task and the "$(BuildConfiguration)" setting in the pack task are not needed and were removed.  In both cases it is because they used the variable $(BuildConfiguration) and it isn't used anywhere else.

### DIFF
--- a/.azure/pipelines/azure-pipelines-external-release.yml
+++ b/.azure/pipelines/azure-pipelines-external-release.yml
@@ -158,8 +158,7 @@ jobs:
 
   - task: GitHubRelease@0
     displayName: 'Create the GitHub release'
-# Debug - set to True after done testing
-    enabled: False
+    enabled: True
     inputs:
       action: 'create'
       gitHubConnection: ADO_to_Github_ServiceConnection
@@ -184,8 +183,7 @@ jobs:
 
   - task: NuGetCommand@2
     displayName: 'Push to NuGet.org'
-# Debug - set to True after done testing
-    enabled: False
+    enabled: True
     inputs:
       command: push
       packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg'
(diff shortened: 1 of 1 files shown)


.azure/pipelines/azure-pipelines-external-release.yml
```diff
@@ -104,19 +104,12 @@ jobs:
...
-    inputs:
-      Contents: '**/bin/AnyCPU/$(BuildConfiguration)/**/*'
-      TargetFolder: $(build.artifactstagingdirectory)
-
   - task: NuGetCommand@2
     displayName: nuget pack Garnet
     enabled: True
     inputs:
       command: custom
-      arguments: pack Garnet.nuspec -OutputDirectory $(Build.ArtifactStagingDirectory) -Properties Configuration=$(BuildConfiguration) -Symbols -SymbolPackageFormat snupkg -version $(Build.BuildNumber) -Verbosity Detailed
```
badrishc, line 119: we need to verify that the nupkg produced does indeed use Release build, and they are signed (by checking the contents of the .nupkg).

darrenge: 1) Confirmed DLLs in nuget package are signed - looking at properties of file
2) Verified files are release and double checked output of build:
MSBuild version 17.9.8+b34f75857 for .NET
  Determining projects to restore...
  Restored /home/vsts/work/1/s/libs/common/Garnet.common.csproj (in 1.49 sec).
  Restored /home/vsts/work/1/s/libs/client/Garnet.client.csproj (in 1.49 sec).
  Restored /home/vsts/work/1/s/metrics/HdrHistogram/HdrHistogram.csproj (in 5 ms).
  Restored /home/vsts/work/1/s/libs/storage/Tsavorite/cs/src/core/Tsavorite.core.csproj (in 5 ms).
  HdrHistogram -> /home/vsts/work/1/s/metrics/HdrHistogram/bin/Release/net8.0/HdrHistogram.dll
etc
etc

badrishc (approved):

Issue #180, opened 2024-03-28
API Coverage - Implement SINTER and SINTERSTORE


TalZaccai:
Here's your opportunity to implement a couple of RESP commands in Garnet! 

## SINTER

**Syntax:** 
`SINTER key [key ...]`

**Command description:**
Returns the members of the set resulting from the intersection of all the given sets.

For example:

```
key1 = {a,b,c,d}
key2 = {c}
key3 = {a,c,e}
SINTER key1 key2 key3 = {c}
```

Keys that do not exist are considered to be empty sets. With one of the keys being an empty set, the resulting set is also empty (since set intersection with an empty set always results in an empty set).

**Response:**

_Array reply:_ a list with members of the resulting set.

## SINTERSTORE

**Syntax:** 
`SINTERSTORE destination key [key ...]`

**Command description:**
This command is equal to `SINTER`, but instead of returning the resulting set, it is stored in `destination`.

If `destination` already exists, it is overwritten.

**Response:**

_Integer reply_: the number of elements in the resulting set.

---

**How to approach this task:**
1. Add the new command info to the setCommandsInfoMap in the RespCommandsInfo class _(Garnet.server/Resp/RespCommandsInfo.cs)_
2. Add the new command string to the returned HashSet in the RespInfo.GetCommands method _(Garnet.server/Resp/RespInfo.cs)_
3. Add the appropriate fast parsing logic for the new command in the RespCommand.FastParseArrayCommand method (use other commands as reference) _(Garnet.server/Resp/RespCommand.cs)_
4. Implement the wrapper method to the storage layer in StorageSession _(Garnet.server/Storage/Session/ObjectStore/SetOps.cs)_
5. Define a new method in IGarnetAPI and implement it in GarnetApiObjectCommands, calling the method you have implemented in step #4 _(Garnet.server/API/IGarnetAPI.cs, Garnet.server/API/GarnetApiObjectCommands.cs)_
6. Add a new method to the RespServerSession class which will call the storage API and return the appropriate response _(Garnet.server/Resp/Objects/SetCommands.cs)_
8. Add a new method to SetObjImpl which will perform the required operation on the hash object _(Garnet.server/Objects/Set/SetObjectImpl.cs)_
9. Add a new enum value to SetOperation and add the appropriate case to the switch in SetObject.Operate, in which you will call the method defined in step #8 _(Garnet.server/Objects/Set/SetObject.cs)_
[...]

PR #334, opened 2024-04-26
API Coverage - Implement SINTER and SINTERSTORE (#180)

lapkarol:


lapkarol:

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -95,6 +95,117 @@ private unsafe bool SetAdd<TGarnetApi>(int count, byte* ptr, ref TGarnetApi stor
...
+            {
+                var bufSpan = new ReadOnlySpan<byte>(recvBufferPtr, bytesRead);
+                if (!DrainCommands(bufSpan, count)) return false;
+                return true;
+            }
+
+            storageApi.SetIntersect(keys, out var result);
+
+            // write the size of result
+            var resultCount = result.Count;
```
TalZaccai, line 135: Verify that the status returned by SetIntersect is GarnetStatus.OK
Also verify that result is not null here

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -422,6 +422,144 @@ internal unsafe GarnetStatus SetMove(ArgSlice sourceKey, ArgSlice destinationKey
...
+        /// <summary>
+        /// Returns the members of the set resulting from the intersection of all the given sets.
+        /// Keys that do not exist are considered to be empty sets.
+        /// </summary>
+        /// <param name="keys"></param>
+        /// <param name="output"></param>
+        /// <returns></returns>
+        public GarnetStatus SetIntersect(ArgSlice[] keys, out HashSet<byte[]> output)
+        {
+            output = new HashSet<byte[]>(new ByteArrayComparer());
```
TalZaccai, line 435: No need to instantiate a HashSet here, just set it to default

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -422,6 +422,144 @@ internal unsafe GarnetStatus SetMove(ArgSlice sourceKey, ArgSlice destinationKey
...
+            for (var i = 1; i < keys.Length; i++)
+            {
+                // intersection of anything with empty set is empty set
+                if (result.Count == 0)
+                {
+                    return result;
+                }
+
+                status = GET(keys[i].ToArray(), out var next, ref objectContext);
+                if (status == GarnetStatus.OK)
```
TalZaccai, line 551: If a key is not found, its value is being treated as an empty set, so you should check for GarnetStatus.NOTFOUND and then just return an empty set

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -422,6 +422,144 @@ internal unsafe GarnetStatus SetMove(ArgSlice sourceKey, ArgSlice destinationKey
...
+                {
+                    if (next.garnetObject is SetObject nextObject)
```
TalZaccai, line 553: If you encounter an object that is not a SetObject you should return an error (CmdStrings.RESP_ERR_WRONG_TYPE)

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -422,6 +422,144 @@ internal unsafe GarnetStatus SetMove(ArgSlice sourceKey, ArgSlice destinationKey
...
+        /// <summary>
+        /// This command is equal to SINTER, but instead of returning the resulting set, it is stored in destination.
+        /// If destination already exists, it is overwritten.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="keys"></param>
+        /// <param name="count"></param>
+        /// <returns></returns>
+        public GarnetStatus SetIntersectStore(byte[] key, ArgSlice[] keys, out int count)
+        {
```
TalZaccai, line 476: If keys.Length == 0 you can just return, since you don't need to intersect the set at key with anything

test/Garnet.test/RespSetTest.cs
```diff
@@ -351,6 +351,35 @@ public void CanDoSetUnionStore(string key)
             Assert.IsTrue(expectedResult.OrderBy(t => t).SequenceEqual(strResult.OrderBy(t => t)));
         }
 
+        [Test]
+        public void CanDoSetInterStore()
```
TalZaccai, line 355: Please add a test for SINTER as well (that uses SE.Redis)

test/Garnet.test/RespSetTest.cs
```diff
@@ -351,6 +351,35 @@ public void CanDoSetUnionStore(string key)
...
+            var db = redis.GetDatabase(0);
+
+            var key1 = "key1";
+            var key1Value = new RedisValue[] { "a", "b", "c" };
+
+            var key2 = "key2";
+            var key2Value = new RedisValue[] { "c", "d", "e" };
+
+            var addResult = db.SetAdd(key1, key1Value);
+            Assert.AreEqual(3, addResult);
```
TalZaccai, line 368: Probably better to use key*value.Length instead of 3, in case the values change

test/Garnet.test/RespSetTest.cs
```diff
@@ -351,6 +351,35 @@ public void CanDoSetUnionStore(string key)
...
+
+            var result = (int)db.Execute("SINTERSTORE", key, key1, key2);
+            Assert.AreEqual(1, result);
+
+            var membersResult = db.SetMembers(key);
+            Assert.AreEqual(1, membersResult.Length);
+            var strResult = membersResult.Select(m => m.ToString()).ToArray();
+            var expectedResult = new[] { "c" };
+            Assert.IsTrue(expectedResult.SequenceEqual(strResult));
+        }
```
TalZaccai, line 380: Other good test cases to add:
- Check that if `key` exists, it is being overridden
- Check that if `key1` or `key2` don't exist, the returned set is empty
- Check that if no keys are given we get a "wrong number of arguments" error
- Check that if `key1` or `key2` don't hold a set object, we get a "WRONGTYPE" error

test/Garnet.test/RespSetTest.cs
```diff
@@ -910,6 +939,36 @@ public void CanDoSdiffLC()
...
+
+
+        [Test]
+        public void CanDoSinterStoreLC()
+        {
+            var lightClientRequest = TestUtils.CreateRequest();
+            lightClientRequest.SendCommand("SADD key1 a b c d");
+            lightClientRequest.SendCommand("SADD key2 c");
+            lightClientRequest.SendCommand("SADD key3 a c e");
+            var response = lightClientRequest.SendCommand("SINTERSTORE key key1 key2");
```
TalZaccai, line 963: I think you forgot key3 here :)

libs/server/Resp/RespInfo.cs
```diff
@@ -38,7 +38,7 @@ public static HashSet<string> GetCommands()
                 // Pub/sub
                 "PUBLISH", "SUBSCRIBE", "PSUBSCRIBE", "UNSUBSCRIBE", "PUNSUBSCRIBE",
                 // Set
-                "SADD", "SREM", "SPOP", "SMEMBERS", "SCARD", "SSCAN", "SRANDMEMBER", "SISMEMBER", "SUNION", "SUNIONSTORE", "SDIFF", "SDIFFSTORE", "SMOVE",
+                "SADD", "SREM", "SPOP", "SMEMBERS", "SCARD", "SSCAN", "SRANDMEMBER", "SISMEMBER", "SUNION", "SUNIONSTORE", "SDIFF", "SDIFFSTORE", "SMOVE", "SINTER", "SINTERSTORE",
```
TalZaccai, line 41: General request:
If you can add SINTER and SINTERSTORE to the list of supported commands on the website, that would be great! :)
Look at: website/docs/commands/api-compatibility.md and website/docs/commands/data-structures.md
Much appreciated!

TalZaccai (changes requested):
Thank you very much for your contribution to Garnet!
Added a few comments on your PR, but great work overall!

libs/server/Storage/Session/ObjectStore/SetOps.cs
lapkarol: As far as I can see, at this point there is no easy way to implement this. There is a separate but not yet merged branch in which there is initial support for RESP_ERR_WRONG_TYPE
branch name :obj-store-wrongtype

test/Garnet.test/RespSetTest.cs
lapkarol: Tests added without last point due to lack of WRONGTYPE error support

TalZaccai (approved):
Sorry for taking a while reviewing this one, I was out last week... Looks good to me! :)

PR #341, opened 2024-04-27
Seal more Tsavorite internals

PaulusParssinen:
* Good type hygiene & potential minor perf. benefits

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/Device/AsyncPool.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Device/AsyncPool.cs
@@ -13,7 +13,7 @@ namespace Tsavorite.core
     /// Supports sync get (TryGet) for fast path
     /// </summary>
     /// <typeparam name="T"></typeparam>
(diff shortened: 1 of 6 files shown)


libs/storage/Tsavorite/cs/src/core/Device/AsyncPool.cs
```diff
@@ -13,7 +13,7 @@ namespace Tsavorite.core
     /// Supports sync get (TryGet) for fast path
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public class AsyncPool<T> : IDisposable where T : IDisposable
+    internal sealed class AsyncPool<T> : IDisposable where T : IDisposable
```
PaulusParssinen, line 16: Made this implementation detail ``internal``. 

I had quick stab at trying to make this use ``System.Threading.Channels`` but I think I'll try again when/if I have better understanding of the complex nature of the sync & async APIs of Tsavorite.. 😅

TedHartMS (approved):

Issue #350, opened 2024-05-02
CLUSTERGETKEYSINSLOT continues to return the same keys even after they've been migrated


priyanjgupta:
### Describe the bug

CLUSTERGETKEYSINSLOT returns an array of keys stored in the contacted node which are mapped to the specified hash slot. But when the node is in MIGRATING state, if some keys have migrated and the CLUSTERGETKEYSINSLOT function is called again, it returns the same keys, including those that have been migrated.

### Steps to reproduce the bug

Repro:
1. Create a cluster comprising of two nodes, N1 and N2.
2. Add some slots to N1 
3. Add keys mapped to a slot X served by N1
4. Set the target node N2 in importing state to import slot X from node N1
5. Set the source node N1 in migrating state to migrate slot X to node N2
6. Call ClusterGetKeysInSlot for slot X on N1
7. Migrate an existing key, K1, mapped to slot X from N1 to N2
8.  Call ClusterGetKeysInSlot for slot X on N1
9.  K1, along with any other existing keys, will be included in the returned result.

### Expected behavior

Following the migration of key K1 from the source node, when invoking ClusterGetKeysInSlot, the returned result should not include the migrated key K1.

### Screenshots

![image](https://github.com/microsoft/garnet/assets/137490230/658f18dc-54c0-4a0a-95c2-f9c9b03268cd)

PR #352, opened 2024-05-02
Fix: Delete keys locally if  _copyOption is set to false 

priyanjaligupta:
The PR resolves issue #350


libs/cluster/Server/Migration/MigrateSessionSlots.cs
```diff
@@ -56,7 +56,7 @@ private bool MigrateSlotsDataDriver()
         /// </summary>
         public void DeleteKeysInSlot()
```
vazois, line 57: nit; Can you update the summary message of the method?

vazois (changes requested):
Good catch!

priyanjaligupta:

PR #359, opened 2024-05-05
Enforce object store total memory size constraint during recovery

yrajas:
**Problem**
Whenever object store operations are performed, corresponding heap size impact is computed and tracked to ensure that the total memory size is within the user configured limit. However, this constraint is not currently enforced during recovery. This change addresses that.

**Fix description**
* Ability to register a deserialization observer, `OnDeserializationObserver`, that can compute and use the `CacheSizeTracker` to track heap usage when pages are read in during recovery. `LogOperationObserver` is added to perform this.
* During recovery, if total object store size needs to be tracked, the log memory is not fully saturated by reading pages from disk rather done sequentially.
* If reading in a page causes the memory limit to be breached, earlier pages are evicted `EvictPage` one at a time until the total memory size is back within the limit.
* If pages were freed to constrain memory limits, `DoPostRecovery` updates `headAddress` to account for the freed pages.


libs/storage/Tsavorite/cs/src/core/Index/Recovery/Recovery.cs
```diff
@@ -840,24 +919,29 @@ private void RecoverHybridLogFromSnapshotFile(long scanFromAddress, long recover
...
+        private async ValueTask<long> RecoverHybridLogFromSnapshotFileAsync(long scanFromAddress, long recoverFromAddress, long untilAddress, long snapshotStartAddress, long snapshotEndAddress, long nextVersion, Guid guid, RecoveryOptions options, DeltaLog deltaLog, long recoverTo, CancellationToken cancellationToken)
         {
+            long lastFreedPage = NoPageFreed;
             GetSnapshotPageRangesToRead(scanFromAddress, untilAddress, snapshotStartAddress, snapshotEndAddress, guid, out long startPage, out long endPage, out long snapshotEndPage, out int capacity, out var recoveryStatus, out int numPagesToReadPerIteration);
 
             for (long page = startPage; page < endPage; page += numPagesToReadPerIteration)
             {
-                await ReadPagesForRecoveryAsync(snapshotEndAddress, recoveryStatus, snapshotEndPage, capacity, numPagesToReadPerIteration, page, cancellationToken).ConfigureAwait(false);
+                var result = await ReadPagesForRecoveryAsync(snapshotEndAddress, recoveryStatus, snapshotEndPage, capacity, numPagesToReadPerIteration, page, cancellationToken).ConfigureAwait(false);
```
TedHartMS, line 932: 'result' was added but is not used; it does not appear needed in which case it can be '_'

TedHartMS (approved):

PR #362, opened 2024-05-06
Cluster Mode SELECT Command 

vazois:
Return error SELECT not supported with cluster mode

### DIFF
--- a/libs/server/Resp/ArrayCommands.cs
+++ b/libs/server/Resp/ArrayCommands.cs
@@ -668,15 +668,25 @@ namespace Garnet.server
 
             readHead = (int)(ptr - recvBufferPtr);
 
-            if (string.Equals(result, "0"))
+            if (storeWrapper.serverOptions.EnableCluster)
             {
-                while (!RespWriteUtils.WriteDirect(CmdStrings.RESP_OK, ref dcurr, dend))
+                // Cluster mode does not allow DBID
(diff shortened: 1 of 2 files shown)


libs/server/Resp/ArrayCommands.cs
```diff
@@ -668,15 +668,25 @@ private bool NetworkSELECT(byte* ptr)
...
+                // Cluster mode does not allow DBID
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_SELECT_CLUSTER_MODE, ref dcurr, dend))
                     SendAndReset();
             }
             else
             {
-                while (!RespWriteUtils.WriteError("ERR invalid database index."u8, ref dcurr, dend))
-                    SendAndReset();
+
+                if (!storeWrapper.serverOptions.EnableCluster && string.Equals(result, "0"))
```
badrishc, line 680: why do we need to check for EnableCluster here?

vazois: Currently Redis does not allow for multiple databases in Cluster mode. While awaiting for the work to add support for this, I thought it would be good to at least inform clients with an error message that for cluster mode this feature is not available.

vazois: Good catch, forgot to remove the second check. Removed it now!

badrishc (approved):

PR #368, opened 2024-05-08
Pass comparer when instantiating new result set in SortedSet diff operation

PaulusParssinen:
Missed in #344 

Also now avoids rehashing the keys on the fast-path (other comparisong operand is empty set) as the comparers match.

8pf8vl


libs/server/Objects/SortedSet/SortedSetObject.cs
```diff
@@ -353,9 +353,9 @@ public override unsafe void Scan(long start, out List<byte[]> items, out long cu
                 return [];
 
             if (dict2 == null)
-                return new Dictionary<byte[], double>(dict1);
+                return new Dictionary<byte[], double>(dict1, dict1.Comparer);
 
-            Dictionary<byte[], double> result = [];
+            var result = new Dictionary<byte[], double>(ByteArrayComparer.Instance);
```
PaulusParssinen, line 358:
```suggestion
            var result = new Dictionary<byte[], double>(dict1.Comparer);
```

badrishc (approved):

Issue #366, opened 2024-05-08
API Coverage - Lua scripting or workaround


KeithHenry:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

Garnet does not support Lua, which prevents it from being able to replace Redis.

However, we mainly use Lua to workaround issues in Redis, most significantly that `MSET` does not support expiry.

We had issues in production due to queued up `SETEX` when batch storing. Our workaround looks like this:

```c#
namespace ...
using StackExchange.Redis;

public static class RedisDatabaseExtension {

    /// <summary>Set a batch of keys to the string values with the given expiry.
    /// <para>MSET does not support setting expiry, but is better if values don't expire.</para>
    /// <para>EVAL a Lua script to call SETEX for every record in the batch.</para>
    /// <para>In any batches fail this will return false, and not execute further batches.</para>
    /// <seealso cref="https://redis.io/commands/eval/"/></summary>
    /// <param name="db">Redis DB to extend.</param>
    /// <param name="batchSize">Number of records in each batch.</param>
    /// <param name="expiresSeconds">Number of seconds to set as the expiry.</param>
    /// <param name="items">Collection of key-value pairs to set.</param>
    /// <returns>True if all scripts evaluated successfully, false if any failed to match the number of records input in a batch.</returns>
    public static async Task<bool> StringSetLuaBatch(
        this IDatabase db, 
        int batchSize, 
        int expiresSeconds,
        params (RedisKey key, RedisValue value)[] items) {

        // ARGV[1] is the expiry, so all the values are off-by-one compared to their keys
        const string lua = """
            local expiry = tonumber(ARGV[1])
            local count = 0
            for i, key in ipairs(KEYS) do
                redis.call('SETEX', key, expiry, ARGV[i+1])
                count = count + 1
            end
            return count
            """;

        List<RedisKey> keys = new(batchSize);
        List<RedisValue> values = new(batchSize + 1) { expiresSeconds };

        foreach(var (key, value) in items)
        {
            keys.Add(key);
            values.Add(value);

            if(keys.Count == batchSize)
            {
                if(keys.Count != (int) await db.ScriptEvaluateAsync(lua, [.. keys], [.. values]))
                    return false; // Count doesn't match, exit early with failure

                // Reset for next batch
                keys.Clear();
                values.Clear();
                values.Add(expiresSeconds);
            }
        }

        if (keys.Count != 0) // If any left over execute final batch
[...]

PR #369, opened 2024-05-08
Custom stored procedures with variable number of parameters

badrishc:
Support custom commands (stored procedures) with a variable number of parameters. This can be used to implement commands such as a custom `MSET` variation that takes expiration as an argument.

Example of `MSETPX` is included in the PR here: https://github.com/microsoft/garnet/blob/ecbddd2c09b196cb19cdb67005f7ccc92192c214/main/GarnetServer/Extensions/MSetPx.cs

Example usage: set 5 key-values pairs with an expiration of 60 seconds (60000 milliseconds):

`MSETPX 60000 key1 value1 key2 value2 key3 value3 key4 value4 key5 value5`

image

Contributes to https://github.com/microsoft/garnet/issues/366


yrajas (approved):

badrishc:
Added example for multi-get, where values match a specified prefix. Example:

```
127.0.0.1:3278> mset key1 com/value1 key2 org/value2 key3 edu/value3 key4 com/value4
OK
127.0.0.1:3278> mgetifpm com key1 key2 key3 key4
1) "key1"
2) "com/value1"
3) "key4"
4) "com/value4"
```

PR #370, opened 2024-05-08
BenchmarkDotNet snippet to compare Span<byte> to raw pointer

badrishc:


TedHartMS (approved):

benchmark/BDN.benchmark/SpanVsPointer.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+
+namespace BDN.benchmark
+{
+    public unsafe class SpanVsPointer
```
lmaas, line 9: I would suggest adding a comment that briefly summarizes the benchmarks. Something like this:

```suggestion
    /// <summary>
    /// Benchmarks to quantify the impact of using pointer vs. span accesses.
    /// </summary>
    public unsafe class SpanVsPointer
```

badrishc: perhaps in a later PR as this just merged :)

PaulusParssinen:
There's a lot more nuance in comparing them like this, not exactly fair comparison 😄

badrishc:
This was just to check the claim/hope that `Span<byte>` could actually even outperform pointers. It seems they are close, but the gap still exists. What other nuances do you have in mind for server code such as Garnet, where we can ensure that we are being careful with pointers?

PaulusParssinen:
If you _already_ have a pointer, it's of course generally better to stay in the pointer world. And as the team seems clearly comfortable in using them and if they feel intuitive, absolutely it's the choice for Garnet. Especially in the APIs that are not directly exposed to the library consumers.

The strength of `Span<T>` is really in its flexibility to represent any contiguous memory "slice". This means where library author previously needed to expose ``T* ptr, int length`` for their high-perf consumers and then ``T[] array, int start, int length`` in addition to the just ``T[] array``. All of these are now merged in one safe .NET construct (Long live [Midori](https://en.wikipedia.org/wiki/Midori_(operating_system))). This means that often you can just stay in the `(RO)Span<T>` world and delegate the work to the BCL's already vectorized workhorse methods. What this means is that _when_ we need a pointer, we can make the pinning __short lived as possible__ which "costs" only a stack spill (couple `mov`s. this can and will be improved: https://github.com/dotnet/runtime/issues/63397) and if no GC occurs during this pin, it's essentially free.

In Garnet's case, one has to consider the trade-offs of maintaining all the manual parsing logic where we think we can beat BCL. If something we do is measurably faster, why not contribute it upstream for all to benefit (yes, we can always make more assumptions than BCL, so we can be faster) and how to "allocate" the efforts.

I really want to have great benchmarking tooling in Garnet repo to quickly iterate and see real effects of performance changes against different workloads. Something along the lines of ASP.NET's [Crank](https://github.com/dotnet/crank) but configurable RESP workloads to see tangible RPS/Memory/Latency/CPU metrics. These don't need to be maximum theoretical special lab-hardware measurement, but something to give indication how much RPS/Latency we lose if, for example, we just used `Utf8Parser` & `Utf8Formatter` APIs instead of handrolling our own.

Sorry for the wall of text rambling😅

badrishc:
We are in vehement agreement here. In fact, the above BDN seems to suggest that even for this rather contrived case, `Span<byte>` is very close (if not sometimes equal) to `byte*` - as long as the user (Garnet code in our case) does not need to re-`fixed` the `Span<byte>` just to get a pointer because they need a pointer for some reason.

Issue #377, opened 2024-05-12
ACL support with Microsoft Entra tokens


msft-paddy14:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

Today the authentication methods include password and Microsoft Entra, however when paired with ACL, only passwords can be used. Rotation of passwords can be cumbersome and pose security risks while we manage them.
Azure Cache for Redis already supports Microsoft Entra+ ACL in non Enterprise tiers  - [Azure Redis](https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-azure-active-directory-for-authentication).

### Describe the solution you'd like

Based on this, I recommend a similar approach which at it's core will continue to use ACL based auth but will allow a configurable authenticator which can be ACL based. For parity with Azure Redis, it might make sense to use Oid as username with token as password.
```
User = Object ID of your managed identity or service principal
Password = Microsoft Entra token that you acquired using MSAL

```

At code level:
Currently ACL auth by default does username and password validation with ACL entries. We compose it with an IGarnetAuthenticator instance to inject authentication behavior of username and  password and then just validate the permissions against the ACL list of the user. This approach is more favorable as it minimizes changes and avoids redundant code needed to combine behaviors of AclAuthenticator and AADAuthenticator. Rather than inheriting these behavior, we compose AclAuthenticator with an IGarnetAuthenticator.

PR #378, opened 2024-05-12
ACL with Microsoft Entra token support

msft-paddy14:
This PR proposes changes for #377.
The changes are done 

- Currently ACL auth by default does username and password validation with ACL entries. As part of this PR I've added IAuthenticator instance to inject authentication behavior of username and  password and then just validate the permissions against the ACL list of the user rather than default behavior of password.
-  This approach favours composing the existing AclAuthenticator with AADAuthenticator rather than creating a new AclAadAuthenticator to keep the code clean and avoid class explosion

Testing:
There were no AAD tests. Have added a basic test to validate AAD + ACL and that it works with cluster auth.
I had to add support for a new constructor for testing with IssuerSigningToken  - it might be easier to put this behind an abstraction but didn't do it to avoid changing scope of this PR


libs/server/Auth/GarnetACLAuthenticator.cs
```diff
@@ -25,15 +25,19 @@ class GarnetACLAuthenticator : IGarnetAuthenticator
...
+
+        private IGarnetAuthenticator _authenticator;
+
         /// <summary>
         /// Initializes a new ACLAuthenticator instance.
         /// </summary>
         /// <param name="accessControlList">Access control list to authenticate against</param>
         /// <param name="logger">The logger to use</param>
-        public GarnetACLAuthenticator(AccessControlList accessControlList, ILogger logger)
+        public GarnetACLAuthenticator(AccessControlList accessControlList, ILogger logger, IGarnetAuthenticator wrapperAuthenticator = null)
```
vazois, line 36: Please always add ILogger as the last parameter. Also, don't forget to update the summary for the new parameter.
```suggestion
public GarnetACLAuthenticator(AccessControlList accessControlList, IGarnetAuthenticator wrapperAuthenticator = null, ILogger logger = null)
```

libs/server/Auth/GarnetACLAuthenticator.cs
```diff
@@ -66,6 +70,22 @@ public bool Authenticate(ReadOnlySpan<byte> password, ReadOnlySpan<byte> usernam
                 string uname = Encoding.ASCII.GetString(username);
                 User user = string.IsNullOrEmpty(uname) ? _acl.GetDefaultUser() : _acl.GetUser(uname);
 
+                // Use injected authenticator if configured.
+                if (_authenticator != null)
```
vazois, line 74: I am not very familiar with this part of the code but I think that this part does not look right? 
Aren't we supposed to override Authenticate based on the implementation of the authentication that is defined at startup?
Also, have you seen the existing implementation of GarnetAadAuthenticator? Maybe we need an extension of that implementation, instead of injecting it here?

vazois (changes requested):

libs/server/Auth/Aad/IssuerSigningTokenProvider.cs
```diff
@@ -31,15 +31,19 @@ public class IssuerSigningTokenProvider : IDisposable
 
         private readonly ILogger _logger;
 
-        private IssuerSigningTokenProvider(string authority, IReadOnlyCollection<SecurityKey> signingTokens, ILogger logger)
+        private IssuerSigningTokenProvider(string authority, IReadOnlyCollection<SecurityKey> signingTokens, ILogger logger, bool refreshTokens = true)
         {
             _authority = authority;
-            _refreshTimer = new Timer(RefreshSigningTokens, null, TimeSpan.Zero, TimeSpan.FromDays(1));
+            if (refreshTokens)
```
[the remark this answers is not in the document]
msft-paddy14: I created this one for test support since we kick off remote calls. There were no existing AAD tests to validate this.  Can see if maybe putting this behind an interface helps.

libs/server/Auth/AuthenticationSettings.cs
```diff
@@ -153,13 +158,27 @@ public AadAuthenticationSettings(string[] authorizedAppIds, string[] audiences,
             _signingTokenProvider = signingTokenProvier;
         }
 
+        public AadAuthenticationSettings WithUsernameValidation()
+        {
+            _validateUsername = true;
```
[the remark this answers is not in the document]
msft-paddy14: I'm trying to avoid having multiple configuration settings in the constructor and allowing flexibility incase we add more settings in future so preferring the Builder pattern here. Also modifying constructor breaks binray compatibility since it's a public class, I'm not sure if someone had a dependency on it

libs/server/Auth/GarnetACLAuthenticator.cs
msft-paddy14: So the way I see it - we're trying to extend authz with another form of authn rather than other way around.
I have added another option for this mode AclWithAad which configures the ACLAuthenticator with an IGarnetAuthenticator and short circuits the response if such a Authenticator exists. 
Currently, the codebase has checks like ``` if (_authenticator is GarnetACLAuthenticator aclAuthenticator)``` and  ```(_authenticator.GetType() != typeof(GarnetACLAuthenticator))```
casting logic making it dependent on the GarnetACLAuthenticator, adding a GarnetACLWithAADAuthenticator would break a lot of things. There's ACL refresh logic also as part of this which would need to be ported to other classes. So above challenges made a strong case to reuse existing GarnetACLAuthenticator.

Now we can inject a default multi pass authenticator to cover default logic of scanning multiple passwords behind an IGarnetAuthenticator too but again on ACL refresh, this set will change. As a result, I stuck with default implementation with a short circuit if an IGarnetAuthenticator is defined. Incase of ACLWithAAD we're validating we're not relying on ACL password so don't need a refresh on ACL as well (only users which are scoped to ACLAuthenticator). An alternate approach I can think of is that we use an IACLAuthenticator instance which exposes two methods ```Authenticate(username, password)  and Refresh(ACLList)``` to with implementations being AclPasswordAuthenticator and AclAADAuthenticator. ACLPasswordAuthenticator can have default logic of looking at all passwords whereas in AclAADAuthenticator we can keep this current logic.

vazois: The confusion is that you are treating GarnetACLAuthenticator as a GarnetAuthenticatorManager. 
I think it is a valid point that you want to use existing implementation, that is why I suggested creating an extension of GarnetACLAuthenticator.

``` csharp
    class GarnetACLWithAadAuthenticator : GarnetACLAuthenticator
    {
        public GarnetACLWithAadAuthenticator(AccessControlList accessControlList, ILogger logger) : base(accessControlList, logger)
        {

        }
```
It seems messy and hard to follow when interjecting another authenticator inside GarnetACLAuthenticator.
Your suggestion makes sense and I believe you don't need an IACLAuthenticator. You can add Refresh to IGarnetAuthenticator and just make it so that it throws unimplemented exception for any case that it is not needed.
I know it may be getting out of scope for the PR but I will let others chime in on how to proceed.
My vote is to address it now.

libs/server/Auth/AuthenticationSettings.cs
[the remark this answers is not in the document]
vazois: It should be fine if you add it with default value. So, you won't have to change it everywhere. I personally feel it is much easier to reason about.

libs/server/Auth/GarnetAadAuthenticator.cs
```diff
@@ -35,6 +45,7 @@ class GarnetAadAuthenticator : IGarnetAuthenticator
         private readonly IReadOnlyCollection<string> _audiences;
         private readonly IReadOnlyCollection<string> _issuers;
         private readonly IssuerSigningTokenProvider _signingTokenProvider;
+        private readonly bool ValidateUsername;
```
vazois, line 48: nit
```suggestion
private readonly bool _validateUsername;
```

libs/server/Auth/GarnetAadAuthenticator.cs
```diff
@@ -49,6 +60,17 @@ class GarnetAadAuthenticator : IGarnetAuthenticator
...
+            _logger = logger;
+        }
+
+        internal GarnetAadAuthenticator(AadValidationConfig aadValidationConfig, ILogger logger)
+        {
+            _authorizedAppIds = aadValidationConfig._authorizedAppIds;
+            _signingTokenProvider = aadValidationConfig._signingTokenProvider;
+            _audiences = aadValidationConfig._audiences;
+            _issuers = aadValidationConfig._issuers;
+            ValidateUsername = aadValidationConfig.ValidateUsername;
```
vazois, line 73: nit
```suggestion
_validateUsername = aadValidationConfig._validateUsername;
```

libs/server/Auth/GarnetAadAuthenticator.cs
```diff
@@ -14,13 +14,23 @@
 
 namespace Garnet.server.Auth
 {
+    internal class AadValidationConfig
+    {
+        internal IReadOnlyCollection<string> _authorizedAppIds { get; set; }
+        internal IReadOnlyCollection<string> _audiences { get; set; }
+        internal IReadOnlyCollection<string> _issuers { get; set; }
+        internal IssuerSigningTokenProvider _signingTokenProvider { get; set; }
+        internal bool ValidateUsername { get; set; }
```
vazois, line 23: nit
```suggestion
 internal bool _validateUsername { get; set; }
```

vazois (changes requested):

libs/server/Auth/AuthenticationSettings.cs
```diff
@@ -116,6 +120,7 @@ public class AadAuthenticationSettings : IAuthenticationSettings
         private readonly IReadOnlyCollection<string> _audiences;
         private readonly IReadOnlyCollection<string> _issuers;
         private IssuerSigningTokenProvider _signingTokenProvider;
+        private bool _validateUsername;
         private bool _disposed;
```
vazois, line 125: This is not the scope of this PR but this file is getting too big. We might need to refactor it into smaller files. If you are up to it please separate the AuthenticationSettings under Settings/ folder and then for each class have a separate file.

vazois (changes requested):

vazois (approved):
Great work!

Issue #380, opened 2024-05-13
100% mem and disk usage, when use "save" cmd


sun51586:
### Describe the bug

when use "save" cmd in Another Redis Desktop Manager, 100% mem(from 6.1G up to 15.1G) and 100% disk usage(write c:\windows\system32\store\checkpointxxx,c:\windows\system32\objectstore\xxx).
my config：
{
	"Port" : 6379,
	"MemorySize" : "1g",
	"LogDir": "c:\\temp\\_garnet\\",
	"EnableStorageTier": true,
	/*"EnableAOF": true*/
}
use nuget v1.0.8 ,in .net 8 woker service project, host in windows service.
Is that my config error or bug?

### Steps to reproduce the bug

1,start garnet service.the cache is empty.
2,open Another Redis Desktop Manager V1.5.2.
3,input cmd "set ka aaa" enter,"set kb bbb" enter,then input “save" enter。
4,garnet process write c:\windows\system32\stroe\checkpoint** crazy ,and memory up to the max.

### Release version

V1.0.8

### OS version

windows 10 pro 22h2

badrishc:
1. The default hash table size is unfortunately set to very high (`8g`) so this is an expensive flush operation. Please use `--index` to reduce the hash table size to a reasonable size such as `1m`. See [this link](https://microsoft.github.io/garnet/docs/getting-started/memory#index-size) for how to configure hash size for your workload. We will update Garnet to use a better default.
2. You have not specified checkpointdir so the checkpoint will get written in the default location. Suggest you use `--checkpointdir` to specify the checkpoint folder.

sun51586:
Very Thanks..I try it.

It would be even better to give a default configuration file for diffrent usage scenarios。

PR #381, opened 2024-05-14
Handle large writes in SendAndReset, adjust index default size

badrishc:
Network Sends
* Handle unexpected large writes in `SendAndReset` by throwing an exception so that the session can end gracefully
* Support larger echo responses by splitting the response through a `WriteDirectLarge` wrapper

Index Defaults
* Change default main store index size to `128m` (optimized for up to ~8m distinct keys)
* Change default object store index size to `16m` (optimized for up to ~1m distinct keys)
* Add logger informational message to indicate what the index is optimized for

Misc
* Set CheckpointDir to LogDir (or current directory if LogDir is null) if CheckpointDir is not directly provided.

Fixes https://github.com/microsoft/garnet/issues/380.

### DIFF
--- a/libs/common/GarnetException.cs
+++ b/libs/common/GarnetException.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Garnet.common
 {
@@ -33,5 +34,13 @@ namespace Garnet.common
         public GarnetException(string message, Exception innerException) : base(message, innerException)
         {
         }
+
+        /// <summary>
+        /// Throw helper that throws a GarnetException.
+        /// </summary>
+        /// <param name="message">Exception message.</param>
+        [DoesNotReturn]
+        public static void Throw(string message) =>
+            throw new GarnetException(message);
     }
 }
--- a/libs/server/Resp/RespServerSession.cs
+++ b/libs/server/Resp/RespServerSession.cs
@@ -231,8 +231,6 @@ namespace Garnet.server
                 }
                 recvBufferPtr = null;
             }
-
-
             catch (Exception ex)
             {
                 sessionMetrics?.incr_total_number_resp_server_session_exceptions(1);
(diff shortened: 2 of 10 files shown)


libs/server/Resp/RespServerSession.cs
```diff
@@ -211,28 +211,23 @@ public override int TryConsumeMessages(byte* reqBuffer, int bytesReceived)
...
-                    networkSender.Dispose();
-                }
+                ProcessMessages();
                 recvBufferPtr = null;
             }
+            catch (RespParsingException ex)
+            {
+                sessionMetrics?.incr_total_number_resp_server_session_exceptions(1);
+                logger?.LogCritical($"Aborting open session due to RESP parsing error: {ex.Message}");
+                logger?.LogDebug(ex, "RespParsingException in ProcessMessages:");
```
vazois, line 221: nit; I know it was not introduced in this PR but we do not need both log messages
```suggestion
 logger?.LogCritical(ex, "Aborting open session due to RESP parsing error:");
```

libs/common/GarnetException.cs
```diff
@@ -33,5 +34,13 @@ public GarnetException(string message) : base(message)
         public GarnetException(string message, Exception innerException) : base(message, innerException)
         {
         }
+
+        /// <summary>
+        /// Throw helper that throws a GarnetException.
+        /// </summary>
+        /// <param name="message">Exception message.</param>
+        [DoesNotReturn]
+        public static void Throw(string message) =>
```
vazois, line 43: nit; there is a warning in the inherited method from RespParsingException. 
https://github.com/microsoft/garnet/blob/951cf82c120d4069e940e832db03bfa018c688ea/libs/common/Parsing/RespParsingException.cs#L82

Can you explicitly override the inherited method as below
```csharp 
        [DoesNotReturn]
        public static new void Throw(string message) =>
            throw new RespParsingException(message);
```

vazois (approved):

PR #386, opened 2024-05-15
More Complete ACL Implementation

kevin-montrose:
This sketches out a more proper ACL implementation - one with categories beyond @admin, and where individual commands can be ACL'd.

Basic idea is to use `RespCommandsInfo` as a mechanism to discover command ACLs, and attach a bitfield to each `User` to track individual commands rather than categories.

Outstanding TODOs:
 - [x] Test rest of commands w.r.t. ACL categories
 - [x] Implement individual command ACL'ing (ie. `+get`)
 - [x] Implement sub-command ACL'ing (ie. `-config|get`)
 - [x] Figure out what to do with unlisted/internal/cluster commands
 - [x] Figure out what to do with custom extensions
   * Introduced a @garnet ACL category that all commands not in Redis are now tagged with
 - [x] (Probably) Remove `CommandCategory` - it is redundant
 - [x] Clean up command parsing, now that subcommand bytes are a "later in the pipeline'-thing
   * As part of this, probably DRY up ACL checking - it's a bit error prone to add a new command with this code so "wet"
   * `SET` (with EX, NX, etc. but _not_ `SETEX`) are still special cased as they weren't realized as subCmds
 - [x] Cluster subcommands are weird, but need ACL testing too
 - [x] CustomCmd, CustomTnx, CustomObjCmd are weird and need to be ACL'able in some way too
   * Introduced a @custom ACL category that blocks all of these, and the individual commands can also be ACL'd (that is `+customcmd`  works)
   * Longer term we may want to allow ACL'ing the commands by their actual registered name, but IMO that should be a separate enhancement

This is **_explicitly not_** going to implement keyspace patterns.  I feel like that should be a separate enhancement, as it's pretty substantial on it's own.

---

The biggest incidental change here is a substantial refactoring of `RespCommand` and the various switches over it.  This is, IMO, the cleanest way to get a canonical value to test against ACLs with.  As a consequence, `RespCommand` has a bunch more entries as most things that were previously "sub commands" (but not in the RESP sense) are now actual enum values.

There's one additional implementation detail to call out, which is how ACL descriptions are maintained.  So that ACL descriptions behave more or less like you'd expect (that is `+@foo +bar -bar` becomes `+@foo`) a greedy approach is taken where, upon update, a trial removal of each token is made to see if it impacts the effective ACL set.  This is pretty allocate-y, and while that could be improved I think ACL manipulation is rare and locked down enough that clarity was a bit more important.  It's worth a stringent review, however.  Code is in [User.RationalizeACLDescription](https://github.com/microsoft/garnet/blob/ce2703d2a2ff89008ef88577fcfd37efd27ed2ac/libs/server/ACL/User.cs#L454C31-L454C56).


kevin-montrose:
There's still some more to do here (more testing, subcommands, some cleanup) , but I'm going on vacation for a week and this is pretty dang close to how it's going to actually look when done.  So taking it off draft, and marking ready for review (if not mergin).

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -108,7 +108,7 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
         /// The bit is either set or cleared depending on value, which can be either 0 or 1.
         /// When key does not exist, a new key is created.The key is grown to make sure it can hold a bit at offset.
         /// </summary>
-        private bool StringSetBit<TGarnetApi>(byte* ptr, ref TGarnetApi storageApi)
+        private bool StringSetBit<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storageApi)
```
vazois, line 111: I don't think we need to introduce a count here. Ideally, the ACL check should happen at the parsing level or right after. See  RespCommand.cs

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -432,10 +431,73 @@ private bool StringBitPosition<TGarnetApi>(byte* ptr, int count, ref TGarnetApi
...
+            {
+                return false;
+            }
+
+            readHead = (int)(ptr - recvBufferPtr);
+
+            count--;
+
+            BitmapOperation bitop;
+            if (op.SequenceEqual(CmdStrings.AND) || op.SequenceEqual(CmdStrings.and))
```
vazois, line 457: Matching of the operation type should happen at parsing layer. I think we treat this as subcommand plus for parameters that are fixes and few bytes it might be better to match them using MemoryMarshal.Read

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -432,10 +431,73 @@ private bool StringBitPosition<TGarnetApi>(byte* ptr, int count, ref TGarnetApi
         /// <summary>
         /// Performs bitwise operations on multiple strings and store the result.
         /// </summary>
-        private bool StringBitOperation<TGarnetApi>(int count, byte* ptr, BitmapOperation bitop, ref TGarnetApi storageApi)
+        private bool StringBitOperation<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storageApi)
             where TGarnetApi : IGarnetApi
         {
+            // <AND | OR | XOR | NOT> destkey key
+            if (count < 3)
```
vazois, line 438: I think this is good work, but arity checks are not part of this PR. Even though we need to perform these checks earlier, maybe by capturing a parsing state.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -15,7 +15,7 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
         /// <summary>
         /// TryRENAME
         /// </summary>
-        private bool NetworkRENAME<TGarnetApi>(byte* ptr, ref TGarnetApi storageApi)
+        private bool NetworkRENAME<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storageApi)
```
vazois, line 18: Similar to above, I am not sure why count is needed here other than performing an arity or ACL permission check. We should do this outside given that at this point we have identified the CommandFlag.

libs/server/Resp/Objects/HashCommands.cs
```diff
@@ -67,6 +67,15 @@ private unsafe bool HashSet<TGarnetApi>(int count, byte* ptr, HashOperation hop,
 
                 var inputCount = (count - 1) / 2;
 
+                HashOperation hop = 0;
+                switch (command)
```
vazois, line 71: Seems like many of these changes do not have to do with the ACL implementation. Unless I am mistaken but pushing the Hash operation type inside the method it will make it less clean and harder to do the ACL check globally. Similar to above the operation type should pass as parameter and if we want to do arity checks or ACL checks we should do it before reaching this point.

libs/server/Resp/Objects/HashCommands.cs
```diff
@@ -114,18 +123,17 @@ private unsafe bool HashSet<TGarnetApi>(int count, byte* ptr, HashOperation hop,
...
         /// <returns></returns>
-        private unsafe bool HashGet<TGarnetApi>(int count, byte* ptr, HashOperation op, ref TGarnetApi storageApi)
+        private unsafe bool HashGet<TGarnetApi>(RespCommand command, int count, byte* ptr, ref TGarnetApi storageApi)
             where TGarnetApi : IGarnetApi
         {
-            if ((op == HashOperation.HGETALL && count != 1) ||
-                (op == HashOperation.HRANDFIELD && count < 1) ||
-                (op == HashOperation.HGET && count != 2) ||
-                (op == HashOperation.HMGET && count < 2))
+            if ((command == RespCommand.HGETALL && count != 1) ||
```
vazois, line 131: HashOperation type should be passed as parameter and any ACL checks or mapping should happen at an earlier point.

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
```diff
@@ -160,10 +160,6 @@ public int GetRMWModifiedValueLength(ref SpanByte t, ref SpanByte input)
                         length += HyperLogLog.DefaultHLL.MergeGrow(srcHLL + sizeof(int), dstHLL);
                         return length + t.MetadataSize;
 
-                    case RespCommand.SETKEEPTTLXX:
```
vazois, line 163: Is this change related somehow to the ACL implementation?

vazois (commented):
I think this PR has too many changes to vet, maybe we need to break it into smaller PRs. I think the first is good but very hard to review because the surface area is huge. I think the idea of having a bitmap to match command permissions is good. There might I couple alternatives to tradeoff between CPU cycles or memory. We need to separate general fixes from clear ACL enchantments and avoid pushing parsing inside the execution methods.

libs/cluster/Session/ClusterSession.cs
```diff
@@ -86,20 +86,10 @@ public bool ProcessClusterCommands(RespCommand command, ReadOnlySpan<byte> bufSp
                 }
                 else if (command == RespCommand.FAILOVER)
                 {
-                    if (!CheckACLAdminPermissions(bufSpan, count, out bool success))
```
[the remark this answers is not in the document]
lmaas: From what I can tell this is now checked in `ProcessBasicCommands`, which is better. Ideally none of the command executors should have to check for permissions. The flow should be Parsing -> Permission Checks -> Command Dispatch. That should give us more flexibility, including giving users access to just a subset of the cluster commands. `@admin` should just translate to a predefined command mask.

libs/cluster/Session/ClusterSession.cs
```diff
@@ -86,20 +86,10 @@ public bool ProcessClusterCommands(RespCommand command, ReadOnlySpan<byte> bufSp
...
-                    {
-                        return success;
-                    }
-
                     result = TryFAILOVER(count, recvBufferPtr + readHead);
                 }
                 else if ((command == RespCommand.REPLICAOF) || (command == RespCommand.SECONDARYOF))
                 {
-                    if (!CheckACLAdminPermissions(bufSpan, count, out bool success))
```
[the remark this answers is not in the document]
lmaas: I think this is properly covered in the new implementation (see the comment for `RespCommand.FAILOVER`).

libs/cluster/Session/RespClusterBasicCommands.cs
```diff
@@ -23,7 +23,7 @@ private bool NetworkClusterBumpEpoch(ReadOnlySpan<byte> bufSpan, int count, out
         {
             invalidParameters = false;
             // Check admin permissions for command
-            if (!CheckACLAdminPermissions(bufSpan, count, out var success))
+            if (!CheckACLPermissions(RespCommand.CLUSTER, RespCommandsInfo.SubCommandIds.ClusterBumpEpoch, count, out bool success))
```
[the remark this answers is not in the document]
lmaas: Small note here: the bug is in `CheckACLPermissions`. Technically, this function signature is fine - as long as `CheckACLPermissions` drains starting from the read head. But I agree with @vazois that we should consider still passing the `bufSpan`.

libs/cluster/Session/ClusterSession.cs
```diff
@@ -171,20 +161,23 @@ public void SetUser(User user)
...
+        /// (NOTE: This function does not check keyspaces)
         /// </summary>
-        /// <param name="bufSpan">Buffer containing the current command in RESP3 style.</param>
+        /// <param name="cmd">Command be processed</param>
+        /// <param name="subCommand">Id (w.r.t. <see cref="RespCommandsInfo"/>) of any sub command - will be 0 if no sub command is being processed</param>
         /// <param name="count">Number of parameters left in the command specification.</param>
         /// <param name="processingCompleted">Indicates whether the command was completely processed, regardless of whether execution was successful or not.</param>
         /// <returns>True if the command execution is allowed to continue, otherwise false.</returns>
-        bool CheckACLAdminPermissions(ReadOnlySpan<byte> bufSpan, int count, out bool processingCompleted)
+        bool CheckACLPermissions(RespCommand cmd, byte subCommand, int count, out bool processingCompleted)
```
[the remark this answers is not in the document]
kevin-montrose: I had taken this route because such a buffer wasn't available everywhere I needed to slap an ACL check.  Now that I've DRY'd that up a bit, I agree taking a buffer is cleaner - I'll take a whack at it.

libs/server/ACL/User.cs
```diff
@@ -14,6 +14,258 @@ namespace Garnet.server.ACL
     /// </summary>
     public class User
     {
+        /// <summary>
+        /// Wraps up command permissions behind a reference so it can be atomically swapped.
+        /// </summary>
+        private sealed class CommandPermissionSet
```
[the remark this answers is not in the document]
kevin-montrose: @badrishc suggested fully parsing subcommands up in `RespServerSession`, which would remove the whole notion of `MaxSubCommands` and get rid of some of the other TODOs in this PR.  I think that's a more promising angle, though I can re-order the bitmask here if that doesn't pan out.

libs/cluster/Session/ClusterSession.cs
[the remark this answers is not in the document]
kevin-montrose: That is correct, the intent is to check permissions in one place.  Subcommands currently require special handling (since they aren't parsed "up front"), but that's something that will ideally be fixed (I left a comment above about a suggestion I was given about how to persue that).

One clarifying statement though, assuming Garnet is looking to mimic Redis behavior, the idea that a session has `@admin` privileges is a misnomer.  Sessions have permission to run certain (sub-)commands, possibly further constrained by key filters.  ACL categories are just a convenient shorthand for sets of commands, with the exception of `@all` which includes commands not yet loaded.  Accordingly, something like `+@admin -failover` is possible and `FAILOVER` should be denied in that case.

libs/server/Resp/Objects/HashCommands.cs
kevin-montrose: This (and similar checks for List, SortedSet, and Geo operations) came out of removing the subcommand `byte` (not to be confused with Redis's notion of a subcommand) and hoisting those operations up into `RespCommand` to centralized ACL checking.

There are still some places where the `(List|Hash|Set|whatever)Operation` is needed, so while the "true" value so far as parsing is concerned is the `RespCommand` some new `switch`es (and changed `if`s) were added.

kevin-montrose:
> I think this PR has too many changes to vet, maybe we need to break it into smaller PRs. I think the first is good but very hard to review because the surface area is huge. I think the idea of having a bitmap to match command permissions is good. There might I couple alternatives to tradeoff between CPU cycles or memory. We need to separate general fixes from clear ACL enchantments and avoid pushing parsing inside the execution methods.

Yeah, it's gigantic - ACLs touch literally everything.  Excluding tests, it is trending smaller as things DRY up.

A bunch of random-ish fixes are somewhat unfortunate, as a big part of testing this is making sure that failed ACLs don't bork the command stream - and I encountered a number of cases of fairly "brittle" command implementations.  I'm open to breaking those out into separate PRs once ACLs are in a good place, but I needed some confidence that the ACL design would work for all commands upfront.

test/Garnet.test/CredentialManager.cs
```diff
@@ -35,9 +35,36 @@ public void GenerateCredentials(string TestFolder, ServerCredential[] customCred
...
+                    aclConfig += $"user {user} on " +
+                    $"{(cred.IsClearText ? ">" : "#")}{password} ";
+
+                    aclConfig += "+@all";
+
+                    if (!cred.IsAdmin)
+                    {
+                        aclConfig += " -@admin";
+                    }
+                }
```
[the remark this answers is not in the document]
kevin-montrose: Good catch, this is a hold-over from moving from the old implementation to the new one where the default user was still under-implemented.  This doesn't affect correctness (until/unless key or channel filters are implemented), but is pointless code since the default user (if not otherwise configured) gets this exact ACL anyway.

libs/server/ACL/User.cs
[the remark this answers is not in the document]
lmaas: I support folding the command and subcommand IDs into a single enum. It's cleaner and that way we can also (hopefully) get rid of explicitly assigned IDs.

libs/cluster/Session/ClusterSession.cs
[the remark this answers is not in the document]
badrishc: Good point. The special case of @admin might be removable and replaced with a category.

libs/server/Resp/Bitmap/BitmapCommands.cs
badrishc: Agreed in general, but parsing state should probably be out of scope for this PR. Adding checks where we notice them missing is fine for now.

badrishc:
Great work overall. Shaping up really nicely, thanks!

libs/server/Resp/Objects/HashCommands.cs
kevin-montrose: I want to push back gently on this (ow that ACL checks are all DRY'd up).  The command name ends up serialized in the failure cases, so we have a number of options here:

 1. Don't include the command name in error message, just pass `HashOperation`.
     * This is sort a reduction in quality but it is generic errors are used in some places already.
 2. Pass just the `RespCommand` and lookup the `HashOperation` in the method.
     * This is what I'm currently doing.
 3. Pass just the `HashOperation` and lookup the `RespCommand` on failure.
     * This feels backwards, but is about equivalent to 2.
 4. Pass just the `HashOperation` and assume its name matches `RespCommand`.
     * This is what the old code did, and feels wrong.  There's an implicit contract here that's pretty easy to accidentally break.
 5. Pass both the `HashOperation` and the `RespCommand`.
     * This feels redundant, and adds the chance of calling with mismatched parameters.  We could add a check to avoid that, but then we're doing the lookup anyway... so might as well do 2 or 3.

Of these I clearly then number 2 is most appropriate.

I don't feel that strongly, but I would like to hear why a different option is preferrable.

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
kevin-montrose: Hmm, thought I already replied to this one...

It's not - it was a change that snuck in as I was trying to figure out how the various `SETxxx` values on `RespCommand` were used.  I'm reverting it.

kevin-montrose:
@lmaas @vazois @badrishc This is now in a complete state from my perspective, and ready for full review.

libs/cluster/Session/ClusterCommands.cs
```diff
@@ -138,155 +138,191 @@ public void UnsafeWaitForConfigTransition()
...
         }
 
-        private bool ProcessClusterCommands(ReadOnlySpan<byte> bufSpan, int count)
+        /// <summary>
+        /// Handle cluster subcommands.
+        /// </summary>
+        /// <param name="command">Subcommand to execute.</param>
+        /// <param name="bufSpan">Remaining parameters in the command buffer.</param>
+        /// <param name="count">Number of parameters in teh command buffer</param>
+        /// <returns>True if command is fully processed, false if more processesing is needed.</returns>
```
vazois, line 147: nit
```suggestion
        /// <returns>True if command is fully processed, false if more processing is needed.</returns>
```

badrishc:
Benchmark: `Embedded.perftest --op-workload PING --op-percent 100`

## Branch: main

Running with 1 threads...
Avg. throughput: 29309.33 Kops/sec; 29309.33 Kops/sec/thread (Max: 29309.33, Min: 29309.33)

Running with 2 threads...
Avg. throughput: 58775.95 Kops/sec; 29387.97 Kops/sec/thread (Max: 29476.37, Min: 29299.58)

Running with 4 threads...
Avg. throughput: 110845.18 Kops/sec; 27711.3 Kops/sec/thread (Max: 27735.68, Min: 27651.08)

Running with 8 threads...
Avg. throughput: 222910.64 Kops/sec; 27863.83 Kops/sec/thread (Max: 28094.35, Min: 27739.37)


## Branch: aclImprovements

Running with 1 threads...
Avg. throughput: 23647.38 Kops/sec; 23647.38 Kops/sec/thread (Max: 23647.38, Min: 23647.38)

Running with 2 threads...
Avg. throughput: 48188.8 Kops/sec; 24094.4 Kops/sec/thread (Max: 24138.88, Min: 24049.92)

Running with 4 threads...
Avg. throughput: 91461.15 Kops/sec; 22865.29 Kops/sec/thread (Max: 23044.1, Min: 22646.23)

Running with 8 threads...
Avg. throughput: 180223.81 Kops/sec; 22527.98 Kops/sec/thread (Max: 22673.81, Min: 22344.31)

## Summary

Looks like we are at around 20% performance regression at the moment. Need to drill deeper into why.

badrishc:
Benchmark (Release mode): `Embedded.perftest --op-workload PING --op-percent 100 -t 1,2,4,8`

## Framework: net8.0

## Branch: main

Running with 1 threads...
Avg. throughput: 135719.46 Kops/sec; 135719.46 Kops/sec/thread (Max: 135719.46, Min: 135719.46)

Running with 2 threads...
Avg. throughput: 280635.8 Kops/sec; 140317.9 Kops/sec/thread (Max: 140504.16, Min: 140131.64)

Running with 4 threads...
Avg. throughput: 518444.34 Kops/sec; 129611.09 Kops/sec/thread (Max: 130067.32, Min: 129333.93)

Running with 8 threads...
Avg. throughput: 1038346.73 Kops/sec; 129793.34 Kops/sec/thread (Max: 130582.93, Min: 129158.74)

## Branch: aclImprovements

Running with 1 threads...
Avg. throughput: 62371.18 Kops/sec; 62371.18 Kops/sec/thread (Max: 62371.18, Min: 62371.18)

Running with 2 threads...
Avg. throughput: 127748.11 Kops/sec; 63874.06 Kops/sec/thread (Max: 64003.18, Min: 63744.93)

Running with 4 threads...
Avg. throughput: 236809.68 Kops/sec; 59202.42 Kops/sec/thread (Max: 59403.47, Min: 58968.43)

Running with 8 threads...
Avg. throughput: 473081.86 Kops/sec; 59135.23 Kops/sec/thread (Max: 59271.6, Min: 58980.22)

## Summary

Looks like we are at around 54% performance regression at the moment. Need to drill deeper into why.

vazois (changes requested):
There are some issues related to code style. Please use compact switch statements where possible.
There are some protocol compatibility issues with the responses generated for invalid input in cluster and bitmap commands.
Thanks for refactoring the cluster commands to make them compatible with the improved ACL implementation.

libs/server/Resp/Objects/HashCommands.cs
kevin-montrose: > I was leaning towards 4 because write now we are doing one extra conversion from RespCommand to HashOperation.
I am not sure how things could break unless you are talking about adding more commands in this method which is not what I would recommend going forward
It is unfortunate that this method includes all HGET variants, it should be one NetworkOp method for each command.
This way we won't need have this conversion issue.
I believe with some refactoring work that we are doing this will eventually change in the future.
Overall, I don't feel too strongly against leaving it as is as long as we restructure it in the future.
Maybe others can chime in about it.

<sub><i>Moving above to this thread to keep context</i></sub>

The breaks I imagined were either adding new methods (which is admittedly rare, but Redis has like... 300 commands?  it's not impossible), or refactors that change the names of `HashOperation`s.

I do agree that changing these to one method per operation is the cleanest option.

libs/server/Resp/RespCommand.cs
```diff
@@ -993,136 +1219,499 @@ static RespCommand MatchedNone(RespServerSession session, int oldReadHead)
...
             else if (command.SequenceEqual(CmdStrings.HELLO))
             {
-                return (RespCommand.HELLO, 0);
+                return RespCommand.HELLO;
             }
             else if (command.SequenceEqual(CmdStrings.CLUSTER))
             {
-                return (RespCommand.CLUSTER, 0);
+                bufSpan = new(recvBufferPtr + readHead, bytesRead);
+                ReadOnlySpan<byte> subCommand = GetCommand(bufSpan, out bool gotSubCommand);
```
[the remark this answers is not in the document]
kevin-montrose: Sure.  Style throughout Garnet seems pretty inconsistent - I've tried to match the, like, "ambient" style of a file, but I'll correct anywhere I've misstepped.

I assume long term these style things will be enforced with .editorconfig error rules?

libs/server/Resp/RespCommand.cs
```diff
@@ -993,136 +1219,499 @@ static RespCommand MatchedNone(RespServerSession session, int oldReadHead)
...
+                    ReadOnlySpan<byte> subCommand = GetCommand(bufSpan, out bool gotSubCommand);
+                    if (!gotSubCommand)
+                    {
+                        success = false;
+                        return RespCommand.NONE;
+                    }
+
+                    count--;
+
+                    if (subCommand.EqualsUpperCaseSpanIgnoringCase(CmdStrings.HELP))
```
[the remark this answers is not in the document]
kevin-montrose: It's not an exhaustive set of conditions, and the fallthrough isn't trivial (have to drain and conditionally set `success`)... so I don't really think that'd be cleaner.

libs/server/Resp/RespCommand.cs
```diff
@@ -993,136 +1219,499 @@ static RespCommand MatchedNone(RespServerSession session, int oldReadHead)
...
+                ReadOnlySpan<byte> subCommand = GetCommand(bufSpan, out bool gotSubCommand);
+                if (!gotSubCommand)
+                {
+                    success = false;
+                    return RespCommand.NONE;
+                }
+
+                count--;
+
+                if (subCommand.EqualsUpperCaseSpanIgnoringCase(CmdStrings.CAT))
```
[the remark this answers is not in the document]
kevin-montrose: Per prior comment, ifs aren't exhaustive and the fallthrough isn't trivial.

test/Garnet.test/GarnetBitmapTests.cs
```diff
@@ -2302,7 +2302,7 @@ public void BitmapOperationInvalidOption()
             }
             catch (Exception ex)
             {
-                Assert.AreEqual("syntax error", ex.Message);
+                Assert.AreEqual("ERR unknown command", ex.Message);
```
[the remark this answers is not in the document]
kevin-montrose: Sure, I'll make the attempt though it will complicate parsing a bit now that subcommands are handled at the parsing layer.

I will note that the old test was incorrect (Redis's actual response is `-ERR syntax error`, not `-syntax error`), and that  Redis does not document the text of errors (and it can change between minor versions). See the [RESP2 documentation on errors](https://github.com/redis/redis-specifications/blob/master/protocol/RESP2.md#resp-errors).

libs/server/Resp/RespCommand.cs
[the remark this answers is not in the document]
vazois: It is true that at release there was no official style guide. We have been going through the code and fixing these issues. I am not sure if there any plans to add these rules into .editoconfig. @TalZaccai can comment on this?!
However, going forward for everything we check-in we should not introduce any more of these inconsistencies.

test/Garnet.test/GarnetBitmapTests.cs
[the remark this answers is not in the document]
vazois: > Sure, I'll make the attempt though it will complicate parsing a bit now that "subcommands" are handled at the parsing layer.
> 
> I will note that the old test was incorrect (Redis's actual response is `-ERR syntax error`, not `-syntax error`), and that Redis does not document the text of errors (and it can change between minor versions). See the [RESP2 documentation on errors](https://github.com/redis/redis-specifications/blob/master/protocol/RESP2.md#resp-errors).

just for your reference,

https://github.com/microsoft/garnet/blob/f864b0b3e573e8da847413f197a20615bb625e0e/libs/server/Resp/CmdStrings.cs#L120

libs/server/Resp/CmdStrings.cs
```diff
@@ -181,5 +182,58 @@ static partial class CmdStrings
...
+        public static ReadOnlySpan<byte> GETKEYSINSLOT => "GETKEYSINSLOT"u8;
+        public static ReadOnlySpan<byte> KEYSLOT => "KEYSLOT"u8;
+        public static ReadOnlySpan<byte> SETSLOT => "SETSLOT"u8;
+        public static ReadOnlySpan<byte> SLOTS => "SLOTS"u8;
+        public static ReadOnlySpan<byte> REPLICAS => "REPLICAS"u8;
+        public static ReadOnlySpan<byte> REPLICATE => "REPLICATE"u8;
+        public static ReadOnlySpan<byte> DELKEYSINSLOT => "DELKEYSINSLOT"u8;
+        public static ReadOnlySpan<byte> SETSLOTSRANGE => "SETSLOTSRANGE"u8;
+        public static ReadOnlySpan<byte> SLOTSTATE => "SLOTSTATE"u8;
+        public static ReadOnlySpan<byte> INITIATE_REPLICA_SYNC => "INITIATE_REPLICA_SYNC"u8;
```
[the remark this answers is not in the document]
kevin-montrose: I'll make that change, though I will have to fix a number of places that send these internal commands in upper case already.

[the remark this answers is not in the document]
vazois: > I'll make that change, though I will have to fix a number of places that send these internal commands in upper case already.
> 
> Wait, perhaps you mean the property names should be lower case?

Yes, I meant the property name. Perhaps it is not a big deal, I was just looking for consistency between the PR and main.
It is personal preference, if you have time to address it.

libs/server/Resp/BasicCommands.cs
```diff
@@ -1046,20 +1046,78 @@ private bool NetworkSTRLEN<TGarnetApi>(byte* ptr, ref TGarnetApi storageApi)
...
+        /// <returns>true if parsing succeeded correctly, false if not all tokens could be consumed and further processing is necessary.</returns>
+        private bool NetworkCOMMAND(ReadOnlySpan<byte> bufSpan, int count)
+        {
+            // No additonal args allowed
+            if (count != 0)
+            {
+                if (!DrainCommands(bufSpan, count))
                     return false;
+
+                while (!RespWriteUtils.WriteError($"ERR Unknown subcommand or wrong number of arguments for COMMAND.", ref dcurr, dend))
```
TalZaccai, line 1092: Error should be in the format of CmdStrings.GenericErrUnknownSubCommand

libs/server/Resp/BasicCommands.cs
```diff
@@ -1046,20 +1046,78 @@ private bool NetworkSTRLEN<TGarnetApi>(byte* ptr, ref TGarnetApi storageApi)
...
+        {
+            // No additonal args allowed
+            if (count != 0)
+            {
+                if (!DrainCommands(bufSpan, count))
+                    return false;
 
-            // Handle COMMAND COUNT
-            if (count > 0 && (subCommand.SequenceEqual(CmdStrings.COUNT) || subCommand.SequenceEqual(CmdStrings.count)))
+                while (!RespWriteUtils.WriteError($"ERR Unknown subcommand or wrong number of arguments for COMMAND COUNT.", ref dcurr, dend))
```
TalZaccai, line 1117: Error should be in the format of CmdStrings.GenericErrWrongNumArgs

test/Garnet.test/RespCommandTests.cs
```diff
@@ -45,6 +46,89 @@ public void TearDown()
             TestUtils.DeleteDirectory(Directory.GetParent(extTestDir)?.FullName);
         }
 
+        /// <summary>
+        /// Verify that all existing combinations of RespCommand and subcommand byte (if relevant)
+        /// have a matching RespCommandInfo objects defined in RespCommandsInfo
+        /// </summary>
+        [Test]
+        public void CommandsInfoCoverageTest()
```
TalZaccai, line 54: This test was removed from main, we decided not to have it enforced

TalZaccai (changes requested):

badrishc:
> I reproduce this regression on my latest too, I'll spend some time digging (now that all feedback is addressed) and update when I have an explanation/fix.

* Is it possible something is inadvertently getting boxed?
* I made this into a BDN testcase here: https://github.com/microsoft/garnet/compare/main...badrishc/resp-parse-stress - here are the results comparing main to this PR:

main:
| Method     | Job    | EnvironmentVariables | Runtime  | Mean     | Error     | StdDev    |
|----------- |------- |--------------------- |--------- |---------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 | 4.292 us | 0.0446 us | 0.0395 us |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 2.449 us | 0.0052 us | 0.0040 us |

PR:
| Method     | Job    | EnvironmentVariables | Runtime  | Mean     | Error     | StdDev    |
|----------- |------- |--------------------- |--------- |---------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 | 4.937 us | 0.0462 us | 0.0409 us |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 3.958 us | 0.0443 us | 0.0393 us |

kevin-montrose:
> Is it possible something is inadvertently getting boxed?
I made this into a BDN testcase here: https://github.com/microsoft/garnet/compare/main...badrishc/resp-parse-stress - here are the results comparing main to this PR:

Doesn't appear to be allocations based on local profiling.  Might be useful to add `[MemoryDiagnoser]` to that benchmark, which I haven't had a chance to poke at yet.

Doing some quick and dirty benchmarking with the Embedded.perftest example, it looks like it's mostly the cost of doing the `CheckACLPermissions` on every request now.

On my box, .NET 8 with PGO disabled I'm getting about...
  - `main` ~62K `PING`s / s
  - this branch ~43K `PING`s / s
  - this branch, but remove the call to `if (...)` with the `CheckACLPermissions` call ~58K `PING`s / s

Total loss ~30%, and a solid ~25% (~15K `PING`s / s) is lost in checking ACLs.

I'll noodle on speeding this up in the morning.

badrishc:
> Doesn't appear to be allocations based on local profiling. Might be useful to add `[MemoryDiagnoser]` to that benchmark, which I haven't had a chance to poke at yet.

I have updated the branch with `[MemoryDiagnoser]` and added Get/Set in addition to Ping, for completeness. PR for this is here: https://github.com/microsoft/garnet/pull/446

kevin-montrose:
Mkay, because I can't leave well enough alone despite it being after 11 PM here...

One relatively simple approach would be to slap a `int[]` (trading space for time) on User that's used to cache the last ACL check indexed by each command (skipping any ACL checks on a 1, only transitioning from 0 to 1 on a successful ACL check).  We just allocate a new `int[]` and swap it out whenever the user's ACL is modified.  A really terrible hacked version of this gets in the ~59K range again locally, but that's with no testing so huge grain of salt.

I need to think on if there's any point in keeping the bit-packed version around; if not, we could just initialize the int[] with the "truth" and then there's not even an update.

Another, much less simple, approach would be to lean in to `RespCommand` being expanded and change the various `ProcessXXXCommands` methods into a single lookup into an array of function pointers / delegates.  The array starts initialized with some function that does the ACL check and replaces the appropriate entry with a pointer to the corresponding `NetworkXXX(...)` method for the command on successful ACL.  Subsequent invocations would jump straight to the command's actual implementation.  On any user permission change, we reset the array.

That might ultimately be a bit cleaner, but it'd require a bunch of additional changes on top of this already gigantic change.  There's also an indirection in there which makes me not so confident the perf would be any better.

So I'll keep pursuing the first approach.  Of course, if there are any other clever ideas, I'd be happy to hear them.

badrishc:
> Mkay, because I can't leave well enough alone despite it being after 11 PM here...
> 
> One relatively simple approach would be to slap a `int[]` (trading space for time) on User that's used to cache the last ACL check indexed by each command (skipping any ACL checks on a 1, only transitioning from 0 to 1 on a successful ACL check). We just allocate a new `int[]` and swap it out whenever the user's ACL is modified. A really terrible hacked version of this gets in the ~59K range again locally, but that's with no testing so huge grain of salt.
> 
> I need to think on if there's any point in keeping the bit-packed version around; if not, we could just initialize the int[] with the "truth" and then there's not even an update.
> 
> Another, much less simple, approach would be to lean in to `RespCommand` being expanded and change the various `ProcessXXXCommands` methods into a single lookup into an array of function pointers / delegates. The array starts initialized with some function that does the ACL check and replaces the appropriate entry with a pointer to the corresponding `NetworkXXX(...)` method for the command on successful ACL. Subsequent invocations would jump straight to the command's actual implementation. On any user permission change, we reset the array.
> 
> That might ultimately be a bit cleaner, but it'd require a bunch of additional changes on top of this already gigantic change. There's also an indirection in there which makes me not so confident the perf would be any better.
> 
> So I'll keep pursuing the first approach. Of course, if there are any other clever ideas, I'd be happy to hear them.

I'm not sure we need to do this. The bitmap is nice and tight. There might be a simpler way to get back the performance. Can look into it tomorrow.

kevin-montrose:
@badrishc some exploratory work in [my latest commit](https://github.com/microsoft/garnet/pull/386/commits/7ebd963a6bb7bfa9b3837203be424616d9c6da30) clawed some of the perf back, down to a ~10% loss (62K vs 56K `PING`s / sec on my local).

This doesn't introduce any caching or anything, it's a revision on the existing approach.  Most "interesting" thing is it introduces an unauthorized user and checks against that instead of first checking `IsAuthenticated` on an interface - we're still covered by another call to `IsAuthenticated` later in the pipeline.

Going off my previous math, there's still may 2K loss to the ACL check

libs/server/Resp/AdminCommands.cs
```diff
@@ -711,35 +711,62 @@ void ProcessHelloCommand(byte? respProtocolVersion, ReadOnlySpan<byte> username,
...
 
         /// <summary>
-        /// Performs @admin command group permission checks for the current user and the given command.
-        /// (NOTE: This function is temporary until per-command permissions are implemented)
+        /// Performs permission checks for the current user and the given command.
+        /// (NOTE: This function does not check keyspaces)
         /// </summary>
-        /// <param name="bufSpan">Buffer containing the current command in RESP3 style.</param>
+        /// <param name="cmd">Command be processed</param>
+        /// <param name="bufSpan">Buffer with remaining parts of command to process.</param>
```
vazois, line 718: nit: remove bufSpan param summary if not used any more.

kevin-montrose:
Pushed up one more exploratory commit that remembers `CanAuthenticate` so we can avoid an interface method call on each command.

---

Switching to BDN now that it's in `main`.

`main` (as of fcf880a0f5ef27384768dca43d89144fa30ae5fb)

| Method     | Job    | EnvironmentVariables | Runtime  | Mean      | Error     | StdDev    | Allocated |
|----------- |------- |--------------------- |--------- |----------:|----------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 |  2.249 us | 0.0338 us | 0.0316 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | 23.161 us | 0.2178 us | 0.1931 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | 13.875 us | 0.0973 us | 0.0910 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  1.983 us | 0.0138 us | 0.0122 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 18.985 us | 0.0977 us | 0.0914 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 12.280 us | 0.1564 us | 0.1463 us |         - |

this (as of 18d7d133e08b60726972db697f91c6ae761aee49), now with a UseACLs option

| Method     | Job    | EnvironmentVariables | Runtime  | UseACLs | Mean      | Error     | StdDev    | Allocated |
|----------- |------- |--------------------- |--------- |-------- |----------:|----------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 | False   |  2.378 us | 0.0232 us | 0.0217 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | False   | 23.721 us | 0.2094 us | 0.1959 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | False   | 14.102 us | 0.0550 us | 0.0514 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   |  2.033 us | 0.0154 us | 0.0144 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 18.883 us | 0.1771 us | 0.1656 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 11.194 us | 0.0779 us | 0.0728 us |         - |
| InlinePing | .NET 6 | Empty                | .NET 6.0 | True    |  2.493 us | 0.0105 us | 0.0098 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | True    | 23.715 us | 0.2381 us | 0.1989 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | True    | 14.485 us | 0.0909 us | 0.0806 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    |  2.306 us | 0.0254 us | 0.0237 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 19.279 us | 0.1247 us | 0.1106 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 12.000 us | 0.0741 us | 0.0693 us |         - |

full logs here: https://gist.github.com/kevin-montrose/8145ddfffa201735b50c78e7edf8f39f

Which is awful close to a wash when ACLs are disable, and maybe a 0.5us loss when enabled.

Still need to look at Badrish's proposal, so that's up next.

badrishc:
> Still need to look at Badrish's proposal, so that's up next.

Adding a link to this thread for reference - https://github.com/microsoft/garnet/commit/f20925a381c47050ac02ce0b75504852d70707a4

badrishc:
> exploratory commit that remembers CanAuthenticate

I don't think this will work with authenticators that might time out, e.g., the Aad authenticator.

kevin-montrose:
> > exploratory commit that remembers CanAuthenticate
> 
> I don't think this will work with authenticators that might time out, e.g., the Aad authenticator.

`IsAuthenticated` is used for timeouts, `CanAuthenticate` (which I cached) is a constant for all current implementations so I _think_ it'd be safe.

That said, I think your proposal here is cleaner - getting a perf run together then I'll update.

kevin-montrose:
I've pushed up a version that merges some of my tweaks and @badrishc's proposal.

The `CanAuthenticate` cache is gone, but other changes remain: `_user` will never be null (so we can elide some other checks), `NoAuth` commands aren't special cased anymore (they're set in the bitmap like everything else), some less likely paths are `NoInlining`]'d, and some hot paths are `AggressiveInlining`'d.

Latest BDN (as of 1ed1635f17efef7100a733c498257f31a5abbd6d)

| Method     | Job    | EnvironmentVariables | Runtime  | UseACLs | Mean      | Error     | StdDev    | Allocated |
|----------- |------- |--------------------- |--------- |-------- |----------:|----------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 | False   |  2.229 us | 0.0194 us | 0.0172 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | False   | 23.134 us | 0.1162 us | 0.1030 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | False   | 13.650 us | 0.0628 us | 0.0557 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   |  1.890 us | 0.0082 us | 0.0077 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 18.742 us | 0.1390 us | 0.1160 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 11.005 us | 0.0559 us | 0.0495 us |         - |
| InlinePing | .NET 6 | Empty                | .NET 6.0 | True    |  2.502 us | 0.0201 us | 0.0178 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | True    | 23.270 us | 0.0814 us | 0.0722 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | True    | 14.275 us | 0.0563 us | 0.0527 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    |  1.913 us | 0.0154 us | 0.0144 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 18.779 us | 0.1348 us | 0.1126 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 11.582 us | 0.0520 us | 0.0486 us |         - |

Full log: https://gist.github.com/kevin-montrose/a1a7f9ef472c68fbb2b3b61c1c3f190d

Suggests a w/o ACL improvement, and about the same for perf with ACLs.

I did a sanity check with `Embedded.perftest` again, and while I trust it less than BDN numbers it also shows a gain of ~4K `PING`s / sec (for about 6% gain, upt o ~66K total `PING`s / sec; .NET 8, PGO disabled) in the no-ACL case.

libs/server/ACL/CommandPermissionSet.cs
```diff
@@ -0,0 +1,197 @@
...
+        public string Description { get; set; }
+
+        /// <summary>
+        /// Returns true if the given command + subCommand pair can be run.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CanRunCommand(RespCommand command)
+        {
+            // Special case "everything is permitted" 
+            if (this == All)
```
badrishc, line 60: You had the `AllState` enum earlier which seemed to be a useful concept and caused only a byte comparison here. Any reason to use this refactoring? I.e. does it help performance wise?

libs/server/ACL/CommandPermissionSet.cs
```diff
@@ -0,0 +1,197 @@
...
+            : this(new ulong[CommandListLength], description)
+        {
+        }
+
+        private CommandPermissionSet(ulong[] commandList, string description)
+        {
+            this._commandList = commandList;
+            this.Description = description;
+
+            // Certain commands can always be run, so initialize with them
```
badrishc, line 35: Probably revert this since cmd.IsNoAuth needs to happen at CheckACL site anyway ...

libs/server/Resp/AdminCommands.cs
```diff
@@ -711,35 +711,62 @@ void ProcessHelloCommand(byte? respProtocolVersion, ReadOnlySpan<byte> username,
...
         /// <returns>True if the command execution is allowed to continue, otherwise false.</returns>
-        bool CheckACLAdminPermissions(ReadOnlySpan<byte> bufSpan, int count, out bool processingCompleted)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        bool CheckACLPermissions(RespCommand cmd, byte* ptr, int count, out bool processingCompleted)
         {
-            Debug.Assert(!_authenticator.IsAuthenticated || (_user != null));
+            Debug.Assert(_user != null);
 
-            if (!_authenticator.IsAuthenticated || (!_user.CanAccessCategory(CommandCategory.Flag.Admin)))
+            if (!_user.CanAccessCommand(cmd))
```
badrishc, line 727: I think this needs to be:
```
if ((!_authenticator.IsAuthenticated || !_user.CanAccessCommand(cmd)) && !cmd.IsNoAuth())
```

Assuming:
* if Auth fails, we want to proceed ONLY IF the command is a NoAuth command
* if ACL fails, we want to proceed ONLY IF the command is a NoAuth command

This means the NoAuth check has to be in the `&&` clause.

badrishc: Also note that `_authenticator.IsAuthenticated` needs to run no matter what code path, which is why it is the first check in the `if` condition. That way we are protected by both Auth and ACL in one place.

libs/cluster/Session/ClusterCommands.cs
```diff
@@ -138,155 +138,86 @@ public void UnsafeWaitForConfigTransition()
...
+                    RespCommand.CLUSTER_RESET => NetworkClusterReset(bufSpan, count, out invalidParameters),
+                    RespCommand.CLUSTER_SEND_CKPT_FILE_SEGMENT => NetworkClusterSendCheckpointFileSegment(count, out invalidParameters),
+                    RespCommand.CLUSTER_SEND_CKPT_METADATA => NetworkClusterSendCheckpointMetadata(count, out invalidParameters),
+                    RespCommand.CLUSTER_SETCONFIGEPOCH => NetworkClusterSetConfigEpoch(bufSpan, count, out invalidParameters),
+                    RespCommand.CLUSTER_SETSLOT => NetworkClusterSetSlot(bufSpan, count, out invalidParameters),
+                    RespCommand.CLUSTER_SETSLOTSRANGE => NetworkClusterSetSlotsRange(bufSpan, count, out invalidParameters),
+                    RespCommand.CLUSTER_SHARDS => NetworkClusterShards(count, out invalidParameters),
+                    RespCommand.CLUSTER_SLOTS => NetworkClusterSlots(bufSpan, count, out invalidParameters),
+                    RespCommand.CLUSTER_SLOTSTATE => NetworkClusterSlotState(bufSpan, count, out invalidParameters),
+                    _ => throw new Exception($"Unexpected cluster subcommad: {command}")
```
vazois, line 196: Is it possible to add a CLUSTER_INVALID case to deal e.g. cluster foo arg1 arg2 arg3?
I think we had support for it in main.

libs/server/Resp/ACLCommands.cs
```diff
@@ -18,53 +18,25 @@ namespace Garnet.server
...
     {
         /// <summary>
-        /// Processes subcommands for the "ACL" RESP-Command.
-        /// NOTE: This function assumes the ACL token was already parsed.
+        /// Processes ACL LIST subcommand.
         /// </summary>
         /// <param name="bufSpan">The remaining command bytes</param>
         /// <param name="count">The number of arguments remaining in bufSpan</param>
         /// <returns>true if parsing succeeded correctly, false if not all tokens could be consumed and further processing is necessary.</returns>
-        private bool ProcessACLCommands(ReadOnlySpan<byte> bufSpan, int count)
```
vazois, line 27: Great work separating these into individual methods. 
I think we are missing ACL GENPASS, LOG and DRYRUN.
Do you think any of those will be useful to add now or later?
https://redis.io/docs/latest/commands/acl-genpass/

badrishc:
Idea to avoid the interface call to authenticator in common case (not necessary for this PR): 

https://github.com/microsoft/garnet/commit/17402f340806980a378682b807f1e4bb142268da

libs/cluster/Session/RespClusterBasicCommands.cs
```diff
@@ -22,11 +22,6 @@ internal sealed unsafe partial class ClusterSession : IClusterSession
         private bool NetworkClusterBumpEpoch(ReadOnlySpan<byte> bufSpan, int count, out bool invalidParameters)
```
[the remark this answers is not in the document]
badrishc: They should eventually go away, but it is orthogonal to this PR imo.

libs/server/Resp/ACLCommands.cs
kevin-montrose: I think we wait, as tempted as I am.  PR is big enough already.

libs/cluster/Session/ClusterCommands.cs
kevin-montrose: I can put a specific error message back, but it'll go in the parsing layer.

Redis responds with `-ERR unknown subcommand 'foo'. Try CLUSTER HELP.` ; so I'll mimic that.

libs/server/ACL/ACLParser.cs
```diff
@@ -191,6 +241,72 @@ public static void ApplyACLOpToUser(ref User user, string op)
             {
                 throw new ACLUnknownOperationException(op);
             }
+
+            // There's some fixup that has to be done when parsing a command
+            static bool TryParseCommandForAcl(string commandName, out RespCommand command)
```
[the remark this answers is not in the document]
kevin-montrose: Ah, I feel almost exactly the opposite - I prefer (static) local methods when the method is only useful in one context, since it doesn't pollute autocomplete and clearly signals "this is a helper for this _single_ method".

But when in Rome and all, I'll move it out.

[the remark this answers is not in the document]
badrishc: Maybe we should start following this style in future. I've seen this in plenty of places, and your points make sense. @TedHartMS?

libs/server/ACL/CommandPermissionSet.cs
badrishc: Sounds good, both are fine so up to you, thanks.

libs/server/ACL/ACLParser.cs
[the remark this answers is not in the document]
vazois: > Ah, I feel almost exactly the opposite - I prefer (static) local methods when the method is only useful in one context, since it doesn't pollute autocomplete and clearly signals "this is a helper for this _single_ method".
> 
> But when in Rome and all, I'll move it out.

Makes sense, and I was expecting the pushback for the exact reasons you mentioned.
I wish there was a way to declare it as local outside the encapsulating method using a keyword or something ( .net people are you listening? 😄)
Maybe a cleaner way to use it will be to declare it at the top of the method.
I would not want to declare more than a couple in any method, @badrishc if you want to follow that style in other places.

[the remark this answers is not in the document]
TedHartMS: I'm a fan of local functions for the reasons @kevin-montrose states. Out of curiosity I just BDN'd a trivial local function vs inlined member and the local function showed a couple % faster.

In this case though I can see @vazois' point that it is kind of hidden due to the size of the calling function. But I still go with the "I only use it here so I only want to see it here" principle.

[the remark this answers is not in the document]
badrishc: Okay then! Let's stick with what Kevin has, for this case.

libs/server/ACL/CommandPermissionSet.cs
kevin-montrose: Looks like avoiding the pointer is still a bit better locally.  Preloading isn't needed or beneficial with the `IsNoAuth()` check returning, so I'll remove that.

kevin-montrose:
Latest perf numbers

BDN
===

main (0acff38b7f343cf26c58a20995b2675f8f42b634)
---

| Method     | Job    | EnvironmentVariables | Runtime  | Mean      | Error     | StdDev    | Allocated |
|----------- |------- |--------------------- |--------- |----------:|----------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 |  2.246 us | 0.0275 us | 0.0244 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | 23.058 us | 0.1166 us | 0.1090 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | 13.873 us | 0.0944 us | 0.0837 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  1.986 us | 0.0089 us | 0.0084 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 19.673 us | 0.1043 us | 0.0975 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 11.025 us | 0.0482 us | 0.0377 us |         - |

aclImprovements (f5e16848e96aa152583e9410a750b1e0f31cf4d9)
---

 Method     | Job    | EnvironmentVariables | Runtime  | UseACLs | Mean      | Error     | StdDev    | Allocated |
|----------- |------- |--------------------- |--------- |-------- |----------:|----------:|----------:|----------:|
| InlinePing | .NET 6 | Empty                | .NET 6.0 | False   |  2.323 us | 0.0440 us | 0.0646 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | False   | 23.092 us | 0.1001 us | 0.0887 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | False   | 13.588 us | 0.0956 us | 0.0848 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   |  2.184 us | 0.0163 us | 0.0153 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 18.372 us | 0.1050 us | 0.0931 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | False   | 11.196 us | 0.0495 us | 0.0439 us |         - |
| InlinePing | .NET 6 | Empty                | .NET 6.0 | True    |  2.586 us | 0.0142 us | 0.0133 us |         - |
| Set        | .NET 6 | Empty                | .NET 6.0 | True    | 23.417 us | 0.0952 us | 0.0891 us |         - |
| Get        | .NET 6 | Empty                | .NET 6.0 | True    | 14.442 us | 0.0901 us | 0.0843 us |         - |
| InlinePing | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    |  2.102 us | 0.0268 us | 0.0251 us |         - |
| Set        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 19.076 us | 0.1852 us | 0.1641 us |         - |
| Get        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | True    | 11.330 us | 0.0473 us | 0.0419 us |         - |

Which looks like around 0.2us loss.

I did grab Embedded.perftest numbers again, but they are wildly variable for me latest `main` - swinging by up to 30K PINGs / sec so I really don't trust them.

kevin-montrose:
> Idea to avoid the interface call to authenticator in common case (not necessary for this PR):
> 
> [17402f3](https://github.com/microsoft/garnet/commit/17402f340806980a378682b807f1e4bb142268da)

This is interesting (as is punching the authenticator into RespServerSession as a generic argument, to de-virtualize everything) but I want push back a little bit on the idea that NoAuth is the common case.

While it certainly isn't _unusual_ to deploy Redis (and thus I'd assume Garnet) with NoAuth and rely on network isolation for security, there are big cases where that's not true.  Top of my mind is if Garnet is ever shipped as a service itself, or in any environment where defense in depth is a big thing.

---

Thinking out loud a bit, it seems to me a remaining perf killer is that we have to call into the `IGarnetAuthenticator` on every command to validate our `User` is still good - a poll model, if you will.  Given that the common case is that the `User` is still good (because invalidation is rare, AD/Kerberos sessions get measured in hours, password rotations happen on the order of months, etc. etc.) it feels a push model is more appropriate.

One way to `IGarnetAuthenticator` into a push model, shooting from the hip:
```csharp
public interface IGarnetAuthenticator
{
    /// <summary>
    /// Can authenticator authenticate
    /// </summary>
    bool CanAuthenticate { get; }

    /// <summary>
    /// Whether this authenticator can be used with the ACL
    /// </summary>
    bool HasACLSupport { get; }

    /// <summary>
    /// Authenticate the incoming username and password from AUTH command. Username is optional
    /// </summary>
    bool TryAuthenticate(ReadOnlySpan<byte> password, ReadOnlySpan<byte> username, out AuthenticationLifetime authLifetime);
}

public sealed class AuthenticationLifetime
{
    // for Authenticators that support auth but not invalidation, or we use null as a special case
    public static AuthenticationLifetime Infinite = ... ;

    // we call this in RespServerSession, it get devirtualized into a simple bool check
    public bool IsValid { get; }

    // IGarnetAuthenticators call this when a previously auth'd user (from a call to IGarnetAuthenticator.TryAuthenticate) is de-auth'd
    public void Invalidate() { ... }
}
```

_I'm not proposing to do this as part of this PR._

To be clear, the `IGarnetAuthenticator` implementation would have to remember the `AuthenticationLifetime`s it hands back to invalidate them - but I think that's fine, it already has to remember stuff about the user to implement `IsAuthenticated`.

This would keep the `User` and authentication separation we have, get the `IGarnetAuthenticator` interface out of the hotpath, and I think cover all reasonable schemes.

PaulusParssinen:
Just noting that in real-world workload, the .NET 8 DynamicPGO (which is disabled in our BDN _micro_benchmarks to reduce noise) would likely specialize the virtualized calls for common types here and we shouldn't have to do it ourselves. Whether it kicks in here is a bit harder to verify but something to keep in mind.

badrishc:
> Thinking out loud a bit, it seems to me a remaining perf killer is that we have to call into the `IGarnetAuthenticator` on every command to validate our `User` is still good - a poll model, if you will. Given that the common case is that the `User` is still good (because invalidation is rare, AD/Kerberos sessions get measured in hours, password rotations happen on the order of months, etc. etc.) it feels a push model is more appropriate.

Yeah, it makes sense to consider a push-based approach to update the auth status in the rare case of it being revoked instead of having to poll the provider every single time. cc @yangmsft whose team uses the auth feature. (definitely out of scope for this PR)

badrishc (approved):

TalZaccai (approved):

vazois (approved):

PR #396, opened 2024-05-17
[bug fix] The bin path is null caused the program startup failure

BlueHtml:
Version 1.0.9 uses `Assembly.GetExecutingAssembly().Location` to get the directory where Garnet is located, but it return an empty string in SingleFile mode ([source link](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location#remarks)), leading to program startup failure with the error: `Unable to initialize server due to exception: Value cannot be null. (Parameter 'path1')`. Using `AppContext.BaseDirectory` works well.

> `Assembly.Location Property`: In .NET 5 and later versions, for bundled assemblies, the value returned is an empty string. ([source link](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location#remarks))

The test code is as follows:
```cs
var binPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
Console.WriteLine($"AppContext.BaseDirectory:{AppContext.BaseDirectory}, is null:{AppContext.BaseDirectory is null}");
Console.WriteLine($"Assembly.GetExecutingAssembly().Location:{Assembly.GetExecutingAssembly().Location}, is null:{Assembly.GetExecutingAssembly().Location is null}");
Console.WriteLine($"binPath:{binPath}, is null:{binPath is null}");
```

The output is as follows:
```
PS D:\tmp\test\bb> \code\tmp\garnet\main\GarnetServer\bin\Release\net8.0\publish\linux-arm64\GarnetServer.exe
    _________
   /_||___||_\      Garnet 1.0.9 64 bit; standalone mode
   '. \   / .'      Port: 3278
     '.\ /.'        https://aka.ms/GetGarnet
       '.'

AppContext.BaseDirectory:D:\code\tmp\garnet\main\GarnetServer\bin\Release\net8.0\publish\linux-arm64\, is null:False
Assembly.GetExecutingAssembly().Location:, is null:False
binPath:, is null:True
Unable to initialize server due to exception: Value cannot be null. (Parameter 'path1')
PS D:\tmp\test\bb>
```

### DIFF
--- a/main/GarnetServer/Program.cs
+++ b/main/GarnetServer/Program.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Reflection;
 using System.Threading;
 using Garnet.common;
 using Garnet.server;
@@ -49,7 +48,7 @@ namespace Garnet
         /// </summary>
         static bool TryRegisterExtensions(GarnetServer server)
         {
-            var binPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var binPath = AppContext.BaseDirectory;
 
             if (!TryGetRespCommandsInfo(Path.Combine(binPath!, CustomRespCommandInfoJsonFileName), out var customCommandsInfo))
                 return false;


jiripospisil:
I can confirm that this fixes the start-up failure. Not related to this PR directly, but what I don't understand is why it's even attempting to load the file like this in the first place when the path to the JSON file could be specified in the config file (or better yet, specify a directory in the config file and all JSON files from there could be automatically loaded). Thoughts?

badrishc:
However we handle path here should be analogous to the way we default paths when LogDir or ChechpointDir is null.

badrishc:
Temporary fix is in this PR: https://github.com/microsoft/garnet/pull/397
We will improve the custom commands info handling next week.

Issue #346, opened 2024-04-30
API Coverage - Implement HELLO


mattholy:
### Feature request type

sample request

### Is your feature request related to a problem? Please describe

When using example from https://microsoft.github.io/garnet/docs/welcome/releases
```bash
redis-cli
127.0.0.1:6379> set key value
OK
127.0.0.1:6379> get key
"value"
127.0.0.1:6379> hello
(error) ERR unknown command
```

But for a actual redis:
```bash
127.0.0.1:6379> hello
[
    "server",
    "redis",
    "version",
    "7.2.4",
    "proto",
    2,
    "id",
    8,
    "mode",
    "standalone",
    "role",
    "master",
    "modules",
    []
]
```
It matters because some python frameworks like Celery are using `hello`. It's a pity if they can not use such a k-v database.

### Describe the solution you'd like

Implement https://redis.io/docs/latest/commands/hello/

### Additional context

https://github.com/celery/celery/blob/90ff2e1290ef058e8dc59dc1d86ff38f668abc50/celery/app/control.py#L388-L389

Issue #179, opened 2024-03-28
Incorrect PONG issued on subscriber channel


mgravell:
The `PING` command is one of only a few that are permitted once a connection has entered subscription mode; however, it has a different semantic:

> If the client is subscribed to a channel or a pattern, it will instead return a multi-bulk with a "pong" in the first position and an empty bulk in the second position, unless an argument is provided in which case it returns a copy of the argument.

A typical reply might therefore be:

``` txt
*2
$4
pong
$0

```

However, Garnet issues instead:

``` txt
+PONG
```

To repro, try (looking in the obvious directory):

``` txt
using StackExchange.Redis;
using StackExchange.Redis.Configuration;

var options = ConfigurationOptions.Parse("127.0.0.1:3278");
LoggingTunnel.LogToDirectory(options, @"c:\Code\RedisLog");
using var muxer = ConnectionMultiplexer.Connect(options);
var sub = muxer.GetSubscriber();
var ttl = sub.Ping();
Console.WriteLine(ttl);
```

This *may* impact clients that aren't expecting it; I *think* SE.Redis gracefully survives this, but I haven't checked for niche cases.

It may also be worth checking whether Garnet is responding to any other commands in subscriber mode; there are *very few* things it should allow while there are subscriptions **assuming RESP2**:

- `PING` (with different behavior)
- `[P|S][UN]SUBSCRIBE`
- `RESET`
- `QUIT`

PR #398, opened 2024-05-18
Basic implementation of HELLO

badrishc:
Towards https://github.com/microsoft/garnet/pull/387
Fixes https://github.com/microsoft/garnet/issues/346
Fixes https://github.com/microsoft/garnet/issues/179


libs/server/Resp/AdminCommands.cs
```diff
@@ -270,6 +270,75 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+            {
+                int? protocolVersion = null;
+                ReadOnlySpan<byte> authUsername = default, authPassword = default;
+                string clientName = null;
+
+                if (count > 0)
+                {
+                    var ptr = recvBufferPtr + readHead;
+                    int _protocolVersion;
+                    if (!RespReadUtils.ReadIntWithLengthHeader(out _protocolVersion, ref ptr, recvBufferPtr + bytesRead))
```
TalZaccai, line 283: nit: Inconsistent naming convention for local variable (preceding underscore)

libs/server/Resp/AdminCommands.cs
```diff
@@ -270,6 +270,75 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+
+                    protocolVersion = _protocolVersion;
+                    count--;
+                    while (count > 0)
+                    {
+                        var param = GetCommand(bufSpan, out bool success1);
+                        if (!success1) return false;
+                        var paramStr = Encoding.ASCII.GetString(param);
+                        count--;
+                        if (paramStr.Equals("AUTH", StringComparison.OrdinalIgnoreCase))
```
TalZaccai, line 295: Can use CmdStrings.AUTH and then do a SequenceEquals instead of converting to ASCII & doing a string comparison

libs/server/Resp/AdminCommands.cs
```diff
@@ -270,6 +270,75 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+                        {
+                            if (count < 2)
+                            {
+                                if (!DrainCommands(bufSpan, count))
+                                    return false;
+                                count = 0;
+                                errorFlag = true;
+                                errorCmd = "hello";
```
TalZaccai, line 303: I'd use nameof(RespCommand.HELLO)

libs/server/Resp/AdminCommands.cs
```diff
@@ -488,6 +557,79 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
         }
 
+
+        void ProcessHelloCommand(int? protocolVersion, ReadOnlySpan<byte> username, ReadOnlySpan<byte> password, string clientName)
+        {
+            if (protocolVersion != null)
+            {
+                if (protocolVersion.Value is < 2 or > 3)
+                {
+                    while (!RespWriteUtils.WriteError("ERR Unsupported protocol version"u8, ref dcurr, dend))
```
TalZaccai, line 567: I think our convention is to have the error messages defined in CmdStrings

libs/server/Resp/AdminCommands.cs
```diff
@@ -488,6 +557,79 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+                this.clientName = clientName;
+            }
+
+            (string, string)[] helloResult =
+                [
+                    ("server", "redis"),
+                    ("version", storeWrapper.redisProtocolVersion),
+                    ("garnet_version", storeWrapper.version),
+                    ("proto", $"{this.protocolVersion}"),
+                    ("id", "63"),
```
TalZaccai, line 605: Shouldn't this return a client ID?

libs/server/Resp/AdminCommands.cs
```diff
@@ -488,6 +557,79 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
             return true;
         }
 
+
+        void ProcessHelloCommand(int? protocolVersion, ReadOnlySpan<byte> username, ReadOnlySpan<byte> password, string clientName)
+        {
+            if (protocolVersion != null)
+            {
+                if (protocolVersion.Value is < 2 or > 3)
```
TalZaccai, line 565: Should we allow a protocol switch to RESP3?

TalZaccai (changes requested):

libs/server/Resp/AdminCommands.cs
```diff
@@ -270,6 +270,75 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
                     errorCmd = "ping";
                 }
             }
+            else if (command == RespCommand.HELLO)
```
TalZaccai, line 273: General comment: probably good to add a test for this command

TalZaccai (changes requested):

libs/server/Resp/AdminCommands.cs
TalZaccai: Probably also worthwhile to comment somewhere that we're not actually doing any protocol version switching, and that it should be implemented if we support multiple protocols at some point....

libs/server/Resp/AdminCommands.cs
badrishc: Added a test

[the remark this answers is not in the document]
badrishc: Other commands like PING reside here. I suggest we tackle the correct command splitting in a separate PR.

libs/server/Resp/AdminCommands.cs
```diff
@@ -488,6 +557,79 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
...
+            }
+
+            if (clientName != null)
+            {
+                this.clientName = clientName;
+            }
+
+            (string, string)[] helloResult =
+                [
+                    ("server", "redis"),
```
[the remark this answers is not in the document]
badrishc: It needs to be redis here, apparently, for clients to read that.

libs/server/Resp/AdminCommands.cs
badrishc: The async PR requires RESP3 so we do have to support protocol switching.

libs/server/Resp/AdminCommands.cs
badrishc: We don't support a client ID concept, so I just return a constant here.

libs/server/Resp/AdminCommands.cs
badrishc: I added a new `EqualsIgnoreCase` method - pls check carefully. And we can make a work item to convert other uses of CmdStrings to use this method so we can remove lowercase versions in CmdStrings.

badrishc:
Made this PR have HELLO work only with RESP2, the other open PR will add RESP3 for async.

libs/server/Resp/AdminCommands.cs
badrishc: I have updated this PR to only support RESP2. RESP3 will be added in the "async" PR.

badrishc:
fixed warnings

libs/server/Resp/AdminCommands.cs
```diff
@@ -488,6 +564,104 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
             return true;
         }
 
+        bool EqualsIgnoreCase(ReadOnlySpan<byte> left, ReadOnlySpan<byte> right)
```
yrajas, line 567: >EqualsIgnoreCase [](http://example.com/codeflow?start=13&length=16)

This could move to a utils class under libs\common.

badrishc: done, made extension method for even better clarity. pls check.

libs/server/Resp/AdminCommands.cs
badrishc: See `EqualsUpperCaseSpanIgnoringCase`

libs/server/Resp/AdminCommands.cs
TalZaccai: Shouldn't we return something like -1 or 0 then? Might be less confusing

libs/common/ConvertUtils.cs
```diff
@@ -53,5 +54,38 @@ public static void MakeUpperCase(Span<byte> command)
...
+
+        /// <summary>
+        /// Check if two byte spans are equal, where right is an all-upper-case span, ignoring case if there are ASCII bytes.
+        /// </summary>
+        /// <param name="left"></param>
+        /// <param name="right"></param>
+        /// <returns></returns>
+        public static bool EqualsUpperCaseSpanIgnoringCase(this ReadOnlySpan<byte> left, ReadOnlySpan<byte> right)
+        {
+            if (left.SequenceEqual(right))
```
TalZaccai, line 66: Isn't it wasteful to do a SequenceEqual here? Wouldn't it be better to just iterate over the array once the way we're doing at the bottom?

libs/server/Resp/AdminCommands.cs
```diff
@@ -250,8 +250,16 @@ private bool ProcessAdminCommands<TGarnetApi>(RespCommand command, ReadOnlySpan<
             {
```
TalZaccai, line 250: Looks like we have 2 implementations for PING for some reason (here and in BasicCommands), I don't thing this one is ever called... Can we remove it altogether?

libs/server/Resp/AdminCommands.cs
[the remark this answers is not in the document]
TalZaccai: PING is implemented both here and in BasicCommands for some reason... Probably better to put this in BasicCommands in the first place, just so that people aren't confused, but up to you.

TalZaccai (changes requested):
Looks good to me, added a couple of questions / comments

libs/common/ConvertUtils.cs
badrishc: this is optimizing for the common case where sequences are equal, as span.sequenceequal does some tricks to make this very fast.

libs/server/Resp/AdminCommands.cs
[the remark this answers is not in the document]
badrishc: Many non-admin commands are in this file anyway, worth a separate dev refactor.

libs/server/Resp/AdminCommands.cs
badrishc: just followed another server's model where it returned 64. can revisit in a polishing PR later.

TalZaccai (approved):

Issue #395, opened 2024-05-17
CLUSTER KEYSLOT mistake


tianlileer:
### Describe the bug

The result of the CLUSTER KEYSLOT command is inconsistent with the result of the redis command

### Steps to reproduce the bug

do the command CLUSTER KEYSLOT CLUSTER KEYSLOT "Hm{W\x13\x1c" in garnet and redis, you can see the result
the garnet:
<img width="367" alt="image" src="https://github.com/microsoft/garnet/assets/5744425/427bb0d0-061f-4951-a6c0-1ed58626d67d">

the redis:
<img width="359" alt="image" src="https://github.com/microsoft/garnet/assets/5744425/ced7f23a-1a96-40de-9aaa-f963666da41f">

### Expected behavior

Looking forward to the same outcome as Redis

PR #198, opened 2024-03-29
Optimize NumUtils.CRC16

PaulusParssinen:
This PR changes the `NumUtils.CRC16` method to use RVA instead of `readonly static` for the precomputed CRC16 table.

```diff
-; Total bytes of code 121, prolog size 11, PerfScore 53.06, instruction count 41, allocated bytes for code 121 (MethodHash=13cde191) for method Garnet.common.NumUtils:CRC16(ulong,int):ushort (FullOpts)
+; Total bytes of code 65, prolog size 0, PerfScore 36.06, instruction count 21, allocated bytes for code 65 (MethodHash=13cde191) for method Garnet.common.NumUtils:CRC16(ulong,int):ushort (FullOpts)
```

### Full diff
<details>
<summary>Garnet.common.NumUtils:CRC16</summary>

```diff
 method Garnet.common.NumUtils:CRC16(ulong,int):ushort (FullOpts)
 ; Emitting BLENDED_CODE for X64 with AVX - Windows
 ; FullOpts code
 ; optimized code
 ; rsp based frame
 ; fully interruptible
 ; No PGO data
+; 0 inlinees with PGO data; 2 single block inlinees; 0 inlinees without PGO data
 ; Final local variable assignments
 ;
 ;  V00 arg0         [V00,T01] (  7, 16   )    long  ->  registers  
-;  V01 arg1         [V01,T04] (  3,  3   )     int  ->  rdx         single-def
-;  V02 loc0         [V02,T02] (  5, 14   )  ushort  ->  rsi        
-;  V03 loc1         [V03,T03] (  3,  6   )    long  ->  rdi         single-def
-;  V04 OutArgs      [V04    ] (  1,  1   )  struct (32) [rsp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
-;  V05 tmp1         [V05,T00] (  3, 24   )    long  ->  rbx         "impSpillLclRefs"
-;* V06 cse0         [V06,T06] (  0,  0   )    long  ->  zero-ref    hoist "CSE #01: aggressive"
-;  V07 cse1         [V07,T05] (  2,  4.25)    long  ->  rax         hoist "CSE #02: aggressive"
+;  V01 arg1         [V01,T05] (  3,  3   )     int  ->  rdx         single-def
+;  V02 loc0         [V02,T02] (  5, 14   )  ushort  ->  rax        
+;  V03 loc1         [V03,T04] (  3,  6   )    long  ->  rdx         single-def
[...]

PR #399, opened 2024-05-18
Fix CRC16 Hashslot Calculation

vazois:
This PR addresses the hash slot calculation issue #395 by ensuring that key strings without both left and right brackets (e.g., "Hm{W\x13\x1c") are used in their entirety within the CRC16 hash function.
I also added more tests and separated CRC16 hashlot implementation into its own folder.


PaulusParssinen:
Not directly related to the PR but thought this would be good place to ask 😄 

Why was CRC16 chosen as the key slot hash function? There's much more effort spent in industry to optimize CRC32 variants. We have CRC32C available as intrinsic and there's a lot of even faster tricks with SIMD (The intrinsic would probably win against the SIMD for the smaller data). Any background on this?

vazois:
> Not directly related to the PR but thought this would be good place to ask 😄
> 
> Any background on why was CRC16 chosen as the key slot hash function? There's much more effort spent in industry to optimize CRC32 variants. We have CRC32C available as intrinsic and there's a lot of even faster tricks with SIMD. The intrinsic would probably win against the SIMD for the smaller data and _very likely_ against the current LUT approach which will incur L1-L2 misses (would have to measure this).
> 
> cc @badrishc



> Not directly related to the PR but thought this would be good place to ask 😄
> 
> Any background on why was CRC16 chosen as the key slot hash function? There's much more effort spent in industry to optimize CRC32 variants. We have CRC32C available as intrinsic and there's a lot of even faster tricks with SIMD. The intrinsic would probably win against the SIMD for the smaller data and _very likely_ against the current LUT approach which will incur L1-L2 misses (would have to measure this).
> 
> cc @badrishc

I believe it is done as precaution since people expect it to be crc16 like what Redis is using. 
If there is no reason for the consumers of Garnet to expect CRC16 mapping, we can use whatever hash function (even non crc based one.
I didn't know that people expect it to be explicitly CRC16 until an issue was opened, which actually relates to a problem with how we extract the substring with the hashtags and not the hash function itself.

PaulusParssinen:
> I believe it is done as precaution since people expect it to be crc16 like what Redis is using. If there is no reason for the consumers of Garnet to expect CRC16 mapping, we can use whatever hash function (even non crc based one. I didn't know that people expect it to be explicitly CRC16 until an issue was opened, which actually relates to a problem with how we extract the substring with the hashtags and not the hash function itself.

Ah right, I didn't know it was specced out to Redis. Looking at it now it makes sense as there's 16384 hash slots. Disregard my message 😄

libs/common/HashSlotUtils.cs
```diff
@@ -0,0 +1,121 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Garnet.common
+{
+    public static unsafe class HashSlotUtils
+    {
+        /// <summary>
+        /// This table is based on the CRC-16-CCITT polynomial (0x1021)
```
PaulusParssinen, line 13: I had it wrong when I wrote this comment in https://github.com/microsoft/garnet/pull/198
```suggestion
        /// This table is based on the CRC-16-XMODEM polynomial (0x1021)
```

badrishc:
> Not directly related to the PR but thought this would be good place to ask 😄
> 
> Any background on why was CRC16 chosen as the key slot hash function? There's much more effort spent in industry to optimize CRC32 variants. We have CRC32C available as intrinsic and there's a lot of even faster tricks with SIMD. The intrinsic would probably win against the SIMD for the smaller data and _very likely_ against the current LUT approach which will incur L1-L2 misses (would have to measure this).
> 
> cc @badrishc

Hi Paulus, can you confirm that the changes you made to CRC logic do not change the hash computation compared to what was there before, so that we return the compatibility with what Redis uses?

PaulusParssinen:
There was no behaviour change in that minor optimization.

## CRC16 Loop body (https://github.com/microsoft/garnet/pull/198)

### Before 
```assembly
;  V01 arg1         [V01,T04] (  3,  3   )     int  ->  rdx         single-def
;  V02 loc0         [V02,T02] (  5, 14   )  ushort  ->  rsi        
;  V03 loc1         [V03,T03] (  3,  6   )    long  ->  rdi         single-def
;  V04 OutArgs      [V04    ] (  1,  1   )  struct (32) [rsp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;  V05 tmp1         [V05,T00] (  3, 24   )    long  ->  rbx         "impSpillLclRefs"
;  V07 cse1         [V07,T05] (  2,  4.25)    long  ->  rax         hoist "CSE #02: aggressive"

; ushort result = 0;
       xor      esi, esi
; byte* end = data + len;
       movsxd   rdi, edx
       add      rdi, rbx
; while (data < end)
       cmp      rbx, rdi
       jae      SHORT LOOP_EXIT
; ...
       mov      LUT_PTR, 0x7FFC59808C18    ; LUT_PTR = 0x7FFC59808C18;
LOOP_ENTRY:
       lea      rcx, [rbx+0x01]            ; rcx = rbx + 1; 
       mov      rbp, rcx                   ; rbp = rcx;
	  
; tmp = (((result >> 8) ^ *data++) & 0xff)
       mov      rcx, qword ptr [LUT_PTR]   ; rcx = [LUT_PTR];
       mov      edx, esi                   ; edx = esi; 
       sar      edx, 8                     ; edx >>= 8; 
       movzx    r8, byte  ptr [rbx]        ; r8 = [rbx]; Load byte from address in rbx into r8
       xor      edx, r8d                   ; edx = edx ^ r8d;
       movzx    rdx, dl                    ; rdx = dl;

; result = (ushort)(*(crc_table_ptr + tmp) ^ (result << 8));
       movzx    rcx, word  ptr [rcx+2*rdx] ; rcx = [rcx + 2*rdx];
       shl      esi, 8                     ; esi <<= 8;
       xor      ecx, esi                   ; ecx = ecx ^ esi;
       movzx    rsi, cx                    ; rsi = cx;

       cmp      rbp, rdi
       mov      rbx, rbp
       jb       SHORT LOOP_ENTRY
LOOP_EXIT:
; ...
```

### After
```assembly
; arg1  int     ->  rdx         single-def
; loc0  ushort  ->  rax        
; loc1  long    ->  rdx         single-def
; loc2  long    ->  rcx        
; tmp1  long    ->  rcx         "impSpillLclRefs"
; cse0  long    ->   r8         hoist "CSE #01: aggressive"

; ushort result = 0;
       xor      eax, eax
; byte* end = data + len;
       movsxd   rdx, edx
       add      rdx, rcx
; while (data < end)
       cmp      rcx, rdx
       jae      SHORT EXIT

       mov      LUT_PTR, 0x2195CA42EB0      ; LUT_PTR = 0x2195CA42EB0;
LOOP_ENTRY:  ;; offset=0x0017
; tmp = data++
       lea      r10, [rcx+0x01]

; nuint index = (nuint)(uint)((result >> 8) ^ *(data + 1)) & 0xff;
       mov      r9d, eax                    ; r9d = eax;
       sar      r9d, 8                      ; r9d >>= 8
       movzx    rcx, byte  ptr [rcx]        ; rcx = [rcx]  
       xor      ecx, r9d                    ; ecx = ecx ^ r9d
       movzx    rcx, cl                     ; zero extend

; result = (ushort)(Unsafe.Add(LUT_PTR, rcx) ^ (result << 8));
       movzx    rcx, word  ptr [LUT_PTR+2*rcx] ; rcx = [LUT_PTR + 2*rcx];
       shl      eax, 8                         ; eax <<= 8; 
       xor      eax, ecx                       ; eax = eax ^ ecx;
       movzx    rax, ax                        ; rax = ax;

       cmp      r10, rdx
       mov      rcx, r10
       jb       SHORT LOOP_ENTRY
EXIT:
; ...
```

badrishc:
> The intrinsic would probably win against the SIMD for the smaller data and very likely against the current LUT approach which will incur L1-L2 misses (would have to measure this).

The cluster spec requires us to map keys to the 16384 hash slot space. This is a very common server side operation and has measurable perf, so if there is an opportunity to speed it up that would be good.

But I think existing clients expect the specific CRC mapping chosen by Redis as they might use this to route keys? (@mgravell can confirm).

badrishc:
Unrelated, speaking of hash code logic, our server side key hash logic is here: https://github.com/microsoft/garnet/blob/main/libs/storage/Tsavorite/cs/src/core/Utilities/Utility.cs#L179

Any way to improve its speed, while maintaining (or improving) the good hash distribution/spread property, would be interesting as well.

vazois:
> > The intrinsic would probably win against the SIMD for the smaller data and very likely against the current LUT approach which will incur L1-L2 misses (would have to measure this).
> 
> The cluster spec requires us to map keys to the 16384 hash slot space. This is a very common server-side operation and has measurable perf impact, so if there is an opportunity to speed it up that would be good.
> a
> But I think existing clients expect the specific CRC mapping chosen by Redis as they might use this to route keys? (@mgravell can confirm).

I guess the issue is that when routing requests, the client sees only the key and it has to perform the hashslot calculation to figure out which shard is responsible for that key.
We can use a different hash function but the client needs to be aware of it to perform the correct hashslot calculation and route the request appropriately. 
I am wondering if there is a way in C# to provide a string code snippet and execute it at runtime. Could be a solution for the client to acquire the HashSlot  calculation implementation. Though it sounds risky from security perspective.

badrishc:
Have you verified that clients are doing this CRC computation, or as they just querying some server and using the redirect to determine the mapping?

vazois:
> Have you verified that clients are doing this CRC computation, or are they just querying some server and using the redirect to determine and cache the mapping?

I did not check this at first, just assumed this is how they would be doing it. However, I quick search reveals that SERedis is doing something like what I said above because I can see they have their own CRC16 implementation

https://github.com/StackExchange/StackExchange.Redis/blob/61c13c21844ff3e92eb077523dc876688878ba25/src/StackExchange.Redis/ServerSelectionStrategy.cs#L63

mgravell:
I can 100% confirm that cluster-aware clients try to perform the hash
locally for routing. The alternatives are:

1. perform dumb routing and blindly respond to `-MOVED` (or issue the
command that returns the slot) - adds latency, makes a mess of ordering
2. use a cluster-aware proxy, which just kicks the can up a step to the
proxy, since the proxy still has to route somehow

The client must know the hashing algorithm to route efficiently. At the
moment *only* the predefined crc16 with the specified seed data is
supported. Could another algorithm be negotiated? I mean, maybe, if it is
demonstrably faster, but asking client libraries to support it would be
painfully slow - it basically blocks all clients until they choose to
implement that hash (if they ever do), plus we'd need to come up with a
mechanism to even convey that difference.

PaulusParssinen:
I think cost of switching the hash function greatly outweights any benefits of slightly faster, given it's apparently _not_ a implementation detail. Also I think naive CRC16 lookup table loop is doing just good enough here given the inputs are small.

> Unrelated, speaking of hash code logic, our server side key hash logic is here: https://github.com/microsoft/garnet/blob/main/libs/storage/Tsavorite/cs/src/core/Utilities/Utility.cs#L179
> 
> Any way to improve its speed, while maintaining (or improving) the good hash distribution/spread property, would be interesting as well.

This on the other hand probably could be improved but with good benchmark/profiling data as proof of course.

badrishc (approved):

PR #402, opened 2024-05-21
LTRIM Fix

TalZaccai:
Fixing LTRIM logic & adding some tests
Related to bug #400


test/Garnet.test/RespListTests.cs
```diff
@@ -97,24 +96,66 @@ public void MultiLPUSHAndLTRIM()
...
+            new object[] {3, 5, new[] {3, 4, 5} },
+            new object[] {-12, 0, new[] {0} },
+            new object[] {-12, 2, new[] {0, 1, 2} },
+            new object[] {-12, -7, new[] {0, 1, 2, 3} },
+            new object[] {-15, -11, Array.Empty<int>() },
+            new object[] {8, 8, new[] {8} },
+            new object[] {8, 12, new[] {8, 9} },
+            new object[] {9, 12, new[] {9} },
+            new object[] {10, 12, Array.Empty<int>() },
+        };
```
TedHartMS, line 130: Add a start > end case (code handles it) and per docs the key should be removed in that case or if start > end of list

libs/server/Objects/List/ListObjectImpl.cs
```diff
@@ -217,55 +216,55 @@ private void ListRange(byte* input, ref SpanByteAndMemory output)
...
+                            this.UpdateSize(value, false);
                         }
-                        _output->opsDone = numDeletes;
+                        outputHeader->opsDone = numDeletes;
                     }
                     else
                     {
-                        int i = 0;
+                        var i = 0;
                         IList<byte[]> readOnly = new List<byte[]>(list).AsReadOnly();
```
TedHartMS, line 252: Performance: Is this copy necessary? Can it iterate nodes via First..Next to the start then just remove nodes at that point? Possibly another PR to address

TedHartMS (approved):

libs/server/Objects/List/ListObjectImpl.cs
TalZaccai: Agreed

PR #404, opened 2024-05-21
Fix Bitop Edge Case

vazois:
This PR fixes edge case where bitop will result in null ptr exception when none of the source keys exist.
I added also some more negative tests for bitop.


badrishc:
cc @kevin-montrose

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -450,19 +459,34 @@ private bool StringBitOperation<TGarnetApi>(int count, byte* ptr, BitmapOperatio
...
             }
 
-            var status = storageApi.StringBitOperation(keys, bitop, out long result);
-
-            if (status != GarnetStatus.NOTFOUND)
+            if (sizeof(byte*) * (keyCount - 1) > 512)
             {
-                while (!RespWriteUtils.WriteInteger(result, ref dcurr, dend))
-                    SendAndReset();
+                throw new Exception("Bitop source key limit (64) exceeded");
```
badrishc, line 483: What does this exception surface to client code as?

libs/server/Resp/RespCommand.cs
```diff
@@ -460,47 +460,7 @@ static RespCommand MatchedNone(RespServerSession session, int oldReadHead)
...
-                                                            return (RespCommand.BITOP, (byte)BitmapOperation.NOT);
-                                                        }
-
-                                                        // Reset read head and count if we didn't match operator.
-                                                        readHead -= 9;
-                                                        count += 1;
-                                                    }
-                                                }
-                                            }
+                                            return (RespCommand.BITOP, 0);
```
badrishc, line 463: Why was this change reverted? Our goal going forward is to try and consolidate all the command and subcommand parsing into the parsing phase, rather than depend on individual operations to parse. That way ACL can also be managed centrally after the (command, subcommand) are known.

libs/server/Resp/Bitmap/BitmapCommands.cs
vazois: Good point. I added a RESP response that is surfaced to the client instead of throwing an exception.

libs/server/Resp/RespCommand.cs
vazois: Whereas the most data commands I have seen follow the format <command> <key>, BITOP expects the key after the operation type (i.e. AND,OR, XOR etc.) which could be considered as subcommand, though technically it isn't.
I made the change in order to deal with the event the provided operation type is not valid.
I reverted it back to the original state and added a workaround.
Now if BITOP is matched and then the subcommand is not matched we still end up in BITOP but capture the error using BitmapOperation.NONE flag and drain the buffer.

badrishc (approved):

Issue #405, opened 2024-05-22
INFO modules in redis do not work in garnet.


kyuminMoon:
### Describe the bug

When using INFO modules through redis-cli in redis cluster mode, the values below are returned.
```
#Modules
module:name=graph,ver=20019,api=1,filters=0,usedby=[],using=[],options=[]
module:name=rg,ver=10001,api=1,filters=0,usedby=[],using=[ai],options=[]
module:name=ReJSON,ver=10004,api=1,filters=0,usedby=[],using=[],options=[]
```

However, in garnet cluster mode, the error below is returned.
```
[2024-05-22 12:10:42] ERR Invalid section MODULES. TRY INFO HELP
```
Since client tools are using the command, I think this part needs to be fixed quickly.

### Steps to reproduce the bug

```
redis-cli > INFO modules
```

PR #408, opened 2024-05-22
INFO field for modules

badrishc:
Fixes https://github.com/microsoft/garnet/issues/405


vazois (approved):
Looks good! I added missing help message.

PR #410, opened 2024-05-22
Custom Command Registration Updates

TalZaccai:
- [x] Added option to register custom transaction with (optional) command info
- [x] Added option to dynamically register custom commands / transactions while specifying path for JSON file containing commands info on Garnet server
- [x] Added examples for registering a command + transaction with a RESP command info object in GarnetServer's Main
- [x] Added tests for dynamic command registration while specifying command info JSON path on server
- [x] Updated REGISTERCS website docs to reflect changes


main/GarnetServer/Program.cs
```diff
@@ -46,7 +44,18 @@ static void Main(string[] args)
         static bool TryRegisterExtensions(GarnetServer server)
```
badrishc, line 44: Does not need to return bool any more.

test/Garnet.test.cluster/Garnet.test.cluster.csproj
```diff
@@ -24,6 +24,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" />
+    <PackageReference Include="Microsoft.CodeAnalysis" />
```
badrishc, line 27: why is this package added?

TalZaccai: I moved the CreateTestLibrary method (that uses CodeAnalysis to build the dlls with the custom classes) to TestUtils since I needed it in a separate test class. Because test.cluster has a link to TestUtils, it also needs that package...

badrishc (approved):

PR #417, opened 2024-05-23
Various API bugfixes

TalZaccai:
- [x] TTL / PTTL should return -2 when key does not exist
- [x] MODULE * should return an unknown command error since not yet implemented
- [x] GETRANGE with non-existing key should return an empty string
- [x] SRANDMEMBER with non-existing key and count specified should return an empty array
- [x] [H|S|Z|CO]SCAN with too few parameters should return an error
- [x] GEO commands (except GEODIST) should return an array of null arrays
- [x] GEOADD with CH and NX options should include added items in output count
- [x] RUNTXP with too few parameters should return an error

Added tests for all cases described


badrishc:
cc @kevin-montrose

libs/server/Resp/Objects/SharedObjectCommands.cs
```diff
@@ -25,109 +25,114 @@ private unsafe bool ObjectScan<TGarnetApi>(int count, byte* ptr, GarnetObjectTyp
...
+            pcurr += sizeof(int);
+
+            // Prepare length of header in input buffer
+            var inputLength = (int)(recvBufferPtr + bytesRead - (byte*)inputPtr);
 
+            // Prepare GarnetObjectStore output
+            var outputFooter = new GarnetObjectStoreOutput { spanByteAndMemory = new SpanByteAndMemory(dcurr, (int)(dend - dcurr)) };
+            var status = storageApi.ObjectScan(key, new ArgSlice((byte*)inputPtr, inputLength), ref outputFooter);
+
+            //restore input buffer
```
badrishc, line 114: Nit: "//[space]**R**estore input buffer"

yrajas (approved):

PR #419, opened 2024-05-24
Support WITHSCORE option in ZRANK, ZREVRANK commands

yrajas:
**Problem**
Currently, processing accepts the WITHSCORE parameter to ZRANK and ZREVRANK commands, but the result is not built with the score if the parameter is passed in.

**Fix**
Modified the ZRANK, ZREVRANK implementation to have the ability to provide either a single rank integer output or an array output based on presence of WITHSCORE parameter.


TalZaccai (approved):
LGTM!

badrishc (approved):

PR #392, opened 2024-05-16
Add stronger TLS check for Gossip in Cluster mode

msft-paddy14:
This PR resolves issue #391.

It's a small change that uses ClusterTlsClientTargetHost  and validates the incoming remote cert (during Gossip CLUSTER MEET call) and ensures that we're attempting to join a node that was created as part of current cluster setup. This can prevent issues of unknown clusters getting connected when multiple caches might be running in same IP pool. Previously, no such validation was there and we did an unconditional handshake approval (when ClientCertRequired is false). This will enforce the checks for only gossip where the server node acts as a client

PR #421, opened 2024-05-24
Add flag for backward compatibility with TLS checks

msft-paddy14:
In #392, we added stronger TLS checks for cluster gossip to prevent an unintended node from connecting. The behavior might be unexpected for some of consumers (using older versions of Garnet) as this would mean specifying the ClusterTlsClientTargetHost parameter which they might've missed earlier as it was not validated in all scenarios - it was behind a ClientCertificateRequired flag. This flag also implied that server cert was also not validated when Garnet acted as a client in gossip scenarios.
But to maintain backward compatibility we're adding another flag to enable the stronger TLS checks conditionally. However, it should be recommended that this is turned on to prevent unexpected cluster connections.


badrishc:
For Garnet server, is there a good reason to even expose this flag?

libs/host/Configuration/Options.cs
```diff
@@ -224,6 +224,10 @@ internal sealed class Options
         [Option("cluster-tls-client-target-host", Required = false, HelpText = "Name for the client target host when using TLS connections in cluster mode.")]
         public string ClusterTlsClientTargetHost { get; set; }
 
+        [OptionValidation]
+        [Option("server-certificate-required", Required = false, HelpText = "Whether server TLS certificate is required by clients on the server side, e.g., for cluster gossip and replication.")]
```
vazois, line 228:
```suggestion
[Option("server-certificate-required", Required = false, HelpText = "Whether server TLS certificate is required by clients e.g., for cluster gossip and replication.")]
````

Comment (badrishc) [reply libs/host/Configuration/Options.cs:228]: It would be more precise to say "clients on the server side" else it is confusing how a server is asking for client information.

Comment (vazois) [reply libs/host/Configuration/Options.cs:228]:
```suggestion
[Option("server-certificate-required", Required = false, HelpText = "Whether server TLS certificate is required by clients established on the server side, e.g., for cluster gossip and replication.")]
```

badrishc: OK, please mirror the text change into defaults.conf as well

vazois (approved):

PR #431, opened 2024-05-30
Remove misc. temporary allocations from RESP parsing

PaulusParssinen:
This PR first introduces `RespWriteUtils.TrySliceWithLengthHeader(out ROS<byte>, ..)` <sub>(Better name suggestions are welcomed)</sub> method to ease allocation-free RESP command parsing. This PR then utilizes the new method in various places of command parsing and eliminates various tiny allocations with it.

> [!Warning]
> Like with the existing ``RespReadUtils.ReadPtrWithLengthHeader`` method, caller must be careful when using `TrySlice`. The underlying memory the sliced span points to can be clobbered from the object header tricks or by buffer reallocation (`ObjectUtils.ReallocateOutput`) in many places. In those places, we have no option but to copy to new array due to the current way the buffer is managed.

This PR also moves and adds more common Ascii manipulation logic to `AsciiUtils` and delegates some of them to vectorized BCL implementations.

This method overload will gain more use in .NET 9 where we'll be able to allocation-free lookups against the BCLs hash-based collection types via `IAlternateEqualityComparer<in TAlternate, T> where TAlternate : allow ref struct`.


libs/common/AsciiUtils.cs
```diff
@@ -1,17 +1,87 @@
...
+#endif
+    }
+
+    /// <summary>
+    /// Check if two byte spans are equal, where right is an all-upper-case span, ignoring case if there are ASCII bytes.
+    /// </summary>
+    public static bool EqualsIgnoreCase(ReadOnlySpan<byte> left, ReadOnlySpan<byte> right)
+    {
+#if NET8_0_OR_GREATER
+        return Ascii.EqualsIgnoreCase(left, right);
```
badrishc, line 66: The idea of `EqualsUpperCaseSpanIgnoringCase` is that we assume the common case is both are upper case in which Sequence.Equals is the fast path. Only otherwise do we pay the detailed comparison cost. Also, we do not want to call Ascii.EqualsIgnoreCase even in the slow path, as our loop knows that the right side is upper-case and thus can avoid a few comparisons.

badrishc: Even the vector optimized version [here](https://github.com/gfoidl/dotnet-runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Text/Ascii.Equality.cs#L195) has to do a lot as it is unable to make assumptions about ASCII or the right being upper case.

PaulusParssinen: Alright then, I don't feel strongly about this one. 

We will get JIT to completely unroll the half-const inputs someday though :) I have been pestering one of the JIT developers about it for a while now and thought to use this as motivation 😆

libs/common/AsciiUtils.cs
```diff
@@ -1,17 +1,83 @@
...
+        foreach (ref var c in command)
+            if (c > 96 && c < 123)
+                c -= 32;
+#endif
+    }
+
+    /// <summary>
+    /// Check if two byte spans are equal, where right is an all-upper-case span, ignoring case if there are ASCII bytes.
+    /// </summary>
+    public static bool EqualsIgnoreCase(ReadOnlySpan<byte> left, ReadOnlySpan<byte> right)
```
badrishc, line 63: Can we rename this back to `EqualsUpperCaseSpanIgnoringCase`? Else the consumer may use it in wrong situation where `right` is not known to be uppercase.

badrishc: We had it as an extension method earlier, so the name name more sense. `A.EqualsUpperCaseSpanIgnoringCase(B)`.

PaulusParssinen: > Why did you change it from extension method to normal.

Mostly consistency with other `*Utils` & allowing PR reviewers to know exact method which a line is calling by glance & without having to hover in IDE. 

Reverted to original shape.

badrishc (approved):

PR #433, opened 2024-05-30
Handle ResponseWriteUtils Fail Correctly

vazois:
This PR fixes the code in places where SendAndReset is not called after failure to write into provide response buffer.

### DIFF
--- a/libs/server/Custom/CustomFunctions.cs
+++ b/libs/server/Custom/CustomFunctions.cs
@@ -24,7 +24,7 @@ namespace Garnet.server
         protected static unsafe void WriteSimpleString(ref (IMemoryOwner<byte>, int) output, ReadOnlySpan<char> simpleString)
         {
             // Get space for simple string
-            int len = 1 + simpleString.Length + 2;
+            var len = 1 + simpleString.Length + 2;
             if (output.Item1 != null)
             {
                 if (output.Item1.Memory.Length < len)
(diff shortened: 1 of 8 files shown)


badrishc (approved):

libs/server/Custom/CustomFunctions.cs
```diff
@@ -71,9 +72,9 @@ protected static unsafe void WriteBulkStringArray(ref MemoryResult<byte> output,
             fixed (byte* ptr = output.MemoryOwner.Memory.Span)
             {
                 var curr = ptr;
-                RespWriteUtils.WriteArrayLength(values.Length, ref curr, ptr + totalLen);
-                for (int i = 0; i < values.Length; i++)
-                    RespWriteUtils.WriteBulkString(values[i].Span, ref curr, ptr + totalLen);
+                _ = RespWriteUtils.WriteArrayLength(values.Length, ref curr, ptr + totalLen);
```
lmaas, line 75: I like that you added a note to remember that we are using pre-allocated memory and that `WriteArrayLength` should never fail. I would add that comment here as well, and maybe consume the output of the call to `RespWriteUtils.WriteArrayLength` and check in a debug assert (applies to the other locations as well, of course).

lmaas (changes requested):
Nice cleanup and bug fixes! I have added a small comment below, but this looks good overall.

lmaas (approved):
The code looks good to me now. Thanks for addressing the comment!

PR #436, opened 2024-05-31
Tsavorite session reorganization

TedHartMS:
- Rename IFunctions to ISessionFunctions (in prep for StoreFunctions)
- Clean up ITsavoriteSession to SessionFunctionsWrapper, minimizing the per-Context code to just the locking portion
- remove RUMD (Read/Upsert/rMw/Delete) functionality from ClientSession; users should use one of the 4 Contexts (Basic, Lockable, Unsafe, LockableUnsafe) for all such actions. This changed a lot of tests
- Remove the unused RUMDAsync methods
- Speed up some of the more time-consuming UTs


libs/server/AOF/AofProcessor.cs
```diff
@@ -31,11 +31,13 @@ public sealed unsafe partial class AofProcessor
         /// Session for main store
         /// </summary>
         readonly ClientSession<SpanByte, SpanByte, SpanByte, SpanByteAndMemory, long, MainStoreFunctions> session = null;
```
badrishc, line 33: do we need to keep `session` and `objectStoreSession` here as fields?

libs/storage/Tsavorite/cs/Tsavorite.sln
```diff
@@ -29,12 +29,12 @@ Global
 		Release|x64 = Release|x64
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{33A732D1-2B58-4FEE-9696-B9483496229F}.Debug|Any CPU.ActiveCfg = Debug|x64
-		{33A732D1-2B58-4FEE-9696-B9483496229F}.Debug|Any CPU.Build.0 = Debug|x64
+		{33A732D1-2B58-4FEE-9696-B9483496229F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
badrishc, line 32: is this change to Any CPU intentional?

libs/storage/Tsavorite/cs/benchmark/FixedLenYcsbBenchmark.cs
```diff
@@ -260,34 +215,33 @@ private void RunYcsbSafeContext(int thread_idx)
                 for (long idx = chunk_idx; idx < chunk_idx + YcsbConstants.kChunkSize && !done; ++idx)
                 {
                     if (idx % 512 == 0)
-                        session.CompletePending(false);
+                        bContext.CompletePending(false);
```
badrishc, line 218: does this use of wrapper affect perf numbers?

libs/storage/Tsavorite/cs/benchmark/Functions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Tsavorite.benchmark
 {
-    public struct Functions : IFunctions<Key, Value, Input, Output, Empty>
+    public struct Functions : ISessionFunctions<Key, Value, Input, Output, Empty>
```
badrishc, line 9: rename to `struct SessionFunctions`

badrishc (approved):

libs/server/AOF/AofProcessor.cs
TedHartMS: Changing to add XxxContext.Session; this is primarily used for Dispose() and also to get Session.ID and Session.Version

libs/storage/Tsavorite/cs/Tsavorite.sln
TedHartMS: Yes; we previously built x64 before an earlier csproj reorganization PR. That was a good change so we now always build AnyCPU, but it revealed that this was broken (it's why we were seeing that "Configuration incorrect" message when loading Tsavorite.sln)

libs/storage/Tsavorite/cs/benchmark/FixedLenYcsbBenchmark.cs
TedHartMS: Other way around--this actually removed a wrapper (before, bContext redirected to ClientSession)

PR #439, opened 2024-06-03
Fix some TimeSpan duration code

kevin-montrose:
Noticed while working on ACLs, these uses of `TimeSpan.Seconds` and `TimeSpan.Days` appear incorrect.

Switched to using the `TimeSpan.TotalXXX` equivalents which return total duration in whatever unit, not just the "component" bit of the `TimeSpan`.


vazois (approved):
LGTM!

PR #442, opened 2024-06-04
Missing configuration switch in dotnet test

TalZaccai:
Adding missing configuration switch to dotnet test steps in CI


darrenge (approved):
Thanks for fixing. Nice catch.

PR #443, opened 2024-06-05
Removing key from object store when collection object it empty

TalZaccai:
* Fixing bug where keys of emptied collections are not removed from object store (this currently prevents reuse of keys to store object of different type)
* Fixing additional bug in RecordStatus not using the correct mask


libs/server/Objects/List/ListObject.cs
```diff
@@ -57,6 +57,7 @@ public enum OperationDirection
     /// </summary>
     public partial class ListObject : GarnetObjectBase
     {
+        /// <inheritdoc />
         readonly LinkedList<byte[]> list;
```
TedHartMS, line 61: Inheritdoc from where? Same question for SetObject

libs/server/Objects/SortedSetGeo/SortedSetGeoObjectImpl.cs
```diff
@@ -15,7 +15,7 @@ namespace Garnet.server
     /// <summary>
     /// Sorted Set - RESP specific operations for GEO Commands
     /// </summary>
-    public unsafe partial class SortedSetObject : GarnetObjectBase
+    public unsafe partial class SortedSetObject : IGarnetObject
     {
```
TedHartMS, line 19: why this change?

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -30,7 +30,7 @@ unsafe GarnetStatus RMWObjectStoreOperation<TObjectContext>(byte[] key, ArgSlice
             Debug.Assert(_output.spanByteAndMemory.IsSpanByte);
 
 
-            if (!status.Record.Created && !status.Record.CopyUpdated && !status.Record.InPlaceUpdated)
+            if (status.NotFound || (!status.Record.Created && !status.Record.CopyUpdated && !status.Record.InPlaceUpdated))
                 return GarnetStatus.NOTFOUND;
```
TedHartMS, line 34: We should not need CopyUpdated or InPlaceUpdated to be mentioned here; NotFound cannot be true if they are present. Look at RMW_MainStore (AdvancedOps.cs); it is much more straightforward.

           return status.Found || status.Record.Created ? GarnetStatus.OK : GarnetStatus.NOTFOUND;

libs/server/Objects/List/ListObject.cs
TalZaccai: Sorry, this was related to something I did before & forgot to remove, deleting these...

libs/server/Objects/SortedSetGeo/SortedSetGeoObjectImpl.cs
TalZaccai: Undid this

TedHartMS (approved):

PR #450, opened 2024-06-07
Adding type validation logic to object operations

TalZaccai:
- [x] List commands
- [x] Set commands
- [x] SortedSet commands
- [x] Hash commands
- [x] Geo commands
- [x] Shared commands
- [x] Custom object commands


libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -398,19 +391,57 @@ internal unsafe GarnetStatus SetMove(ArgSlice sourceKey, ArgSlice destinationKey
...
                 }
+                else
+                {
+                    dstSetObject = new SetObject();
+                }
+
+                var arrMember = member.ToArray();
+
+                var removed = srcSetObject.Set.Remove(arrMember);
+                if (!removed) return GarnetStatus.OK;
```
TedHartMS, line 428: Do these *Move operations need a "bool removeSrcKey" if they go empty?

TedHartMS (approved):

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -289,11 +315,9 @@ private unsafe bool SortedSetRange<TGarnetApi>(RespCommand command, int count, b
                     //command partially executed
                     return false;
                 }
-                else
-                {
-                    while (!RespWriteUtils.WriteNull(ref dcurr, dend))
-                        SendAndReset();
-                }
+
+                while (!RespWriteUtils.WriteNull(ref dcurr, dend))
```
yrajas, line 319: >while (!RespWriteUtils.WriteNull(ref dcurr, dend)) [](http://example.com/codeflow?start=16&length=50)

Shouldn't it be wrong # of args?

TalZaccai: Good catch!

libs/server/Storage/Session/ObjectStore/SetOps.cs
TalZaccai: Will resolve this an other similar scenarios in a separate PR

PR #456, opened 2024-06-09
Fix Parsing Regression in GarnetClient and GarnetClientSession

vazois:
Fix regression introduce in 52024990, for which RespBenchOnline was breaking when GET returned null.

Repro:

<server> dotnet run -c Release --framework=net8.0
<client> dotnet run -c Release --framework=net8.0 -- --online --batchsize=1 --op-workload=GET --op-percent=100 --client=GarnetClientSession --threads=1 --runtime=-1

Result:


badrishc (approved):

libs/common/RespReadUtils.cs
```diff
@@ -624,14 +624,14 @@ public static bool ReadStringWithLengthHeader(out string result, ref byte* ptr,
         /// <summary>
         /// Read string with length header
         /// </summary>
-        public static bool ReadStringWithLengthHeader(MemoryPool<byte> pool, out MemoryResult<byte> result, ref byte* ptr, byte* end)
+        public static bool ReadStringWithLengthHeader(MemoryPool<byte> pool, out MemoryResult<byte> result, ref byte* ptr, byte* end, bool allowNull = false, bool isArray = false)
```
lmaas, line 627: I don't think `isArray` should ever be true when reading a string. The parameter is used to indicate whether we are reading a string length header or an array length header, but in this case we are always expecting a string length header.

badrishc: Remove isArray, and any other `if` checks that were added solely for the purpose of code reuse. This is inner loop so keeping minimal is preferred.

vazois: I removed isArray where it was not needed and created a dedicated ReadStringResponseWithLengthHeader for the client

libs/common/RespReadUtils.cs
```diff
@@ -212,15 +211,46 @@ private static bool TryReadInt(ref byte* ptr, byte* end, out int value, out ulon
...
+            if (ptr + 3 > end)
+                return false;
+
+            var readHead = ptr + 1;
+            var negative = *readHead == '-';
+
+            if (!ReadSignedLengthHeader(out length, ref ptr, end, isArray))
+                return false;
+
+            if (negative)
```
TalZaccai, line 234: You can put this before the ReadSignedLengthHeader call

libs/common/RespReadUtils.cs
```diff
@@ -336,8 +362,7 @@ public static bool Read64Int(out long number, ref byte* ptr, byte* end)
         /// <returns>True if a length header was successfully read.</returns>
 
         public static bool ReadArrayLength(out int length, ref byte* ptr, byte* end)
```
TalZaccai, line 364: Can this be called with a null array? Then the length would be -1

libs/common/RespReadUtils.cs
```diff
@@ -1040,5 +923,171 @@ public static bool ReadSerializedData(out byte[] key, out byte[] value, out long
...
+        #region ClientSpecific
+        /// <summary>
+        /// Try to read a RESP formatted bulk string
+        /// NOTE: This is used with client implementation to parse responses that may include a null value (i.e. $-1\r\n)
+        /// </summary>
+        /// <param name="result"></param>
+        /// <param name="ptr"></param>
+        /// <param name="end"></param>
+        /// <returns></returns>
+        public static bool ReadStringResponseWithLengthHeader(out string result, ref byte* ptr, byte* end)
```
TalZaccai, line 936: Looks pretty similar to the next overload except for the setting of result, maybe extract the logic to a helper function?

TalZaccai (changes requested):

libs/common/RespReadUtils.cs
vazois: Good point, I just pushed a fix.

libs/common/RespReadUtils.cs
vazois: I updated this too in the latest commit. Please take a look.

libs/common/RespReadUtils.cs
vazois: Pushed a fix on this

TalZaccai (approved):

PR #461, opened 2024-06-12
Only truncate AOF if all active stores are checkpointed

chenhao-ye:
AOF is truncated after checkpointing. However, truncation should only happen if both stores are checkpointed, since they share the same AOF. The current implementation can lose data if `storeType` is not `StoreType.All`.

This bug so far is not triggered because every caller is using `StoreType.All`.

### DIFF
--- a/libs/server/StoreWrapper.cs
+++ b/libs/server/StoreWrapper.cs
@@ -719,15 +719,19 @@ namespace Garnet.server
                     objectStoreCheckpointResult = await objectStore.TakeHybridLogCheckpointAsync(checkpointType, tryIncremental);
             }
 
-            // If cluster is enabled the replication manager is responsible for truncating AOF
-            if (serverOptions.EnableCluster && serverOptions.EnableAOF)
+            // Only truncate AOF if both stores are checkpointed
+            if (storeType == StoreType.All)
             {
-                clusterProvider.SafeTruncateAOF(storeType, full, CheckpointCoveredAofAddress, storeCheckpointResult.token, objectStoreCheckpointResult.token);
-            }
-            else
-            {
-                appendOnlyFile?.TruncateUntil(CheckpointCoveredAofAddress);
-                appendOnlyFile?.Commit();
+                // If cluster is enabled the replication manager is responsible for truncating AOF
+                if (serverOptions.EnableCluster && serverOptions.EnableAOF)
+                {
+                    clusterProvider.SafeTruncateAOF(storeType, full, CheckpointCoveredAofAddress, storeCheckpointResult.token, objectStoreCheckpointResult.token);
+                }
+                else
+                {
+                    appendOnlyFile?.TruncateUntil(CheckpointCoveredAofAddress);
+                    appendOnlyFile?.Commit();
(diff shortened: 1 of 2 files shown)


chenhao-ye:

libs/server/StoreWrapper.cs
```diff
@@ -693,15 +693,19 @@ private async Task InitiateCheckpoint(bool full, CheckpointType checkpointType,
                     objectStoreCheckpointResult = await objectStore.TakeHybridLogCheckpointAsync(checkpointType, tryIncremental);
             }
 
-            // If cluster is enabled the replication manager is responsible for truncating AOF
-            if (serverOptions.EnableCluster && serverOptions.EnableAOF)
+            // Only truncate AOF if both stores are checkpointed
+            if (storeType == StoreType.All)
```
vazois, line 697: The logic should be slightly more complex, since some people might be running Garnet with only the main store which means we can truncate the log safely even if the storeType == StoreType.MainStore

libs/server/StoreWrapper.cs
```diff
@@ -693,15 +693,19 @@ private async Task InitiateCheckpoint(bool full, CheckpointType checkpointType,
                     objectStoreCheckpointResult = await objectStore.TakeHybridLogCheckpointAsync(checkpointType, tryIncremental);
             }
 
-            // If cluster is enabled the replication manager is responsible for truncating AOF
-            if (serverOptions.EnableCluster && serverOptions.EnableAOF)
+            // Only truncate AOF if both stores are checkpointed
```
vazois, line 696:
```suggestion
            // Checkpoint completed here so we can safely truncate the AOF
            TruncateAOF();

            if (objectStore != null)
            {
                // During the checkpoint, we may have serialized Garnet objects in (v) versions of objects.
                // We can now safely remove these serialized versions as they are no longer needed.
                using (var iter1 = objectStore.Log.Scan(objectStore.Log.ReadOnlyAddress, objectStore.Log.TailAddress, ScanBufferingMode.SinglePageBuffering, includeSealedRecords: true))
                {
                    while (iter1.GetNext(out _, out _, out var value))
                    {
                        if (value != null)
                            ((GarnetObjectBase)value).serialized = null;
                    }
                }
            }

            logger?.LogInformation("Completed checkpoint");

            void TruncateAOF()
            {
                if (!serverOptions.EnableAOF)
                    return;

                if (serverOptions.EnableCluster)
                {
                    clusterProvider.SafeTruncateAOF(storeType, full, CheckpointCoveredAofAddress, storeCheckpointResult.token, objectStoreCheckpointResult.token);
                }
                else
                {
                    // Truncate if all stores have been checkpointed or if main store has been checkpointed and instance is running with disabled object store
                    if (storeType == StoreType.All || (storeType == StoreType.Main && serverOptions.DisableObjects))
                    {
                        appendOnlyFile?.TruncateUntil(CheckpointCoveredAofAddress);
                        appendOnlyFile?.Commit();
                    }
                }
            }
```

chenhao-ye:
Comments are addressed.

PR #462, opened 2024-06-12
Fix SafeAofAddress

chenhao-ye:
1. Fix the typo: get `SafeAofAddress` from the right store type `StoreType.Object`.
2. Fix the conditional logic of `DisableObjects`.


vazois (approved):
Good catch!

PR #463, opened 2024-06-13
Support MODULE LOADCS command

yrajas:
Adds support for MODULE LOADCS command.

This initial version supports .net modules.
Adds ability to register a module with its name and version number.
As part of initialization, a module can register commands and transactions that it implements.


libs/server/Module/ModuleRegistrar.cs
```diff
@@ -0,0 +1,207 @@
...
+    /// Common interface that all Garnet modules must implement.
+    /// </summary>
+    public interface IModule
+    {
+        /// <summary>
+        /// Called when the module is loaded.
+        /// </summary>
+        /// <param name="context">Module load context</param>
+        /// <param name="moduleArgs">Module load args</param>
+        void OnLoad(ModuleLoadContext context, List<string> moduleArgs);
```
badrishc, line 24: Should be `string[] args` - zero-length if there are no args in the common case.

libs/server/Resp/ArrayCommands.cs
```diff
@@ -795,16 +806,64 @@ private bool NetworkTYPE<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storag
...
                 return AbortWithWrongNumberOfArguments("MODULE", count);
 
-            // TODO: pending implementation for module support.
-            while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_UNK_CMD, ref dcurr, dend))
-                SendAndReset();
+            // Read sub-command
+            if (!RespReadUtils.TrySliceWithLengthHeader(out var subCommand, ref ptr, recvBufferPtr + bytesRead))
+                return false;
 
+            if (subCommand.SequenceEqual(CmdStrings.LOAD) || subCommand.SequenceEqual(CmdStrings.load))
```
badrishc, line 819: Should we call this `MODULE LOAD` since existing clients might try to load non-compliant modules from other systems with this call? Should it instead be `MODULE LOADCS`?

test/Garnet.test/RespModuleTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                Path.GetFullPath(@"../main/GarnetServer/Extensions/SetIfPM.cs", TestUtils.RootTestsProjectPath),
+                Path.GetFullPath(@"../main/GarnetServer/Extensions/ReadWriteTxn.cs", TestUtils.RootTestsProjectPath)};
+            TestUtils.CreateTestLibrary(null, referenceFiles, filesToCompile, modulePath);
+            return modulePath;
+        }
+
+        [Test]
+        public void TestModuleLoad()
+        {
+            var onLoad =
```
badrishc, line 92: Use verbatim string here for readability.

test/Garnet.test/RespModuleTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                    "            context.Initialize(\"TestModule\", 1); " +
+                    "            context.RegisterCommand(\"TestModule.SetIfPM\", 2, CommandType.ReadModifyWrite, new SetIfPMCustomCommand()," +
+                    "            new RespCommandsInfo{ Name = \"TestModule.SETIFPM\", Arity = 4, FirstKey = 1, LastKey = 1, Step = 1," +
```
badrishc, line 95: Why is Arity set to 4, and the second argument earlier is set to 2?

libs/server/Module/ModuleRegistrar.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.Logging;
+
+namespace Garnet.server.Module
+{
+    /// <summary>
+    /// Common interface that all Garnet modules must implement.
+    /// </summary>
+    public interface IModule
```
badrishc, line 17: Use base class instead of interface for future extensibility?

badrishc (changes requested):

TalZaccai (changes requested):

test/Garnet.test/RespModuleTests.cs
yrajas: Arity includes command name and key as well. That's why the difference in count.

libs/server/Resp/ArrayCommands.cs
yrajas: Changed the command to LOADCS.

libs/server/Resp/ArrayCommands.cs
```diff
@@ -795,16 +806,64 @@ private bool NetworkTYPE<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storag
...
+            {
+                if (count < 2) // At least module path is required
+                    return AbortWithWrongNumberOfArguments("MODULE LOAD", count);
+
+                // Read path to module file
+                if (!RespReadUtils.TrySliceWithLengthHeader(out var modulePath, ref ptr, recvBufferPtr + bytesRead))
```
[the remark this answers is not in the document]
yrajas: Moved to using parseState

libs/server/Resp/ArrayCommands.cs
```diff
@@ -795,16 +806,64 @@ private bool NetworkTYPE<TGarnetApi>(int count, byte* ptr, ref TGarnetApi storag
...
+                    Debug.Assert(assembliesList.Count == 1, "Only one assembly per module load");
+
+                    if (ModuleRegistrar.Instance.LoadModule(customCommandManager, assembliesList[0], moduleArgs, logger, out errorMsg))
+                    {
+                        while (!RespWriteUtils.WriteDirect(CmdStrings.RESP_OK, ref dcurr, dend))
+                            SendAndReset();
+                    }
+                }
+
+                if (errorMsg != default)
```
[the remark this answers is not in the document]
yrajas: Accounts for error from the inner LoadModule call.

TalZaccai (approved):

badrishc (approved):

Issue #460, opened 2024-06-11
Unpin PoolEntry items in LimitedFixedBufferPool class?


AliKhalili:
### Describe the bug

I recently came across the `LimitedFixedBufferPool` class and noticed that its [`Dispose`](https://github.com/microsoft/garnet/blob/5916135b507087a7f3b83d35b4ed464b8695d468/libs/common/Memory/LimitedFixedBufferPool.cs#L107) method doesn't dispose of individual `PoolEntry` items after [dequeuing ](https://github.com/microsoft/garnet/blob/5916135b507087a7f3b83d35b4ed464b8695d468/libs/common/Memory/LimitedFixedBufferPool.cs#L126)them. This isn't a major issue, but if we're using handles to pin entry instances, it would be better to unpin these handles when disposing.
Is there a reason behind that?

### Steps to reproduce the bug

n/a

vazois:
Thanks for bringing this too our attention. We are not currently using the handles option for the PoolEntry.
I created PR #465 to remove it all together to avoid any confusion and possible misuse in the future.

PR #465, opened 2024-06-13
Remove unused useHandlesForPin

vazois:
This PR resolves issue #460 by removing useHandlesForPin option.
This option is not actively being used. Instead, we rely on GC.AllocateArray to pin memory.

### DIFF
--- a/libs/common/Memory/PoolEntry.cs
+++ b/libs/common/Memory/PoolEntry.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 
 namespace Garnet.common
 {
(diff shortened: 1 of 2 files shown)


libs/common/Memory/PoolEntry.cs
```diff
@@ -23,38 +23,27 @@ public unsafe class PoolEntry : IDisposable
         /// </summary>
         public byte* entryPtr;
 
-        readonly bool useHandlesForPin;
         GCHandle handle;
```
TalZaccai, line 26: Note the warnings about this not being used

TalZaccai (changes requested):
Added one comment

libs/common/Memory/PoolEntry.cs
vazois: Pushed an update

TalZaccai (approved):

PR #466, opened 2024-06-14
Empty collection key deletion - continued

TalZaccai:
* Implemented key deletion for emptied collection when RMW is not called (and underlying object is directly altered)
* Extended tests to test that keys are being removed for each command that may empty the object collection

(Continues the fixes introduced in PR #443)

### DIFF
--- a/libs/server/Storage/Session/ObjectStore/SetOps.cs
+++ b/libs/server/Storage/Session/ObjectStore/SetOps.cs
@@ -718,7 +718,7 @@ namespace Garnet.server
                         _ = EXPIRE(destination, TimeSpan.Zero, out _, StoreType.Object, ExpireOption.None,
                             ref lockableContext, ref setObjectStoreLockableContext);
                     }
-                    
+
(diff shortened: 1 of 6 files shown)


libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -686,13 +702,23 @@ public GarnetStatus SetUnionStore(byte[] key, ArgSlice[] keys, out int count)
 
                 if (status == GarnetStatus.OK)
                 {
-                    var newSetObject = new SetObject();
-                    foreach (var item in members)
+                    if (members.Count > 0)
```
badrishc, line 705: How can union result in an empty set?

TalZaccai: The internal SetUnion method will return an empty HashSet if all the keys don't exist

badrishc (approved):

PR #468, opened 2024-06-17
Update default port to 6379

badrishc:
* Update default port to 6379
* Change Dockerfile to make entrypoint devoid of override arguments as this causes confusion when reading from conf
* Fix Object Store read methods to unify SingleReader and ConcurrentReader

### DIFF
--- a/libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
+++ b/libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
@@ -15,25 +15,6 @@ namespace Garnet.server
         /// <inheritdoc />
         public bool SingleReader(ref byte[] key, ref SpanByte input, ref IGarnetObject value, ref GarnetObjectStoreOutput dst, ref ReadInfo readInfo)
         {
-            if (value.Expiration > 0 && value.Expiration < DateTimeOffset.Now.Ticks)
-                return false;
-
-            var header = (RespInputHeader*)input.ToPointer();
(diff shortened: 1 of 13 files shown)


yrajas (approved):

libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
```diff
@@ -58,5 +39,9 @@ public bool ConcurrentReader(ref byte[] key, ref SpanByte input, ref IGarnetObje
             dst.garnetObject = value;
             return true;
         }
+
+        /// <inheritdoc />
+        public bool ConcurrentReader(ref byte[] key, ref SpanByte input, ref IGarnetObject value, ref GarnetObjectStoreOutput dst, ref ReadInfo readInfo, ref RecordInfo recordInfo)
```
TalZaccai, line 44: Why not consolidate to one single method if they're both doing the same thing?

TalZaccai (changes requested):
Added one question

libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
badrishc: It is consolidated into one method (SingleReader). ConcurrentReader calls into SingleReader now.

badrishc: Tsavorite calls into these under different circumstances, so they both need to be in here. They are part of the interface.

PR #471, opened 2024-06-18
Fix CLUSTER REPLICATE error handling

kevin-montrose:
`CLUSTER REPLICATE` would return `+OK` even if an error was encountered.

Adds a test for the failure, which replicates the conditions under which it was found (scripting spinning up a two node cluster).

### DIFF
--- a/test/Garnet.test.cluster/ClusterReplicationTests.cs
+++ b/test/Garnet.test.cluster/ClusterReplicationTests.cs
@@ -1009,7 +1009,7 @@ namespace Garnet.test.cluster
 
         [Test, Order(21)]
         [Category("REPLICATION")]
-        public void FailedReplication()
+        public void ClusterReplicateFails()
         {
             const string UserName = "temp-user";
             const string Password = "temp-password";
--- a/test/Garnet.test.cluster/ClusterReplicationTests.cs
+++ b/test/Garnet.test.cluster/ClusterReplicationTests.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


test/Garnet.test.cluster/ClusterReplicationTests.cs
```diff
@@ -1005,5 +1006,40 @@ void ClusterDivergentReplicasTest(bool performRMW, bool disableObjects, bool ckp
...
+        [Category("REPLICATION")]
+        public void ClusterReplicateFails()
+        {
+            const string UserName = "temp-user";
+            const string Password = "temp-password";
+
+            const string ClusterUserName = "cluster-user";
+            const string ClusterPassword = "cluster-password";
+
+            // setup a cluster (mimicing the style in which this bug was first found)
```
vazois, line 1020:
```suggestion
// Setup a cluster (mimicking the style in which this bug was first found)
```

test/Garnet.test.cluster/ClusterReplicationTests.cs
```diff
@@ -1005,5 +1006,40 @@ void ClusterDivergentReplicasTest(bool performRMW, bool disableObjects, bool ckp
...
+            using var primaryConnection = ConnectionMultiplexer.Connect($"{primaryEndpoint.Address}:{primaryEndpoint.Port},user={UserName},password={Password}");
+            var primaryServer = primaryConnection.GetServer(primaryEndpoint);
+
+            Assert.AreEqual("OK", (string)primaryServer.Execute("CLUSTER", ["ADDSLOTSRANGE", "0", "16383"], flags: CommandFlags.NoRedirect));
+            Assert.AreEqual("OK", (string)primaryServer.Execute("CLUSTER", ["MEET", replicaEndpoint.Address.ToString(), replicaEndpoint.Port.ToString()], flags: CommandFlags.NoRedirect));
+
+            using var replicaConnection = ConnectionMultiplexer.Connect($"{replicaEndpoint.Address}:{replicaEndpoint.Port},user={UserName},password={Password}");
+            var replicaServer = replicaConnection.GetServer(replicaEndpoint);
+
+            // try to replicate from a server that doesn't exist
```
vazois, line 1040:
```suggestion
            // Try to replicate from a server that doesn't exist
```

vazois (approved):

PR #479, opened 2024-06-20
Memory related improvements

badrishc:
- [x] Do not free pages in the TsavoriteLog if not needed to do so, in circular buffer, during the Reset operation
- [x] Make sure NetworkHandler grows the network buffer only if absolutely needed
- [x] Avoid spinning on sync path while waiting for async TLS stream processing to complete


libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -617,14 +617,15 @@ private bool StringBitField<TGarnetApi>(int count, byte* ptr, ref TGarnetApi sto
 
             for (int i = 0; i < secondaryCmdCount; i++)
             {
+                /* Commenting due to excessive verbosity
```
yrajas, line 620: May be these could be changed to more verbose log levels like Trace.

badrishc: even Trace is too much for these, seems like. we cannot log every interaction.

PR #482, opened 2024-06-21
Add compaction docs, update website, add compaction switch

badrishc:
- [x] Update website
- [x] Add compaction docs
- [x] Add switch option for `CompactionForceDelete`


libs/server/StoreWrapper.cs
```diff
@@ -465,6 +470,24 @@ void DoCompaction(int mainStoreMaxSegments, int objectStoreMaxSegments, int numS
             }
         }
 
+        void CompactionCommitAof()
+        {
+            // If we are the primary, we commit the AOF
+            // If we are the replica, we commit the AOF if fast commit is disabled
```
yrajas, line 476: Could help with readability to have a comment here on the why part of this.

yrajas (approved):

PR #489, opened 2024-06-23
Remove `Worker.ReplicationOffset` in `ClusterConfig`

chenhao-ye:
Remove the field `ReplicationOffset` from `struct Worker` for the following reasons:
1. It is barely used: Only the local worker (`ClusterConfig.workers[1]`) carries a meaningful replication offset, other workers have `ReplicationOffset = 0` upon initialization but never being updated later (this field is not updated during gossip).
2. It is redundant: the local replication offset is already available in replication info (by command `INFO`).
3. Most importantly, it does not match the semantics of `ClusterConfig`: the cluster configuration is not expected to be updated in a steady state (i.e., no cluster membership changes) while replication offset can change rapidly as long as there is new data. The cluster epoch should not be updated simply because there is new data being replicated.

### DIFF
--- a/libs/cluster/Server/ClusterConfig.cs
+++ b/libs/cluster/Server/ClusterConfig.cs
@@ -575,7 +575,7 @@ namespace Garnet.cluster
 
         private string CreateFormattedNodeInfo(int workerId)
         {
-            var nodeInfo = "*12\r\n";
+            var nodeInfo = "*10\r\n";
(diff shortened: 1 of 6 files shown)


chenhao-ye:
Redis `CLUSTER SHARDS` will return the replication offset, so this change may break the compatibility with Redis.

PR #490, opened 2024-06-24
Support cancellation for cache size resizer task.

yrajas:
Object store size resizer task is currently not cancellable when the server is getting disposed.
This change fixes that by having the resizer task accepting a cancellation token to which cancellation is requested when the server object is disposed.


libs/server/Storage/SizeTracker/CacheSizeTracker.cs
```diff
@@ -75,10 +76,10 @@ public CacheSizeTracker(TsavoriteKV<byte[], IGarnetObject> store, LogSettings lo
             }
         }
 
-        public void Start()
+        public void Start(CancellationToken token)
         {
-            mainLogTracker.Start();
-            readCacheTracker?.Start();
+            mainLogTracker.Start(token);
```
badrishc, line 81: why is there a main log tracker always? arent' there cases where the main log is fixed in size for the server in which case we can avoid having the tracking running.

libs/storage/Tsavorite/cs/src/core/Index/Common/LogSizeTracker.cs
```diff
@@ -151,16 +152,16 @@ public void IncrementSize(long size)
         /// Performs resizing by waiting for an event that is signaled whenever memory utilization changes.
         /// This is invoked on the threadpool to avoid blocking calling threads during the resize operation.
         /// </summary>
-        async Task ResizerTask()
+        async Task ResizerTask(CancellationToken token)
         {
-            while (true)
+            while (!token.IsCancellationRequested)
```
badrishc, line 157: when we leave the while loop, we usually set some variable, that the main dispose can check to make sure dispose is completed. i believe this is done in a few other places.

libs/server/Storage/SizeTracker/CacheSizeTracker.cs
yrajas: The containing class CacheSizeTracker is instantiated only when Object Store total memory size config is set. Here, the mainLogTracker variable is to track the size of the object store size whereas the readCacheTracker is for its corresponding read cache that is optional based on if read cache has been enabled. So, whenever object store managed memory size is to be tracked that is done using the tracker called 'mainLogTracker'.

libs/storage/Tsavorite/cs/src/core/Index/Common/LogSizeTracker.cs
yrajas: Updated to have the StoreWrapper dispose method wait for the resizer task cancellation to complete.

badrishc (approved):

PR #508, opened 2024-07-02
Consolidate Cluster Slot Verification

vazois:
This PR is a first attempt of moving slot verification outside the command execution methods


libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -321,17 +321,52 @@ cmd switch
...
+                RespCommand.MIGRATE => false,
+                RespCommand.DBSIZE => false,
+                RespCommand.MEMORY_USAGE => false,
+                RespCommand.FLUSHDB => false,
+                _ => cmd >= FirstReadCommand() && cmd <= LastWriteCommand()
+            };
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsWriteOnly(this RespCommand cmd)
```
TalZaccai, line 354: This approach seems fairly error-prone to me. Have we thought about alternative approaches? Maybe adding a read-only flag?

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -30,24 +30,40 @@ bool NetworkSingleKeySlotVerify(ReadOnlySpan<byte> key, bool readOnly)
...
+
+            csvi.readOnly = cmd.IsReadOnly();
             csvi.sessionAsking = SessionAsking;
-            csvi.firstKey = firstKey;
-            csvi.lastKey = lastKey < 0 ? parseState.count + 1 + lastKey : lastKey;
-            csvi.step = step;
-            return clusterSession.NetworkMultiKeySlotVerify(ref parseState, ref csvi, ref dcurr, ref dend);
+            csvi.firstKey = cmd switch
+            {
+                RespCommand.ZDIFF => 1, // ZDIFF first key comes after keyCount parameter
```
TalZaccai, line 56: This info of first key and last key can come from the command info. ZDIFF is not the only command in which the first key is in the 2nd position... We should pre-populate this information prior to calling this method.

TalZaccai (changes requested):

libs/server/Resp/Parser/RespCommand.cs
vazois: This piece was there from the ACL refactor; I just moved it under a common method for people to re-use it. I am actually using the IsReadOnly method myself. Can we have IsReadOnly as part of the command info instead?

libs/server/Resp/RespServerSessionSlotVerify.cs
vazois: RespCommandsInfo.json seems to be wrong for ZDIFF, it indicates that the first key is at position zero. I assumed that the data in the json file is auto-generated so I opted out to do it at runtime. If we can manually edit I can fix the issue. Please confirm.

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -30,24 +31,99 @@ bool NetworkSingleKeySlotVerify(ReadOnlySpan<byte> key, bool readOnly)
...
+                    throw new GarnetException("KeySpecification unknown count");
+            }
+            csvi.readOnly = cmd.IsReadOnly();
             csvi.sessionAsking = SessionAsking;
-            csvi.firstKey = firstKey;
-            csvi.lastKey = lastKey < 0 ? parseState.count + 1 + lastKey : lastKey;
-            csvi.step = step;
-            return clusterSession.NetworkMultiKeySlotVerify(ref parseState, ref csvi, ref dcurr, ref dend);
+
+            //csvi.firstKey = cmd switch
```
TalZaccai, line 114: Can we remove these comments?

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -30,24 +31,99 @@ bool NetworkSingleKeySlotVerify(ReadOnlySpan<byte> key, bool readOnly)
...
+                // This only happens if we failed to parse the json file
                 return false;
-            csvi.readOnly = readOnly;
+
+            // The provided command is not a data command
+            // so we can serve without any slot restrictions
+            if (commandInfo == null)
+                return true;
+
+            var specs = commandInfo.KeySpecifications;
```
TalZaccai, line 54: Might be neater to have GetFirstKey/GetLastKey/GetKeyStep abstract methods as part of FindKeysKeySpecMethodBase and implement them for each FindKeys type. WDYT?

TalZaccai (changes requested):

libs/server/Resp/RespServerSessionSlotVerify.cs
vazois: I am closing this, since it no longer applies since we are using KeySpecifications

libs/server/Resp/RespServerSessionSlotVerify.cs
vazois: Done, thanks!

libs/server/Resp/RespServerSessionSlotVerify.cs
vazois: We could do that, but it will require evaluating the switch statement multiple times per Get call. If you want to move it towards the FindKeysBase it might be best to have GetSlotVerificationInput instead of having different calls.

TalZaccai (approved):

PR #511, opened 2024-07-05
Remove AOF commit blocking on State Invariant or Aof Independent Commands

hamdaankhalid:
This PR removes the need for commands of a certain type to wait for commits on the AOF log before sending response back to client.

The commands that qualify for no needing to be blocked on AOF are commands that are either:
1. State Invariant Commands that return static data (PING, ECHO, etc)
2.  Commands that are a part of a system that does not depend on the AOF logging at all. (ACL, Latency, etc)

These commands do not need to be blocked on WaitForCommit to complete if there are other RespServerSessions concurrently enqueueing commits to the shared AOF across the sessions.

Benchmark #s:
## My Branch
| Method       | Job    | EnvironmentVariables | Runtime  | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------- |------- |--------------------- |--------- |-----------:|----------:|----------:|-------:|----------:|
| InlinePing   | .NET 6 | Empty                | .NET 6.0 |   1.846 us | 0.0303 us | 0.0349 us |      - |         - |
| Set          | .NET 6 | Empty                | .NET 6.0 |  21.414 us | 0.3717 us | 0.3477 us |      - |         - |
| SetEx        | .NET 6 | Empty                | .NET 6.0 |  27.847 us | 0.5397 us | 0.5049 us |      - |         - |
| Get          | .NET 6 | Empty                | .NET 6.0 |  13.138 us | 0.2342 us | 0.2077 us |      - |         - |
| ZAddRem      | .NET 6 | Empty                | .NET 6.0 | 137.701 us | 1.5651 us | 1.3874 us | 0.2441 |   23552 B |
| LPushPop     | .NET 6 | Empty                | .NET 6.0 |  13.344 us | 0.1122 us | 0.0937 us |      - |    1080 B |
| SAddRem      | .NET 6 | Empty                | .NET 6.0 | 120.131 us | 1.8166 us | 1.6104 us | 0.2441 |   16384 B |
| HSetDel      | .NET 6 | Empty                | .NET 6.0 | 153.112 us | 3.0034 us | 3.4588 us |      - |   55297 B |
| MyDictSetGet | .NET 6 | Empty                | .NET 6.0 | 198.118 us | 2.4954 us | 2.2121 us | 0.4883 |   30720 B |
| InlinePing   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |   1.855 us | 0.0202 us | 0.0189 us |      - |         - |
| Set          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  14.418 us | 0.1550 us | 0.1450 us |      - |         - |
| SetEx        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  21.934 us | 0.4370 us | 0.8420 us |      - |         - |
| Get          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  10.417 us | 0.0508 us | 0.0475 us |      - |         - |
| ZAddRem      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 118.714 us | 1.6122 us | 1.3463 us | 0.3662 |   23552 B |
| LPushPop     | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  13.024 us | 0.1262 us | 0.1119 us | 0.0153 |    1080 B |
| SAddRem      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 101.434 us | 1.3466 us | 1.1938 us | 0.2441 |   16384 B |
| HSetDel      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 132.911 us | 2.1731 us | 1.8146 us |      - |   55297 B |
| MyDictSetGet | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 177.361 us | 1.7671 us | 1.6530 us | 0.4883 |   30720 B |

## Main Branch 
| Method       | Job    | EnvironmentVariables | Runtime  | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------- |------- |--------------------- |--------- |-----------:|----------:|----------:|-------:|----------:|
| InlinePing   | .NET 6 | Empty                | .NET 6.0 |   1.919 us | 0.0382 us | 0.0870 us |      - |         - |
| Set          | .NET 6 | Empty                | .NET 6.0 |  21.689 us | 0.4274 us | 0.5248 us |      - |         - |
| SetEx        | .NET 6 | Empty                | .NET 6.0 |  28.035 us | 0.4881 us | 0.5223 us |      - |         - |
| Get          | .NET 6 | Empty                | .NET 6.0 |  13.688 us | 0.2669 us | 0.2855 us |      - |         - |
| ZAddRem      | .NET 6 | Empty                | .NET 6.0 | 140.423 us | 2.2635 us | 2.6946 us | 0.2441 |   23552 B |
| LPushPop     | .NET 6 | Empty                | .NET 6.0 |  14.147 us | 0.2563 us | 0.2272 us | 0.0153 |    1080 B |
| SAddRem      | .NET 6 | Empty                | .NET 6.0 | 117.170 us | 2.1330 us | 1.8909 us | 0.2441 |   16384 B |
| HSetDel      | .NET 6 | Empty                | .NET 6.0 | 157.688 us | 2.2019 us | 1.9520 us |      - |   55297 B |
| MyDictSetGet | .NET 6 | Empty                | .NET 6.0 | 205.448 us | 2.2899 us | 1.9122 us | 0.4883 |   30720 B |
| InlinePing   | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |   1.928 us | 0.0380 us | 0.0438 us |      - |         - |
| Set          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  14.968 us | 0.1696 us | 0.1416 us |      - |         - |
| SetEx        | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  21.868 us | 0.4071 us | 0.3808 us |      - |         - |
| Get          | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  11.177 us | 0.2232 us | 0.2388 us |      - |         - |
| ZAddRem      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 126.285 us | 2.5125 us | 5.0754 us | 0.2441 |   23552 B |
| LPushPop     | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 |  13.180 us | 0.1857 us | 0.1646 us | 0.0153 |    1080 B |
| SAddRem      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 100.532 us | 1.0271 us | 0.9105 us | 0.2441 |   16384 B |
| HSetDel      | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 135.303 us | 1.2437 us | 1.0385 us |      - |   55297 B |
| MyDictSetGet | .NET 8 | DOTNET_TieredPGO=0   | .NET 8.0 | 177.669 us | 1.4129 us | 1.1799 us | 0.4883 |   30720 B |


hamdaankhalid:
> @hamdaankhalid please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -343,6 +350,17 @@ public static ulong OneIfRead(this RespCommand cmd)
...
         }
 
+        /// <summary>
+        /// Returns True if <paramref name="cmd"/> does not change state of database
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static bool IsNonDbStateMutatingCommand(this RespCommand cmd, CustomCommand customCommand, CustomObjectCommand customObjectCommand)
+        {
+            return (cmd == RespCommand.CustomCmd && customCommand?.type == CommandType.Read) || (cmd == RespCommand.CustomObjCmd && customObjectCommand?.type == CommandType.Read) ||
+                cmd <= RespCommand.ZSCORE || // command is from read section
```
pradeepyadavmsft, line 360: This can lead to a case where we return a value in the get command but the set for that key has not persisted. So if the server crashes before persisting it that key will become unavailable.

libs/server/Resp/RespServerSession.cs
```diff
@@ -336,6 +342,8 @@ private void ProcessMessages()
 
             var _origReadHead = readHead;
 
+            // Each message assumes the default behavior of committing the AOF log, given that AOF is enabled on the session
+            aofCommitCurrentCommand = true;
```
pradeepyadavmsft, line 346: If my understanding is correct, we are doing it at the message level.

1 message can have multiple commands. It should be done at the command level as the command can be mutating or non-mutating. But in current logic, if one of the commands in a message is mutating, it will wait to commit. It will not be helpful if it has an info command after a set command. It will block the info as well. I may be wrong but that's my understanding.

libs/server/Resp/Parser/RespCommand.cs
hamdaankhalid: Good point, query-like commands won't be lost if the "enqueued" AOF items that need to be flushed are not related to the key the queries are related to.
However, for the same key, not waiting for persistence could return a value for a Key-Val that is failed to persist since it will be in memory.

I guess I could look into having some sort of way to know if something is a dirty/new (not flushed) that is being queried and then proceed off of that.

libs/server/Resp/RespServerSession.cs
hamdaankhalid: this was initially on purpose, but I think now that I have changed my logic to only perform this  step on very lkimited set of commands we should be safe to move this to command level.

libs/server/Resp/Parser/RespCommand.cs
hamdaankhalid: Dirty Read handling will be done in separate PR, limited the set of Commands in this first PR

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -286,6 +286,78 @@ public static class RespCommandExtensions
...
+            foreach (RespCommand cmd in Enum.GetValues(typeof(RespCommand)))
+            {
+                if (Array.IndexOf(AofIndependentCommands, cmd) == -1)
+                    continue;
+
+                // mark the command as an AOF independent command for lookups later by setting the bit in bit vec
+                int bitIdxToUse = (int)cmd / sizeOfLong;
+                // set the respCommand's bit to indicate
+                int bitIdxOffset = (int)cmd % sizeOfLong;
+                long bitmask = 1U << bitIdxOffset;
```
TedHartMS, line 343: 1UL

libs/server/Resp/RespServerSession.cs
```diff
@@ -901,7 +943,7 @@ private void Send(byte* d)
             if ((int)(dcurr - d) > 0)
             {
                 // Debug.WriteLine("SEND: [" + Encoding.UTF8.GetString(new Span<byte>(d, (int)(dcurr - d))).Replace("\n", "|").Replace("\r", "!") + "]");
-                if (storeWrapper.appendOnlyFile != null && storeWrapper.serverOptions.WaitForCommit)
+                if (runningWithAOFWaitForCommitMode && waitForAofBlocking)
```
TedHartMS, line 946: This can test just waitForAofBlocking, as it is not set true if runningWithAOFWaitForCommitMode is false

TedHartMS (changes requested):

libs/server/Resp/RespServerSession.cs
```diff
@@ -133,6 +133,17 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
...
+        /// <summary>
+        /// Flag indicating whether any of the commands in one message
+        /// requires us to block on AOF before sending response over the network
+        /// </summary>
+        bool waitForAofBlocking = false;
+
+        /// <summary>
+        /// Flag to indicate if server is running in AOF mode along with setting to wait for commits
+        /// </summary>
+        bool runningWithAOFWaitForCommitMode = false;
```
TedHartMS, line 145: can be readonly

libs/server/Resp/RespServerSession.cs
hamdaankhalid: nice! completely missed the redundancy. Thanks

TedHartMS (approved):

pradeepyadavmsft (approved):

libs/server/Resp/RespServerSession.cs
```diff
@@ -299,6 +312,9 @@ public override int TryConsumeMessages(byte* reqBuffer, int bytesReceived)
             }
             finally
             {
+                // reset the session's flag for AOF blocking to default value after processing all commands
+                waitForAofBlocking = false;
```
badrishc, line 316: Move this to https://github.com/microsoft/garnet/pull/511/files#diff-b42721ea9aeab99ac7f5177323136d172b8d0c441c03df3f3634f96aa1774ac6L907

libs/server/Resp/RespServerSession.cs
```diff
@@ -175,6 +186,8 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
             clusterSession = storeWrapper.clusterProvider?.CreateClusterSession(txnManager, this._authenticator, this._user, sessionMetrics, basicGarnetApi, networkSender, logger);
             clusterSession?.SetUser(this._user);
 
+            this.runningWithAOFWaitForCommitMode = storeWrapper.appendOnlyFile != null && storeWrapper.serverOptions.WaitForCommit;
```
badrishc, line 189: no need to introduce another bool variable for this and take up space in the instance. just directly refer to `storeWrapper.appendOnlyFile != null && storeWrapper.serverOptions.WaitForCommit`

libs/server/Resp/RespServerSession.cs
```diff
@@ -353,6 +369,29 @@ private void ProcessMessages()
                 // Check ACL permissions for the command
                 if (cmd != RespCommand.INVALID && CheckACLPermissions(cmd))
                 {
+
+                    if (runningWithAOFWaitForCommitMode)
```
badrishc, line 373: This is polluting the main ProcessMessages. Can it be moved to CheckACLPermissions and tighten the long comment.

hamdaankhalid: can be moved, but not completely sure by what you mean by tightening the comment. Do you just want me to get rid of the whitespaces between paragraphs?

AOF/aof.log.0
badrishc, line 1: do not check in this file

hamdaankhalid: not a part of latest commit

libs/server/Resp/RespServerSession.cs
badrishc: do we really need the full truth table there? seems implicit from the boolean logic ... i am on the fence w.r.t. this so if you think it adds significant value, it can be kept. just a small nit/preference.

hamdaankhalid: I usually tend to leave the work in if something took me longer than 2 minutes to come up with. Although looking back, I think we can get rid of the truth table since the comment above tells you they "why" pretty well.

hamdaankhalid: removed

libs/server/Resp/RespServerSession.cs
```diff
@@ -883,16 +889,19 @@ private void WriteDirectLarge(ReadOnlySpan<byte> src)
...
+                Send(networkSender.GetResponseObjectHead(), true);
                 networkSender.GetResponseObject();
                 dcurr = networkSender.GetResponseObjectHead();
                 dend = networkSender.GetResponseObjectTail();
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private void Send(byte* d)
+        private void Send(byte* d, bool isChunked = false)
```
badrishc, line 900: We call `SendAndReset` in many places after partial command processing, and those are not caught by this code change. It seems it can be made correct by moving the setting to false of waitForAofBlocking, back to the finally in `TryConsumeMessages`.

But unfortunately it is a bit conservative, because we might have 100 commands in the batch, the first command made waitForAofBlocking true and we finished flushing its result after blocking, yet the bool will continue to stay true and block the AOF flush for all the remaining 99 commands even if they did not need to block.

libs/server/Resp/RespServerSession.cs
hamdaankhalid: moved it back to conservative approach, after SendAndReset was discussed with you. MArking this as resolved.

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -1803,6 +1875,18 @@ private RespCommand ParseCommand(out bool success)
...
+            if (storeWrapper.appendOnlyFile != null && storeWrapper.serverOptions.WaitForCommit)
+            {
+                /* 
+                    keeping the expensive call inside the conditional only adds ~4 MSIL instructions in hotpath
+                    W.r.t AOF  Blocking
+                    If a previous command marked AOF for blocking we should not change AOF blocking flag.
+                    If no previous command marked AOF for blocking, then we only change AOF flag to block
+                    if the current command is AOF dependent.
+                */
+                waitForAofBlocking = waitForAofBlocking || !cmd.IsAofIndependent();
```
badrishc, line 1887: Adding this above line 1887 should make it optimal:

```
  // Reset waitForAofBlocking if there is no pending unsent data on the network
  if (dcurr == networkSender.GetResponseObjectHead()) waitForAofBlocking = false;
```

libs/server/Resp/RespServerSession.cs
```diff
@@ -299,6 +305,9 @@ public override int TryConsumeMessages(byte* reqBuffer, int bytesReceived)
             }
             finally
             {
+                // reset the session's flag for AOF blocking to default value after processing all commands
+                waitForAofBlocking = false;
```
badrishc, line 309: If above comment makes sense, then this is no longer necessary.

libs/server/Resp/RespServerSession.cs
badrishc: Can undo this change too.

libs/server/Resp/RespServerSession.cs
```diff
@@ -901,10 +910,12 @@ private void Send(byte* d)
             if ((int)(dcurr - d) > 0)
             {
                 // Debug.WriteLine("SEND: [" + Encoding.UTF8.GetString(new Span<byte>(d, (int)(dcurr - d))).Replace("\n", "|").Replace("\r", "!") + "]");
-                if (storeWrapper.appendOnlyFile != null && storeWrapper.serverOptions.WaitForCommit)
+                if (waitForAofBlocking)
                 {
                     var task = storeWrapper.appendOnlyFile.WaitForCommitAsync();
                     if (!task.IsCompleted) task.AsTask().GetAwaiter().GetResult();
+                    // once the set of commands so for has been processed we can safely reset the session flag to not need AOF blocking
+                    if(!isChunked) waitForAofBlocking = false;
```
badrishc: And this ...

libs/server/Resp/RespServerSession.cs
hamdaankhalid: Looking at how the RespSession objects are created, I see that they're held in Network Handler, and may be reused. So it makes sense to set the state of the session back to default value for the flag, right?

you would still want to reset the session flag at the very end of all commands, since the last command may set it to block OnAof and then after we exit command processing we may have a Session be left in state where the flag is set to true.

badrishc: It doesn't matter because the next time you receive a chunk of data, this line will fire and reset the bool. Because at that point there is no data yet injected into the response buffer.

`if (dcurr == networkSender.GetResponseObjectHead()) waitForAofBlocking = false;`

badrishc (approved):

PR #518, opened 2024-07-10
Fix issue with resetting stats throwing null exception

Mathos1432:
Currently, calling INFO RESET with the object store disabled or clustering disabled results in a null reference exception and subsequent queries for the info result in invalid information.

This change resolves the issue by adding a couple of null checks in the ResetStats method of the GarnetServerMonitor.

Tested by adding a test that was failing before the change and that is now passing.


TalZaccai (approved):
Thank you for finding and fixing this issue!

PR #519, opened 2024-07-10
Fix SortedSetPop to return popped elements.

### LINKED ISSUE #514: GarnetApi.SortedSetPop doesn't return popped members
### Describe the bug

The `GarnetAPI.SortedSetPop` method is not returning the popped members as expected when called from a custom transaction. This issue appears to be specific to the following method signature:

```c# 
GarnetStatus GarnetAPI.SortedSetPop(
    ArgSlice key, 
    out (ArgSlice score, ArgSlice member)[] pairs, 
    int count = 1, 
    bool lowScoresFirst = true
)
```
defined in `libs/server/API/IGarnetApi.cs`.
 
Interestingly, this behavior is not observed when testing with a RESP client and ZPOPMIN command, as it seems to utilize a different method signature:
```c#
GarnetStatus GarnetApi.SortedSetPop(byte[] key, ArgSlice input, ref GarnetObjectStoreOutput outputFooter)
```
 
The root cause may be in `libs/server/Storage/Session/ObjectStore/SortedSetOps.cs`, where the `StorageSession.SortedSetPop` function [doesn't appear to handle the output of ProcessRespArrayOutput](https://github.com/microsoft/garnet/blob/2166147f12bdae143fad9cc8dc52c62702c0fff1/libs/server/Storage/Session/ObjectStore/SortedSetOps.cs#L327).

1. Implement a custom transaction that calls the SortedSetPop function:
  ```c#
  using Garnet.common;
  using Garnet.server;
  using Tsavorite.core;
  
  namespace Garnet {
      sealed class ZPopMinViaApi : CustomTransactionProcedure
      {
          public override bool Prepare<TGarnetReadApi>(TGarnetReadApi api, ArgSlice input)
          {
              int offset = 0;
              AddKey(GetNextArg(input, ref offset), LockType.Exclusive, true);
              return true;
          }
      
          public override void Main<TGarnetApi>(TGarnetApi api, ArgSlice input, ref MemoryResult<byte> output)
          {
              int offset = 0;
              var key = GetNextArg(input, ref offset);
      
              var status = api.SortedSetPop(key, out var pairs);
      
              if (status == GarnetStatus.OK)
              {
                  if (pairs != null)
                  {
                      WriteBulkStringArray(ref output, pairs[0].score, pairs[0].member);
                  }
                  else
                  {
                      WriteSimpleString(ref output, "No pairs!");
                  }
              }
          }
      }
  }
  ```
2. Execute the following commands with a RESP client:
  ```
  127.0.0.1:6379> ZADD myzset 12 b
  (integer) 1
  127.0.0.1:6379> ZPOPMINVIAAPI myzset
  No pairs!
  ```

### Expected behavior

I think the expected output for the above commands should be:
```
127.0.0.1:6379> ZADD myzset 12 b
(integer) 1
127.0.0.1:6379> ZPOPMINVIAAPI myzset
1) "b"
2) "12"
```
Instead, we receive "No pairs!" indicating that the `pairs` array in the example code is null.

### Screenshots

_No response_

### Release version

[2166147](https://github.com/microsoft/garnet/commit/2166147f12bdae143fad9cc8dc52c62702c0fff1), i.e., current `main` (2024-07-09).

### IDE

_No response_

### OS version

MacOS 14.5 (23F79)

### Additional context

```shell
$ dotnet --version
8.0.104
```

yrajas:
Fixes #514 

Fixed SortedSetPop operation to parse the output buffer array and extract pairs of elements as scores & member information to be returned to the caller.

### DIFF
--- a/test/Garnet.test/Garnet.test.csproj
+++ b/test/Garnet.test/Garnet.test.csproj
@@ -21,6 +21,7 @@
     <Compile Include="..\..\main\GarnetServer\Extensions\SampleDeleteTxn.cs" Link="SampleDeleteTxn.cs" />
     <Compile Include="..\..\main\GarnetServer\Extensions\SetIfPM.cs" Link="SetIfPM.cs" />
     <Compile Include="..\..\main\GarnetServer\Extensions\SetWPIfPGT.cs" Link="SetWPIFPGT.cs" />
+    <Compile Include="..\..\playground\Embedded.perftest\DummyNetworkSender.cs" Link="DummyNetworkSender.cs" />
   </ItemGroup>
 
 	<ItemGroup>
--- a/libs/server/Storage/Session/ObjectStore/Common.cs
+++ b/libs/server/Storage/Session/ObjectStore/Common.cs
@@ -102,7 +102,7 @@ namespace Garnet.server
         }
 
         /// <summary>
-        /// Converts an array of elements in RESP format to ArgSlice[] type
+        /// Converts an array of result in RESP format to ArgSlice[] type
         /// </summary>
         /// <param name="outputFooter">The RESP format output object</param>
         /// <param name="error">A description of the error, if there is any</param>
@@ -113,7 +113,7 @@ namespace Garnet.server
             ArgSlice[] elements = default;
             error = default;
 
-            // For reading the elements in the outputFooter
+            // For reading the result in the outputFooter
             byte* element = null;
             int len = 0;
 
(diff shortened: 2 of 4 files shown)


test/Garnet.test/Garnet.test.csproj
```diff
@@ -21,6 +21,7 @@
     <Compile Include="..\..\main\GarnetServer\Extensions\SampleDeleteTxn.cs" Link="SampleDeleteTxn.cs" />
     <Compile Include="..\..\main\GarnetServer\Extensions\SetIfPM.cs" Link="SetIfPM.cs" />
     <Compile Include="..\..\main\GarnetServer\Extensions\SetWPIfPGT.cs" Link="SetWPIFPGT.cs" />
+    <Compile Include="..\..\playground\Embedded.perftest\DummyNetworkSender.cs" Link="DummyNetworkSender.cs" />
```
TalZaccai, line 24: Do we want to move this to Common perhaps?

TalZaccai (changes requested):

TalZaccai (approved):

PR #529, opened 2024-07-23
Pool latency histograms for sessions to avoid GC hit

badrishc:


yrajas:
Instance of LongHistogram in LatencyMetricsEntry needs a Return as well to return the shared ArrayPool?

libs/server/Metrics/Latency/GarnetLatencyMetricsSession.cs
```diff
@@ -24,6 +24,16 @@ public GarnetLatencyMetricsSession(GarnetServerMonitor monitor)
             Init();
         }
 
+        public void Return()
+        {
+            foreach (var cmd in defaultLatencyTypes)
+            {
```
yrajas, line 30: Better to have a Return method within LatencyMetricsEntrySession that this method can invoke to maintain the abstraction? That way, invoking Return will become obvious to any future users of LatencyMetricsEntrySession.

badrishc: Done

badrishc:
> Instance of LongHistogram in LatencyMetricsEntry needs a Return as well to return the shared ArrayPool?

Added.

yrajas (approved):

PR #530, opened 2024-07-25
Fix disposal of migration sessions.

yrajas:
**Problem**
When using KEYS option for slot migration, the migration session removal isn't disposing the session object. This leads to an out of memory exception when migrating a large number of keys.

**Fix**
- [x] Set slots correctly for the migrate operation. This helps to ensure the slots are correctly reset on removal of migration session.
- [x] Also, removal of multiple outstanding session objects when node is removed for the FORGET command, will dispose each of those session objects.

### DIFF
--- a/libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
+++ b/libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
@@ -129,23 +129,51 @@ namespace Garnet.cluster
             return success;
         }
 
+        public bool TryRemove(MigrateSession mSession)
+        {
+            try
+            {
+                _lock.WriteLock();
+                if (_disposed) return false;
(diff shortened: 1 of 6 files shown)


libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
```diff
@@ -129,23 +129,51 @@ public int GetNumSessions()
...
+        }
+
         public bool TryRemove(string targetNodeId)
         {
             try
             {
                 _lock.WriteLock();
                 if (_disposed) return false;
-                MigrateSession mSession = null;
+                HashSet<MigrateSession> mSessions = [];
```
badrishc, line 164: consider allocating lazily if common case is that the hash set will be empty.

badrishc (approved):

PR #531, opened 2024-07-25
Helm Chart: Add support for configuration from a secret

MartinHell:
Hi,

I noticed there's no support for using garnet.conf instead of command line arguments. This aims to fix that by adding support for a secret or using an existingSecret.

If you configure Auth and Passwords currently you can't really store the values.yaml in version control since it would contain the password in clear text. This will allow you to add sensitive information (password) to a secret, you could store it using SealedSecrets or something similar in git and apply that secret alongside the helm chart for example.

Hopefully this is appreciated :)

### DIFF
--- a/charts/garnet/templates/secret.yaml
+++ b/charts/garnet/templates/secret.yaml
@@ -1,0 +1,13 @@
+{{- if not .Values.garnet.existingSecret }}
+{{- if .Values.garnet.config }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "garnet.fullname" . }}-config
+  labels:
+    {{- include "garnet.labels" . | nindent 4 }}
+type: Opaque
+data:
+  garnet.conf: {{ .Values.garnet.config | b64enc | quote }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


MartinHell:

badrishc:
cc @babykart @nicholih for review

babykart:
We are already in a helm chart called _garnet_ so let's avoid any potential confusion or repetition (?).

```yaml
config:
  # -- Garnet secret (if you want to use an existing secret)
  # Make sure the key in the secret is 'garnet.conf'
  existingSecret: ""
  # -- The garnet.conf data content.
  garnetConf: ""
  # garnetConf: |
  #   {
  #     "AuthenticationMode": "Password",
  #     "Password": ""
  #   }
```

msft-paddy14:
The idea seems right but we should use existing charts as pointed out by @babykart . Generally how we do it in Kubernetes is to have a sidecar init container where we fetch the secrets (say from Azure Key Vault) based on secret name and keyvault name. We use the ACL mode to apply it as a file mount during startup. For just password usage, something similar can be done by updating the garnet.conf in place. It might be something you want to do using dependent helm charts so that you can add your chart while referring garnet one?

badrishc:
We have a Helm chart package release now thanks to @babykart. Would be great to see support for configuration files and secrets added to that, as discussed in #594. Closing this one for now. Thank you!

PR #541, opened 2024-07-27
Validate arguments of public methods#CA1062

Zombach:
Source [CA1062](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062)

Cause
An externally visible method dereferences one of its reference arguments without verifying whether that argument is null (Nothing in Visual Basic).

You can [configure](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#configure-code-to-analyze) this rule to exclude certain types and parameters from analysis. You can also [indicate null-check validation methods](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#null-check-validation-methods).

Rule description
All reference arguments that are passed to externally visible methods should be checked against null. If appropriate, throw an [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) when the argument is null.

If a method can be called from an unknown assembly because it is declared public or protected, you should validate all parameters of the method. If the method is designed to be called only by known assemblies, mark the method internal and apply the [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute) attribute to the assembly that contains the method.

I implemented a small part so that you could evaluate and point out the necessary corrections.
It would also be nice if you indicated a list of classes (projects) that need to be covered with these tests.

I would suggest creating a issue so that you can solve the problem in parts and also for tracking

### DIFF
--- a/Garnet.sln
+++ b/Garnet.sln
@@ -1,5 +1,5 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
+# 17
 VisualStudioVersion = 17.0.31808.319
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "benchmark", "benchmark", "{346A5A53-51E4-4A75-B7E6-491D950382CE}"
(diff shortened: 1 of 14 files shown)


Zombach:

yrajas:
Hi @Zombach 
Thanks for your contribution in adding these defensive checks.
However, as performance is a key differentiator for Garnet, evaluating the impact of these checks across the wide source base would be a non-trivial amount of effort. Also, in case of a valid scenario that causes a null reference exception because of a missing null check wouldn't be difficult to troubleshoot and root cause the issue. So, as this project generally skews towards performance, will close this PR.
We encourage you to keep contributing and sharing your ideas. Thank you again for your understanding.

Zombach:
> Hi @Zombach Thanks for your contribution in adding these defensive checks. However, as performance is a key differentiator for Garnet, evaluating the impact of these checks across the wide source base would be a non-trivial amount of effort. Also, in case of a valid scenario that causes a null reference exception because of a missing null check wouldn't be difficult to troubleshoot and root cause the issue. So, as this project generally skews towards performance, will close this PR. We encourage you to keep contributing and sharing your ideas. Thank you again for your understanding.

Hi
Thanks for the explanation. OK, no problem!
Good luck

PR #542, opened 2024-07-29
Add StoreFunctions and Allocator wrapper to TsavoriteKV

TedHartMS:
StoreFunctions are the KV-level equivalent of SessionFunctions, providing functionality such as Key comparison, serialization, and record disposal. This allows many of these operation to be inlined rather than interface calls.

The Allocator wrapper is a struct specific to each allocator type, that wraps the class implementing that type. This is a workaround to allow these class-implementation calls to be inlined.

TStoreFunctions and TAllocator are new type arguments on TsavoriteKV and ClientSession and its Contexts.

Detailed documentation of this change is in .\website\docs\dev\tsavorite\storefunctions.md.


benchmark/BDN.benchmark/Resp/RespGetSetStress.cs
```diff
@@ -0,0 +1,62 @@
...
+        public void GlobalCleanup()
+        {
+            session.Dispose();
+            server.Dispose();
+        }
+
+        [Benchmark]
+        public void Get()
+        {
+            fixed (byte* ptr = SINGLE_GET)
```
badrishc, line 47: Remove the `fixed` here and rerun - see `RespParseStress` for how we structure the BDN.

libs/host/GarnetServer.cs
```diff
@@ -14,6 +14,11 @@
 
 namespace Garnet
 {
+    using MainStoreFunctions = StoreFunctions<SpanByte, SpanByte, SpanByteComparer, SpanByteRecordDisposer>;
+
```
badrishc, line 18: For symmetry add:
`using MainStoreAllocator = SpanByteAllocator<MainStoreFunctions>`

libs/server/API/GarnetApi.cs
```diff
@@ -8,16 +8,21 @@
 
 namespace Garnet.server
 {
+    using ObjectStoreAllocator = GenericAllocator<byte[], IGarnetObject, StoreFunctions<byte[], IGarnetObject, ByteArrayKeyComparer, DefaultRecordDisposer<byte[], IGarnetObject>>>;
+    using ObjectStoreFunctions = StoreFunctions<byte[], IGarnetObject, ByteArrayKeyComparer, DefaultRecordDisposer<byte[], IGarnetObject>>;
+
+    using SpanByteStoreFunctions = StoreFunctions<SpanByte, SpanByte, SpanByteComparer, SpanByteRecordDisposer>;
```
badrishc, line 14: Use same name across all files: `MainStoreFunctions`

libs/server/Storage/Session/MainStore/CompletePending.cs
```diff
@@ -6,6 +6,8 @@
 
 namespace Garnet.server
 {
+    using MainStoreFunctions = StoreFunctions<SpanByte, SpanByte, SpanByteComparer, SpanByteRecordDisposer>;
```
badrishc, line 9: Add `using MainStoreAllocator` here (and other places).

badrishc (changes requested):
See comments.

badrishc (approved):

Issue #509, opened 2024-07-03
Lua support - EVAL and friends


mmclure-msft:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

Our team's code is heavily dependent on Redis and we're looking into the possible performance benefits of using Garnet. However, we're heavily dependent on the Redisson library which itself is a heavy user of Lua scripting with EVAL/EVALSHA, for example
https://github.com/redisson/redisson/blob/670c7dbdc0187801d55c8a457fe294cc9ce43020/redisson/src/main/java/org/redisson/RedissonMapCache.java#L535
As such, we can't replace Redis with Garnet until EVAL support is added.

### Describe the solution you'd like

Support for Lua scripting and the EVAL family of commands.

### Describe alternatives you've considered

Our only alternative would be to implement the functionality of Redisson as Garnet extensions. This would be a heavy lift.

### Additional context

I understand that adding Lua support will be a big feature request - I'm just opening the conversation to see if this is on the roadmap.

mardukbp:
Celery, a popular Python library for background jobs also depends on EVALSHA.

eve8110:
BullMQ, a popular library in Node.js, also relies on Lua scripts.

mardukbp:
Awesome! Thanks @badrishc for implementing this. Unfortunately, it does not work as expected. Here is a minimal Python script that fails with `redis.exceptions.ResponseError: NOSCRIPT No matching script. Please use EVAL.`:

```python
import redis

lua = """
local value = redis.call('GET', KEYS[1])
value = tonumber(value)
return value * ARGV[1]
"""

r = redis.Redis()
multiply = r.register_script(lua)
r.set('foo', 2)
multiply(keys=['foo'], args=[5])
```

Actually I tested Garnet with Celery and I got the above error. So I searched for a minimal example to reproduce it.

badrishc:
> Awesome! Thanks @badrishc for implementing this. Unfortunately, it does not work as expected. Here is a minimal Python script that fails with `redis.exceptions.ResponseError: NOSCRIPT No matching script. Please use EVAL.`:
> 
> ```python
> import redis
> 
> lua = """
> local value = redis.call('GET', KEYS[1])
> value = tonumber(value)
> return value * ARGV[1]
> """
> 
> r = redis.Redis()
> multiply = r.register_script(lua)
> r.set('foo', 2)
> multiply(keys=['foo'], args=[5])
> ```
> 
[...]

PR #550, opened 2024-08-02
Basic Lua scripting support

badrishc:
This PR adds very basic support for Lua scripts in Garnet (EVAL, EVALSHA, SCRIPT commands) using the [NLua](https://github.com/NLua/NLua) library. We hope for contributions from the community to extend as per their needs. Consider this a research prototype. Note that performance is not expected to be as good as C# custom procedures and transactions, due to the overheads of Lua scripts and the conversions from C# to Lua and back.

Tasks:
- [x] Basic tests passing for GET, SET, ZADD
- [x] BDN tests to see performance limits when using NLua
- [x] Optimize script invocation code path for end-to-end performance
- [x] Add support for all API calls in Lua scripts (by formatting into RESP command)
- [x] Error handling, support more types, math
- [x] Add switch (`--lua` and `EnableLua`) to enable Lua script on server (turned off by default)
- [x] Enable transactions (opt in at server-level) via `--lua-transaction-mode` and `LuaTransactionMode` boolean option
- [x] Support ACL

Future work:
- Better script cache management (currently all scripts are cached with no upper bound)
- Support `LuaTransactionMode` on a per-script basis
- Support shared vs exclusive key locks (currently we pessimistically lock all keys as exclusive)
- Add cluster slot verification
- Support read-only scripts (`EVAL_RO`)

Fix https://github.com/microsoft/garnet/issues/509


badrishc:
### End-to-End Performance

Increasing number of client sessions,  8-byte keys and values, using `Resp.benchmark` with batch size 4096, no transactions.

1. Script: `return KEYS[1]`

How to run: `Resp.benchmark --op SCRIPTRETKEY -t 1,2,4,8,16 --dbsize 1024`

<p align="center">
  
</p>

2. Script: `return redis.call('get', KEYS[1])`

How to run: `Resp.benchmark --op SCRIPTGET -t 1,2,4,8,16 --dbsize 1024`

<p align="center">
  
</p>

libs/server/Lua/LuaRunner.cs
```diff
@@ -0,0 +1,396 @@
...
+            this.scratchBufferNetworkSender = scratchBufferNetworkSender;
+            this.scratchBufferManager = respServerSession?.scratchBufferManager;
+            this.logger = logger;
+
+            state = new Lua();
+            state.State.Encoding = Encoding.UTF8;
+            if (txnMode)
+            {
+                this.txnKeyEntries = new TxnKeyEntries(16, respServerSession.storageSession.lockableContext, respServerSession.storageSession.objectStoreLockableContext);
+                state.RegisterFunction("garnet_call", this, this.GetType().GetMethod("garnet_call_txn"));
```
Meir017, line 46:
```suggestion
                state.RegisterFunction("garnet_call", this, this.GetType().GetMethod(nameof(garnet_call_txn)));
```

libs/server/Lua/LuaRunner.cs
```diff
@@ -0,0 +1,396 @@
...
+                        return resultStr;
+                    break;
+
+                case (byte)'*':
+                    if (RespReadUtils.ReadStringArrayResponseWithLengthHeader(out var resultArray, ref ptr, ptr + length))
+                        return resultArray;
+                    break;
+
+                default:
+                    throw new Exception("Unexpected response: " + Encoding.UTF8.GetString(new Span<byte>(ptr, length)).Replace("\n", "|").Replace("\r", "") + "]");
```
Meir017, line 228: consider introducing a regex for `\n\r` and then calling `regex.Replace`

libs/server/Lua/SessionScriptCache.cs
```diff
@@ -0,0 +1,104 @@
...
+            foreach (var runner in scriptCache.Values)
+            {
+                runner.Dispose();
+            }
+            scriptCache.Clear();
+        }
+
+        static ReadOnlySpan<byte> HEX_CHARS => "0123456789abcdef"u8;
+
+        public byte[] GetScriptDigest(ReadOnlySpan<byte> source)
```
Meir017, line 92: can we use `System.Security.Cryptography.SHA1` here?

test/Garnet.test/LuaScriptRunnerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                var ex = Assert.Throws<LuaScriptException>(() => runner.Run());
+                Assert.AreEqual("[string \"dofile();\"]:1: attempt to call a nil value (global 'dofile')", ex.Message);
+            }
+
+            // Try to call a windows executable
+            using (var runner = new LuaRunner("require \"notepad\""))
+            {
+                runner.Compile();
+                var ex = Assert.Throws<LuaScriptException>(() => runner.Run());
+                Assert.AreEqual("[string \"require \"notepad\"\"]:1: attempt to call a nil value (global 'require')", ex.Message);
```
Meir017, line 46: does this test-case make sense when running on linux?

libs/server/Lua/SessionScriptCache.cs
badrishc: `GetScriptDigest` uses `SHA1.HashData` which is indeed part of `System.Security.Cryptography.SHA1`.

test/Garnet.test/LuaScriptRunnerTests.cs
badrishc: it doesn't hurt so there is no strong need to special case it.

libs/server/Lua/LuaRunner.cs
badrishc: This could be done across the codebase. Leaving for future optimization as it is a non-critical code path.

libs/server/Lua/LuaCommands.cs
```diff
@@ -0,0 +1,227 @@
...
+            }
+            else
+            {
+                result = ExecuteScript(count - 1, runner);
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// SCRIPT Commands (load, exists, flush, kills)
```
TedHartMS, line 94: KILL is not implemented; maybe add a TODO case for it

TedHartMS (changes requested):

libs/server/Lua/LuaRunner.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            string[] keys = null;
+            if (nKeys > 0)
+            {
+                keys = new string[nKeys + 1];
+                for (int i = 0; i < nKeys; i++)
+                {
+                    if (txnMode)
+                    {
+                        var key = parseState.GetArgSliceByRef(offset);
```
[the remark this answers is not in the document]
badrishc: That will unnecessarily create the key ArgSlice entries from string[] with an additional copy into scratch buffer in Run(string[], string[]) overload.

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -1086,6 +1102,13 @@ private RespCommand FastParseArrayCommand(ref int count, ref ReadOnlySpan<byte>
                             case 7:
                                 switch ((ushort)ptr[4])
                                 {
+                                    case 'E':
+                                        if (*(ulong*)(ptr + 2) == MemoryMarshal.Read<ulong>("\r\nEVALSHA\r\n"u8))
```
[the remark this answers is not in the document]
badrishc: no, because the equality test includes `\r\n`

libs/server/Lua/LuaCommands.cs
badrishc: removed KILL from comment, can be added in future.

TedHartMS (approved):

PR #552, opened 2024-08-02
Migration Misc Fixes

vazois:
This PR fixes the following bugs related to migration operation

- [x]  When adding creating a new migration session check if associated slot is already owned by an existing migration session.
- [x] Fix SLOTSRANGE option.
- [x] When removing migrate session by targetNodeId dispose session immediately to avoid collection allocations.
- [x] Fix issue with migration window tracking when copyOption is not set.
- [x] Refactor logger messages to reduce verbosity.


yrajas (commented):
A test needed to catch the SLOTSRANGE issue?

yrajas (approved):

PR #563, opened 2024-08-07
Bugfix: Fix AOF commit mode when not in Transactions

hamdaankhalid:
Per my last PR I introduced a bug which made the AOF related test flaky. The error was coming from the fact that AOF commit was not being called when we were executing from outside the context of a transaction. This was meant to not invoke the IsAoFIndependent method when we are in NetworkSkip mode.

### DIFF
--- a/libs/server/Resp/Parser/RespCommand.cs
+++ b/libs/server/Resp/Parser/RespCommand.cs
@@ -1902,9 +1902,7 @@ namespace Garnet.server
             if (dcurr == networkSender.GetResponseObjectHead())
                 waitForAofBlocking = false;
 
-            // If we are in NetworkSkip's state we do not need to be changing the AOF blocking flag
-            // since we do not interact with AOF in that path.
-            if (txnManager.state != TxnState.None && txnManager.state == TxnState.Running)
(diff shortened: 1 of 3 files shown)


libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -1902,9 +1902,7 @@ private void HandleAofCommitMode(RespCommand cmd)
             if (dcurr == networkSender.GetResponseObjectHead())
                 waitForAofBlocking = false;
 
-            // If we are in NetworkSkip's state we do not need to be changing the AOF blocking flag
```
badrishc, line 1905: please add a comment here for why this if is needed and correct.

badrishc (approved):

libs/server/Resp/Parser/RespCommand.cs
badrishc: In particular, we need to make sure the Commit of the transaction (e.g., EXEC) will not respond until a commit is successfully done, if we are blocking on AOF.

badrishc (changes requested):
See comments.

badrishc (approved):

badrishc (approved):

PR #564, opened 2024-08-07
chore: logger enforce "CA2254: Template should be a static expression"

Meir017:


Meir017:
there are currently 173 warnings

benchmark/Resp.benchmark/RespOnlineBench.cs
```diff
@@ -323,7 +323,9 @@ public void Run()
                             $"{"total_ops;",pad}" +
                             $"{"iter_tops;",pad}" +
                             $"{"tpt (Kops/sec)",pad}";
+#pragma warning disable CA2254 // Template should be a static expression
```
badrishc, line 326: instead of disabling, could we rewrite this (and the other disable sites) as `logger?.LogInformation("{msg}", msg)`

Meir017: nice!

badrishc (approved):

PR #570, opened 2024-08-07
Fix directory permission issue in linux containers

akoken:
Fixes #325

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,26 +30,18 @@ RUN dotnet publish -a $TARGETARCH -c Release -o /app --no-restore --self-contain
 # Final stage/image
 FROM mcr.microsoft.com/dotnet/runtime:8.0 AS runtime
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -30,27 +30,19 @@ RUN dotnet publish -a $TARGETARCH -c Release -o /app --no-restore --self-contain
 # Final stage/image
 FROM mcr.microsoft.com/dotnet/runtime:8.0 AS runtime
 
-# Determine the distribution and install libaio1
```
badrishc, line 33: why was this part removed in the PR?

akoken: The default .NET 8 base image is [Debian](https://github.com/dotnet/dotnet-docker/discussions/5709), and this will remain the case at least until .NET 10. There is no reason to keep that `if statements` in Dockerfile.

badrishc (approved):

Issue #578, opened 2024-08-09
Remove support for dotnet 6


Meir017:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

dotnet 6 will reach end-of-life by Nov 12, 2024 (see https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core)

### Describe the solution you'd like

support only the latest LTS version of dotnet

### Additional context

this blocks #379

akoken:
I don’t understand why .NET 6 support is being removed, especially since it’s an LTS version with time remaining before it’s out of support. For many companies, switching between .NET versions isn’t as straightforward in practice.

If there is an API diff, [preprocessor directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) can be used.

Additionally, I wouldn’t recommend adding .NET 9 support before its final release. You could end up questioning whether a bug is in Garnet or .NET 9.

badrishc:
If there are any open source users of Garnet who are still on .NET 6, please let us know.

PR #580, opened 2024-08-09
Remove support for .net 6 and 7, and add support for .net 9

TedHartMS:
Fixes https://github.com/microsoft/garnet/issues/578


TedHartMS:
Abandoning in favor of #582 which only removes support for downlevel .net versions; support for .net9 will be added later

rektide:
.net6 support now removed, in #582

PR #585, opened 2024-08-11
feature: introduce source generator for enum code generation

### LINKED ISSUE #568: Replace EnumUtils with a source generator
### Feature request type

enhancement

The logic in the `EnumUtils` is heavily relying on Dictionaries for optimizing the performance of converting the description attribute of an enum value to the enum and back

We could do better and have a source generator generate static methods that have a fast switch-case that will perform better than looking up in a dictionary (and in some flows having additional logic)

_No response_

### Additional context

we could use https://github.com/andrewlock/NetEscapades.EnumGenerators (see https://andrewlock.net/netescapades-enumgenerators-a-source-generator-for-enum-performance/) which also support the `Description` attribute

or maybe https://www.nuget.org/packages/Microsoft.Extensions.EnumStrings but it's not clear what the state of this is due to https://github.com/dotnet/extensions/pull/4639

Meir017:
closes #568

viewing the generated files 
image


Meir017:
example generated enum-utils class for `KeySpecificationFlags`

```cs
// <auto-generated>
// This code was generated by the EnumsSourceGenerator source generator.
using System;
using System.ComponentModel;
using System.Numerics;

namespace Garnet;

public static partial class EnumUtils
{
    public static bool TryParseKeySpecificationFlagsFromDescription(string description, out KeySpecificationFlags result)
    {
        result = default;
        switch (description)
        {
            case "RW":
                result = KeySpecificationFlags.RW;
                return true;
            case "RO":
                result = KeySpecificationFlags.RO;
                return true;
            case "OW":
                result = KeySpecificationFlags.OW;
                return true;
            case "RM":
                result = KeySpecificationFlags.RM;
                return true;
            case "access":
                result = KeySpecificationFlags.Access;
                return true;
            case "update":
                result = KeySpecificationFlags.Update;
                return true;
            case "insert":
                result = KeySpecificationFlags.Insert;
                return true;
            case "delete":
                result = KeySpecificationFlags.Delete;
                return true;
            case "not_key":
                result = KeySpecificationFlags.NotKey;
                return true;
            case "incomplete":
                result = KeySpecificationFlags.Incomplete;
                return true;
            case "variable_flags":
                result = KeySpecificationFlags.VariableFlags;
                return true;
        }

        return false;
    }



    public static string[] GetKeySpecificationFlagsDescriptions(KeySpecificationFlags value)
    {
        var setFlags = BitOperations.PopCount((uint)value);
        if (setFlags == 0) return Array.Empty<string>();
        if (setFlags == 1)
        {
            return value switch
            {
                KeySpecificationFlags.RW => [ "RW" ],
                KeySpecificationFlags.RO => [ "RO" ],
                KeySpecificationFlags.OW => [ "OW" ],
                KeySpecificationFlags.RM => [ "RM" ],
                KeySpecificationFlags.Access => [ "access" ],
                KeySpecificationFlags.Update => [ "update" ],
                KeySpecificationFlags.Insert => [ "insert" ],
                KeySpecificationFlags.Delete => [ "delete" ],
                KeySpecificationFlags.NotKey => [ "not_key" ],
                KeySpecificationFlags.Incomplete => [ "incomplete" ],
                KeySpecificationFlags.VariableFlags => [ "variable_flags" ],
                _ => Array.Empty<string>(),
            };
        }

        var descriptions = new string[setFlags];
        var index = 0;
        if ((value & KeySpecificationFlags.RW) != 0) descriptions[index++] = "RW";
        if ((value & KeySpecificationFlags.RO) != 0) descriptions[index++] = "RO";
        if ((value & KeySpecificationFlags.OW) != 0) descriptions[index++] = "OW";
        if ((value & KeySpecificationFlags.RM) != 0) descriptions[index++] = "RM";
        if ((value & KeySpecificationFlags.Access) != 0) descriptions[index++] = "access";
        if ((value & KeySpecificationFlags.Update) != 0) descriptions[index++] = "update";
        if ((value & KeySpecificationFlags.Insert) != 0) descriptions[index++] = "insert";
        if ((value & KeySpecificationFlags.Delete) != 0) descriptions[index++] = "delete";
        if ((value & KeySpecificationFlags.NotKey) != 0) descriptions[index++] = "not_key";
        if ((value & KeySpecificationFlags.Incomplete) != 0) descriptions[index++] = "incomplete";
        if ((value & KeySpecificationFlags.VariableFlags) != 0) descriptions[index++] = "variable_flags";

        return descriptions;
    }

}

```

libs/gen/Garnet.gen.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
Meir017, line 4: the `TargetFrameworks` from the `Directory.Build.props` is being injected here which breaks this

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,151 @@
...
+        method.AppendLine("    }");
+        method.AppendLine();
+
+        return method.ToString();
+    }
+    private static string GenerateGetEnumDescriptionsMethod(string enumName, List<(string Name, string? Value, AttributeSyntax Description)> values)
+    {
+        var method = new StringBuilder();
+        method.AppendLine($"    /// <summary>");
+        method.AppendLine($"    /// Gets the descriptions of the set flags. Assumes the enum is a flags enum.");
```
Meir017, line 102: I noticed that only 3 enums use this utility class and all 3 have the `Flags` attribute

Meir017:
I just ran a small benchmark to compare:

```

BenchmarkDotNet v0.14.0, Windows 10 (10.0.19045.4651/22H2/2022Update)
Intel Core i7-6700HQ CPU 2.60GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET SDK 8.0.303
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2


```
| Method                                     | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------------------------------- |-----------:|----------:|----------:|-------:|----------:|
| GetEnumDescriptions_SourceGenFlag1         |  10.541 ns | 0.2041 ns | 0.1809 ns | 0.0102 |      32 B |
| GetEnumDescriptions_GarnetFlag1            | 177.819 ns | 1.5695 ns | 1.3106 ns | 0.0713 |     224 B |
| GetEnumDescriptions_SourceGenFlag2         |   6.413 ns | 0.1210 ns | 0.1073 ns | 0.0102 |      32 B |
| GetEnumDescriptions_GarnetFlag2            | 181.967 ns | 1.9685 ns | 1.7450 ns | 0.0713 |     224 B |
| GetEnumDescriptions_SourceGenFlag3         |  10.860 ns | 0.0698 ns | 0.0545 ns | 0.0127 |      40 B |
| GetEnumDescriptions_GarnetFlag3            | 321.769 ns | 6.1681 ns | 5.7696 ns | 0.1349 |     424 B |
| GetEnumDescriptions_SourceGenFlag4         |  12.912 ns | 0.2870 ns | 0.2684 ns | 0.0179 |      56 B |
| GetEnumDescriptions_GarnetFlag4            | 524.886 ns | 5.7487 ns | 5.0961 ns | 0.2470 |     776 B |
| GetEnumDescriptions_SourceGenFlag5         |   5.994 ns | 0.0429 ns | 0.0358 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_GarnetFlag1    |  50.095 ns | 1.4312 ns | 3.9419 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_SourceGenFlag1 |   2.481 ns | 0.0325 ns | 0.0288 ns |      - |         - |
| TryParseEnumFromDescription_GarnetFlag2    |  43.657 ns | 0.6137 ns | 0.5740 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_SourceGenFlag2 |   1.801 ns | 0.0694 ns | 0.0902 ns |      - |         - |
| TryParseEnumFromDescription_GarnetFlag3    |  44.333 ns | 0.4689 ns | 0.3916 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_SourceGenFlag3 |   1.707 ns | 0.0293 ns | 0.0260 ns |      - |         - |
| TryParseEnumFromDescription_GarnetFlag4    |  53.670 ns | 0.6426 ns | 0.7142 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_SourceGenFlag4 |   1.592 ns | 0.0332 ns | 0.0311 ns |      - |         - |
| TryParseEnumFromDescription_GarnetFlag5    |  41.455 ns | 0.4447 ns | 0.3714 ns | 0.0102 |      32 B |
| TryParseEnumFromDescription_SourceGenFlag5 |   1.866 ns | 0.0247 ns | 0.0219 ns |      - |         - |


with the following code:

```cs
[MemoryDiagnoser]
public class Benchmarks
{
    private readonly RespAclCategories flag1 = RespAclCategories.Garnet;
    private readonly RespAclCategories flag2 = RespAclCategories.Read;
    private readonly RespAclCategories flag3 = RespAclCategories.Read | RespAclCategories.Write;
    private readonly RespAclCategories flag4 = RespAclCategories.Read | RespAclCategories.Write | RespAclCategories.SortedSet | RespAclCategories.Transaction;
    private readonly RespAclCategories flag5 = RespAclCategories.All;

    [Benchmark]
    public string[] GetEnumDescriptions_SourceGenFlag1() => EnumUtils.GetRespAclCategoriesDescriptions(flag1);

    [Benchmark]
    public string[] GetEnumDescriptions_GarnetFlag1() => GarnetEnumUtils.GetEnumDescriptions(flag1);

    [Benchmark]
    public string[] GetEnumDescriptions_SourceGenFlag2() => EnumUtils.GetRespAclCategoriesDescriptions(flag2);

    [Benchmark]
    public string[] GetEnumDescriptions_GarnetFlag2() => GarnetEnumUtils.GetEnumDescriptions(flag2);

    [Benchmark]
    public string[] GetEnumDescriptions_SourceGenFlag3() => EnumUtils.GetRespAclCategoriesDescriptions(flag3);

    [Benchmark]
    public string[] GetEnumDescriptions_GarnetFlag3() => GarnetEnumUtils.GetEnumDescriptions(flag3);

    [Benchmark]
    public string[] GetEnumDescriptions_SourceGenFlag4() => EnumUtils.GetRespAclCategoriesDescriptions(flag4);

    [Benchmark]
    public string[] GetEnumDescriptions_GarnetFlag4() => GarnetEnumUtils.GetEnumDescriptions(flag4);

    [Benchmark]
    public string[] GetEnumDescriptions_SourceGenFlag5() => EnumUtils.GetRespAclCategoriesDescriptions(flag5);

    [Benchmark]
    public bool TryParseEnumFromDescription_GarnetFlag1() => GarnetEnumUtils.TryParseEnumFromDescription("Garnet", out RespAclCategories garnetParsedFlag1);

    [Benchmark]
    public bool TryParseEnumFromDescription_SourceGenFlag1() => EnumUtils.TryParseRespAclCategoriesFromDescription("Garnet", out RespAclCategories sourceGenParsedFlag1);

    [Benchmark]
    public bool TryParseEnumFromDescription_GarnetFlag2() => GarnetEnumUtils.TryParseEnumFromDescription("Read", out RespAclCategories garnetParsedFlag2);

    [Benchmark]
    public bool TryParseEnumFromDescription_SourceGenFlag2() => EnumUtils.TryParseRespAclCategoriesFromDescription("Read", out RespAclCategories sourceGenParsedFlag2);

    [Benchmark]
    public bool TryParseEnumFromDescription_GarnetFlag3() => GarnetEnumUtils.TryParseEnumFromDescription("Read, Write", out RespAclCategories garnetParsedFlag3);

    [Benchmark]
    public bool TryParseEnumFromDescription_SourceGenFlag3() => EnumUtils.TryParseRespAclCategoriesFromDescription("Read, Write", out RespAclCategories sourceGenParsedFlag3);

    [Benchmark]
    public bool TryParseEnumFromDescription_GarnetFlag4() => GarnetEnumUtils.TryParseEnumFromDescription("Read, Write, SortedSet, Transaction", out RespAclCategories garnetParsedFlag4);

    [Benchmark]
    public bool TryParseEnumFromDescription_SourceGenFlag4() => EnumUtils.TryParseRespAclCategoriesFromDescription("Read, Write, SortedSet, Transaction", out RespAclCategories sourceGenParsedFlag4);

    [Benchmark]
    public bool TryParseEnumFromDescription_GarnetFlag5() => GarnetEnumUtils.TryParseEnumFromDescription("All", out RespAclCategories garnetParsedFlag5);

    [Benchmark]
    public bool TryParseEnumFromDescription_SourceGenFlag5() => EnumUtils.TryParseRespAclCategoriesFromDescription("All", out RespAclCategories sourceGenParsedFlag5);
}
```

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Garnet;
+
+[Generator]
+public class EnumsSourceGenerator : ISourceGenerator
```
PaulusParssinen, line 8: ``ISourceGenerators`` are deprecated in favor of [incremental generators](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md). We should use them instead. They're slightly harder to write as you need to develop a value type model to minimize performance hit to the IDE experience (i.e. there should be no syntax nodes or compilation objects which would break the incremental caching).

See more https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/

Meir017: nice catch! done!

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+        return method.ToString();
+    }
+
+    static bool IsPow2(object? value)
+    {
+        if (value is int x) return x != 0 && (x & (x - 1)) == 0;
+        return false;
+    }
+
+    private record struct EnumDetails(string Namespace, string EnumName, List<(string Name, object? Value, AttributeSyntax Description)> Values);
```
PaulusParssinen, line 163: I fear this value type data model is not cacheable and thus breaks the incremental generation.
- .NET collection types rarely implement structural equality 
  - See more about the topic here https://github.com/dotnet/runtime/issues/77183 and see ["4. Watch out for collection types"](https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/#4-watch-out-for-collection-types) in Andrew Lock's article
  - I recommend using [`EquatableArray<T>`](https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/EquatableArray%7BT%7D.cs) by Sergio Pedri (and add 3rd party notice to the NOTICE.md) for all collections in a source generation pipeline
- The `AttributeSyntax` instance in the pipeline is not equatable and such breaks the caching too.

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+                    .SelectMany(al => al.Attributes).Where(a => a.Name.ToString() == "Description")
+                    .SingleOrDefault()
+            ))
+            .ToList();
+        return new EnumDetails(namespaceDeclaration!.Name.ToString(), enumName, values);
+    }
+
+    private static (string EnumName, string ClassSource) Execute(EnumDetails details)
+    {
+        var classBuilder = new StringBuilder();
```
PaulusParssinen, line 45: I highly recommend using [IndentedTextWriter](https://learn.microsoft.com/en-us/dotnet/api/system.codedom.compiler.indentedtextwriter?view=net-8.0) for syntax writing and get rid of the manual indenting which tends to get unmaintainable really fast.

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+
+[Generator]
+public class EnumsSourceGenerator : IIncrementalGenerator
+{
+    const string GeneratedClassName = "EnumUtils";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        var enumDetails = context.SyntaxProvider
+            .CreateSyntaxProvider(
```
PaulusParssinen, line 15: We should use [`ForAttributeWithMetadataName`](https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/#1-use-the-net-7-api-forattributewithmetadataname) which is specially optimized in the Roslyn.

PaulusParssinen (changes requested):
Great job with the work of making the source generator incremental, as they're _very hard_ to get right (I could link dozens of instances where the most experienced .NET developers get them wrong too). 

Some little suggestions left to fix couple issues in the pipeline caching (very, very common pitfalls), but otherwise, lgtm!

<sub>The `ISourceGenerator` interface should have never been made public, as it is confusing everyone now 😅 </sub>

libs/gen/EnumsSourceGenerator.cs
badrishc: I am not sure I follow, whether there is a viable fix to each of these issues raised here?

PaulusParssinen: The fact that this seems to work already is _weird_. The source generator targeting should be limited by the Visual Studio being stuck in .NET FX but I've seen Visual Studio & Roslyn folks work on making the source generator run out-of-proc so maybe that's already working then? Cool!

libs/gen/EnumsSourceGenerator.cs
```diff
@@ -0,0 +1,184 @@
...
+public class EnumsSourceGenerator : IIncrementalGenerator
+{
+    const string GeneratedClassName = "EnumUtils";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        var enumDetails = context.SyntaxProvider
+            .ForAttributeWithMetadataName(
+                "Garnet.common.GenerateEnumDescriptionUtilsAttribute",
+                predicate: static (_, _) => true,
```
PaulusParssinen, line 17:
```suggestion
                predicate: static (node, _) => node is EnumDeclarationSyntax,
```

fast check to make the pipeline little bit more rigorous :)

Meir017: I'm not sure about this, `GenerateEnumDescriptionUtilsAttribute` is already limited to be used only on enums

PaulusParssinen: Sure, that works too.

TalZaccai:
Code generally looks good to me, however when we pull this locally we're having issues specifically in VS. I wasn't able to build Garnet.server (but dotnet build runs fine). This is the warning I'm seeing in the build output:
`warning CS8032: An instance of analyzer Garnet.EnumsSourceGenerator cannot be created from D:\Git\GarnetOSS\libs\gen\bin\Debug\net8.0\Garnet.gen.dll : Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified..`
I asked another dev in the team to check and he said that VS won't even load the solution properly.
Any ideas on why we're seeing these issues here?

PaulusParssinen:
I think the source generator should target NS 2.0, not .NET 8.

Meir017:
> I think the source generator should target NS 2.0, not .NET 8.

This could explain why using the dotnet cli and VS doesn't. 
I'll test using VS and migrate to netstandard2

badrishc:
Source generators in my experience create challenges with debugging and stability across platforms. We have to consider whether this will help us in critical paths. Adding it for non-critical regions of the code might bring more challenges than benefits. Also, we want to stick to net8.0 for the codebase for now. Adding ns2.0 might be overkill just for this capability. Just my 2 cents.

Tornhoof:
> Adding ns2.0 might be overkill just for this capability. 
@badrishc fwiw, only the actual srcgen project needs to be netstandard 2.0, so VS can load it.

The question here is more, is build time srcgen the right approach? How often do these enums change? Is something like a t4 template maybe better? 
( t4 templates suffer other problems, mostly bad tooling in . NET 5+, as VS still runs t4 in netfx)

Note: for larger enums (100+ values) the performance might get worse than dictionary for certain switch/case codepaths.  Especially things like string->enum.

badrishc:
Makes sense. At the end of the day, the open question is (1) will this work seamlessly on all platforms; (2) are enums used in this way anywhere critical in Garnet, where this optimization will have a meaningful benefit.

badrishc:
> > I think the source generator should target NS 2.0, not .NET 8.
> 
> This could explain why using the dotnet cli and VS doesn't. I'll test using VS and migrate to netstandard2

was this change pushed?

Meir017:
pushed now, working in VS


https://github.com/user-attachments/assets/94cb4fca-7df2-4fb9-9baa-5296ac202127

badrishc:
Closing as we have addressed this for the existing enums without using source generators, for now. Thank you!

PR #587, opened 2024-08-12
chore: optimize RegGen.WriteOp flow when opType is not BITOP

Meir017:


benchmark/Resp.benchmark/ReqGenUtils.cs
```diff
@@ -95,10 +85,19 @@ private bool WriteOp(ref byte* curr, byte* vend, OpType opType)
                         return false;
                     break;
                 case OpType.BITOP_AND:
+                    if (!WriteStringBytes(ref curr, vend, Encoding.ASCII.GetBytes("AND")))
```
badrishc, line 88: This would be more efficient as it avoids calling GetBytes every time:

```
while (!RespWriteUtils.WriteDirect(CmdStrings.AND, ref dcurr, dend))
    SendAndReset();
```
and
```
public static ReadOnlySpan<byte> AND => "AND"u8;
```

But also note that is just pre-generation of data in client benchmark, so it is not perf critical.

Meir017:
@badrishc I missed that this is just part of the benchmarks and not the actual server

PR #590, opened 2024-08-14
feature: support azure MangedIdentity TokenCredential

### LINKED ISSUE #558: Allow using managed identity when using Azure Page Blobs
### Feature request type

enhancement

Using a ManagedIdentity instead of a connection-string is more secure when interacting with Azure resources.

see https://learn.microsoft.com/en-us/azure/storage/blobs/authorize-access-azure-active-directory#access-data-with-a-microsoft-entra-account

Introduce a new options of `"storage-managed-identity"` that will be the managed-identity id.

_No response_

### Additional context

_No response_

Meir017:
closes #558

Tests:

- [x] test managed identity in azure 
- [ ] Unit tests


libs/storage/Tsavorite/cs/src/devices/AzureStorageDevice/AzureStorageNamedDeviceFactory.cs
```diff
@@ -29,6 +30,17 @@ public AzureStorageNamedDeviceFactory(string connectionString, ILogger logger =
         {
         }
 
+        /// <summary>
+        /// Create instance of factory for Azure devices
+        /// </summary>
+        /// <param name="serviceUri"></param>
+        /// <param name="credential"></param>
+        /// <param name="logger"></param>
+        public AzureStorageNamedDeviceFactory(string serviceUri, TokenCredential credential, ILogger logger = null)
```
Meir017, line 39: I think that we don't need to explicitly support the managed-identity here, this way we also depend on `Azure.Core` and not `Azure.Identity` here, and any future work for additional `TokenCredential` implementations will be able to use this

badrishc:
How does this feature work exactly? Is it compatible with the older connection strings? And how does it interface with unit tests?

Meir017:
> How does this feature work exactly? 
Assuming the garnet application will run inside of azure (ex: vm / aks) and managed identity will be able to be assigned to the environment 

> Is it compatible with the older connection strings? 
I think this us instead of using a connection string, so any Metadata provided in the connection string will need to be provided in a different way

> And how does it interface with unit tests?
We might consider using `DefaultAzureCredential`  https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet then for unit tests we could run `AZ login` in order to populate the AzureCli credential layer

Currently the unit tests don't use the managed identity credential to perform the login

test/Garnet.test/GarnetServerConfigTests.cs
```diff
@@ -188,41 +188,96 @@ public void ImportExportRedisConfigLocal()
...
         public void ImportExportConfigAzure()
         {
+            if (!TestUtils.IsRunningAzureTests)
+            {
+                Assert.Ignore("Azure tests are disabled.");
+            }
+
             var AzureTestDirectory = $"{TestContext.CurrentContext.Test.MethodName.ToLowerInvariant()}";
             var configPath = $"{AzureTestDirectory}/test1.config";
             var AzureEmulatedStorageString = "UseDevelopmentStorage=true;";
```
badrishc, line 198: Add a unit test that uses the new version of identity.

badrishc: Any progress on this? We would need it to be confident that the feature works.

Meir017: I'm hoping to get to this by the end of this week

badrishc: Ping on this, thank you!

badrishc: Any update on this? Feel free to close and re-open when you have made progress on it, thanks.

Meir017: @badrishc I'm not sure how to test this locally, I got garnet to run but I'm not sure how to force garnet to save into the blob storage

Meir017: OK, I got garnet to interact with the azure blob storage I created.
image

Meir017:
Somthing that might be concerning, I noticed that the constructor of `public AzureStorageNamedDeviceFactory(string serviceUri, TokenCredential credential, ILogger logger = null)` is being called multiple times with the exact same argument when starting the server

running locally with 
```
.\bin\Debug\net8.0\GarnetServer.exe "--use-azure-storage" "--storage-managed-identity" "fake-managed-identity" "--storage-service-uri" "https://garnettest.blob.core.windows.net" "--storage-tier" "--logdir" "garnettestcontainer"
```

called the constructor 5 times, this causes the time until the service prints `* Ready to accept connections` be to ~20s

badrishc:
> Somthing that might be concerning, I noticed that the constructor of `public AzureStorageNamedDeviceFactory(string serviceUri, TokenCredential credential, ILogger logger = null)` is being called multiple times with the exact same argument when starting the server
> 
> running locally with
> 
> ```
> .\bin\Debug\net8.0\GarnetServer.exe "--use-azure-storage" "--storage-managed-identity" "fake-managed-identity" "--storage-service-uri" "https://garnettest.blob.core.windows.net" "--storage-tier" "--logdir" "garnettestcontainer"
> ```
> 
> called the constructor 5 times, this causes the time until the service prints `* Ready to accept connections` be to ~20s
> 
> created #739

apart from this delay issue, is this PR ready yet?

Meir017:
> apart from this delay issue, is this PR ready yet?

Adding unit tests for managed identity is tricky as they would need to run inside azure (not locally)

Other than that and the conflicts this us ready

badrishc:
OK. @Meir017, if you could rebase to the latest in main, I think we can go ahead with reviewing and merging this. thank you.

badrishc:
> Somthing that might be concerning, I noticed that the constructor of `public AzureStorageNamedDeviceFactory(string serviceUri, TokenCredential credential, ILogger logger = null)` is being called multiple times with the exact same argument when starting the server
> 
> running locally with
> 
> ```
> .\bin\Debug\net8.0\GarnetServer.exe "--use-azure-storage" "--storage-managed-identity" "fake-managed-identity" "--storage-service-uri" "https://garnettest.blob.core.windows.net" "--storage-tier" "--logdir" "garnettestcontainer"
> ```
> 
> called the constructor 5 times, this causes the time until the service prints `* Ready to accept connections` be to ~20s
> 
> created #739

Could you try https://github.com/microsoft/garnet/pull/970 to see if the startup delay is fixed by this refactor?

priyanjgupta:
@Meir017 is this PR ready to be merged? We need this feature, to connect with Azure Storage using ManagedIdentity/WorkloadIdentity.
Thank you

badrishc:
@priyanjgupta, if @Meir017 is busy then feel free to directly create a cleaned up PR on this targeting main directly, thanks.

PR #591, opened 2024-08-14
chore: update nodejs version to 20

Meir017:
some places already use nodejs v20 and the v4 action

https://github.com/microsoft/garnet/blob/d6d9a3c4c218c8503a29c254c05a10137e03f17b/.github/workflows/nightly.yml#L68-L71

https://github.com/microsoft/garnet/blob/d6d9a3c4c218c8503a29c254c05a10137e03f17b/.github/workflows/ci.yml#L134-L137


Meir017:
> Is there a specific reason to update nodejs to ver 20?

https://github.com/nodejs/Release version 20 is the current LTS, version 18 is already in Maintenance

darrenge (approved):
Don't see an issue with this since all tests passed.

PR #596, opened 2024-08-16
Migration Addition Logging & Fixes

vazois:
This PR adds more logging messages to migration


Meir017:
Consider using the logging source generator https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

libs/client/ClientSession/GarnetClientSessionMigrationExtensions.cs
```diff
@@ -462,5 +429,31 @@ private bool WriteSerializedKeyValueByteArray(byte[] key, byte[] value, long exp
...
+        /// </summary>
+        /// <param name="keyCount"></param>
+        /// <param name="size"></param>
+        /// <param name="isMainStore"></param>
+        /// <param name="completed"></param>
+        public void LogMigrateProgress(int keyCount, int size, bool isMainStore, bool completed = false)
+        {
+            totalKeyCount += keyCount;
+            totalPayloadSize += size;
+            if (completed || lastLog == 0 || TimeSpan.FromTicks(Stopwatch.GetTimestamp() - lastLog).Seconds >= 1)
```
badrishc, line 448: Will this output a log message every second? If so, can that be made less frequent or configurable?

vazois: Instead of introducing another variable I can tie it to the timeout argument provided by the MIGRATE command.
That way we can see in the logs that we are making progress at least within the timeout period.

badrishc (approved):

PR #597, opened 2024-08-16
Invoke custom commands from custom proc/txn

yrajas:
Adds ability to invoke custom raw string commands and object commands from within custom procedures and transactions.

- [x] New APIs in CustomFunctions to invoke custom commands from custom procedures and transactions
- [x] CustomProcedure registration updated to take in a delegate similar to custom transaction registration
- [x] RespServerSession reference added to CustomFunctions to enable parsing of custom commands and invoking underlying API
- [x] Sample custom procedure and transaction that invoke custom raw string and object commands
- [x] Corresponding tests.


test/Garnet.test/Extensions/TxnCustomCmd.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            var myDictKey = GetNextArg(ref procInput, ref offset);
+            var myDictField = GetNextArg(ref procInput, ref offset);
+            var myDictValue = GetNextArg(ref procInput, ref offset);
+
+            var args = new ArgSlice[2];
+            args[0] = myDictField;
+            args[1] = myDictValue;
+
+            ExecuteCustomCommand(api, "MYDICTSET", myDictKey, args, out var _output);
```
badrishc, line 44: Looks like this will have to match and retrieve the command every single time main is called, which is on the critical path. Users should be able to, in Prepare, get the actual ID so that the critical path call is very light.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -12,25 +12,40 @@ internal sealed class CustomCommandManagerSession
...
             sessionTransactionProcMap = new (CustomTransactionProcedure, int)[CustomCommandManager.MaxRegistrations];
+            sessionCustomProcMap = new CustomProcedure[CustomCommandManager.MaxRegistrations];
+        }
+
+        public CustomProcedure GetCustomProcedure(int id, RespServerSession respServerSession)
+        {
+            if (sessionCustomProcMap[id] == null)
+            {
+                var entry = customCommandManager.customProcedureMap[id] ?? throw new GarnetException($"Custom procedure {id} not found");
+                sessionCustomProcMap[id] = entry.CustomProcedure();
```
badrishc, line 30: name this appropriately, as CustomProcedureFactory

libs/server/Resp/RespServerSession.cs
```diff
@@ -769,7 +769,7 @@ bool NetworkCustomProcedure()
                     return true;
                 }
 
-                TryCustomProcedure(currentCustomProcedure.CustomProcedureImpl);
+                TryCustomProcedure(customCommandManagerSession.GetCustomProcedure(currentCustomProcedure.Id, this));
```
badrishc, line 772: every call to this creates a new factory instance? that does not seem efficient at all.

yrajas: The factory invocation happens only once per session. After that it gets cached and reused similar to custom transaction.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -12,25 +12,40 @@ internal sealed class CustomCommandManagerSession
     {
         readonly CustomCommandManager customCommandManager;
         public readonly (CustomTransactionProcedure, int)[] sessionTransactionProcMap;
+        public readonly CustomProcedure[] sessionCustomProcMap;
```
TalZaccai, line 15: These 2 arrays can be made private, right? Also, might be good to add a comment that the arrays are indexed by the IDs.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -12,25 +12,40 @@ internal sealed class CustomCommandManagerSession
...
         public CustomCommandManagerSession(CustomCommandManager customCommandManager)
         {
             this.customCommandManager = customCommandManager;
             sessionTransactionProcMap = new (CustomTransactionProcedure, int)[CustomCommandManager.MaxRegistrations];
+            sessionCustomProcMap = new CustomProcedure[CustomCommandManager.MaxRegistrations];
+        }
+
+        public CustomProcedure GetCustomProcedure(int id, RespServerSession respServerSession)
+        {
+            if (sessionCustomProcMap[id] == null)
```
TalZaccai, line 27: Should we add a bounds check for the id? Also, are we sure we want to throw if the id is not found? Maybe just return null?

libs/server/Custom/CustomRespCommands.cs
```diff
@@ -194,5 +195,152 @@ private bool TryCustomObjectCommand<TGarnetApi>(GarnetObjectType objType, byte s
...
+        /// <returns>True if successful</returns>
+        public bool InvokeCustomRawStringCommand<TGarnetApi>(ref TGarnetApi storageApi, CustomRawStringCommand customCommand, ArgSlice key, ArgSlice[] args, out ArgSlice output)
+            where TGarnetApi : IGarnetAdvancedApi
+        {
+            ArgumentNullException.ThrowIfNull(customCommand);
+
+            var sbKey = key.SpanByte;
+
+            var inputArg = customCommand.expirationTicks > 0 ? DateTimeOffset.UtcNow.Ticks + customCommand.expirationTicks : customCommand.expirationTicks;
+            var sessionParseState = new SessionParseState();
```
TalZaccai, line 229: No need to instantiate a new parse state, you can use the session's parseState field.

libs/server/Custom/CustomRespCommands.cs
```diff
@@ -194,5 +195,152 @@ private bool TryCustomObjectCommand<TGarnetApi>(GarnetObjectType objType, byte s
...
+        {
+            ArgumentNullException.ThrowIfNull(customObjCommand);
+
+            output = default;
+
+            var keyBytes = key.ToArray();
+
+            // Prepare input
+            var header = new RespInputHeader(customObjCommand.GetObjectType()) { SubId = customObjCommand.subid };
+            var sessionParseState = new SessionParseState();
```
TalZaccai, line 296: Same here, no need to instantiate

TalZaccai (changes requested):

libs/server/Custom/CustomCommandManagerSession.cs
yrajas: Fixed.

libs/server/Custom/CustomCommandManagerSession.cs
yrajas: This id is from matching the commands in the global arrays in CustomCommandManager so if the id isn't found that should be an error condition.

libs/server/Custom/CustomRespCommands.cs
yrajas: The CustomProcedure Execute method's CustomProcedureInput parameter still holds a reference to the session's parseState field and could be used by the CustomProcedure's Execute implementation whenever methods like GetNextArg are called which could happen after the current execution of InvokeCustomRawStringCommand needing to retain the session's parseState field data.

libs/server/Custom/CustomRespCommands.cs
yrajas: Same as above.

libs/server/Custom/CustomRespCommands.cs
TalZaccai: Perhaps we can then have a 2nd parse state field in the session to be lazily instantiated and re-used instead of creating a new instance at each invocation?

TalZaccai (approved):

PR #606, opened 2024-08-22
Upgrading NUnit packages

TalZaccai:


darrenge (approved):

TalZaccai:
> consider adding https://www.nuget.org/packages/NUnit.Analyzers and replacing the old asserts with the NUnit style (`Assert.That(`...). the analyzers will be able to do most of the work as they have some code fixers

I've considered it, I didn't want to make that wide a change just yet, will consult with the team on when's the best timing to do that. Thanks!

PR #611, opened 2024-08-24
Fix Migration for Larger Payloads

vazois:
This PR fixes migration bug when values are larger than the given read buffer and require conversion from SpanByteAndMemory back to SpanByte.
The bug resulted in memory bloating since the acquired buffer to write the SpanByte was larger and thus migration was augmenting the size of the value.

### DIFF
--- a/libs/cluster/Server/Migration/MigrateSessionKeys.cs
+++ b/libs/cluster/Server/Migration/MigrateSessionKeys.cs
@@ -68,29 +68,9 @@ namespace Garnet.cluster
                     }
 
                     // Get SpanByte, check expiration and write to network buffer
-                    SpanByte value = default;
-                    if (o.IsSpanByte)
-                        value = o.SpanByte;
-                    else
-                    {
-                        var metadata = 0L;
-                        var metadataSize = 0;
-
-                        // Extract metadata
-                        fixed (byte* ptr = o.Memory.Memory.Span)
-                        {
-                            metadataSize = *(int*)ptr;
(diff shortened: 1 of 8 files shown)


libs/cluster/Server/Migration/MigrateSessionKeys.cs
```diff
@@ -68,25 +67,16 @@ private bool MigrateKeysFromMainStore()
...
-                    {
-                        memoryHandle = o.Memory.Memory.Pin();
-                        value = SpanByte.FromPinnedMemory(o.Memory.Memory);
-                    }
-                    else
-                        value = o.SpanByte;
+                        value = ref SpanByte.ReinterpretWithoutLength(o.Memory.Memory.Span);
 
                     if (!ClusterSession.Expired(ref value) && !WriteOrSendMainStoreKeyValuePair(ref key, ref value))
                         return false;
```
badrishc, line 76: Make sure all returning code paths dispose the memory here and in other places.

vazois: Good catch! I added a dispose call when write to network buffer fails

badrishc: Try-finally might be a more robust pattern of were know this is a batched operation.

vazois: Check 3b08922 commit. I also re-use the heap buffer across multiple reads now. So, we can avoid multiple allocations if the rented buffer is big enough for the value we are reading.

badrishc (approved):

PR #613, opened 2024-08-26
Updated version from 1.0.18 to 1.0.19

darrenge:
This release has package updates and a migration fix that was needed.


TalZaccai (commented):
Should be another one in GarnetServer.cs?

TalZaccai (approved):

PR #615, opened 2024-08-26
Misc Fixes

vazois:
This PR fixes the following:
- [x]  FlushALL error during slot verification.
- [x] Cluster SetSlotsRange was validating slots incorrectly for STABLE option.
- [x] Cluster SetSlotsRange invalid option error.

### DIFF
--- a/libs/cluster/Server/ClusterManagerSlotState.cs
+++ b/libs/cluster/Server/ClusterManagerSlotState.cs
@@ -422,7 +422,7 @@ namespace Garnet.cluster
         /// Reset slot state to <see cref="SlotState.STABLE"/>
         /// </summary>
         /// <param name="slot">Slot id to reset state</param>
-        public void ResetSlotState(int slot)
+        public void TryResetSlotState(int slot)
         {
             var current = currentConfig;
             var slotState = current.GetState((ushort)slot);
@@ -447,10 +447,14 @@ namespace Garnet.cluster
         /// <param name="slots">Slot list</param>
         public void ResetSlotsState(HashSet<int> slots)
         {
-            foreach (var slot in slots)
+            while (true)
             {
-                ResetSlotState(slot);
+                var current = currentConfig;
+                var newConfig = currentConfig.ResetMultiSlotState(slots);
+                if (Interlocked.CompareExchange(ref currentConfig, newConfig, current) == current)
+                    break;
             }
+            FlushConfig();
         }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


libs/cluster/Server/ClusterManagerSlotState.cs
```diff
@@ -447,10 +447,14 @@ public void ResetSlotState(int slot)
         /// <param name="slots">Slot list</param>
         public void ResetSlotsState(HashSet<int> slots)
         {
-            foreach (var slot in slots)
+            while (true)
```
TalZaccai, line 450: Not familiar with this code, but is there a scenario where this doesn't break?

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -369,7 +369,7 @@ public GarnetServerOptions(ILogger logger = null) : base(logger)
         /// <summary>
         /// Get KVSettings for the main store log
         /// </summary>
-        public KVSettings<SpanByte, SpanByte> GetSettings(ILogger logger, out INamedDeviceFactory logFactory)
+        public KVSettings<SpanByte, SpanByte> GetSettings(ILoggerFactory loggerFactory, ILogger logger, out INamedDeviceFactory logFactory)
```
TalZaccai, line 372: Probably cleaner to instantiate the logger within this method as you're already supplying the logger factory. Also probably a good idea to either rename the out parameter or add an XML comment, since it might be confusing (loggerFactory vs. logFactory).

playground/MigrateBench/MigrateRequest.cs
```diff
@@ -6,6 +6,19 @@
 
 namespace MigrateBench
 {
+    struct Endpoint
+    {
+        public string Address;
+        public int Port;
+
+        public Endpoint(string endpoint)
```
TalZaccai, line 14: Should we handle illegal input?

TalZaccai (changes requested):

badrishc (approved):

libs/cluster/Server/ClusterManagerSlotState.cs
vazois: I checked again to make sure, but it is a common pattern that we use elsewhere. Because we update config in parallel we need to retry until CAS succeeds. Maybe the confusion is on the name? I will prefix it with Try

playground/MigrateBench/MigrateRequest.cs
vazois: Good point. This is test code, but we can use IPEndpoint instead of custom Endpoint.
Pushed a fix for it.

libs/server/Servers/GarnetServerOptions.cs
vazois: Added an XML comment

TalZaccai (approved):

PR #620, opened 2024-08-28
Implement CLIENT|ID, CLIENT|INFO, CLIENT|LIST, and CLIENT|KILL.

kevin-montrose:
This implements a subset of `CLIENT` commands: `ID`, `INFO`, `LIST`, and `KILL`.  For those commands, any filters are also implemented.

This required a number of ancillary changes:
 - `IClusterProvider` now allows checking if a given node id is a replica, not just the current node
 - `IClusterSession` now tracks the node id, if any, of the _connecting_ node - this is inferred based on `CLUSTER|GOSSIP` messages
 - `INetworkSender` now tracks the local endpoint as well as the remote endpoint of a connection
 - `INetworkSender` now has a `TryClose()` method that closes the underlying `Socket`, if any, on first call
 - `RespServerSession` now has a unique, monotonic, id
 - `RespServerSession` now tracks a reference back to an `IGarnetServer`, so it can enumerate _other_ active sessions

I aimed for minimal changes to enable these commands.  Some refactoring might make them a bit cleaner (in particular, I don't love the various `is Blah`s I've added), but would make the change larger.

Some things differ from Redis behavior in a few ways, all of which I _think_ are fine, but are worth pointing out:
 - [`CLIENT|LIST`](https://redis.io/docs/latest/commands/client-list/) documentation gives the impression that combining `TYPE` and `ID` filters is allowed, but that does not appear to be true.
    * This PR matches the actual Redis behavior I have observed, not the documented behavior.
 - [`CLIENT|KILL`](https://redis.io/docs/latest/commands/client-kill/) is documented as ANDing filters, but that doesn't appear to be true in my testing.
    * This PR instead replies with an error if the same filter is specified multiple times.
 - [`CLIENT|LIST`](https://redis.io/docs/latest/commands/client-list/) and [`CLIENT|INFO`](https://redis.io/docs/latest/commands/client-info/) in Redis return many more fields (this PR just does `id`, `addr`, `laddr`, `age`, `flags`, `resp`, and (optionally) `name` and `user`).
    * Since these can change release to release, I think this is fine. 
 - In Redis [`CLIENT|KILL`](https://redis.io/docs/latest/commands/client-kill/) doesn't "take" until the next command is about to be processed.
    * In this PR, I'm calling `Socket.Close()` immediately - which causes in flight reads and writes to fail. 
    * This is motivated by my particular use case: which is draining client connections before some administrative action (restarting or upgrading Garnet, modifying cluster configs, rebooting VMs, etc.).  Waiting for connections to "naturally" die isn't ideal, plus I don't particularly want to put an `isKillRequested`-check in the hot path.


libs/cluster/Server/ClusterProvider.cs
```diff
@@ -119,6 +119,19 @@ public void Dispose()
...
+            var config = clusterManager?.CurrentConfig;
+            if (config is null || !config.IsKnown(nodeId))
+            {
+                return false;
+            }
+
+            var worker = config.GetWorkerFromNodeId(nodeId);
+            return worker.Role == NodeRole.REPLICA;
+        }
+
```
vazois, line 134: It might be better to use GetNodeRoleFromNodeId(nodeId) to avoid having to do two scans of the Workers array.

```suggestion
        public bool IsReplica(string nodeId)
        {
            var config = clusterManager?.CurrentConfig;
            return config is null || config.GetNodeRoleFromNodeId(nodeId) != NodeRole.REPLICA
        }
```

libs/cluster/Session/RespClusterBasicCommands.cs
```diff
@@ -378,6 +380,10 @@ private bool NetworkClusterGossip(out bool invalidParameters)
                 if (gossipWithMeet || current.IsKnown(other.LocalNodeId))
                 {
                     _ = clusterProvider.clusterManager.TryMerge(other);
+
+                    // Remember that this connection is being used for another cluster node to talk to us
+                    Debug.Assert(RemoteNodeId is null || RemoteNodeId == other.LocalNodeId, "Node Id shouldn't change once set for a connection");
```
vazois, line 385: How does this get impacted by using CLUSTER RESET? Are we killing the connection and resetting RemoteNodeId?

libs/server/Resp/BasicCommands.cs
```diff
@@ -1326,6 +1333,486 @@ private bool NetworkASYNC()
             return true;
         }
 
+        /// <summary>
+        /// CLIENT LIST
+        /// </summary>
+        private bool NetworkCLIENTLIST()
```
vazois, line 1339: Can you add this in a separate file? e.g. ClientCommands.cs

libs/cluster/Server/ClusterProvider.cs
kevin-montrose: Ah, I missed that `workers` is padded with a dead value so the `IsKnown(...)` check is unnecessary.

libs/cluster/Session/RespClusterBasicCommands.cs
kevin-montrose: That's a good question.  Looking at the code, I do think a `CLUSTER RESET HARD` might cause that assert to fail.

It's a little odd to keep the connections alive across a `CLUSTER RESET` (soft or hard).

I'll put together a test to see if it's an issue, and think on fixes if so.

kevin-montrose: Ah, it appears we toss the connections as part of banning them during a `CLUSTER|RESET`.

Existing known nodes [get banned here](https://github.com/microsoft/garnet/blob/ccbb29b2fecb7d6078dc324c8b68a390434abb8c/libs/cluster/Server/ClusterManagerWorkerState.cs#L113) in a reset.
And then the main Gossip task will close them on the next pulse ([`GossipMain`](https://github.com/microsoft/garnet/blob/ccbb29b2fecb7d6078dc324c8b68a390434abb8c/libs/cluster/Server/Gossip.cs#L465) -> [`InitConnections`](https://github.com/microsoft/garnet/blob/ccbb29b2fecb7d6078dc324c8b68a390434abb8c/libs/cluster/Server/Gossip.cs#L326) -> [`DisposeBannedWorkerConnections`](https://github.com/microsoft/garnet/blob/ccbb29b2fecb7d6078dc324c8b68a390434abb8c/libs/cluster/Server/Gossip.cs#L304)) before transmitting another gossip message.

vazois: That makes sense, though banning a node will not drop the connection of that remote node. Even if it did, that node would try to re-establish the connection and keep failing with the below warning message. Right now, we have been relying on the admin to issue a cluster forget on that remote node. But in the case where you issue a CLIENT INFO the RemoteId will be null. Is there a race condition here?

kevin-montrose: It is true the connection from the _not reset_ node will not be dropped, but that nodes id hasn't changed - so it's `CLUSTER|GOSSIP` messages won't fail the assert.

Perhaps I'm misunderstanding, but I thought this would work like so:
 - Node ids can only changes if `CLUSTER|RESET` is run on an instance
  - That act will cause the reset node to dispose outgoing connections to other nodes (per the calls from GossipMain above)
  - So any `RespServerSession` active on nodes _other_ than the reset one will be dropped
  - Re-established connections will get a new `RespServerSesion`, with a new (and null) Remote and the first `CLUSTER|GOSSIP` will set the new node id

I [banged together a small test](https://gist.github.com/kevin-montrose/6844bf8914de33af7f7ba57c7834cb59) with re-init a cluster after a bunch of `CLUSTER|RESET`s and it seems to match this behavior.  Is there some path I'm missing?

---

> ... in the case where you issue a CLIENT INFO the RemoteId will be null. Is there a race condition here?

Sort of, but I think it's correct?

When a connection is established between two nodes in a cluster, that connection doesn't have the "character" of a primary or a replica node yet.  That is, it's not authenticated, it hasn't successfully exchange a `CLUSTER|GOSSIP` command, etc.  Current behavior is to treat it as a normal connection until it successfully participates in clustering, which I would argue is correct.

There's analogous behavior here with pubsub connections, those start as normal connections until authenticated and successfully issuing a `(P)SUBSCRIBE` command.

vazois: > It is true the connection from the _not reset_ node will not be dropped, but that nodes id hasn't changed - so it's `CLUSTER|GOSSIP` messages won't fail the assert.
> 
> Perhaps I'm misunderstanding, but I thought this would work like so:
> 
> * Node ids can only changes if `CLUSTER|RESET` is run on an instance
> * That act will cause the reset node to dispose outgoing connections to other nodes (per the calls from GossipMain above)
> * So any `RespServerSession` active on nodes _other_ than the reset one will be dropped
> * Re-established connections will get a new `RespServerSesion`, with a new (and null) Remote and the first `CLUSTER|GOSSIP` will set the new node id
> 
> I [banged together a small test](https://gist.github.com/kevin-montrose/6844bf8914de33af7f7ba57c7834cb59) with re-init a cluster after a bunch of `CLUSTER|RESET`s and it seems to match this behavior. Is there some path I'm missing?
> 
> > ... in the case where you issue a CLIENT INFO the RemoteId will be null. Is there a race condition here?
> 
> Sort of, but I think it's correct?
> 
> When a connection is established between two nodes in a cluster, that connection doesn't have the "character" of a primary or a replica node yet. That is, it's not authenticated, it hasn't successfully exchanged a `CLUSTER|GOSSIP` command, etc. Current behavior is to treat it as a normal connection until it successfully participates in clustering, which I would argue is correct.

This should be fine. We treat cluster connections as anonymous and classify them correctly as soon as they start gossiping.  The inconsistency might arise if the node is forgotten but the connection is still open and "classified" as a primary or replica. Is RemoteNodeId only for internal use? Is it possible to use CLIENT command to view that value? If so, we might get inconsistent info between that and CLUSTER NODES. In that case we might need to drop the connection on CLUSTER FORGET or RESET.

> 
> There's analogous behavior here with pubsub connections, those start as normal connections until authenticated and successfully issuing a `(P)SUBSCRIBE` command.

kevin-montrose: > The inconsistency might arise if the node is forgotten but the connection is still open and "classified" as a primary or replica. Is RemoteNodeId only for internal use? Is it possible to use CLIENT command to view that value? If so, we might get inconsistent info between that and CLUSTER NODES. In that case we might need to drop the connection on CLUSTER FORGET or RESET.

`RemoteNodeId` is only for internal use, just for determine the "character" of a connection - that is, if a connection is from a primary or replica.  `CLIENT|*` commands don't accept a `node-id` parameter, but `CLIENT|LIST` and `CLIENT|INFO` allow listing/killing by "character".

vazois (approved):
Great work!

badrishc (approved):

badrishc:
+1 very nicely done!

PR #623, opened 2024-08-29
Reuse Large Buffers in MigrateSession

vazois:
This PR tries to improve memory utilization and reduce fragmentation by reusing large buffers that were allocated across different migrate sessions.
The PR includes the following:

- [x] Introduce NetworkBufferSettings which is used to define the network buffer size configuration (i.e. send and receive buffer) for both clients and server implementation.
- [x] Augment the network stack to support user defined instances of LimitedFixedBufferPool (LBFP) and NetworkBufferSettings.
- [x] Define and use a shared instance of LFBP for ReplicationManager and MIgrationManager.
- [x] Introduce PURGEBP command to allow for user to purge the LBFP instances of ReplicationManager, MigrationManager and ServerListener.
- [x] Add BPSTATS section in INFO to retrieve stats about the declared LBFP instances of ReplicationManager, MigrationManager and ServerListener at runtime.

Notes:
- There is an upper limit on the number of entries per level in the LimitedBufferPool which may cause fragmentation of the LOH due to the way we allocate and return buffers to the pool itself (shown below) https://github.com/microsoft/garnet/blob/68a7a85c8f0ca367d921b1caf13a18a89f341c34/libs/common/Memory/LimitedFixedBufferPool.cs#L55-L61 The default limit is 16 entries which should be enough for common scenarios (i.e. up to 16 parallel migrate sessions and up to 16 replication sessions).


libs/server/Metrics/Info/GarnetInfoMetrics.cs
```diff
@@ -224,15 +225,15 @@ private void PopulateObjectStoreStats(StoreWrapper storeWrapper)
                 ];
         }
 
-        public void PopulateStoreHashDistribution(StoreWrapper storeWrapper) => storeHashDistrInfo = [new("", storeWrapper.store.DumpDistribution())];
+        private void PopulateStoreHashDistribution(StoreWrapper storeWrapper) => storeHashDistrInfo = [new("", storeWrapper.store.DumpDistribution())];
```
badrishc, line 228: no need to explicitly call out "private" members in the codebase.

libs/common/NetworkBuffers.cs
```diff
@@ -0,0 +1,91 @@
...
+using System;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Garnet.common
+{
+    /// <summary>
+    /// Create a NetworkBuffers instance
+    /// </summary>
+    public struct NetworkBuffers
```
badrishc, line 13: This struct does not do much, can create challenges with single-copy as it is a struct with value semantics, and generally pervades across the codebase. The main this it does is to use two allocation sizes, which can be maintained by callers or within LFBP itself.

badrishc (changes requested):
See comments.

libs/cluster/Server/Migration/MigrationManager.cs
```diff
@@ -3,27 +3,57 @@
...
         {
-            migrationTaskStore.Dispose();
+            migrationTaskStore?.Dispose();
+            networkBuffers.Dispose();
         }
 
+        /// <summary>
+        /// Used to free up buffer pool
+        /// </summary>
+        public void Purge() => networkBuffers.Purge();
```
badrishc, line 49: It can be dangerous to call potentially mutating methods (purge sounds mutating, although its implementation here is safe as it probably passes through into the LFBP class). As this could be a potential copy of the struct, it can create subtle bugs, as well as correctness and maintenance challenges.

libs/common/LightClient.cs
```diff
@@ -212,7 +211,7 @@ public override void Dispose()
         {
             networkSender.ReturnResponseObject();
             networkHandler?.Dispose();
-            networkPool.Dispose();
+            networkBuffers.Dispose();
```
badrishc, line 214: can't really reliably dispose structs.

libs/common/Memory/LimitedFixedBufferPool.cs
```diff
@@ -98,6 +108,23 @@ public unsafe PoolEntry Get(int size)
...
+        /// </summary>
+        public void Purge()
+        {
+            for (var i = 0; i < numLevels; i++)
+            {
+                if (pool[i] == null) continue;
+                // Keep trying Dequeuing until no items left to free
+                while (pool[i].items.TryDequeue(out var entry))
+                {
+                    entry = null;
```
TedHartMS, line 122: What does this "= null" do, given that it immediately goes out of scope? And it is not done in Dispose().

 I also see a Thread.Yield() in Dispose() that is not here.
 
 Can Dispose() use Purge()?

libs/common/Memory/LimitedFixedBufferPool.cs
```diff
@@ -61,8 +69,8 @@ public void Return(PoolEntry buffer)
                         Interlocked.Decrement(ref pool[level].size);
```
TedHartMS, line 69: This needs commenting. My first thought on seeing .size was to wonder why we can't just use the queue.Count method. Now it looks like .size is total number of allocations? The PoolLevel field name comments are uninformative.

libs/common/Memory/LimitedFixedBufferPool.cs
```diff
@@ -107,59 +134,78 @@ public void Dispose()
...
                     if (++count % 10000 == 0)
                         logger?.LogTrace("Dispose iteration {count}, {activeHandlerCount}", count, activeHandlerCount);
 #endif
                 Thread.Yield();
             }
-            for (int i = 0; i < numLevels; i++)
+            for (var i = 0; i < numLevels; i++)
             {
                 if (pool[i] == null) continue;
                 while (pool[i].size > 0)
```
TedHartMS, line 149: If I'm understanding correctly that .size is total # of allocations which may be > maxEntriesPerLevel, then this can spin forever if there are unReturned items. This might warrant an Assert or logging with an exit. (Holding an unReturned item and calling Dispose() (and Purge()?) is bad anyway, so let's make it easier to catch).

libs/common/NetworkBuffers.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Send buffer size.
+        /// (NOTE: Send buffers are fixed and cannot grow automatically. Caller responsible for allocating correct amount and handling larger payloads.)
+        /// </summary>
+        public int sendBufferSize;
+
+        /// <summary>
+        /// Initial allocation size for receive network buffer.
+        /// (NOTE: Receive buffers can automatically grow to accomodate larger payloads.)
+        /// </summary>
```
TedHartMS: Good comments.  "Accommodate"

libs/server/Resp/PurgeBPCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace Garnet.server
+{
+    public enum ManagerType : byte
+    {
+        /// <summary>
+        /// MigrationManager Buffer Pool
+        /// </summary>
+        MM,
+        /// <summary>
```
TedHartMS, line 16: Why not Migration, Replication, ServerSocket?

libs/server/Resp/PurgeBPCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+                ManagerType.SS => "GC completed for ServerSocket"u8,
+                _ => throw new GarnetException()
+            };
+        }
+    }
+
+    internal sealed unsafe partial class RespServerSession : ServerSessionBase
+    {
+        private bool NetworkBurgeBP()
+        {
```
TedHartMS, line 46: "PurgeBP"

TedHartMS (changes requested):

libs/common/Memory/LimitedFixedBufferPool.cs
vazois: We cannot do Dispose on the entry because it will return it to the pool. 
Dequeuing it without setting it to null should be fine since it goes out of scope the GC will take care of it.

libs/common/Memory/LimitedFixedBufferPool.cs
vazois: This PR does not change the LBFP implementation. Changes to its core operations should be part of a separate PR.

libs/common/Memory/LimitedFixedBufferPool.cs
vazois: I am not sure what the comment is here? LFBP implementation has not change. I just added Purge functionality and GetStats option. There might be a need to redesign the buffer pool but should part of another PR.

libs/cluster/Server/Migration/MigrationManager.cs
vazois: Removed the wrapper and now we have a pool separate from NetworkBufferSettings

libs/server/Resp/PurgeBPCommand.cs
badrishc: It should not use the word socket, as that implies TCP. the same network stack can be used with RDMA etc. Maybe ServerListener

vazois: I updated this here https://github.com/microsoft/garnet/pull/623/commits/8ce18094529c8323dd6c1eb4a14cebbb3723535f

libs/common/NetworkBufferSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+            this.maxReceiveBufferSize = maxReceiveBufferSize;
+        }
+
+        /// <summary>
+        /// Allocate network buffer pool
+        /// </summary>
+        /// <param name="maxEntriesPerLevel"></param>
+        /// <param name="logger"></param>
+        /// <returns></returns>
+        public LimitedFixedBufferPool Create(int maxEntriesPerLevel = 16, ILogger logger = null)
```
badrishc, line 56: Rename to CreateBufferPool, as NetworkBufferSettings has nothing to do with buffer pools.

badrishc (approved):

libs/common/NetworkBufferSettings.cs
vazois: Done!

libs/common/Memory/LimitedFixedBufferPool.cs
```diff
@@ -98,6 +138,20 @@ public unsafe PoolEntry Get(int size)
...
+        /// <summary>
+        /// Purge pool entries from all levels
+        /// </summary>
+        public void Purge()
+        {
+            for (var i = 0; i < numLevels; i++)
+            {
+                if (pool[i] == null) continue;
+                // Keep trying Dequeuing until no items left to free
+                while (pool[i].items.TryDequeue(out var _))
```
TedHartMS, line 150: add comment to clarify the intended difference between Purge and Dispose's wrt removing items

TedHartMS (approved):

PR #625, opened 2024-08-30
chore: skip nightly & locker workflows on forks

Meir017:
both the nightly & locker workflow depend on some configuration that is set only on the main garnet repo (not on forks)

### DIFF
--- a/.github/workflows/locker.yml
+++ b/.github/workflows/locker.yml
@@ -19,6 +19,7 @@ permissions:
 
 jobs:
   main:
+    if: github.repository == 'microsoft/garnet'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Actions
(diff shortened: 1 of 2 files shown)


badrishc:
Why exactly do nightly runs fail on forks, what's the error? Can we fix that instead of disabling?

Meir017:
getting many messages of `Environment variable RunAzureTests is not defined` (see https://github.com/Meir017/garnet/actions/runs/10643408367/job/29506930910)

looks like the job fails due to timing out.

badrishc:
> getting many messages of `Environment variable RunAzureTests is not defined` (see https://github.com/Meir017/garnet/actions/runs/10643408367/job/29506930910)
> 
> looks like the job fails due to timing out.

@TedHartMS - Tsavorite tests failing here, any idea why?

darrenge (approved):
Agree, this would be good to skip on forks

badrishc:
We decided that it would be better to run these actions even on forks so that any errors can be worked on by forking contributors. The nightly job is just a copy of https://github.com/dotnet/orleans/blob/main/.github/workflows/locker.yml where they do not disable it on forks either. Once you fork, you can always disable them on your fork as opt-in. Closing accordingly.

PR #645, opened 2024-09-05
Handle status_reply and error_reply properly

mmclure-msft:
This PR adds the `redis.status_reply` and `redis.error_reply` Lua functions to the code base. It also adds code to handle the return value from these functions (a Lua table) correctly - thanks to @danielchristianschroeter for example code.
To match Redis behavior in the `redis.error_reply` case I added a new exception type that does not add the "ERR " prefix to the return message.
Redis output:
```
127.0.0.1:6379> EVAL "return { ok = 'Success' }" 0
Success
127.0.0.1:6379> EVAL "return { err = 'Failure' }" 0
(error) Failure
127.0.0.1:6379> EVAL "return redis.status_reply('Success')" 0
Success
127.0.0.1:6379> EVAL "return redis.error_reply('Failure')" 0
(error) Failure
```
Garnet output with this code:
```
172.24.80.1:6379> EVAL "return { ok = 'Success' }" 0
Success
172.24.80.1:6379> EVAL "return { err = 'Failure' }" 0
(error) Failure
172.24.80.1:6379> EVAL "return redis.status_reply('Success')" 0
Success
172.24.80.1:6379> EVAL "return redis.error_reply('Failure')" 0
(error) Failure
```
I also added a wrapper around String that can be used to allow methods to specify that their return value should be a SimpleString instead of a BulkString to bring Garnet's replies closer to what Redis returns.

### DIFF
--- a/libs/server/Lua/LuaCommands.cs
+++ b/libs/server/Lua/LuaCommands.cs
@@ -196,7 +196,12 @@ namespace Garnet.server
                 object scriptResult = scriptRunner.Run(count, parseState);
                 if (scriptResult != null)
                 {
-                    if (scriptResult is string s)
+                    if (scriptResult is LuaSimpleString lss)
+                    {
+                        while (!RespWriteUtils.WriteSimpleString(lss.getContents(), ref dcurr, dend))
+                            SendAndReset();
+                    }
+                    else if (scriptResult is string s)
                     {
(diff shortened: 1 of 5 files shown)


mmclure-msft:

danielchristianschroeter (approved):

badrishc:
Thanks! Just implemented a slightly more optimized version based on this PR, here: https://github.com/microsoft/garnet/pull/655

PR #646, opened 2024-09-05
Moving itemBroker to storeWrapper

TalZaccai:


libs/server/Resp/Objects/ListCommands.cs
```diff
@@ -347,7 +347,7 @@ private unsafe bool ListBlockingMove(RespCommand command)
                 return true;
             }
 
-            var result = itemBroker.MoveCollectionItemAsync(command, srcKey.ToArray(), this, timeout,
+            var result = storeWrapper.itemBroker.MoveCollectionItemAsync(command, srcKey.ToArray(), this, timeout,
```
badrishc, line 350: Can itemBroker not be null?

TalZaccai: I assume you're asking if it can be null here, I suppose it can if the object store is disabled and someone's invoked BLMOVE. Let me add a null check here...

badrishc (approved):

PR #648, opened 2024-09-06
Added custom authentication support to Garnet.

Xizt:
Changes to support custom authentication. Extended AuthenticationSettings by adding a custom mode which is expected to be passed during server start if custom auth mode is chosen.


Xizt:
@microsoft-github-policy-service agree [company="microsoft"]

libs/host/Configuration/Options.cs
```diff
@@ -493,7 +493,7 @@ public bool IsValid(out List<string> invalidOptions, ILogger logger)
             return isValid;
         }
 
-        public GarnetServerOptions GetServerOptions(ILogger logger = null)
+        public GarnetServerOptions GetServerOptions(IAuthenticationSettings authenticationSettings = null, ILogger logger = null)
```
vazois, line 496: I am not sure I understand why we need this. Instead of passing authenticationSettings as a parameter one could do the following if they need custom authenticationSettings

this.opts = serverSettings.GetServerOptions(this.loggerFactory.CreateLogger("Options"));
this.opts.AuthSettings = new CustomAuthenticationSettings();

libs/server/Auth/Settings/AuthenticationSettings.cs
```diff
@@ -34,7 +34,12 @@ public enum GarnetAuthenticationMode
         /// <summary>
         /// ACL mode using Aad token instead of password. Here username is expected to be ObjectId or a valid Group's Object Id and token will be validated for claims.
         /// </summary>
-        AclWithAad
+        AclWithAad,
+
+        /// <summary>
+        /// Custom authentication mode. Requires a custom authenticator to be implemented and passed during Garnet server instantiation.
+        /// </summary>
+        Custom
```
vazois, line 42: I don't think we should expose the custom option, because the user cannot provide an IAuthenticationSettings command line parameter. It will be confusing and may create bugs.

libs/host/Configuration/Options.cs
```diff
@@ -663,6 +663,14 @@ private IAuthenticationSettings GetAuthenticationSettings(ILogger logger = null)
                 case GarnetAuthenticationMode.AclWithAad:
                     var aadAuthSettings = new AadAuthenticationSettings(AuthorizedAadApplicationIds?.Split(','), AadAudiences?.Split(','), AadIssuers?.Split(','), IssuerSigningTokenProvider.Create(AadAuthority, logger), AadValidateUsername.GetValueOrDefault());
                     return new AclAuthenticationAadSettings(AclFile, Password, aadAuthSettings);
+                case GarnetAuthenticationMode.Custom:
```
vazois, line 666: maybe we could default this to NoAuth and print a warning message. The caller is responsible for providing authentication settings. It might be better than passing it as a parameter.

vazois (changes requested):
Need some changes before merging

libs/host/Configuration/Options.cs
badrishc: +1

libs/host/GarnetServer.cs
```diff
@@ -76,7 +77,7 @@ public class GarnetServer : IDisposable
         /// </summary>
         /// <param name="commandLineArgs">Command line arguments</param>
         /// <param name="loggerFactory">Logger factory</param>
-        public GarnetServer(string[] commandLineArgs, ILoggerFactory loggerFactory = null, bool cleanupDir = false)
+        public GarnetServer(string[] commandLineArgs, IAuthenticationSettings authenticationSettings = null, ILoggerFactory loggerFactory = null, bool cleanupDir = false)
```
badrishc, line 80: Rename to authenticationSettingsOverride and use that if not null, instead of creating a new enum type.

libs/host/Configuration/Options.cs
Xizt: @vazois , Updated.

libs/host/Configuration/Options.cs
Xizt: Updated

libs/host/Configuration/Options.cs
```diff
@@ -664,8 +664,8 @@ private IAuthenticationSettings GetAuthenticationSettings(ILogger logger = null)
                     var aadAuthSettings = new AadAuthenticationSettings(AuthorizedAadApplicationIds?.Split(','), AadAudiences?.Split(','), AadIssuers?.Split(','), IssuerSigningTokenProvider.Create(AadAuthority, logger), AadValidateUsername.GetValueOrDefault());
                     return new AclAuthenticationAadSettings(AclFile, Password, aadAuthSettings);
                 default:
-                    logger?.LogError("Unsupported authentication mode: {mode}", AuthenticationMode);
-                    throw new Exception($"Authentication mode {AuthenticationMode} is not supported.");
+                    logger?.LogWarning("Defaulting to NoAuth if authenticationSettingsOverride is not provided.");
```
badrishc, line 667: why is this needed? you can send NoAuth and provide override.

Xizt: Resolved, Thanks

libs/host/GarnetServer.cs
```diff
@@ -76,7 +77,7 @@ public class GarnetServer : IDisposable
         /// </summary>
         /// <param name="commandLineArgs">Command line arguments</param>
         /// <param name="loggerFactory">Logger factory</param>
-        public GarnetServer(string[] commandLineArgs, ILoggerFactory loggerFactory = null, bool cleanupDir = false)
+        public GarnetServer(string[] commandLineArgs, IAuthenticationSettings authenticationSettingsOverride = null, ILoggerFactory loggerFactory = null, bool cleanupDir = false)
```
msft-paddy14, line 80: should we not create a new constructor for this or add it at the end atleast?
This will break for anyone already using existing optional parameters ( ex: an ILoggerFactory instance as 2nd parameter)

Xizt: Thanks. Need to keep in mind for compatibility.

badrishc (approved):

vazois (approved):

Issue #638, opened 2024-09-05
[Bug] `ACL LIST` not working


Alirexaa:
### Describe the bug

running `ACL LIST` produces `Error: Server closed the connection` in redis-cli.

Garnet logs:

```
= _________

/_||___||_\ Garnet 1.0.19 64 bit; standalone mode

'. \ / .' Port: 6379

'.\ /.' https://aka.ms/GetGarnet

'.'


* Ready to accept connections


10::09::06 crit: Session[0] [172.17.0.2:6379] [172.17.0.1:49456] [022EA309] ProcessMessages threw an exception: System.InvalidCastException: Unable to cast object of type 'Garnet.server.Auth.GarnetNoAuthAuthenticator' to type 'Garnet.server.Auth.GarnetACLAuthenticator'. at Garnet.server.RespServerSession.NetworkAclList() in /src/libs/server/Resp/ACLCommands.cs:line 34 at Garnet.server.RespServerSession.ProcessAdminCommands(RespCommand command) in /src/libs/server/Resp/AdminCommands.cs:line 53 at Garnet.server.RespServerSession.ProcessOtherCommands[TGarnetApi](RespCommand command, TGarnetApi& storageApi) in /src/libs/server/Resp/RespServerSession.cs:line 739 at Garnet.server.RespServerSession.ProcessArrayCommands[TGarnetApi](RespCommand cmd, TGarnetApi& storageApi) in /src/libs/server/Resp/RespServerSession.cs:line 666 at Garnet.server.RespServerSession.ProcessMessages() in /src/libs/server/Resp/RespServerSession.cs:line 420 at Garnet.server.RespServerSession.TryConsumeMessages(Byte* reqBuffer, Int32 bytesReceived) in /src/libs/server/Resp/RespServerSession.cs:line 299
```

### Steps to reproduce the bug

1-Run Garnet
2-Run `ACL LIST` in redis-cli
3-Checking redis-cli response and garnet console logs.

### Release version

v1.0.19

msft-paddy14:
can you share more details like what parameters you used to start this? Did you configure ACL authenticator correctly?

Alirexaa:
> can you share more details like what parameters you used to start this? Did you configure ACL authenticator correctly?

just using `docker run ghcr.io/microsoft/garnet:1.0`
no parameters. no configuration.

vazois:
[...]

PR #650, opened 2024-09-06
Additional Checks for ACL configuration

vazois:
This PR is in response to issue #638.

- [x] Additional checks when calling ACL commands to ensure the ACL authentication is enabled.
- [x] Additional checks to ensure ACL file path is defined and can be used with ACL LOAD/SAVE

### DIFF
--- a/libs/server/Resp/ACLCommands.cs
+++ b/libs/server/Resp/ACLCommands.cs
@@ -16,7 +16,7 @@ namespace Garnet.server
     /// </summary>
     internal sealed unsafe partial class RespServerSession : ServerSessionBase
     {
-        private bool IsACLAuthenticatorEnabled()
+        private bool ValidateACLAuthenticator()
         {
             if (_authenticator is null or not GarnetACLAuthenticator)
             {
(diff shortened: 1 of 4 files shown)


libs/server/Resp/ACLCommands.cs
```diff
@@ -17,6 +16,37 @@ namespace Garnet.server
     /// </summary>
     internal sealed unsafe partial class RespServerSession : ServerSessionBase
     {
+        private bool IsACLAuthenticatorEnabled()
```
kevin-montrose, line 19: I would not expect an `IsXXX()` method to _write a response_.  Something like `ValidateXXX` or `EnforceXXX` would communicate that a response might be written.

kevin-montrose (changes requested):

libs/server/Resp/ACLCommands.cs
kevin-montrose: Beyond that first blush of confusion, LGTM.

vazois: Makes sense. I just pushed a change using ValidateACLAuthenticator and ValidateACLFileUse names

badrishc (approved):

kevin-montrose (approved):

Issue #651, opened 2024-09-08
[Bug] RENAME removes expiration of the renamed key


Vijay-Nirmal:
### Describe the bug

RENAME command in garnet removes the expiration details from the key after renaming the new key

### Steps to reproduce the bug

Do the following in the Redis CLI to reproduce

```cli
127.0.0.1:6379> SET KeyA ValueA EX 60
OK
127.0.0.1:6379> TTL KeyA
(integer) 54
127.0.0.1:6379> RENAME KeyA KeyB
OK
127.0.0.1:6379> TTL KeyB
(integer) -1
127.0.0.1:6379> GET KeyB
"ValueA"
127.0.0.1:6379> GET KeyA
(nil)
```

### Expected behavior

Here is the behavior from Redis (Memurai)

```cli
127.0.0.1:6379> SET KeyA ValueA EX 60
OK
127.0.0.1:6379> TTL KeyA
(integer) 55
127.0.0.1:6379> RENAME KeyA KeyB
OK
127.0.0.1:6379> TTL KeyB
(integer) 48
127.0.0.1:6379> GET KeyB
"ValueA"
127.0.0.1:6379> GET KeyA
(nil)
```

### Release version

Main branch -> Last Commit SHA: 6c68094c3c0b6f95e4082c5489691e15d17a73a1

### IDE

Visual Studio 2022

### OS version

Windows 11 23H2

### Additional context

I might be able to create a PR provided I can fix the issue. Issue is caused because of the below line not passing the current TTL when calling `SET`

https://github.com/microsoft/garnet/blob/6c68094c3c0b6f95e4082c5489691e15d17a73a1/libs/server/Storage/Session/MainStore/MainStoreOps.cs#L562

PR #652, opened 2024-09-08
[Bug] Fixed RENAME command removing the expiration of the renamed key

Vijay-Nirmal:
This PR fixes RENAME command removing the expiration of the renamed key

Fix for #651

### DIFF
--- a/libs/server/Storage/Session/MainStore/MainStoreOps.cs
+++ b/libs/server/Storage/Session/MainStore/MainStoreOps.cs
@@ -539,14 +539,12 @@ namespace Garnet.server
 
             var context = txnManager.LockableContext;
             var objectContext = txnManager.ObjectStoreLockableContext;
+            SpanByte oldKey = oldKeySlice.SpanByte;
 
             if (storeType == StoreType.Main || storeType == StoreType.All)
             {
                 try
                 {
-                    SpanByte oldKey = oldKeySlice.SpanByte;
-                    SpanByte newKey = newKeySlice.SpanByte;
-
                     SpanByte input = default;
                     var o = new SpanByteAndMemory();
                     var status = GET(ref oldKey, ref input, ref o, ref context);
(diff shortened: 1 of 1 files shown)


libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -558,16 +556,39 @@ public unsafe GarnetStatus RENAME(ArgSlice oldKeySlice, ArgSlice newKeySlice, St
...
+                            var expireMemoryHandle = expireSpan.Memory.Memory.Pin();
+                            var expirePtrVal = (byte*)expireMemoryHandle.Pointer;
+                            RespReadUtils.TryRead64Int(out var expireTimeMs, ref expirePtrVal, expirePtrVal + expireSpan.Length, out var _);
+
+                            // If the key has an expiration, set the new key with the expiration
+                            if (expireTimeMs > 0)
+                            {
+                                SETEX(newKeySlice, new ArgSlice(ptrVal, headerLength), TimeSpan.FromMilliseconds(expireTimeMs), ref context);
+                            }
+                            else if (expireTimeMs == -1) // Its possible to have expire as 0 or -2, in those cases we don't SET the new key
```
badrishc, line 575: Check what expire time of 0 and -2 indicate.

TalZaccai: To my understanding:
0 - Key expired
-1 - No expiry associated with key
-2 - Key does not exist

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -558,16 +556,39 @@ public unsafe GarnetStatus RENAME(ArgSlice oldKeySlice, ArgSlice newKeySlice, St
...
-                        DELETE(ref oldKey, StoreType.Main, ref context, ref objectContext);
 
-                        returnStatus = GarnetStatus.OK;
+                        // Find expiration time of the old key
+                        var expireSpan = new SpanByteAndMemory();
+                        var ttlStatus = TTL(ref oldKey, storeType, ref expireSpan, ref context, ref objectContext, true);
+
+                        if (ttlStatus == GarnetStatus.OK && !expireSpan.IsSpanByte)
+                        {
+                            var expireMemoryHandle = expireSpan.Memory.Memory.Pin();
```
TalZaccai, line 566: Can be wrapped in a using

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -591,19 +612,40 @@ public unsafe GarnetStatus RENAME(ArgSlice oldKeySlice, ArgSlice newKeySlice, St
...
-                        // Delete the old key
-                        DELETE(oldKeyArray, StoreType.Object, ref context, ref objectContext);
+                        byte[] newKeyArray = newKeySlice.ToArray();
+
+                        var expireSpan = new SpanByteAndMemory();
+                        var ttlStatus = TTL(ref oldKey, StoreType.Object, ref expireSpan, ref context, ref objectContext, true);
+                        
+                        if (ttlStatus == GarnetStatus.OK && !expireSpan.IsSpanByte)
+                        {
+                            var expireMemoryHandle = expireSpan.Memory.Memory.Pin();
```
TalZaccai, line 627: Can be wrapped in a using

TalZaccai (commented):
Looks good to me overall. Thanks for pushing this fix! 
Added a couple of tests & a couple of comments.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: @TalZaccai Is right. That is the reason I added a comment saying we shouldn't call `SET` in those cases because it could create a KEY that is not there now (or shouldn't be there now). I will add the reason behind it in the comments to clarify.

Vijay-Nirmal:
@TalZaccai Fixed the review comments. Thanks for adding tests, I will make sure to add them in future PRs. I have a few questions.

1. Should every `.Dispose()` be wrapped in `using` or `try...finally`? I see almost all `.Dispose()` not being wrapped. I thought it could be some performance optimization thing.
2. (Off topic) I am planning to create PRs for adding `RENAMENX`, `RANDOMKEY`, `EXPIRETIME`, `PEXPIRETIME` and others when time permits. Should I need to create an issue first before raising the PR? What is the standard practice here?

TalZaccai:
> @TalZaccai Fixed the review comments. Thanks for adding tests, I will make sure to add them in future PRs. I have a few questions.
> 
> 1. Should every `.Dispose()` be wrapped in `using` or `try...finally`? I see almost all `.Dispose()` not being wrapped. I thought it could be some performance optimization thing.
> 2. (Off topic) I am planning to create PRs for adding `RENAMENX`, `RANDOMKEY`, `EXPIRETIME`, `PEXPIRETIME` and others when time permits. Should I need to create an issue first before raising the PR? What is the standard practice here?

1. Possibly... Would probably be better to put the Dispose calls in the finally in any case (with a null check ofc).
2. No need to create an issue, just create the PR. You can refer to this [guide](https://microsoft.github.io/garnet/docs/dev/garnet-api) on how to add new commands to Garnet to make sure you haven't missed anything (it might be slightly outdated, I need to go over it soon)

TalZaccai (approved):

PR #657, opened 2024-09-12
Tsavorite allocator - tighten the packing of pages

badrishc:
This is a change to the Tsavorite allocator base to make its page-filling algorithm (primarily for TsavoriteLog) deterministic.
 
We change the allocator to enqueue at tail with the invariant that the first record of page `(p+1)` is guaranteed to have not fit in the empty space at the end of the previous page (`p`).
 
This allows replication to independently replay AOF records and guarantee that they fit on the AOF created on the secondary, exactly in the same way as they do on the primary. I.e., they will be in perfect lockstep.
 
Previously it was possible that two threads raced to add at the end of a page, the larger allocation won first and closed that page, but the smaller allocation won on the new page and populated the first entry with a record that could have fitted on the previous page. On the replica, this record would end up on the previous page, resulting in offset mismatch.

We already have a workaround in the replica replay logic to correct for this, but this PR makes that workaround unnecessary.

PR also includes an unrelated fix to Upsert logic from Ted: [Fix InernalUpsert srcRecordInfo setting when found below ReadOnlyAddress](https://github.com/microsoft/garnet/pull/657/commits/1c95a5ac908952225920a2d431a1e8124246b8b2)

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/BlockAllocate.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/BlockAllocate.cs
@@ -28,6 +28,7 @@ namespace Tsavorite.core
                 return true;
             }
 
+            // logicalAddress less than 0 (RETRY_NOW) should already have been handled
             Debug.Assert(logicalAddress == 0);
             // We expect flushEvent to be signaled.
             internalStatus = OperationStatus.ALLOCATE_FAILED;
--- a/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/BlockAllocate.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/BlockAllocate.cs
@@ -20,7 +20,7 @@ namespace Tsavorite.core
                 out OperationStatus internalStatus)
         {
             pendingContext.flushEvent = allocator.FlushEvent;
-            logicalAddress = allocator.TryAllocate(recordSize);
+            logicalAddress = allocator.TryAllocateRetryNow(recordSize);
             if (logicalAddress > 0)
             {
                 pendingContext.flushEvent = default;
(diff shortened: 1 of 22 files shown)


libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/BlockAllocate.cs
```diff
@@ -20,32 +20,28 @@ private static bool TryBlockAllocate<TInput, TOutput, TContext>(
...
-                // We expect flushEvent to be signaled.
-                internalStatus = OperationStatus.ALLOCATE_FAILED;
-                return false;
-            }
-
-            // logicalAddress is < 0 so we do not expect flushEvent to be signaled; return RETRY_LATER to refresh the epoch.
-            pendingContext.flushEvent = default;
-            allocator.TryComplete();
-            internalStatus = OperationStatus.RETRY_LATER;
+            Debug.Assert(logicalAddress == 0);
```
TedHartMS, line 31: Add a message to the Assert e.g. "<0 (RETRY_NOW) should already have been handled"

libs/storage/Tsavorite/cs/src/core/TsavoriteLog/TsavoriteLog.cs
```diff
@@ -872,28 +861,20 @@ private long AllocateBlock(int recordSize)
             while (true)
             {
                 var flushEvent = allocator.FlushEvent;
-                var logicalAddress = allocator.TryAllocate(recordSize);
+                var logicalAddress = allocator.TryAllocateRetryNow(recordSize);
                 if (logicalAddress > 0)
                     return logicalAddress;
-
-                if (logicalAddress == 0)
+                Debug.Assert(logicalAddress == 0);
```
TedHartMS, line 867: Add a message to the Assert e.g. "<0 (RETRY_NOW) should already have been handled"

TedHartMS (approved):

TedHartMS (approved):

vazois (approved):

PR #661, opened 2024-09-14
[Compatibility] Adding RENAMENX command

Vijay-Nirmal:
Adding support for RENAMENX command in Garnet


Vijay-Nirmal:
Is there a way to check the existence of the object without calling `GET` which the current `EXISTS` method does?
Or is there a similar method for object type like `SET NX`?

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -40,6 +40,46 @@ private bool NetworkRENAME<TGarnetApi>(ref TGarnetApi storageApi)
...
+            where TGarnetApi : IGarnetApi
+        {
+            if (parseState.Count != 2)
+            {
+                return AbortWithWrongNumberOfArguments(nameof(RespCommand.RENAMENX));
+            }
+
+            var oldKeySlice = parseState.GetArgSliceByRef(0);
+            var newKeySlice = parseState.GetArgSliceByRef(1);
+            var status = storageApi.RENAMENX(oldKeySlice, newKeySlice);
```
TalZaccai, line 56: GarnetStatus is not meant to be interpreted this way, you should pass an output parameter to this method and write the appropriate output to it.

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -620,6 +621,7 @@ private RespCommand FastParseCommand(out int count)
                         (2 << 4) | 6 when lastWord == MemoryMarshal.Read<ulong>("INCRBY\r\n"u8) => RespCommand.INCRBY,
                         (2 << 4) | 6 when lastWord == MemoryMarshal.Read<ulong>("DECRBY\r\n"u8) => RespCommand.DECRBY,
                         (2 << 4) | 6 when lastWord == MemoryMarshal.Read<ulong>("RENAME\r\n"u8) => RespCommand.RENAME,
+                        (2 << 4) | 8 when lastWord == MemoryMarshal.Read<ulong>("NAMENX\r\n"u8) => RespCommand.RENAMENX,
```
TalZaccai, line 624: You need to read the 'RE' prefix as well

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -618,32 +618,171 @@ public unsafe GarnetStatus RENAME(ArgSlice oldKeySlice, ArgSlice newKeySlice, St
...
+        /// </summary>
+        /// <param name="oldKeySlice">The old key to be renamed.</param>
+        /// <param name="newKeySlice">The new key name.</param>
+        /// <param name="storeType">The type of store to perform the operation on.</param>
+        /// <returns>
+        ///     GarnetStatus OK means rename operation is successful
+        ///     GarnetStatus NOTFOUND means old key does not exists
+        ///     GarnetStatus MOVED means new key already exists
+        /// </returns>
+        public unsafe GarnetStatus RENAMENX(ArgSlice oldKeySlice, ArgSlice newKeySlice, StoreType storeType)
```
TalZaccai, line 651: Since this is majorly similar to RENAME, I would suggest adding a bool parameter to the RENAME method to specify if rename should happen only if key doesn't exist, instead of copying all this code (as any future change to this would need to happen in multiple locations).

TalZaccai (changes requested):
Thanks for your contribution! Great job adding such comprehensive testing. 
Added a few comments...

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Agree I will make it into a single method, was planned but didn't do it at the end

libs/server/Resp/KeyAdminCommands.cs
Vijay-Nirmal: Will do it

TalZaccai:
> Is there a way to check the existence of the object without calling `GET` which the current `EXISTS` method does? Or is there a similar method for object type like `SET NX`?

GET is probably your best bet in terms of performance.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Converted into single method. Not my best work but I think its good enough.

libs/server/Resp/Parser/RespCommand.cs
Vijay-Nirmal: Fixed

libs/server/Resp/KeyAdminCommands.cs
Vijay-Nirmal: Fixed

badrishc (approved):

PR #664, opened 2024-09-15
[Compatibility] Adding EXPIRETIME and PEXPIRETIME command

Vijay-Nirmal:
Adding EXPIRETIME and EXPIRETIME command to garnet

- [x] Add EXPIRETIME and EXPIRETIME command
- [x] Add Test cases
- [x] Add documentation

Open Question:
1. For Object type, Should I create a single GarnetObjectType called `Expiration` instead of `EXPIRETIME` and `PEXPIRETIME`? The advantage will be that it will be a single generic implementation that can be reused for others in the future if required. The disadvantage is that there will be a slight performance penalty but no additional memory needed since values of `Expiration`, `EXPIRETIME` and `PEXPIRETIME` are `long` type, so we can do in-place override the value after recomputing the required value.


TalZaccai:
> Adding EXPIRETIME and EXPIRETIME command to garnet
> 
> * [x]  Add EXPIRETIME and EXPIRETIME command
> * [x]  Add Test cases
> * [x]  Add documentation
> 
> Open Question:
> 
> 1. For Object type, Should I create a single GarnetObjectType called `Expiration` instead of `EXPIRETIME` and `PEXPIRETIME`? The advantage will be that it will be a single generic implementation that can be reused for others in the future if required. The disadvantage is that there will be a slight performance penalty but no additional memory needed since values of `Expiration`, `EXPIRETIME` and `PEXPIRETIME` are `long` type, so we can do in-place override the value after recomputing the required value.

There is already a GarnetObjectType.Expire, why not use that one?

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -302,6 +302,72 @@ public unsafe GarnetStatus TTL<TContext, TObjectContext>(ref SpanByte key, Store
...
+        /// <param name="storeType">The store to operate on</param>
+        /// <param name="output">Span to allocate the output of the operation</param>
+        /// <param name="context">Basic Context of the store</param>
+        /// <param name="objectContext">Object Context of the store</param>
+        /// <param name="milliseconds">when true the command to execute is PEXPIRETIME.</param>
+        /// <returns>Returns the absolute Unix timestamp (since January 1, 1970) in seconds or milliseconds at which the given key will expire.</returns>
+        public unsafe GarnetStatus EXPIRETIME<TContext, TObjectContext>(ref SpanByte key, StoreType storeType, ref SpanByteAndMemory output, ref TContext context, ref TObjectContext objectContext, bool milliseconds = false)
+            where TContext : ITsavoriteContext<SpanByte, SpanByte, SpanByte, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator>
+            where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
+        {
```
TalZaccai, line 320: Why not simply call TTL and convert it to unix-time?

TalZaccai (changes requested):
Thanks for your contribution! Added a couple of comments.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: That was my first implementation, and then I did this implementation and did a microbenchmark to compare the performance. I found this method is much faster than calling editing TTL and converting to Unix time. I don't have the number handy now but it was like 60% difference. TTL + Convert took about 15 ns and this approach was taking about 9 ns. I know its still few `ns`. @TalZaccai If you are fine with this performance difference then I can change it to TTL + Convert logic. Your call

libs/server/Objects/Types/GarnetObjectType.cs
```diff
@@ -29,6 +29,16 @@ public enum GarnetObjectType : byte
         /// </summary>        
         Set,
 
+        /// <summary>
+        /// Special type indicating EXPIRETIME command
+        /// </summary>
+        Expiretime = 0xf9,
```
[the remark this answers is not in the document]
Vijay-Nirmal: As mentioned in the other comment if we are going to use TTL + Convert logic then this is not need. I will wait for you confirmation before changing as I mentioned about minor perf difference.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -302,6 +302,72 @@ public unsafe GarnetStatus TTL<TContext, TObjectContext>(ref SpanByte key, Store
...
+                if (status.Found) return GarnetStatus.OK;
+            }
+
+            if ((storeType == StoreType.Object || storeType == StoreType.All) && !objectStoreBasicContext.IsNull)
+            {
+                var objInput = new ObjectInput
+                {
+                    header = new RespInputHeader
+                    {
+                        cmd = milliseconds ? RespCommand.PEXPIRETIME : RespCommand.EXPIRETIME,
```
TalZaccai, line 350: FYI - You can't set both cmd and type since they're written to the same area of memory, use only type in this case.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
TalZaccai: Interesting, I wonder why it would take more time, as this would still do the conversion, only in a different area of the code. 

@badrishc do you have a preference here?

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Thanks for the info, I didn't notice the `FieldOffset` set to `0` for both the fields, Will have the changes required

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Reason for the performance difference is because TTL will write its response in `SpanByteAndMemory` and change its available length. Since UnixTimestamp is always bigger than TTL or PTTL value we are override in the existing space in `SpanByteAndMemory` which makes the code to `Rent` a new memory for the response.

One way I though of overcome this is by resetting the length of `SpanByteAndMemory` to initial after getting the response from TTL/PTTL but not sure how safe it is. I think it would be safe as I would assume those a pinned memory and won't be disposed till we right the response back from the server. If you are fine with doing this then I can use TTL/PTTL itself or if there is a better way, let me know I can implement in that manner.

Vijay-Nirmal: @TalZaccai @badrishc Any update? I can make the suggested changes if you guys let me know what you prefer.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Fixed it

libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
```diff
@@ -37,6 +37,15 @@ public bool SingleReader(ref byte[] key, ref ObjectInput input, ref IGarnetObjec
                 if ((byte)input.header.type < CustomCommandManager.StartOffset)
                     return value.Operate(ref input, ref dst.spanByteAndMemory, out _, out _);
 
+                if (input.header.type == GarnetObjectType.Expiretime || input.header.type == GarnetObjectType.PExpiretime)
```
badrishc, line 40: Can create a single switch with all variants of TTL and ExpireTime, to consolidate.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
badrishc: Since TTL writes directly to the network, it should be fine to have a separate ExpireTime that writes the time directly to the network buffer as this PR does.

badrishc (approved):

libs/server/Storage/Functions/ObjectStore/ReadMethods.cs
Vijay-Nirmal: Added `switch` for the entire type so that in future if there is more `GarnetObjectType` that needs to be handled in the ObjectSession layer itself, we can use the same switch

Vijay-Nirmal:
@badrishc @TalZaccai Fixed the pending review comments

TalZaccai (approved):
Thank you! Please resolve conflicts and I will merge this in :)

Vijay-Nirmal:
@TalZaccai Merged the latest

TalZaccai (approved):

PR #666, opened 2024-09-16
[Compatibility] Added EXPIREAT and PEXPIREAT command and bug fixes for EXPIRE and PEXPIRE

Vijay-Nirmal:
Adding EXPIREAT and PEXPIREAT command to garnet

- [x] Add EXPIREAT and PEXPIREAT command
- [x] Add Test cases
- [x] Add documentation
- [x] Fix a bug in EXPIRE/PEXPIRE where it doesn't support valid option combination
- [x] Fix a bug in EXPIRE/PEXPIRE where it doesn't treat non-volatile key as an infinite TTL for GT and LT option


libs/server/ExpireOption.cs
```diff
@@ -13,23 +13,31 @@ public enum ExpireOption : byte
...
+        GT = 1 << 2,
         /// <summary>
         /// Set expiry only when the new expiry is less than current one
         /// </summary>
-        LT
+        LT = 1 << 3,
+        /// <summary>
+        /// Set expiry only when the key has an existing expiry and the new expiry is greater than current one
+        /// </summary>
+        XXGT = XX | GT,
```
TalZaccai, line 36: NX and GT / LT are mutually exclusive, so the presence of GT / LT implies XX

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -148,6 +148,36 @@ private bool NetworkEXPIRE<TGarnetApi>(RespCommand command, ref TGarnetApi stora
...
+                if (!parseState.TryGetEnum(3, true, out ExpireOption additionExpireOption) || !additionExpireOption.IsValid(ref parseState.GetArgSliceByRef(3)))
+                {
+                    var optionStr = parseState.GetString(3);
+
+                    while (!RespWriteUtils.WriteError($"ERR Unsupported option {optionStr}", ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+                }
+
+                if (expireOption == ExpireOption.XX && (additionExpireOption == ExpireOption.GT || additionExpireOption == ExpireOption.LT))
```
TalZaccai, line 162: You're correct here that it is legal to specify both LT / GT and XX, and that an error message needs to be presented if there is an illegal combination of parameters, but there's no point in adding XXGT and XXLT, since they're essentially the same as GT, LT

libs/server/Storage/Session/MainStore/MainStoreOps.cs
```diff
@@ -726,20 +726,63 @@ public unsafe GarnetStatus EXPIRE<TContext, TObjectContext>(ArgSlice key, ArgSli
...
+        /// <returns>Return GarnetStatus.OK when key found, else GarnetStatus.NOTFOUND</returns>
+        public unsafe GarnetStatus EXPIREAT<TContext, TObjectContext>(ArgSlice key, long expiryTimestamp, out bool timeoutSet, StoreType storeType, ExpireOption expireOption, ref TContext context, ref TObjectContext objectStoreContext, bool milliseconds = false)
+            where TContext : ITsavoriteContext<SpanByte, SpanByte, SpanByte, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator>
+            where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
+        {
+            var expiryTimestampTicks = milliseconds ? ConvertUtils.UnixTimestampInMillisecondsToTicks(expiryTimestamp) : ConvertUtils.UnixTimestampInSecondsToTicks(expiryTimestamp);
+            return EXPIRE(key, expiryTimestampTicks, out timeoutSet, storeType, expireOption, ref context, ref objectStoreContext, milliseconds ? RespCommand.PEXPIREAT : RespCommand.EXPIREAT);
```
TalZaccai, line 756: You can just use RespCommand.EXPIRE / RespCommand.PEXPIRE here, since they're essentially doing the same thing

TalZaccai (changes requested):
Great additions! Thanks for adding the options to EXPIRE / PEXPIRE! Added a couple of comments.

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Hmm, I think I can make it work with RespCommand.EXPIRE / RespCommand.PEXPIRE by computing the expire time from UnixTime. Thanks for pointing this out

libs/server/ExpireOption.cs
Vijay-Nirmal: It's not the case. Below is an example. when values are not there LT and GT will treat it as infinite expiration time but with `XX` setting it doesn't, it will not apply the expiration time if there is no TTL for the key currently.
image

libs/server/Resp/KeyAdminCommands.cs
Vijay-Nirmal: Answered in the above review comment. @TalZaccai  Let me know if you are fine with having XXGT and XXLT

libs/server/ExpireOption.cs
TalZaccai: > It's not the case. Below is an example. when values are not there LT and GT will treat it as infinite expiration time but with `XX` setting it doesn't, it will not apply the expiration time if there is no TTL for the key currently. image

Oh that's good to know, so I guess only GT implies XX then?

libs/server/Resp/KeyAdminCommands.cs
TalZaccai: Yeah then it's ok. Can we drop XXGT though?

libs/server/ExpireOption.cs
```diff
@@ -13,23 +13,31 @@ public enum ExpireOption : byte
         /// <summary>
```
TalZaccai, line 13: Add a Flags attribute to this enum

libs/server/ExpireOption.cs
Vijay-Nirmal: > so I guess only GT implies XX then?

Technically no but practically yes because nothing can be bigger than infinite.

Can I mark this review comments as resolved without making any change?

libs/server/Resp/KeyAdminCommands.cs
Vijay-Nirmal: Personally, I like to have it to differentiate it even in practice (as mentioned above) there is no difference as its a valid option. Your repo your rules, confirm me that I need to remove I will remove it :)

libs/server/Storage/Session/MainStore/MainStoreOps.cs
Vijay-Nirmal: Fixed it, but still having the new private expire method so that I can avoid unnecessary conversion from UnixTime to TimeSpan and then back to DateTimeTicks

Vijay-Nirmal:
@TalZaccai Fixed the review comments except one outstand question in the review comment

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -148,6 +148,36 @@ private bool NetworkEXPIRE<TGarnetApi>(RespCommand command, ref TGarnetApi stora
                 }
             }
 
+            if (parseState.Count > 3)
+            {
+                if (!parseState.TryGetEnum(3, true, out ExpireOption additionExpireOption) || !additionExpireOption.IsValid(ref parseState.GetArgSliceByRef(3)))
```
TalZaccai, line 153: We're eliminating the use of TryGetEnum since it's highly inefficient, please check out this PR - https://github.com/microsoft/garnet/pull/693

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -639,6 +641,7 @@ private RespCommand FastParseCommand(out int count)
                             >= ((6 << 4) | 2) and <= ((6 << 4) | 5) when lastWord == MemoryMarshal.Read<ulong>("BITPOS\r\n"u8) => RespCommand.BITPOS,
                             >= ((7 << 4) | 2) and <= ((7 << 4) | 3) when lastWord == MemoryMarshal.Read<ulong>("EXPIRE\r\n"u8) && ptr[8] == 'P' => RespCommand.PEXPIRE,
                             >= ((8 << 4) | 1) and <= ((8 << 4) | 4) when lastWord == MemoryMarshal.Read<ulong>("TCOUNT\r\n"u8) && *(ushort*)(ptr + 8) == MemoryMarshal.Read<ushort>("BI"u8) => RespCommand.BITCOUNT,
+                            >= ((8 << 4) | 2) and <= ((8 << 4) | 4) when lastWord == MemoryMarshal.Read<ulong>("PIREAT\r\n"u8) && *(ushort*)(ptr + 8) == MemoryMarshal.Read<ushort>("EX"u8) => RespCommand.EXPIREAT,
```
TalZaccai, line 644: Could you move this to FastParseArrayCommand similar to PEXPIREAT?

TalZaccai (changes requested):

libs/server/Resp/KeyAdminCommands.cs
Vijay-Nirmal: Fixed

Vijay-Nirmal:
@TalZaccai Fixed all review comments

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -1278,6 +1284,10 @@ private RespCommand FastParseArrayCommand(ref int count, ref ReadOnlySpan<byte>
                                 {
                                     return RespCommand.RPOPLPUSH;
                                 }
+                                else if (*(ulong*)(ptr + 4) == MemoryMarshal.Read<ulong>("PEXPIREA"u8) && *(uint*)(ptr + 11) == MemoryMarshal.Read<uint>("AT\r\n"u8))
```
TalZaccai, line 1287: nit: you're verifying 'A' twice

TalZaccai (changes requested):
small nit :)

libs/server/Resp/Parser/RespCommand.cs
Vijay-Nirmal: It's the same for other existing 9 char long command as we don't have a datatype with 3 bytes length

TalZaccai: Oops I missed that! Sorry about that :)

TalZaccai (approved):
Great work, thank you!

TedHartMS:
Idk about this merge... Seems we could be letting the devil loose

Vijay-Nirmal:
@TedHartMS

EvilLaughTheDevilGIF

Issue #635, opened 2024-09-04
NullReferenceException in ExecuteScript Method When Executing Lua Scripts


danielchristianschroeter:
### Describe the bug

When attempting to execute Lua scripts using GarnetServer, a NullReferenceException is thrown in the ExecuteScript method. This issue occurs when using the EVAL or EVALSHA commands.

A NullReferenceException is thrown, indicating that an object reference is not set to an instance of an object.

Error Message:
```
2024-09-04 18:28:21 04::28::21 fail: Session[0] [192.168.80.2:6379] [192.168.80.1:42348] [01400C58] Error executing Lua script callback
2024-09-04 18:28:21 04::28::21 crit: Session[0] [192.168.80.2:6379] [192.168.80.1:42348] [01400C58] ProcessMessages threw an exception: System.NullReferenceException: Object reference not set to an instance of an object.    at Garnet.server.RespServerSession.ExecuteScript(Int32 count, LuaRunner scriptRunner) in /src/libs/server/Lua/LuaCommands.cs:line 235    at Garnet.server.RespServerSession.TryEVAL() in /src/libs/server/Lua/LuaCommands.cs:line 97    at Garnet.server.RespServerSession.ProcessArrayCommands[TGarnetApi](RespCommand cmd, TGarnetApi& storageApi) in /src/libs/server/Resp/RespServerSession.cs:line 664    at Garnet.server.RespServerSession.ProcessMessages() in /src/libs/server/Resp/RespServerSession.cs:line 420    at Garnet.server.RespServerSession.TryConsumeMessages(Byte* reqBuffer, Int32 bytesReceived) in /src/libs/server/Resp/RespServerSession.cs:line 299

```

The LuaRunner object might not be properly initialized or retrieved, leading to a null reference when attempting to execute the script.

This report is related to the Lua implementation in https://github.com/microsoft/garnet/pull/550

### Steps to reproduce the bug

1. Enable Lua scripting in GarnetServer.
2. Attempt to execute a Lua script using the EVAL or EVALSHA command.
Go Code Using RedisLock from https://github.com/bsm/redislock/blob/main/redislock.go:
```
func (rs *RedisStorage) Lock(ctx context.Context, name string) error {

	key := rs.prefixLock(name)

	for {
		// try to obtain lock
		lock, err := rs.locker.Obtain(ctx, key, lockTTL, &redislock.Options{})

		// lock successfully obtained
		if err == nil {
			// store the lock in sync.map, needed for unlocking
[...]

PR #672, opened 2024-09-17
Lua - use native tables to improve code compatibility

badrishc:
Use native LuaTable to populate KEYS and ARGV so that most Lua functions will work correctly out of the box.

Fixes https://github.com/microsoft/garnet/issues/635


TalZaccai (approved):

danielchristianschroeter (approved):
LGTM

PR #673, opened 2024-09-17
[Compatibility] Added LPOS command

Vijay-Nirmal:
Adding LPOS command to garnet

- [x] Add LPOS command
- [x] Add Test cases
- [x] Add documentation
- [x] Optimize the array allocation
- [x] Fix ACL test failure


libs/server/Objects/List/ListObjectImpl.cs
```diff
@@ -418,5 +419,211 @@ private void ListSet(ref ObjectInput input, ref SpanByteAndMemory output)
...
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref output_currptr, output_end))
+                        ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref output_startptr, ref ptrHandle, ref output_currptr, ref output_end);
+                    return;
+                }
+
+                count = count == 0 ? list.Count : count;
+                var totalArrayHeaderLen = 0;
+                var lastFoundItemIndex = -1;
+
+                if (count != 1)
```
yrajas, line 470: Looks like there should be a distinction between COUNT argument being present or not. As long as the argument is provided, the return should be an array. Currently, COUNT being passed in as 1 and COUNT not passed in are treated the same, but the return type should be an array when COUNT is passed in even with 1. Same with null reply.

yrajas:
Thanks for your continued contributions!
These PRs are very helpful in bridging the gap in Garnet.

test/Garnet.test/RespListTests.cs
```diff
@@ -1440,5 +1440,122 @@ public void CheckListOperationsOnWrongTypeObjectSE()
...
+        public void LPOSWithoutOptions(string items, string find, int? expectedIndex)
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+            var key = "KeyA";
+            string[] arguments = [key, .. items.Split(",")];
+
+            db.Execute("RPUSH", arguments);
+
+            var actualIndex = (int?)db.Execute("LPOS", key, find);
```
yrajas, line 1460: May be, the ListPosition(s) API could be used.

libs/server/Objects/List/ListObjectImpl.cs
```diff
@@ -418,5 +419,211 @@ private void ListSet(ref ObjectInput input, ref SpanByteAndMemory output)
...
+                else if (noOfFoundItem > 1)
+                {
+                    var newTotalArrayHeaderLen = 0;
+                    var startOutputStartptr = output_startptr;
+                    RespWriteUtils.WriteArrayLength(noOfFoundItem, ref startOutputStartptr, output_end, out var _, out newTotalArrayHeaderLen);  // ReallocateOutput is not needed here as there should be always be available space in the output buffer as we have already written the max array length
+                    Debug.Assert(totalArrayHeaderLen >= newTotalArrayHeaderLen, "newTotalArrayHeaderLen can't be bigger than totalArrayHeaderLen as we have already written max array lenght in the buffer");
+
+                    if (totalArrayHeaderLen != newTotalArrayHeaderLen)
+                    {
+                        Buffer.MemoryCopy(output_startptr + totalArrayHeaderLen, output_startptr + newTotalArrayHeaderLen, output.Length, output.Length);
```
yrajas, line 563: Only remaining length after array size data needs to be copied

yrajas (changes requested):
Added a couple more comments. Thanks again!

libs/server/Objects/List/ListObjectImpl.cs
Vijay-Nirmal: I missed that part in my testing. Fixed in now

test/Garnet.test/RespListTests.cs
Vijay-Nirmal: Personally, I want to test every combination in the raw input like case difference, different option order or without default values but with ListPosition(s) I can't test these. I have added a separate test to test with ListPosition(s) API. Hope that's fine.

libs/server/Objects/List/ListObjectImpl.cs
Vijay-Nirmal: I am not sure I get this. Am I not already doing that using this code `output_startptr + totalArrayHeaderLen`?

yrajas: Shouldn't the length parameters be adjusted to exclude the bytes corresponding to the array length header as only the remaining bytes will have to be moved?

Vijay-Nirmal: Ohh got it my bad, I will fix it.

Vijay-Nirmal: Fixed it

Vijay-Nirmal:
@yrajas Fixed the review comments

yrajas (approved):

PR #674, opened 2024-09-18
[WIP] Add hash expiration - HEXPIRE and family

mmclure-msft:
Redis 7.4 adds the ability to add a per field TTL to hashes through the new HEXPIRE/HPEXPIRE/HEXPIREAT/HPEXPIREAT/HTTL/HPTTL commands. This could heavily reduce the dependency on LUA for users of Redisson MapCache objects since they could use MapCacheNative that use these commands and only a small amount of LUA.

This is a work in progress to try to add this functionality to Garnet.

TODO
* [x] Figure out why the second time we execute the HEXPIRE command it occurs asynchronously and you get the wrong result back causing simple tests to fail.
* [ ] ACL tests
* [ ] Many more tests
* [x] Implement HTTL
* [ ] Actually make the TTL expire the field


mmclure-msft:
Basic code to store a TTL value is in place. Next steps:
* Figure out why the second time we execute the HEXPIRE command it occurs asynchronously and you get the wrong result back causing simple tests to fail.
* Many more tests
* Implement HTTL
* Actually make the TTL expire the field

badrishc:
@mmclure-msft - there are some requests in open source for this. Do you plan to complete it? A lot has changed in command parsing so this might be challenging to rebase to main. If not, perhaps @Vijay-Nirmal, you might be interested in this?

Vijay-Nirmal:
@badrishc A Lot of group work has already been done by @mmclure-msft. I don't like the effort put into this PR going to waste. Let's wait for @mmclure-msft response. In the meantime, I am working on other commands.

badrishc:
@mmclure-msft - let us know what you think, thank you!

Vijay-Nirmal:
@badrishc How do you like to handle field deletion/tombstone after expiration? I have messaged you on Discord

badrishc:
> @badrishc How do you like to handle field deletion/tombstone after the expiration of a field? I am not sure whether can use tombstone logic here. I have messaged you on Discord

I'm not yet sure how exactly to design an efficient HEXPIRE that does not incur space and processing overhead in the common case where there are no expirations needed. This will need a bit of brainstorming.

Hash is a .net object so there is no need to tombstone existing hash entries, just delete them when we encounter them to be expired during a read.

badrishc:
Closing in favor of a new issue https://github.com/microsoft/garnet/issues/857

PR #678, opened 2024-09-19
Implementing COMMAND DOCS

TalZaccai:
- [x] Added support for COMMAND DOCS command
  - All supported command docs are located in Garnet.server/Resp/RespCommandsDocs.json
- [x] Added support for specifying docs path in command registration using the REGISTERCS command
- [x] Added support for specifying command docs when registering command programatically
- [x] Updated CommandInfoUpdater tool to import existing RESP command docs and unify them with Garnet-specific command docs
  - All Garnet-specific command docs are located in CommandInfoUpdater/GarnetCommandsDocs.json
- [x] Updated RespCommandInfo to use "flattened-style" RespCommand enum for sub-commands
- [x] Fixed COMMAND INFO to return sub-command info when called with sub-command name
- [x] Added COMMAND DOCS tests & updated website docs

Examples:
- Autocompletion of Garnet-specific commands:
image
- Autocompletion of custom commands (requires restarting the client for it to refresh the docs):
image


playground/CommandInfoUpdater/GarnetCommandsDocs.json
```diff
@@ -0,0 +1,360 @@
+[
```
badrishc, line 1: If this is not a built-in resource (in order to save space), should we not include it in the NuGet somehow?

TalZaccai: I'm assuming you're referring to RespCommandsDocs.json (in Garnet.server/Resp), which is an embedded resource (similiar to RespCommandsInfo.json), that get lazily parsed upon request. Is that not what we wanted here?

badrishc (approved):
Please check the binary size compared to main, and if it looks reasonable, we can merge.

TalZaccai:
> Please check the binary size compared to main, and if it looks reasonable, we can merge.

Added a tool to compress the json files (pre-build) and reduce the overall size of Garnet.server.dll from 897KB to 650KB (with both resources embedded)

TalZaccai:
> Please check the binary size compared to main, and if it looks reasonable, we can merge.

I've moved the json files to a separate dll that is being loaded (and unloaded) dynamically. The size of Garnet.server.dll is now down to 635KB. Let me know if you like the current solution. The loading & unloading of the assembly would happen *at most* twice (once for command info init & once for command docs init, on an as-needed basis) and is synchronized as to not cause any issues.

badrishc:
> > Please check the binary size compared to main, and if it looks reasonable, we can merge.
> 
> I've moved the json files to a separate dll that is being loaded (and unloaded) dynamically. The size of Garnet.server.dll is now down to 635KB. Let me know if you like the current solution. The loading & unloading of the assembly would happen _at most_ twice (once for command info init & once for command docs init, on an as-needed basis) and is synchronized as to not cause any issues (Look at RespCommandDataCommon.cs for implementation)
> 
> Note: I packed and tested the nuget package locally and it works.

LGTM

badrishc (approved):

PR #679, opened 2024-09-19
Support loading module from command line 

yrajas:
Add support for --loadmodulecs argument to support loading of modules at startup time.


libs/server/Module/ModuleUtils.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            // If necessary, check that all assemblies are digitally signed
+            if (!allowUnsignedAssemblies)
+            {
+                foreach (var loadedAssembly in loadedAssemblies)
+                {
+                    var publicKey = loadedAssembly.GetName().GetPublicKey();
+                    if (publicKey == null || publicKey.Length == 0)
+                    {
+                        errorMessage = CmdStrings.RESP_ERR_GENERIC_ASSEMBLY_NOT_SIGNED;
```
PaulusParssinen, line 59: I wonder if there's any concerns about not unloading the assembly upon failure of signature verification? 🤔

yrajas: Good point.

The locations will have to be allowed by the admin for the assembly to be loaded through the 'LOAD MODULECS' command or the entire assembly specified using the --loadmodulecs argument. So, there is definitely admin control on what gets loaded. 

If the module initializer invocation cannot be precisely controlled, wondering would even explicitly unloading give any better security guarantees.

PaulusParssinen: Unloading wouldn't help in that case, you're correct. We could avoid loading the assembly for the check by using `System.Reflection.Metadata` to inspect the PE file manually:

 ```csharp
using var fs = File.OpenRead(modulePath);
using var peReader = new PEReader(fs);

var metadataReader = peReader.GetMetadataReader();
var assemblyPublicKeyHandle = metadataReader.GetAssemblyDefinition().PublicKey;

// if (assemblyPublicKeyHandle.IsNil) ..

var publicKeyBytes = metadataReader.GetBlobBytes(assemblyPublicKeyHandle);
```

But this also doesn't give any additional security guarantees AFAIK, as [.NET does not validate strong names at runtime](https://learn.microsoft.com/en-us/dotnet/standard/assembly/strong-named) anymore.

yrajas: Agreed. 
Updated to avoid loading assembly if signing check fails as a defensive measure.

badrishc (approved):

PR #685, opened 2024-09-23
Change Reset behavior to be more consistent with Redis

Mathos1432:
1) In Redis RESET doesn't add the other nodes to a ban list.
- Updated to remove the banlist entirely.
2) Close gossip connections when a RESET comes in
- Added a test to verify that the node isn't getting any gossips after a reset.
3) Don't allow RESET when a node has keys in its assigned hashslots.
- Iterate over the main store and the object store until you find a slot that has a key.

Verified by running the updated test before the fixes (failed) and after the fixes (passed)


libs/cluster/Server/ClusterConfig.cs
```diff
@@ -288,6 +288,26 @@ public List<int> GetLocalPrimarySlots()
             return slots;
         }
 
+        /// <summary>
+        /// Retrieve a list of slots served by this node's primary.
+        /// </summary>
+        /// <returns>List of slots.</returns>
+        public List<int> GetLocalSlots()
```
vazois, line 295: I am confused about the purpose of this method. The description seems to indicate that you want to retrieve the slots served by this node's primary. However, you are initializing primaryId from this.LocalNodeId. Shouldn't be this LocalNodePrimaryId?

Mathos1432: My bad, I copied the comment over from the other method, it's meant to retrieve the slots for the local node.
I'll update.

libs/cluster/Server/ClusterManagerWorkerState.cs
```diff
@@ -100,19 +100,30 @@ public ReadOnlySpan<byte> TryReset(bool soft, int expirySeconds = 60)
...
                 var resp = CmdStrings.RESP_OK;
-
                 while (true)
                 {
                     var current = currentConfig;
+                    var localSlots = current.GetLocalSlots();
+                    if (clusterProvider.storeWrapper.HasKeysInSlots(localSlots))
+                    {
+                        resp = CmdStrings.RESP_ERR_RESET_WITH_KEYS_ASSIGNED;
+                        break;
```
vazois, line 110: I would call return here. We don't want to flush the config unnecessarily.

test/Garnet.test.cluster/ClusterManagementTests.cs
```diff
@@ -172,29 +159,171 @@ public void ClusterResetTest()
...
-            // Add node back to the cluster
-            context.clusterTestUtils.SetConfigEpoch(0, 1, context.logger);
-            context.clusterTestUtils.Meet(0, 1, context.logger);
+            context.clusterTestUtils.ClusterReset(0, soft: true);
+            await Task.Delay(1000);
 
-            context.clusterTestUtils.WaitUntilNodeIsKnownByAllNodes(0, context.logger);
+            // Let's compare and make sure there are no gossip messages coming in to the node that got reset.
+            var server = context.clusterTestUtils.GetServer(0);
+            var gossipsAfterReset = GetStat(server, "Stats", gossip);
```
vazois, line 313: To avoid relying on the right timing below, maybe we can add field in gossipStats that returns number of open gossip connections. You can add this in ClusterProvider.cs:GetGossipStats. This way can be explicit that the connections have closed for this test. I am suprised that we didn't have this field before.

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -288,6 +288,26 @@ public List<int> GetLocalPrimarySlots()
...
+        {
+            var primaryId = this.LocalNodeId;
+            List<int> slots = [];
+
+            if (primaryId != null)
+            {
+                for (var i = 0; i < MAX_HASH_SLOT_VALUE; i++)
+                {
+                    if (slotMap[i].workerId > 0 && workers[slotMap[i].workerId].Nodeid.Equals(primaryId, StringComparison.OrdinalIgnoreCase))
```
vazois, line 304: If you want the slots owned by the local node it suffices to do slotMap[i].workerId == 1.  Besides this, are you aware of   config.GetSlotList(uint workerId)? I think it does what you want for workerId = 1. Can you use that method instead of introducing a new one?

vazois (changes requested):
Thanks for your contributions. Please see comments.

libs/cluster/Server/ClusterConfig.cs
Mathos1432: Thanks for pointing out this one, I missed GetSlotList and wasn't aware of that workerId == 1 meaning local node relationship, I'll update.

vazois (approved):

PR #688, opened 2024-09-24
TsavoriteLog iterator varant based on SingleWaiterAutoResetEvent to eliminate allocations

badrishc:
Also, remove persistent iterators. Use SWARE for commit processing worker as well.


libs/storage/Tsavorite/cs/src/core/TsavoriteLog/TsavoriteLog.cs
```diff
@@ -2002,6 +2012,69 @@ public TsavoriteLogScanIterator Scan(long beginAddress, long endAddress, string
             return iter;
         }
 
+        List<TsavoriteLogScanSingleIterator> activeSingleIterators;
+
+        public void RemoveIterator(TsavoriteLogScanSingleIterator iterator)
+        {
+            lock (this)
```
vazois, line 2019: Can we use here the SingleWriterMultiReaderLock

badrishc: The design is for readers to take no locks (not even read locks) as they are perf-sensitive, while writers are heavier in that they replace the list with a new list each time.

PR #695, opened 2024-09-29
[Compatibility] Added GETEX command

Vijay-Nirmal:
Adding GETEX command to garnet

- [x] Add GETEX command
- [x] Add Test cases
- [x] Add documentation


libs/server/Resp/BasicCommands.cs
```diff
@@ -53,6 +53,100 @@ bool NetworkGET<TGarnetApi>(ref TGarnetApi storageApi)
...
+            {
+                return AbortWithWrongNumberOfArguments(nameof(RespCommand.GETEX));
+            }
+
+            var key = parseState.GetArgSliceByRef(0).SpanByte;
+
+            TimeSpan? expiry = null;
+            if (parseState.Count > 1)
+            {
+                var option = parseState.GetArgSliceByRef(1).SpanByte.AsSpan();
```
TalZaccai, line 72: You can just do parseState.GetArgSliceByRef(1).ReadOnlySpan

libs/server/Resp/BasicCommands.cs
```diff
@@ -53,6 +53,100 @@ bool NetworkGET<TGarnetApi>(ref TGarnetApi storageApi)
...
+                if (option.SequenceEqual(CmdStrings.EX))
```
TalZaccai, line 73: Use EqualsUpperCaseSpanIgnoringCase for case insensitivity

libs/server/Resp/BasicCommands.cs
```diff
@@ -53,6 +53,100 @@ bool NetworkGET<TGarnetApi>(ref TGarnetApi storageApi)
...
+                {
+                    if (!parseState.TryGetLong(2, out var expireTime) || expireTime <= 0)
+                    {
```
TalZaccai, line 76: This logic is common for all EX|PX|EXAT|PXAT, can probably just do this if the option is not PERSIST and then do a switch assignment of expiry depending on the option?

libs/server/API/IGarnetApi.cs
```diff
@@ -13,6 +13,15 @@ namespace Garnet.server
     /// </summary>
     public interface IGarnetApi : IGarnetReadApi, IGarnetAdvancedApi
     {
+        #region GETEX
+
+        /// <summary>
+        /// GETEX
+        /// </summary>
+        GarnetStatus GETEX(ref SpanByte key, TimeSpan? expreTime, ref SpanByteAndMemory output);
```
TalZaccai, line 21: typo - expireTime

libs/server/API/GarnetApi.cs
```diff
@@ -75,6 +75,14 @@ public GarnetStatus GET(byte[] key, out GarnetObjectStoreOutput value)
             => storageSession.GET(key, out value, ref objectContext);
         #endregion
 
+        #region GETEX
+
+        /// <inheritdoc />
+        public GarnetStatus GETEX(ref SpanByte key, TimeSpan? expreTime, ref SpanByteAndMemory output)
```
TalZaccai, line 81: typo - expireTime

TalZaccai (changes requested):
Added some comments, will finish the review later today :)

Vijay-Nirmal:
Fixed the current review comments

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -413,6 +415,28 @@ private bool InPlaceUpdaterWorker(ref SpanByte key, ref SpanByte input, ref Span
                     rmwInfo.Action = RMWAction.ExpireAndStop;
                     return false;
 
+                case RespCommand.GETEX:
+                    CopyRespTo(ref value, ref output);
+
+                    if (input.ExtraMetadata > 0)
+                    {
+                        byte* pbOutput = stackalloc byte[ObjectOutputHeader.Size];
```
TalZaccai, line 423: What's this for?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -413,6 +415,28 @@ private bool InPlaceUpdaterWorker(ref SpanByte key, ref SpanByte input, ref Span
                     rmwInfo.Action = RMWAction.ExpireAndStop;
                     return false;
 
+                case RespCommand.GETEX:
+                    CopyRespTo(ref value, ref output);
+
+                    if (input.ExtraMetadata > 0)
```
TalZaccai, line 421: Shouldn't you check for !isPersist here?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -671,6 +695,27 @@ public bool CopyUpdater(ref SpanByte key, ref SpanByte input, ref SpanByte oldVa
...
+                    {
+                        Debug.Assert(newValue.Length == oldValue.Length + input.MetadataSize);
+                        byte* pbOutput = stackalloc byte[ObjectOutputHeader.Size];
+                        var _ = new SpanByteAndMemory(SpanByte.FromPinnedPointer(pbOutput, ObjectOutputHeader.Size));
+                        EvaluateExpireCopyUpdate(ExpireOption.None, expiryExists: oldValue.MetadataSize > 0, ref input, ref oldValue, ref newValue, ref _);
+                    }
+
+                    oldValue.AsReadOnlySpan().CopyTo(newValue.AsSpan());
+                    var isPersist = *(bool*)(inputPtr + RespInputHeader.Size);
```
TalZaccai, line 710: Is this necessary? If we're only doing GET + PERSIST, would we ever reach CopyUpdate?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -671,6 +695,27 @@ public bool CopyUpdater(ref SpanByte key, ref SpanByte input, ref SpanByte oldVa
                     rmwInfo.Action = RMWAction.ExpireAndStop;
                     return false;
 
+                case RespCommand.GETEX:
+                    CopyRespTo(ref oldValue, ref output);
+
+                    if (input.ExtraMetadata > 0)
+                    {
+                        Debug.Assert(newValue.Length == oldValue.Length + input.MetadataSize);
+                        byte* pbOutput = stackalloc byte[ObjectOutputHeader.Size];
```
TalZaccai, line 704: Again, what is this for?

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
```diff
@@ -185,6 +185,13 @@ public int GetRMWModifiedValueLength(ref SpanByte t, ref SpanByte input)
                         // No additional allocation needed.
                         break;
 
+                    case RespCommand.GETEX:
+                        if (input.ExtraMetadata > 0)
```
TalZaccai, line 189: Assuming this should always be true if we reach here? Could just do a Debug.Assert(input.ExtraMetadata > 0)

test/Garnet.test/RespTests.cs
```diff
@@ -3539,5 +3539,192 @@ private static void AssertExpectedClientFields(string list)
             static void AssertField(string line, string[] fields, string name)
             => ClassicAssert.AreEqual(1, fields.Count(f => f.StartsWith($"{name}=")), $"In {line}, expected single field {name}");
         }
+
+        #region GETEX
+
+        [Test]
+        public void GetEXBasicTestWithSERedisApi()
```
TalZaccai, line 3546: Please rename tests to GetExpiry... just for consistency please :)

TalZaccai (changes requested):
Added a few more comments...

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: It's to call existing EvaluateExpireInPlace method but I don't care about the output param, so created a temp SpanByteAndMemory without any allocation and called the existing method

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Same as above

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Ideally not, add it just in case, if you are sure then i can remove it. Shall I?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: I don't have to because if there is ExtraMetadata then that means the user has provided the expiration time so it can't be a Persist options, so there is no point in checking for `!isPersist`

test/Garnet.test/RespTests.cs
Vijay-Nirmal: Fixed

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
Vijay-Nirmal: Done

Vijay-Nirmal:
@TalZaccai Fixed the review comments, there is one open question there.

vazois (changes requested):
Can you please add a slot verification test in ClusterSlotVeficationTests?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
TalZaccai: But you're not calling EvaluateExpireInPlace with this variable...

libs/server/Storage/Functions/MainStore/RMWMethods.cs
TalZaccai: Yeah I think this should be removed...

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: I am using the below code
```
var _ = new SpanByteAndMemory(SpanByte.FromPinnedPointer(**pbOutput**, ObjectOutputHeader.Size));
```

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Ok, I will remove it, Do you know why the existing `PERSIST` comment also has an entry in CopyUpdater?

Vijay-Nirmal: Fixed it

Vijay-Nirmal:
@vazois Added ClusterSlotVeficationTests

@TalZaccai Fixed review comments

libs/server/Storage/Functions/MainStore/RMWMethods.cs
TalZaccai: Now I'm doubting myself too :) Maybe if it needs to move somewhere else in memory... Let me double check with someone in the team.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
TalZaccai: Mmm ok I wasn't familiar with the syntax...

TalZaccai: Why not something like:
```cs
byte* pbOutput = stackalloc byte[ObjectOutputHeader.Size];
var _output = new SpanByteAndMemory(pbOutput, ObjectOutputHeader.Size);
EvaluateExpireCopyUpdate(ExpireOption.None, expiryExists: oldValue.MetadataSize > 0, ref input, ref oldValue, ref newValue, ref _output);
```

It'll make what you're doing here clearer and more in-line with our code.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Fixed

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: I added it back, but do you think we need to add the if condition back in the `GetRMWModifiedValueLength` from the other review comment?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Fixed

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
Vijay-Nirmal: I am marking it as unresolved as I have an open question. Do we need this `if` check since we have CopyUpdater now?

Vijay-Nirmal:
@TalZaccai Fixed the review comments but I have a question there.

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
Vijay-Nirmal: Hmm interesting, the current `PERSIST` doesn't do that. Anyway, I don't want to get confused even more. I will do it as you said. just return `sizeof(int) + t.Length` when `input.ExtraMetadata == 0`

libs/server/Storage/Functions/MainStore/RMWMethods.cs
badrishc: If the source of the RMW is on disk or in the immutable region of memory, we have to issue a CU (CopyUpdater) even though the size does not increase. So yes, CopyUpdate would be needed.

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
badrishc: If we (1) remove the assert; and (2) simply return `sizeof(int) + t.LengthWithoutMetadata + input.MetadataSize` -- that should cover all cases, right?

Vijay-Nirmal: @badrishc  Wow, yes. Never even thought of it. Will push the change soon, in middle of fixing a bug with GeoSearch.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: @badrishc Thanks for the explanation, it makes sense. The developer doc should be updated with this info. Now it just says, "you will need to implement the initialization and in-place / copy update functionality of the new command"

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
Vijay-Nirmal: @badrishc Can you explain to me why the current `PERSIST` doesn't return any length? Shouldn't it return `sizeof(int) + t.LengthWithoutMetadata`?

badrishc: You are right - this looks like a bug in persist that should be fixed. Good catch.

Vijay-Nirmal: Fixed

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Fixed

badrishc (approved):

PR #631, opened 2024-09-03
Improvements in command handling #5

TalZaccai:
- [x]  Introducing RawStringInput struct to simplify main store command processing
  - [x] BasicCommands
  - [x] ArrayCommands
  - [x] MainStoreOps
  - [x] AdvancedOps
  - [x] BitmapCommands
  - [x] BitmapOps
  - [x] HyperLogLogCommands
  - [x] HyperLogLogOps
  - [x] KeyAdminCommands
  - [x] CustomRawStringFunctions
- [x] Supporting RawStringInput serialization/deserialization for AOF recovery
- [x] Relaying parse state to custom commands / transactions
- [x] Fixing bug in PFCOUNT logic

**Benchmarking results:**
**GeoHashBenchmarks**

main:
| Method                 | Mean      | Error     | StdDev    | Gen0   | Allocated |
|----------------------- |----------:|----------:|----------:|-------:|----------:|
| GeoToLongValue         |  3.233 ns | 0.0589 ns | 0.0551 ns |      - |- |
| GetCoordinatesFromLong |  2.181 ns | 0.0829 ns | 0.0775 ns |      - |- |
| GetGeoHashCode         | 30.134 ns | 0.5252 ns | 0.4913 ns | 0.0001 |      48 B |

#631:
| Method                 | Mean      | Error     | StdDev    | Gen0   | Allocated |
|----------------------- |----------:|----------:|----------:|-------:|----------:|
| GeoToLongValue         |  3.532 ns | 0.1080 ns | 0.1649 ns |      - |- |
| GetCoordinatesFromLong |  2.135 ns | 0.0427 ns | 0.0378 ns |      - |- |
| GetGeoHashCode         | 30.555 ns | 0.6885 ns | 1.0305 ns | 0.0001 |      48 B |

**RecoveryBenchmark**

main:
| Method  | MemorySize | LogDir | Mean     | Error   | StdDev   | P90      | P95      |
|-------- |----------- |------- |---------:|--------:|---------:|---------:|---------:|
| Recover | 100m       | -f     | 131.7 us | 5.50 us | 16.04 us | 153.0 us | 162.9 us |
| Recover | 100m       | *      |       NA |      NA |       NA |       NA |   NA |

#631:
| Method  | MemorySize | LogDir | Mean     | Error   | StdDev  | P90      | P95     |
|-------- |----------- |------- |---------:|--------:|--------:|---------:|---------:|
[...]

PR #474, opened 2024-06-18
Allow Write Operations when Slot is in MIGRATING state

vazois:
This PR resolves issue #354.

Tasks:
- [x] Consolidate slot verification methods.
- [x] Introduce BDN benchmark for cluster.
- [x] Add unit tests to validate cluster redirect functionality.
- [x] Implement key level access control.
- [x] Benchmark standalone cluster and main
 
# Overview of Migration

The process of slot migration involves several stages which are designed to manage data access, ensuring both data integrity and high availability as the corresponding keys are being transferred from the source node to the target node.
The MIGRATE command supports two transfer options, specifically MIGRATE KEYS and MIGRATE SLOTS.
Both options use a common data access control interface, that is logically divided into the following categories:

1.  Slot level access control
      Used to orchestrate migration by changing the state of the associated slot accordingly (i.e. MIGRATING, IMPORTING) at the corresponding source and target nodes
2.  Key level access control
      Used to control access to individual keys in order to ensure high availability and data integrity.

## Key level access control
Each migrate session maintains a dictionary of <keys, KeyMigrationStatus> pairs.
This dictionary is used to control access to keys that are actively being managed by a single running migrate session.
When a slot is in the process of migration and there are no active migration sessions managing a specific key, any requests for that key are handled under the assumption that the key exists. If it doesn’t, a redirect -ASK is created, which points to the endpoint of the target node.

The KeyMigrationStatus will affect individual session readers and writers as follows:

- **QUEUED (QD)**
     Key owned by a specific MigrateSession but is not actively being migrated.
     Reads and writes can be served if the keys exist.
- **MIGRATING (MG)**
      Key is actively being migrated by a specific MigrateSession.
      Writes will be delayed until status goes back to QUEUED or MIGRATED.
      Reads can be served without any restriction.
- **DELETING (DL)**
      Key is being deleted after it was sent to the target node.
      Reads and writes will be delayed.
      We need to delay reads to avoid the scenario where a key exists during validation but was deleted before read executes.
- **MIGRATED (MD)**
      Key owned by a specific MigrateSession and has completed all the steps to be MIGRATED to target node.
      This does not mean that key existed or has not expired, just that all the steps associated with MIGRATED have completed.
      This can happen for a key that was provided as an argument in MIGRATE command but did not exist or expired for both main and object stores.

### Key Transfer State Machine Algorithm

1. Add keys to dictionary and initialize KeyMigrationStatus to QUEUED
2. Perform for main and object store separately
[...]

PR #699, opened 2024-10-02
[Compatibility] Added INCRBYFLOAT command

Vijay-Nirmal:
Adding INCRBYFLOAT command to garnet

- [x] Add INCRBYFLOAT command
- [x] Add Test cases
- [x] Add documentation

Question:

Do you guys have any optimal way to `double` to ASCII` byte`?

Below is a benchmark with 2 approaches.

**DoubleToSpanWithMathAbsRound** => Same code as the PR
**DoubleToSpanWithDiyFp**  => Similar code as `string.Format("{0:0.####}")`. Below are the links to the code I used from .dotnet runtime repo
[Number.DiyFp.cs](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/Number.DiyFp.cs)
[Number.Grisu3.cs - TryRunCounted](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/Number.Grisu3.cs#L429)
**DoubleToSpanWithStringFormat** => Using `doubleValue.ToString("0.###############")` and using buffer copy to move the value to input span
**DoubleToSpanWithUtf8Formatter** => Using Utf8Formatter
**DoubleToSpanWithTryFormat** => Using `doubleValue.TryFormat(buffer, out var bytesWritten, "0.#################")`

| Method                                                 | Value                | Mean      | Error    | StdDev   | Gen0   | Allocated |
|------------------------------------------------------- |--------------------- |----------:|---------:|---------:|-------:|----------:|
| DoubleToSpanWithMathAbsRound                      | 10.23                |  32.27 ns | 0.399 ns | 0.333 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 10.23                |  51.40 ns | 0.619 ns | 0.579 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 10.23                | 121.94 ns | 1.474 ns | 1.307 ns | 0.0017 |      32 B |
| DoubleToSpanWithUtf8Formatter                     | 10.23                | 142.52 ns | 2.637 ns | 2.590 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 10.23                | 122.79 ns | 2.079 ns | 1.843 ns |      - |         - |
| DoubleToSpanWithMathAbsRound                      | 12.023642000155785   | 172.41 ns | 3.402 ns | 4.303 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 12.023642000155785   |  51.86 ns | 0.928 ns | 0.868 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 12.023642000155785   | 111.17 ns | 2.158 ns | 2.309 ns | 0.0029 |      56 B |
| DoubleToSpanWithUtf8Formatter                     | 12.023642000155785   | 188.86 ns | 3.793 ns | 3.895 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 12.023642000155785   | 121.57 ns | 1.504 ns | 2.825 ns |      - |         - |
| DoubleToSpanWithMathAbsRound                      | 85.98498198191919    | 161.47 ns | 1.592 ns | 1.330 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 85.98498198191919    |  50.77 ns | 0.834 ns | 0.740 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 85.98498198191919    | 107.39 ns | 1.761 ns | 1.647 ns | 0.0029 |      56 B |
| DoubleToSpanWithUtf8Formatter                     | 85.98498198191919    | 204.33 ns | 1.950 ns | 1.728 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 85.98498198191919    | 111.06 ns | 1.260 ns | 1.052 ns |      - |         - |
| DoubleToSpanWithMathAbsRound                      | 152.3645             |  59.95 ns | 0.985 ns | 0.922 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 152.3645             |  58.23 ns | 1.142 ns | 1.269 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 152.3645             | 128.90 ns | 2.206 ns | 1.956 ns | 0.0019 |      40 B |
| DoubleToSpanWithUtf8Formatter                     | 152.3645             | 220.70 ns | 4.394 ns | 5.557 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 152.3645             | 136.13 ns | 2.682 ns | 3.193 ns |      - |         - |
| DoubleToSpanWithMathAbsRound                      | 456621336.22         |  82.08 ns | 1.669 ns | 1.855 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 456621336.22         |  44.48 ns | 0.425 ns | 0.355 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 456621336.22         | 119.48 ns | 1.382 ns | 1.293 ns | 0.0024 |      48 B |
| DoubleToSpanWithUtf8Formatter                     | 456621336.22         | 248.25 ns | 4.419 ns | 4.134 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 456621336.22         | 114.32 ns | 2.104 ns | 1.968 ns |      - |         - |
| DoubleToSpanWithMathAbsRound                      | 9.874(...)2E+20 [21] | 249.52 ns | 4.870 ns | 5.980 ns |      - |         - |
| DoubleToSpanWithDiyFp                             | 9.874(...)2E+20 [21] |  37.10 ns | 0.761 ns | 0.712 ns |      - |         - |
| DoubleToSpanWithStringFormat                      | 9.874(...)2E+20 [21] | 126.49 ns | 1.706 ns | 1.513 ns | 0.0033 |      64 B |
| DoubleToSpanWithUtf8Formatter                     | 9.874(...)2E+20 [21] | 322.90 ns | 2.100 ns | 1.753 ns |      - |         - |
| DoubleToSpanWithTryFormat                         | 9.874(...)2E+20 [21] | 138.64 ns | 2.705 ns | 2.778 ns |      - |         - |

Even though the dotnet runtime approach does better in most cases in the benchmark, I have not done it as part of this PR because of 2 reasons, 1) it has lots of code and 2) for smaller numbers it always performs worse. If you guys like dotnet's approach, I can change it quickly.

Let me know if you guys have any other ideas, I can implement and compare the performance.


badrishc:
Why not use Utf8Parser.TryParse? The open PR https://github.com/microsoft/garnet/pull/631 also uses it.
 https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Utf8Parser/Utf8Parser.Float.cs#L63

badrishc:
For the other direction, `double.TryParse(ReadOnlySpan<byte> utf8Text, out double result)`

badrishc:
For other direction, check this:

https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.utf8formatter.tryformat?view=net-8.0

badrishc:
might be best to just convert the double to the ROS<byte> with a sufficiently large buffer, to know how many bytes it uses. hopefully the conversion will not need to be repeated.

Vijay-Nirmal:
@badrishc I have added the benchmark results with Utf8Formatter. In my benchmark, it is the slowest option. I've shared below my benchmark method for using Utf8Formatter, please let me know if I made any mistake in my benchmark. Note: We can't use exponential notation.

```c#
[Benchmark]
public int NumOfCharInDoubleWithUtf8Formatter()
{
    Span<byte> buffer = stackalloc byte[327];  // 309 (max digits) + 1 (dot) + 17 (precision)
    Utf8Formatter.TryFormat(Value, buffer, out var bytesWritten, new StandardFormat('f', 17));
    buffer = buffer.Slice(0, bytesWritten).TrimEnd((byte)'0').TrimEnd((byte)'.');
    return buffer[0];
}
```

badrishc:
NumOfCharInDoubleWithMathAbsRound seems better as high-perf scenarios might involve smaller doubles. cc @PaulusParssinen for thoughts on this.

libs/common/NumUtils.cs
```diff
@@ -75,6 +76,80 @@ public static unsafe void LongToBytes(long value, int length, ref byte* result)
...
+            }
+
+            return totalLen;
+        }
+
+        /// <summary>
+        /// Convert double into sequence of ASCII bytes
+        /// </summary>
+        /// <param name="value">Double value to convert</param>
+        /// <param name="length">Number of digits in value</param>
```
badrishc, line 104: Fix the warnings shown at https://github.com/microsoft/garnet/pull/699/files

libs/common/NumUtils.cs
```diff
@@ -370,6 +484,38 @@ public static int NumDigitsInLong(long v, ref bool fNeg)
...
+                signSize = 0;
+                fractionalDigits = 0;
+                return 1;
+            }
+
+            Debug.Assert(!double.IsNaN(v) && !double.IsInfinity(v));
+
+            signSize = (byte)(v < 0 ? 1 : 0); // Add sign if the number is negative
+            v = Math.Abs(v);
+            integerDigits = (int)Math.Log10(v) + 1;
```
badrishc, line 506: Could you use our `NumDigitsInLong(long)` for this part? And Math.Floor for the long part?

Vijay-Nirmal: We can't, initially I did that but encountered issue in number bigger than log because double.MaxValue is very big than long.MaxValue.

double.MaxValue 1.7976931348623157E+308
long.MaxValue 9223372036854775807

libs/common/NumUtils.cs
Vijay-Nirmal: Fixed

badrishc (approved):

badrishc:
Nicely done PR, approved!

badrishc:
Please resolve the conflicts and we can move forward.

badrishc:
@Vijay-Nirmal - there seems to be a perf regression with this PR, for Increment:

BDN paramters: `-f BDN.benchmark.Resp.RespParseStress.Increment`

main:

| Method    | Mean     | Error    | StdDev   | Allocated |
|---------- |---------:|---------:|---------:|----------:|
| Increment | 34.80 us | 0.354 us | 0.314 us |         - |

PR:

| Method    | Mean     | Error    | StdDev   | Allocated |
|---------- |---------:|---------:|---------:|----------:|
| Increment | 44.36 us | 0.842 us | 0.703 us |         - |

### Analysis

It seems this is related to the `stackalloc` of the larger sized buffer for doubles in the same method (NetworkIncrement). My suggestion is that we separate out NetworkIncrementByFloat into its own method.

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -97,6 +97,7 @@ public enum RespCommand : byte
         HSETNX,
         INCR,
         INCRBY,
+        INCRBYFLOAT,
```
badrishc, line 100: Please add a BDN for `IncrementByFloat` in `benchmark\BDN.benchmark\Resp\RespParseStress.cs` - following the same pattern as `Increment()` in that `RespParseStress` class.

vazois (changes requested):
Can you please add a slot verification test in ClusterSlotVeficationTests?

Vijay-Nirmal:
@badrishc Interesting, For `Increment`, `isFloat` will always be false. Then how come the `stackalloc` is impacting?

@badrishc Is there any thumb rule on when to add RespParseStress?

@vazois That a new things to me, I will add it, also same question Is there any thumb rule on when to add ClusterSlotVeficationTests?

I will do everything this weekend

badrishc:
> @badrishc Interesting, For `Increment`, `isFloat` will always be false. Then how come the `stackalloc` is impacting?

I tried making `MaximumFormatDoubleLength` equal to `MaximumFormatInt64Length` and that seems to bring the perf back to closer to original levels, hence the suspicion.

> @badrishc Is there any thumb rule on when to add RespParseStress?

Ideally for all "fast" commands we add so that we can track future perf regressions. However it is understandably extra work, so it is mostly on a case-by-case basis for now.

vazois:
> @badrishc Interesting, For `Increment`, `isFloat` will always be false. Then how come the `stackalloc` is impacting?
> 
> @badrishc Is there any thumb rule on when to add RespParseStress?
> 
> @vazois That a new things to me, I will add it, also same question Is there any thumb rule on when to add ClusterSlotVeficationTests?
> 
> I will do everything this weekend

I made some updates and additions in this PR (https://github.com/microsoft/garnet/pull/709) for slot verification.
If you are adding a multi key command you can check how MSET and MGET is being tested. Single key commands should be easier to understand by check SET and GET redirection tests.
You can also check the description of this PR (https://github.com/microsoft/garnet/pull/474) if you want to learn more about the redirection logic.

Vijay-Nirmal:
@badrishc Moved `INCRBYFLOAT` to a separate `NetworkIncrementByFloat`

@vazois #474 Has very good info, it will be good if we add the same info in Dev doc for slot migration.

libs/server/Resp/Parser/RespCommand.cs
badrishc: RespIncrementStress is less relevant, we are not really using that at the moment.

badrishc: In RespParseStress:
* Create a separate command type for IncrementByFloat
* Add the setup logic to create the original value of key such as f => 1.1
* The stress will issue a command such as `INCRBYFLOAT f 1.1` or similar

badrishc: Actually, do not worry about adding the BDN for this. There are some subtle issues with it that will be resolved by an open PR and then we can add it.

Vijay-Nirmal:
@vazois Added ClusterSlotVeficationTests. Can you review and approve the PR?

badrishc:
> Can you please add a slot verification test in ClusterSlotVeficationTests?

We will merge this PR now. Vijay - if you can add the ClusterSlotVerificationTest in a separate PR, that would be great.

Vijay-Nirmal:
@badrishc Already added ClusterSlotVerificationTest in this PR itself

vazois (approved):

website/docs/commands/api-compatibility.md
```diff
@@ -387,7 +387,7 @@ Note that this list is subject to change as we continue to expand our API comman
 |  | GETSET | ➖ |  |
 |  | [INCR](raw-string.md#incr) | ➕ |  |
 |  | [INCRBY](raw-string.md#incrby) | ➕ |  |
-|  | INCRBYFLOAT | ➖ |  |
+|  | [INCRBYFLOAT](raw-string.md#incrbyfloat) | ➖ |  |
```
nightroman, line 390: @Vijay-Nirmal Presumably "-" meant to be changed to "+".

Currently `INCRBYFLOAT` may look like not implemented with the minus sign
https://microsoft.github.io/garnet/docs/commands/api-compatibility

nightroman: This PR should fix this, if accepted https://github.com/microsoft/garnet/pull/757

PR #700, opened 2024-10-03
API to clear scratch buffer

yrajas:
When repeated large values are retrieved from custom commands, it is possible that the scratch buffer could overflow.
In such scenarios, the GetScratchBufferOffset and ResetScratchBuffer APIs could be invoked to free up the space occupied by the last retrieved data.

### DIFF
--- a/libs/server/ArgSlice/ScratchBufferManager.cs
+++ b/libs/server/ArgSlice/ScratchBufferManager.cs
@@ -29,7 +29,7 @@ namespace Garnet.server
         /// <summary>
         /// Current offset in scratch buffer
         /// </summary>
-        int scratchBufferOffset;
+        internal int scratchBufferOffset { get; private set; }
 
         public ScratchBufferManager()
         {
@@ -53,6 +53,20 @@ namespace Garnet.server
                 return true;
             }
             return false;
+        }
+
+        /// <summary>
+        /// Resets scratch buffer offset to the specified offset.
+        /// </summary>
+        /// <param name="offset">Offset to reset to</param>
+        /// <returns>True if successful, else false</returns>
+        public bool ResetScratchBuffer(int offset)
+        {
+            if (offset < 0 || offset > scratchBufferOffset)
+                return false;
+
+            scratchBufferOffset = offset;
+            return true;
         }
 
         /// <summary>
--- a/libs/server/ArgSlice/ScratchBufferManager.cs
+++ b/libs/server/ArgSlice/ScratchBufferManager.cs
@@ -29,7 +29,10 @@ namespace Garnet.server
         /// <summary>
         /// Current offset in scratch buffer
         /// </summary>
-        internal int scratchBufferOffset { get; private set; }
+        int scratchBufferOffset;
+
+        /// <summary>Current offset in scratch buffer</summary>
+        internal int ScratchBufferOffset => scratchBufferOffset;
 
         public ScratchBufferManager()
         {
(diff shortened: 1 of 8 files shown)


test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -10,13 +10,62 @@
...
+            for (int i = 0; i < 120_000; i++)
+            {
+                garnetApi.GET(key, out var outval);
+                garnetApi.FreeBuffer(ref outval);
+            }
+
+            garnetApi.GET(key, out var outval1);
+            garnetApi.GET(key, out var outval2);
+            garnetApi.FreeBuffer(ref outval2);
+            garnetApi.FreeBuffer(ref outval1); // Ensure graceful handling of earlier allocated buffer
```
badrishc, line 37: test what happens if they first free 1 and then 2

test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -10,13 +10,62 @@
...
+    public class LargeGet : CustomProcedure
+    {
+        public override bool Execute(IGarnetApi garnetApi, ArgSlice input, ref MemoryResult<byte> output)
+        {
+            int offset = 0;
+            var key = GetNextArg(input, ref offset);
+            for (int i = 0; i < 120_000; i++)
+            {
+                garnetApi.GET(key, out var outval);
+                garnetApi.FreeBuffer(ref outval);
```
badrishc, line 31: seems wasteful to need to call for every single item. can they directly get the old status and rewind to that?

test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -10,13 +10,62 @@
...
 {
+    public class LargeGet : CustomProcedure
+    {
+        public override bool Execute(IGarnetApi garnetApi, ArgSlice input, ref MemoryResult<byte> output)
+        {
+            int offset = 0;
+            var key = GetNextArg(input, ref offset);
+            for (int i = 0; i < 120_000; i++)
+            {
+                garnetApi.GET(key, out var outval);
```
badrishc, line 30: Will all the returned ArgSlice still be valid after the 120k iterations? seems like if the scratch buffer doubled, the existing ArgSlices are invalidated?

test/Garnet.test/RespCustomCommandTests.cs
yrajas: That is getting handled gracefully as the offsets and length check within the scratchbuffer manager will fail. Added a test to verify that.

test/Garnet.test/RespCustomCommandTests.cs
yrajas: Added APIs to retrieve offsets and reset to any specific offset in the past. 
The test will now avoid invoking this every time and do it only periodically to avoid expanding the buffer too much.

test/Garnet.test/RespCustomCommandTests.cs
yrajas: The new APIs to record offset and reset the buffer back to any specific offset will help with this.

libs/server/ArgSlice/ScratchBufferManager.cs
```diff
@@ -29,7 +29,7 @@ internal sealed unsafe class ScratchBufferManager
         /// <summary>
         /// Current offset in scratch buffer
         /// </summary>
-        int scratchBufferOffset;
+        internal int scratchBufferOffset { get; private set; }
```
badrishc, line 32: keep this a private int field instead of a property as it is perf critical. can add a separate getter.

badrishc (approved):

PR #701, opened 2024-10-03
[Compatibility] Added NOVALUE option for HSCAN

Vijay-Nirmal:
- [x] Add NOVALUE option for HSCAN
- [x] Updated Test cases
- [x] Updated documentation
- [x] Updated StackExchange.Redis to `2.8.16` to get the new `HashScanNoValues` method in the test project


libs/server/Objects/Hash/HashObject.cs
```diff
@@ -228,7 +228,10 @@ public override unsafe void Scan(long start, out List<byte[]> items, out long cu
                 if (patternLength == 0)
                 {
                     items.Add(item.Key);
-                    items.Add(item.Value);
+                    if (!isNoValue)
```
yrajas, line 231: The count value will have to be adjusted as well if NOVALUES present as it is currently always multiplying it by two to account for key and value.

yrajas (changes requested):
Count to be adjusted for hscan.

libs/server/Objects/Hash/HashObject.cs
Vijay-Nirmal: Fixed

Vijay-Nirmal:
@yrajas Fixed the review command and added one more case to test this case

Vijay-Nirmal:
@yrajas Fixed failed test case

PR #709, opened 2024-10-09
Cleanup Slot Verification

vazois:
- [x] Refactor object commands to use cluster slot verification
- [x] Add comments to slot verification testing
- [x] Fix calculation of keynum for commands containing number of keys as part of the parameters


libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -65,9 +55,10 @@ bool CanServeSlot(RespCommand cmd)
                             csvi.step = findRange.KeyStep;
                             break;
                         case FindKeysKeyNum:
+                            var beginSearch = (BeginSearchIndex)specs[0].BeginSearch;
                             var findKeysKeyNum = (FindKeysKeyNum)specs[0].FindKeys;
                             csvi.firstKey = searchIndex.Index + findKeysKeyNum.FirstKey - 1;
-                            csvi.lastKey = searchIndex.Index + parseState.GetInt(0);
+                            csvi.lastKey = searchIndex.Index + parseState.GetInt(beginSearch.Index + findKeysKeyNum.KeyNumIdx - 1);
```
badrishc, line 61: this is wrong, need to merge from the recently merged PR.

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -65,9 +55,10 @@ bool CanServeSlot(RespCommand cmd)
                             csvi.step = findRange.KeyStep;
                             break;
                         case FindKeysKeyNum:
+                            var beginSearch = (BeginSearchIndex)specs[0].BeginSearch;
```
badrishc, line 58: Why is there a beginSearch and a searchIndex above?

libs/server/Resp/RespServerSessionSlotVerify.cs
badrishc: This line should be: `csvi.lastKey = csvi.firstKey + parseState.GetInt(searchIndex.Index + findKeysKeyNum.KeyNumIdx - 1);`

vazois: Fixed the issue of wrong merge

libs/server/Resp/RespServerSessionSlotVerify.cs
vazois: Removed this because it is not needed. Forgot that searchIndex was available.

badrishc (approved):

PR #712, opened 2024-10-09
Fix Slot Validation for Custom Proc

vazois:
This PR fixes slot verification for custom proc.
It introduces the following:

- [x] Iterative slot verification interface.
- [x] Slot verification fix for WATCH command.
- [x] Separation of WATCH_MS and WATCH_OS as separate commands instead of WATCH subcommands
- [x] Added BDN for custom proc

RespClusterBench
| Method | Mean      | Error    | StdDev   | Allocated |
|------- |----------:|---------:|---------:|----------:|
| Get    |  22.13 us | 0.028 us | 0.026 us |         - |
| Set    |  21.47 us | 0.041 us | 0.039 us |         - |
| MGet   |  18.13 us | 0.024 us | 0.022 us |         - |
| MSet   |  15.55 us | 0.032 us | 0.030 us |         - |
| CPBSET | 178.25 us | 0.398 us | 0.372 us |         - |

RespClusterDisableSlotVerificationBench
| Method | Mean      | Error    | StdDev   | Allocated |
|------- |----------:|---------:|---------:|----------:|
| Get    |  18.58 us | 0.027 us | 0.025 us |         - |
| Set    |  17.19 us | 0.026 us | 0.023 us |         - |
| MGet   |  17.77 us | 0.042 us | 0.039 us |         - |
| MSet   |  15.67 us | 0.034 us | 0.032 us |         - |
| CPBSET | 152.47 us | 0.159 us | 0.149 us |         - |


libs/server/Transaction/TransactionManager.cs
```diff
@@ -172,6 +175,13 @@ internal bool RunTransactionProc(byte id, ArgSlice input, CustomTransactionProce
                     return false;
                 }
 
+                if (state == TxnState.Aborted)
```
Xizt, line 178: I'm wondering if there should another state (CROSSLOTABORT?) when slot verification fails.

vazois: Is it not difficult to add but how is this useful? When txn is aborted the client will receive a message that validation failed due slot verification issue.

badrishc (approved):

libs/server/Transaction/TransactionManager.cs
Xizt: Currently, transaction fail possible only due to slot verification issue.  Can there be more reasons later? If yes, it would keep showing slot verification issue hence thought of another state.

badrishc: We return a state of aborted and the error message to the caller, so the caller can handle any type of abort.

PR #727, opened 2024-10-16
[Bug] Get modified length method returns wrong length for PERSIST commands

Vijay-Nirmal:
Modified the `GetRMWModifiedValueLength` method to return the wrong length for `PERSIST`.

PR created for https://github.com/microsoft/garnet/pull/695#discussion_r1802116909


Vijay-Nirmal:
@badrishc How do I test this? Current test case for PERSIST doesn't call `GetRMWModifiedValueLength`

badrishc:
I have added a new type of test for this, including the fix. Please check it out. closing this PR as i had to add the fix there to get it to pass.

PR #730, opened 2024-10-16
Fix Persist RMW size logic, add fine grained address testing

badrishc:
Adds a new type of test, as discussed in https://github.com/microsoft/garnet/pull/727


Vijay-Nirmal (approved):
Tested in local, looks good to me

PR #732, opened 2024-10-17
[Compatibility] Added ZDIFFSTORE command

Vijay-Nirmal:
Adding the ZDIFFSTORE command to garnet

- [x] Add ZDIFFSTORE command
- [x] Add documentation


vazois:
> @vazois Can you help fix ClusterSlotVerificationTests issue? I am getting `ERR Garnet Exception: FindKeysUnknown range`

Check PR #734 and let me know if that fixes your issue

libs/server/API/GarnetApiObjectCommands.cs
```diff
@@ -129,6 +129,10 @@ public GarnetStatus SortedSetRange(ArgSlice key, ArgSlice min, ArgSlice max, Sor
         public GarnetStatus SortedSetDifference(ArgSlice[] keys, out Dictionary<byte[], double> pairs)
             => storageSession.SortedSetDifference(keys, out pairs);
 
+        /// <inheritdoc />
+        public GarnetStatus SortedSetDifferenceStore(ArgSlice[] keys, ArgSlice destinationKey, out int count)
```
TalZaccai, line 133: For consistency with other similar commands, let's do: ```SortedSetDifferenceStore(byte[] key, ...)``` (see: SetDiffStore, SetIntersectStore etc.)

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -827,5 +827,61 @@ private unsafe bool SortedSetDifference<TGarnetApi>(ref TGarnetApi storageApi)
...
+        /// </summary>
+        /// <param name="storageApi"></param>
+        /// <returns></returns>
+        /// <exception cref="GarnetException"></exception>
+        private unsafe bool SortedSetDifferenceStore<TGarnetApi>(ref TGarnetApi storageApi)
+             where TGarnetApi : IGarnetApi
+        {
+            if (parseState.Count < 3)
+            {
+                return AbortWithWrongNumberOfArguments("ZDIFFSTORE");
```
TalZaccai, line 844: nit: ```nameof(RespCommand.ZDIFFSTORE)```

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -827,5 +827,61 @@ private unsafe bool SortedSetDifference<TGarnetApi>(ref TGarnetApi storageApi)
...
+            }
+
+            //number of keys
```
TalZaccai, line 847: nit: // Number of keys

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -827,5 +827,61 @@ private unsafe bool SortedSetDifference<TGarnetApi>(ref TGarnetApi storageApi)
...
+
+            if (parseState.Count - 2 != nKeys)
+            {
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            var destination = parseState.GetArgSliceByRef(0);
+            var keys = new ArgSlice[nKeys];
```
TalZaccai, line 863: There's no reason to allocate another array for these keys as they already exist in the parse state.
You can create the ObjectInput instance here and hand that off to the storageApi instead.

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -989,12 +990,59 @@ public void CheckSortedSetOperationsOnWrongTypeObjectSE()
             RespTestsUtils.CheckCommandOnWrongTypeObjectSE(() => db.SortedSetRandomMember(keys[1]));
             // ZDIFF
             RespTestsUtils.CheckCommandOnWrongTypeObjectSE(() => db.SortedSetCombine(SetOperation.Difference, keys));
+            // ZDIFFSTORE
+            RespTestsUtils.CheckCommandOnWrongTypeObjectSE(() => db.SortedSetCombineAndStore(SetOperation.Difference, destionationKey, keys));
```
TalZaccai, line 994: typo: destinationKey (multiple instances in this file)

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -989,12 +990,59 @@ public void CheckSortedSetOperationsOnWrongTypeObjectSE()
...
             // ZSCAN
             RespTestsUtils.CheckCommandOnWrongTypeObjectSE(() => db.SortedSetScan(keys[1], new RedisValue("*")).FirstOrDefault());
             //ZMSCORE
             RespTestsUtils.CheckCommandOnWrongTypeObjectSE(() => db.SortedSetScores(keys[1], values[1]));
         }
 
+        [Test]
+        public void CheckSortedSetDifferenceStoreSE()
```
TalZaccai, line 1002: Please also test storing into an existing destination key

TalZaccai (changes requested):
Thanks! Added a few comments

libs/server/API/GarnetApiObjectCommands.cs
Vijay-Nirmal: If I do that, it won't be consistent with `SortedSetDifference`. Is that fine :-)

libs/server/Resp/Objects/SortedSetCommands.cs
Vijay-Nirmal: Fixed

libs/server/Resp/Objects/SortedSetCommands.cs
Vijay-Nirmal: Makes sense, the only reason I did is because the existing `SortedSetDifference` method does the same thing. Instead of using `ObjectInput` I used `Span<ArgSlice>` because `Span<ArgSlice>` will be easier to use if some one need to call this programmatically using GarnetApi. Let me know if you are fine with this change.

test/Garnet.test/RespSortedSetTests.cs
Vijay-Nirmal: Is it not tested already using the below line? Do you mean **non**-existing destination key?
`db.SortedSetAdd(destionationKey, key1Values); // Set up destination key to overwrite if exists`

Vijay-Nirmal: Added test case for those the case

test/Garnet.test/RespSortedSetTests.cs
Vijay-Nirmal: Fixed

libs/server/API/GarnetApiObjectCommands.cs
TalZaccai: SortedSetDifference doesn't have a dstKey...

libs/server/Resp/Objects/SortedSetCommands.cs
TalZaccai: Yes you're correct, that's because we didn't have ObjectInput when SSD was written ;) At some point it needs to be updated as well. And yes, your solution seems fine to me.

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -604,6 +604,69 @@ public unsafe GarnetStatus SortedSetDifference(ArgSlice[] keys, out Dictionary<b
...
+                foreach (var item in keys)
+                    txnManager.SaveKeyEntryToLock(item, true, LockType.Shared);
+                _ = txnManager.Run(true);
+            }
+
+            var setObjectStoreLockableContext = txnManager.ObjectStoreLockableContext;
+
+            try
+            {
+                var status = SortedSetDifference(keys, out var result);
```
TalZaccai, line 637: Note that SortedSetDifference is also doing locking (that you don't need at that point), there should be a common method that extracts only the common logic that both SSD & SSDS call. See SetDiff & SetDiffStore for reference.

libs/server/API/GarnetApiObjectCommands.cs
Vijay-Nirmal: Ohh... Are you asking me to move the `destinationKey` to the first param and convert it to `byte[]` from `ArgSlice`?

I have previously seen that. I don't understand why we are passing it as `byte[]` instead of `ArgSlice` because we are converting `ArgSlice` to `byte[]` in the network layer and then converting `byte[]` to `ArgSlice` in the GarnetApi layer to create the transaction. To convert it back to `ArgSlice` we need additional memory, for that we are using scratchBufferManager. This looks inefficient to me when compared to directly passing it as `ArgSlice`. IMO, the existing places should be changed to use `ArgSlice`.

I don't prefer passing it as `byte[]` for this reason but if you like passing it as `byte[]` I will change it once you confirm it again :)

libs/server/Resp/Objects/SortedSetCommands.cs
Vijay-Nirmal: > And yes, your solution seems fine to me.

Marking this as resolved

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: Fixed

libs/server/API/GarnetApiObjectCommands.cs
TalZaccai: Oh no no need for that, I just referred to the order of parameters.

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -561,45 +561,116 @@ public unsafe GarnetStatus SortedSetDifference(ArgSlice[] keys, out Dictionary<b
...
-                    }
+                    return GarnetStatus.WRONGTYPE;
+                }
 
-                    if (keys.Length == 1)
-                    {
-                        pairs = firstSortedSet.Dictionary;
-                        return GarnetStatus.OK;
-                    }
+                count = pairs is null ? 0 : pairs.Count;
```
TalZaccai, line 610: nit: result?.Count ?? 0

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -561,45 +561,116 @@ public unsafe GarnetStatus SortedSetDifference(ArgSlice[] keys, out Dictionary<b
...
+        }
+
+        /// <summary>
+        /// Computes the difference between the first and all successive sorted sets and store resulting pairs in the destination key.
+        /// </summary>
+        /// <param name="keys"></param>
+        /// <param name="destinationKey"></param>
+        /// <param name="count"></param>
+        /// <returns></returns>
+        public unsafe GarnetStatus SortedSetDifferenceStore(ReadOnlySpan<ArgSlice> keys, ArgSlice destinationKey, out int count)
```
TalZaccai, line 580: nit: unsafe not needed

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -561,45 +561,116 @@ public unsafe GarnetStatus SortedSetDifference(ArgSlice[] keys, out Dictionary<b
...
+                return status;
             }
             finally
             {
                 if (createTransaction)
                     txnManager.Commit(true);
             }
+        }
+
+        private unsafe GarnetStatus SortedSetDifference<TObjectContext>(ReadOnlySpan<ArgSlice> keys, ref TObjectContext objectContext, out Dictionary<byte[], double> pairs)
```
TalZaccai, line 636: Please move to under all public methods in this file

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: Done

libs/server/API/GarnetApiObjectCommands.cs
Vijay-Nirmal: Changed the order of the param

TalZaccai (approved):

PR #735, opened 2024-10-18
Rate limiter using sorted sets.

yrajas:
Rater limiter unit test.


main/GarnetServer/Program.cs
```diff
@@ -94,6 +94,8 @@ static void RegisterExtensions(GarnetServer server)
 
             server.Register.NewProcedure("SUM", new Sum());
             server.Register.NewProcedure("SETMAINANDOBJECT", new SetStringAndList());
+
+            server.Register.NewTransactionProc("RATELIMIT", () => new RateLimiterTxn(), new RespCommandsInfo { Arity = 4 });
```
badrishc, line 98: move to tests

yrajas: Moved to test project.

test/Garnet.test/Extensions/RateLimiterTxn.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Text;
+using Garnet.common;
+using Garnet.server;
+using Tsavorite.core;
+
+namespace Garnet
+{
+    sealed class RateLimiterTxn : CustomTransactionProcedure
+    {
+        public override bool Prepare<TGarnetReadApi>(TGarnetReadApi api, ref CustomProcedureInput procInput)
```
Xizt, line 14: Could you help me understand the difference between CustomProcedureInput and ArgSlice input. Based on docs I could use ArgSlice.

test/Garnet.test/Extensions/RateLimiterTxn.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            int offset = 0;
+            AddKey(GetNextArg(ref procInput, ref offset), LockType.Exclusive, true);
```
Xizt, line 17: Wanted to understand if there is a way to have configurable server side timeouts. There might be a scenario with 1000s of requests for same key waiting to obtain lock. In such cases, want to have timeouts so that load can be reduced on server.

test/Garnet.test/Extensions/RateLimiterTxn.cs
```diff
@@ -0,0 +1,81 @@
...
+            return true;
+        }
+
+        public override unsafe void Main<TGarnetApi>(TGarnetApi api, ref CustomProcedureInput procInput, ref MemoryResult<byte> output)
```
Xizt, line 21: Do we need a way for measuring latency and KPS metrics for these custom transactions?

test/Garnet.test/Extensions/RateLimiterTxn.cs
badrishc: You can override this field from CustomTransactionProcedure.cs base class --

```cs
        /// <summary>
        /// Timeout for acquiring key locks. 100 ms by default
        /// </summary>
        public virtual TimeSpan KeyLockTimeout => TimeSpan.FromMilliseconds(
#if DEBUG
            5000    /* Things are slower in Debug so give it more time */
#else
            100
#endif
        );
```

test/Garnet.test/Extensions/RateLimiterTxn.cs
yrajas: That changed as part of a recent refactoring.
Good point. Will update the documentation as well.

test/Garnet.test/Extensions/RateLimiterTxn.cs
yrajas: TX_PROC_LAT helps track latency for custom transactions.

badrishc (approved):

PR #737, opened 2024-10-19
Cleanup BDN organization

badrishc:


benchmark/BDN.benchmark/Cluster/ClusterContext.cs
```diff
@@ -40,18 +27,18 @@ public void Dispose()
...
         {
             var opt = new GarnetServerOptions
             {
                 QuietMode = true,
-                EnableCluster = enableCluster,
-                Port = Port,
+                EnableCluster = !disableSlotVerification,
+                Port = port,
                 CleanClusterConfig = true,
             };
```
vazois, line 38: Can you add the following? It will make @darrenge life easier because in his CI benchmark, it fails to create the cluster config file due to permission issue. 
```suggestion
            };
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                opt.CheckpointDir = "/tmp";
```

vazois (approved):

PR #740, opened 2024-10-20
chore: make garnet server options DeviceFactoryCreator creation be lazy

### LINKED ISSUE #739: Initialize azure blob storage client once during startup
### Describe the bug

In `Garnet.Options` we use a `Func<INamedDeviceFactory>` that each time called returns a new instance.

We could use a `Lazy<INamedDeviceFactory>` here to reduce the instances used to 1.

1. added log message in `AzureStorageNamedDeviceFactory` contructor, 
ex: 
```cs
logger?.LogInformation("AzureStorageNamedDeviceFactory: serviceUri: {serviceUri}", serviceUri);
```

2. Start the server and see this message multiple times

### Expected behavior

`AzureStorageNamedDeviceFactory` should be created only once

### Screenshots

_No response_

### Release version

_No response_

### IDE

_No response_

### OS version

_No response_

### Additional context

_No response_

Meir017:
closes #739

### DIFF
--- a/libs/server/Servers/GarnetServerOptions.cs
+++ b/libs/server/Servers/GarnetServerOptions.cs
@@ -246,10 +246,21 @@ namespace Garnet.server
         /// </summary>
         public int ThreadPoolMaxThreads = 0;
 
+        private Func<INamedDeviceFactory> _deviceFactoryCreator;
+        private Lazy<INamedDeviceFactory> _deviceFactoryLazyCreator;
+
         /// <summary>
         /// Creator of device factories
         /// </summary>
-        public Func<INamedDeviceFactory> DeviceFactoryCreator = null;
+        public Func<INamedDeviceFactory> DeviceFactoryCreator
+        {
+            get => _deviceFactoryCreator;
+            set
+            {
+                _deviceFactoryCreator = value;
+                _deviceFactoryLazyCreator = new Lazy<INamedDeviceFactory>(_deviceFactoryCreator);
+            }
+        }
 
         /// <summary>
         /// Whether and by how much should we throttle the disk IO for checkpoints (default = 0)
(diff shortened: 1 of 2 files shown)


libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -246,10 +246,21 @@ public class GarnetServerOptions : ServerOptions
...
+        private Func<INamedDeviceFactory> _deviceFactoryCreator;
+        private Lazy<INamedDeviceFactory> _deviceFactoryLazyCreator;
+
         /// <summary>
         /// Creator of device factories
         /// </summary>
-        public Func<INamedDeviceFactory> DeviceFactoryCreator = null;
+        public Func<INamedDeviceFactory> DeviceFactoryCreator
+        {
+            get => _deviceFactoryCreator;
```
Meir017, line 257: I'm wondering if we should only expose the `set` without a `get`

badrishc:
I don't see how this solution would work. The reason we create a different factory each time is that we need to initialize the _factory_ with a different baseName for each factory. That way, the instances created by the factory will all use the same base name. See GarnetServerOptions->GetInitializedDeviceFactory for how this is used.

Meir017:
> I don't see how this solution would work. The reason we create a different factory each time is that we need to initialize the _factory_ with a different baseName for each factory. That way, the instances created by the factory will all use the same base name. See GarnetServerOptions->GetInitializedDeviceFactory for how this is used.

I see, I was confused here the scope of `AzureStorageNamedDeviceFactory`, I think we need another layer for the AzureStorageNamedDeviceFactory as I think for a single blob storage we should create at most a single client, but a single client could interact with multiple blob containers - and currently the code supports a 1:1 relationship between storage-account and blob-container

badrishc:
Agreed, something is off re: the abstraction right now. We need to be able to define the base name independently of the storage library instance.

badrishc:
Closing in favor of https://github.com/microsoft/garnet/pull/970

PR #744, opened 2024-10-23
Enabling limited parameter access to SessionParseState

TalZaccai:
TL;DR: This change is to enable limited access to SessionParseState parameters without adding offset computation overhead.

The command processing layer gets an input (ObjectInput / RawStringInput) which holds a parse state. The start index at which to start reading the parameters and the amount of parameters to read can vary (for instance, if it's being called programatically vs from the network buffer). Before, we had to pass a start and end indexes to offset the existing parse state in the input, which was not ideal (poor readability, computation overhead).
In this PR we're creating a sliced version of the parse state that points to the already offset parameter buffer and has an updated argument count, so the valid indexes are always between 0 and Count. 

Benchmark results:
main:
| Method      | Params | Mean     | Error    | StdDev   | Allocated |
|------------ |------- |---------:|---------:|---------:|----------:|
| Set         | None   | 21.78 us | 0.428 us | 0.627 us |         - |
| SetEx       | None   | 27.99 us | 0.536 us | 0.574 us |         - |
| SetNx       | None   | 26.70 us | 0.058 us | 0.051 us |         - |
| SetXx       | None   | 29.30 us | 0.190 us | 0.178 us |         - |
| GetFound    | None   | 19.43 us | 0.198 us | 0.185 us |         - |
| GetNotFound | None   | 13.79 us | 0.067 us | 0.056 us |         - |
| Increment   | None   | 30.27 us | 0.041 us | 0.034 us |         - |
| Decrement   | None   | 30.19 us | 0.166 us | 0.139 us |         - |
| IncrementBy | None   | 36.91 us | 0.066 us | 0.062 us |         - |
| DecrementBy | None   | 37.13 us | 0.414 us | 0.367 us |         - |

#744:
| Method      | Params | Mean     | Error    | StdDev   | Allocated |
|------------ |------- |---------:|---------:|---------:|----------:|
| Set         | None   | 20.20 us | 0.214 us | 0.200 us |         - |
| SetEx       | None   | 27.06 us | 0.513 us | 0.455 us |         - |
| SetNx       | None   | 25.23 us | 0.016 us | 0.013 us |         - |
| SetXx       | None   | 26.95 us | 0.106 us | 0.094 us |         - |
| GetFound    | None   | 19.74 us | 0.046 us | 0.041 us |         - |
| GetNotFound | None   | 13.85 us | 0.067 us | 0.056 us |         - |
| Increment   | None   | 29.22 us | 0.070 us | 0.059 us |         - |
| Decrement   | None   | 29.81 us | 0.061 us | 0.048 us |         - |
| IncrementBy | None   | 36.78 us | 0.502 us | 0.470 us |         - |
| DecrementBy | None   | 35.76 us | 0.068 us | 0.060 us |         - |


libs/server/InputHeader.cs
```diff
@@ -358,22 +333,20 @@ public RawStringInput(ushort cmd, byte flags = 0, long arg1 = 0) :
...
         /// <param name="arg1">General-purpose argument</param>
         /// <param name="flags">Flags</param>
-        public RawStringInput(RespCommand cmd, ref SessionParseState parseState, int parseStateFirstArgIdx = 0,
-            int parseStateLastArgIdx = -1, long arg1 = 0, RespInputFlags flags = 0) : this(cmd, flags, arg1)
+        public RawStringInput(RespCommand cmd, ref SessionParseState parseState, int startIdx = 0, int argCount = -1, long arg1 = 0, RespInputFlags flags = 0) : this(cmd, flags, arg1)
         {
             this.parseState = parseState;
-            this.parseStateFirstArgIdx = parseStateFirstArgIdx;
-            this.parseStateLastArgIdx = parseStateLastArgIdx;
+            this.parseState.Slice(startIdx, argCount);
```
badrishc, line 343: this.parseState = parseState.Slice(...) - i dont think we should be modifying fields of already created structs. These fields are to be treated as effectively read-only.

libs/server/Resp/Parser/SessionParseState.cs
```diff
@@ -158,10 +169,26 @@ public void InitializeWithArguments(ArgSlice[] args)
...
             }
         }
 
+        /// <summary>
+        /// Limit access to the argument buffer to start at a specified index
+        /// and end after a specified number of arguments.
+        /// </summary>
+        /// <param name="idxOffset">Offset value to the underlying buffer</param>
+        /// <param name="count">Argument count (default: -1 for all remaining arguments)</param>
+        public void Slice(int idxOffset, int count = -1)
```
badrishc, line 182: Should return the new SPS.
Shoud not have to incur the overhead of array lookup + pointer conversion.
Instead, it can do: return new SPS(bufferPtr + start, etc)

libs/server/Resp/Parser/SessionParseState.cs
```diff
@@ -158,10 +182,26 @@ public void InitializeWithArguments(ArgSlice[] args)
...
             }
         }
 
+        /// <summary>
+        /// Limit access to the argument buffer to start at a specified index
+        /// and end after a specified number of arguments.
+        /// </summary>
+        /// <param name="idxOffset">Offset value to the underlying buffer</param>
+        /// <param name="count">Argument count (default: -1 for all remaining arguments)</param>
+        public SessionParseState Slice(int idxOffset, int count = -1)
```
badrishc, line 195: Use separate overloads to avoid `if` overhead?

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ReadOnlySpan.cs#L357

badrishc (commented):
1. ArrayCommands.cs - NetworkArrayPING does not use parseState in WriteDirect?
2. Same for NetworkECHO

`// On success, readHead is left at the start of the command payload for legacy operators`

3. The fields of `readHead` and `endReadHead` can be eliminated/unified and make sure nobody uses them

TedHartMS (approved):
Very nice cleanup!

Approved with one possible necessary change--see comment

TalZaccai:
> 1. ArrayCommands.cs - NetworkArrayPING does not use parseState in WriteDirect?
> 2. Same for NetworkECHO
> 
> `// On success, readHead is left at the start of the command payload for legacy operators`
> 
> 3. The fields of `readHead` and `endReadHead` can be eliminated/unified and make sure nobody uses them

readHead and endReadHead are stilll used in RespCommand and RespServerSession

libs/server/Objects/List/ListObjectImpl.cs
```diff
@@ -582,15 +578,15 @@ private void ListPosition(ref ObjectInput input, ref SpanByteAndMemory output)
...
-            var currTokenIdx = input.parseStateFirstArgIdx;
-
             rank = 1; // By default, LPOS takes first match element
             count = 1; // By default, LPOS return 1 element
             isDefaultCount = true;
             maxlen = 0; // By default, iterate to all the item
 
             error = default;
 
+            var currTokenIdx = 1;
```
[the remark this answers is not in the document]
TalZaccai: It was incremented previously in the caller (ListPosition)

libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -20,9 +20,9 @@ private void SetAdd(ref ObjectInput input, byte* output)
             var _output = (ObjectOutputHeader*)output;
             *_output = default;
 
-            for (var currTokenIdx = input.parseStateFirstArgIdx; currTokenIdx < input.parseState.Count; currTokenIdx++)
+            for (var i = 0; i < input.parseState.Count; i++)
```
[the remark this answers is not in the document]
TalZaccai: Good to know, thanks!!

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -735,7 +735,7 @@ public unsafe GarnetStatus SetIsMember<TObjectContext>(ArgSlice key, ArgSlice[]
             {
                 type = GarnetObjectType.Set,
                 SetOp = SetOperation.SMISMEMBER,
-            }, ref parseState, 0);
```
badrishc, line 738: SetIsMember does not need to construct a new local parseState:
 
            var parseState = new SessionParseState();
            parseState.InitializeWithArguments(members);

libs/server/Resp/Parser/SessionParseState.cs
```diff
@@ -22,33 +22,55 @@ public unsafe struct SessionParseState
...
 
         /// <summary>
         /// Get a Span of the parsed parameters in the form an ArgSlice
         /// </summary>
         public ReadOnlySpan<ArgSlice> Parameters => new(bufferPtr, Count);
 
+        /// <summary>
+        /// Create a new instance of SessionParseState
+        /// </summary>
+        public SessionParseState()
```
badrishc, line 52: Remove default constructur as it would no longer be used.

libs/common/RespWriteUtils.cs
```diff
@@ -289,6 +289,22 @@ public static bool WriteDirect<T>(ref T item, ref byte* curr, byte* end) where T
             return true;
         }
 
+        /// <summary>
+        /// Write bulk string
```
badrishc, line 293: incorrect comment

badrishc (approved):

badrishc (approved):

PR #736, opened 2024-10-19
Expanding RespCommand enum to ushort

TalZaccai:
Benchmark results:

GeoHashBenchmarks

main:
| Method                 | Mean      | Error     | StdDev    | Median    | Gen0   | Allocated |
|----------------------- |----------:|----------:|----------:|----------:|-------:|----------:|
| GeoToLongValue         |  3.839 ns | 0.0661 ns | 0.0552 ns |  3.820 ns |      - |         - |
| GetCoordinatesFromLong |  4.860 ns | 0.0939 ns | 0.0784 ns |  4.845 ns |      - |         - |
| GetGeoHashCode         | 15.136 ns | 0.8173 ns | 2.3186 ns | 14.469 ns | 0.0001 |      48 B |

#736:
| Method                 | Mean      | Error     | StdDev    | Gen0   | Allocated |
|----------------------- |----------:|----------:|----------:|-------:|----------:|
| GeoToLongValue         |  4.001 ns | 0.1029 ns | 0.1374 ns |      - |         - |
| GetCoordinatesFromLong |  4.820 ns | 0.0594 ns | 0.0496 ns |      - |         - |
| GetGeoHashCode         | 15.688 ns | 0.1285 ns | 0.1139 ns | 0.0001 |      48 B |

RecoveryBenchmark

main:
| Method  | MemorySize | LogDir | Mean     | Error   | StdDev   | Median   | P90      | P95      |
|-------- |----------- |------- |---------:|--------:|---------:|---------:|---------:|---------:|
| Recover | 100m       | -f     | 117.9 us | 3.98 us | 11.53 us | 114.3 us | 134.7 us | 139.4 us |
| Recover | 100m       | *      |       NA |      NA |       NA |       NA |       NA |       NA |

#736:
| Method  | MemorySize | LogDir | Mean     | Error    | StdDev   | Median   | P90      | P95      |
|-------- |----------- |------- |---------:|---------:|---------:|---------:|---------:|---------:|
| Recover | 100m       | -f     | 215.4 us | 31.29 us | 90.77 us | 181.9 us | 388.3 us | 407.1 us |
| Recover | 100m       | *      |       NA |       NA |       NA |       NA |       NA |       NA |

SpanVsPointer

main:
| Method        | Mean       | Error    | StdDev   | Median     | Ratio | RatioSD |
|-------------- |-----------:|---------:|---------:|-----------:|------:|--------:|
[...]

PR #748, opened 2024-10-24
[Bug] Fixed CustomCommand test case falure after ushort conversion

Vijay-Nirmal:
This bug is introduced as part of #736


TalZaccai (approved):
Good catch! Thanks!

PR #752, opened 2024-10-26
[Compatibility] Added GEOSEARCHSTORE command and bug fix in GEOSEARCH command

Vijay-Nirmal:
Adding the GEOSEARCHSTORE command to garnet and a bug fix in the response structure of the GEOSEARCH command

- [x] Add GEOSEARCHSTORE command
- [x] Add documentation
- [x] Bug fix in the response structure of GEOSEARCH command where it returns data in unexpected RESP format
- [x] Add new test case to GEOSEARCH command using SE.Redis


libs/common/RespReadUtils.cs
```diff
@@ -1033,6 +1079,32 @@ public static bool ReadString(out string result, ref byte* ptr, byte* end)
             return false;
         }
 
+        /// <summary>
+        /// Read ASCII string as span without header until string terminator ('\r\n').
+        /// </summary>
+        public static bool ReadAsSapn(out ReadOnlySpan<byte> result, ref byte* ptr, byte* end)
```
badrishc, line 1085: nit: Sapn -> Span

Vijay-Nirmal: Fixed

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+            var isMemory = false;
+            MemoryHandle ptrHandle = default;
+            var ptr = output.SpanByte.ToPointer();
+            var curr = ptr;
+            var end = curr + output.Length;
+
+            try
+            {
+                var isStoreDist = false;
+                Span<ArgSlice> geoSearchParseState = stackalloc ArgSlice[input.parseState.Count + 1]; // Current MaxSize of this stactalloc will be 144 bytes ( 12 (Max Args) x 12 (sizeof ArgSlice) )
```
TalZaccai, line 80: Why are we regenerating the parse state here?
Also, STOREDIST can only appear in the last argument position

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+                    }
+                    i++;
+                }
+                geoSearchParseState[currArgIdx++] = isStoreDist ? ArgSlice.FromPinnedSpan(CmdStrings.WITHDIST) : ArgSlice.FromPinnedSpan(CmdStrings.WITHHASH);
+
+                var sourceKey = key.ToArray();
+                var parseState = new SessionParseState();
+                parseState.InitializeWithArguments(geoSearchParseState.Slice(0, currArgIdx));
+
+                var searchInput = new ObjectInput(new RespInputHeader
```
TalZaccai, line 103: For consistency:
```cs
var header = new RespInputHeader(GarnetObjectType.SortedSet) { SortedSetOp = SortedSetOperation.GEOSEARCH };
var searchInput = new ObjectInput(header, ref parseState);
```

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+                        break;
+                    }
+                    else
+                    {
+                        geoSearchParseState[currArgIdx] = input.parseState.GetArgSliceByRef(i);
+                        currArgIdx++;
+                    }
+                    i++;
+                }
+                geoSearchParseState[currArgIdx++] = isStoreDist ? ArgSlice.FromPinnedSpan(CmdStrings.WITHDIST) : ArgSlice.FromPinnedSpan(CmdStrings.WITHHASH);
```
TalZaccai, line 97: Can add this logic in the implementation-level, no need to allocate space for an extra argument

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+                        ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr, ref end);
+                    return GarnetStatus.OK;
+                }
+
+                Debug.Assert(!searchOutMem.IsSpanByte, "Output should not be in SpanByte format when the status is OK");
+
+                var searchOutHandler = searchOutMem.Memory.Memory.Pin();
+                try
+                {
+
```
TalZaccai, line 132: nit: spacing

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+                searchOutMem = searchOut.spanByteAndMemory;
+
+                if (status == GarnetStatus.WRONGTYPE)
+                {
+                    return GarnetStatus.WRONGTYPE;
+                }
+
+                if (status == GarnetStatus.NOTFOUND)
+                {
+                    _ = EXPIRE(destination, TimeSpan.Zero, out _, StoreType.Object, ExpireOption.None, ref lockableContext, ref objectStoreLockableContext);
```
TalZaccai, line 121: Why do we need this expire call here?

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -41,5 +44,160 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+
+                    // Prepare the input
+                    var zAddInput = new ObjectInput(new RespInputHeader
+                    {
+                        type = GarnetObjectType.SortedSet,
+                        SortedSetOp = SortedSetOperation.ZADD,
+                    }, ref zParseState);
+
+                    var zAddOutput = new GarnetObjectStoreOutput { spanByteAndMemory = new SpanByteAndMemory(null) };
+                    RMWObjectStoreOperationWithOutput(destinationKey, ref zAddInput, ref objectStoreLockableContext, ref zAddOutput);
```
TalZaccai, line 181: Why do another call to storage? Why not have 1 call that does both?

TalZaccai (changes requested):
Some initial comments

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: Sadly its not, it can appear in middle, that is the first thing I tested it
image

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: To delete the value in the existing destination key if the source key is not there

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: I didn't get what you mean by "implementation-level", can you explain it a bit more? "allocate space" is just a 12 byte stackalloc

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: What do you mean by "both"? Delete and Insert? I am not sure how I can do that because the key we need to delete can be of a different type (Anything other than SortedSet) than the one we need to insert (SortedSet)

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: Hmmm ok so I think the last tokens are in this group ([ASC|DESC],[COUNT count [ANY]],[STOREDIST]) but can be at any order...

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: I mean pass the parseState as-is and handle iterating through it in the object implementation.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: Thanks, a comment here would be useful.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: Ah yes it's a multi-key operation, you're right.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: > handle iterating through it in the object implementation.

Does that mean I can't use GEOSEARCH from the API level? I have to created a separate `SortedSetOperation`, also I can't create an operation called GEOSEARCHSTORE because that operation doesn't perform store.  Also, I shouldn't modify the existing `GEOSEARCH` operation to accept `STOREDIST` option because if I do that then the existing `GEOSEARCH` command will treat `STOREDIST` as a valid input. I am not sure what is the right way to "handle iterating through it in the object implementation"

Can you provide the high-level implementation plan for this as your prefer then I can implement it in that manner?

TalZaccai: Let's just fix the other small comments, I can revise at a later time and tag you as an FYI :)

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: I have changed the implementation not to use additional stackalloc

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: I have changed the implementation so that it does not use additional stackalloc. Still I have not done the logic in "object implementation". Let me know if this is fine

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
Vijay-Nirmal: Done

Vijay-Nirmal:
Fixed review comment

Vijay-Nirmal:
@TalZaccai Please review it again when possible

TalZaccai:
> @TalZaccai Please review it again when possible

Added some improvements to incur less allocations. Feel free to take a look.

TalZaccai (approved):

PR #753, opened 2024-10-27
Fix HINCRBYFLOAT precision, use double, not float

nightroman:
**Problem:**
`HashObject.HashIncrement` uses `float` as `NumUtils.TryParse` output type, instead of the presumably intended `double`.
This results in floating precision loss, e.g. 2.1 decremented by 1.2 gives 0.89999986 instead of expected 0.9 +/- 1e-15.

**Solution:**
Use `double` as the output type in `NumUtils.TryParse` calls , one for the input and one for the existing value.

**Test:**
Added the test `CheckHashIncrementDoublePrecision`, it fails with the original code, passes with the amended.

### DIFF
--- a/libs/server/Objects/Hash/HashObjectImpl.cs
+++ b/libs/server/Objects/Hash/HashObjectImpl.cs
@@ -393,7 +393,7 @@ namespace Garnet.server
                 }
                 else
                 {
-                    if (!NumUtils.TryParse(incrSlice.ReadOnlySpan, out float incr))
+                    if (!NumUtils.TryParse(incrSlice.ReadOnlySpan, out double incr))
                     {
                         while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_NOT_VALID_FLOAT, ref curr, end))
                             ObjectUtils.ReallocateOutput(ref output, ref isMemory, ref ptr, ref ptrHandle, ref curr,
@@ -405,7 +405,7 @@ namespace Garnet.server
 
                     if (valueExists)
                     {
-                        if (!NumUtils.TryParse(value, out float result))
+                        if (!NumUtils.TryParse(value, out double result))
                         {
                             while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_HASH_VALUE_IS_NOT_FLOAT, ref curr,
                                        end))
--- a/test/Garnet.test/RespHashTests.cs
+++ b/test/Garnet.test/RespHashTests.cs
@@ -274,6 +274,16 @@ namespace Garnet.test
         }
 
         [Test]
+        public void CheckHashIncrementDoublePrecision()
+        {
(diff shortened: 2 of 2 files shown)


TalZaccai (approved):
Good catch! Thank you for the fix!

libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -393,7 +393,7 @@ private void HashIncrement(ref ObjectInput input, ref SpanByteAndMemory output)
                 }
                 else
                 {
-                    if (!NumUtils.TryParse(incrSlice.ReadOnlySpan, out float incr))
+                    if (!NumUtils.TryParse(incrSlice.ReadOnlySpan, out double incr))
```
Vijay-Nirmal, line 396: If you change the datatype from `float` to `double` you have to change the logic on how the double is converted into a string. Because `result.ToString(CultureInfo.InvariantCulture)` will give value in explorational notation eg: 0.00001 double value will be shown as `1E-5` but when we do `HGET` it should show the value as `0.00001` not `1E-5`. You have to make changes in the below 2 places for that.

https://github.com/microsoft/garnet/blob/b2448121ce2ddc3c0850cfd12dea25c42fb3e529/libs/server/Objects/Hash/HashObjectImpl.cs#L419

https://github.com/microsoft/garnet/blob/b2448121ce2ddc3c0850cfd12dea25c42fb3e529/libs/server/Objects/Hash/HashObjectImpl.cs#L426

Vijay-Nirmal (changes requested):

Vijay-Nirmal:
@TalZaccai Just added a review comment :)

libs/server/Objects/Hash/HashObjectImpl.cs
TalZaccai: Thanks @Vijay-Nirmal! I'll have a look separately.

Vijay-Nirmal: @TalZaccai I will create a PR for this

TalZaccai: image

TalZaccai: What's the issue with HGET? It should show the number in this format

Vijay-Nirmal: @TalZaccai What's happening in Garnet
image

Whats expected in Redis
image

TalZaccai: Ah ok so there's a culture difference, I see both as #2

Vijay-Nirmal: This shouldn't be a culture difference. Did you use exactly `0.0000001`? Or you can notice this in any number even non decimal like just type a very big number with more than 32 digits and check

PR #755, opened 2024-10-29
Removing Newtonsoft.Json dependency from Garnet.host and .nuspec file

TalZaccai:


badrishc (approved):

badrishc:
World be good to update the release version as well.

wdscxsj:
Garnet v1.0.35 rejects a conf file with a trailing comma for the last item. That's not strictly valid JSON, but is convenient to write and allowed by v1.0.34. Is it possible to relax the rule a little bit?

badrishc:
> Garnet v1.0.35 rejects a conf file with a trailing comma for the last item. That's not strictly valid JSON, but is convenient to write and allowed by v1.0.34. Is it possible to relax the rule a little bit?

https://github.com/microsoft/garnet/pull/761

PR #758, opened 2024-10-29
Implementing BDN Perf pipeline

darrenge:
There are three areas of this PR:
1) run_bdnperftest.ps1
     a) Runs individual bdn test based on the associated config file (in /config) directory
     b) Compares "allocated" value found from the run to the expected value in config file and if it more than 10% from expected, it will pop a fail
2) yml file for Garnet CI BDN.benchmark Github action
    a) Goes through the matrix of Linux, WIndows and runs each BDN test on those two (total of 10 runs)
    b) Each run saves results file and error log to GH Actions artifacts (for debugging if needed)
    c) Each run takes the BDN results (from the ps1 file run) and compares MEAN to previous and if changes by 50% or more (represented as 150% in yml file) then it will comment in commit
    d) Each run has task that adds to the "Summary" part of the GH Action run (table that shows previous run to current run) 
    e) Each run adds to charts (data is stored in continuousbenchmarking branch in the website/static/charts folder of that branch)
3) Config files each has:
    a) Name of BDN filter it is running
    b) Expected "ground truth" for Allocated values
    b) Acceptable range % for allocated ("acceptableAllocatedRange": 10) before flags a fail


test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
```diff
@@ -0,0 +1,31 @@
+{
+    "configuration": "Release",
+    "framework": "net8.0",
+    "filter": "BDN.benchmark.Cluster.ClusterOperations.*",
+
+    "expectedGet_DSV_AllocatedValue_linux": 0,
```
badrishc, line 6: i think we should only have to add files and entries non-zero allocated. the script should assume, if it does not find a benchmark entry, that the allocated should be zero. that way this entire file gets deleted. and, many lines in other files get deleted as well. will make maintenance a tad easier.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
```diff
@@ -0,0 +1,31 @@
...
+    "expectedMGet_DSV_AllocatedValue_win": 0,
+    "expectedMSet_DSV_AllocatedValue_win": 0,
+    "expectedCPBSet_DSV_AllocatedValue_win": 0,
+    "expectedGet_None_AllocatedValue_win": 0,
+    "expectedSet_None_AllocatedValue_win": 0,
+    "expectedMGet_None_AllocatedValue_win": 0,
+    "expectedMSet_None_AllocatedValue_win": 0,
+    "expectedCPBSet_None_AllocatedValue_win": 0,
+
+    "acceptableAllocatedRange": 10,
```
badrishc, line 28: make these three lines part of the script, rather than put it in every file. it should be very stable. this change will be needed in order to get rid of the entire file anyway, per earlier comment.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Operations.ObjectOperations.json
```diff
@@ -0,0 +1,53 @@
+{
+    "configuration": "Release",
+    "framework": "net8.0",
+    "filter": "BDN.benchmark.Operations.ObjectOperations.*",
+
+    "expectedZAddRem_ACL_AllocatedValue_linux": 18400,
```
badrishc, line 6: allocated values should not be different across windows and linux, so can we unify this?

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Lua.LuaScripts.json
```diff
@@ -0,0 +1,19 @@
+{
+    "configuration": "Release",
```
badrishc, line 2: remove configuration (it will always be release), framework (it will always be .net8 - or change to .net9 when we are ready)

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Lua.LuaScripts.json
```diff
@@ -0,0 +1,19 @@
...
+    "framework": "net8.0",
+    "filter": "BDN.benchmark.Lua.*",
+
+    "expectedScript1_None_AllocatedValue_linux": 24,
```
badrishc, line 6: these names should just be of the minimal form `"None": 24` ... we need to make this easy to understand and maintain. hopefully the script will automatically find the right entry by looking for the Param name, so the script does not need to change every time we add a benchmark. we only have to add a single line into the json file, that's it.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Operations.BasicOperations.json
```diff
@@ -0,0 +1,17 @@
+{
+    "configuration": "Release",
```
badrishc, line 2: Merge everything into a single JSON file called bdn-allocated-targets.json with one nested section for each filter.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
darrenge: I see where you are coming from but not sure I agree with this.  Maintenance is easiest if it is consistent so we know one config file per BDN run. Also, assuming it should be zero if not there is an assumption and it might be better to specifically state to show we know it should be zero.  Currently, the BDN name/ filter is in the config file as well, so that would be a change to only just use the test name.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
darrenge: This gives us some flexibility to have the acceptable range linked to the specific BDN. If we move it to the script, then all BDN runs use the same value. I guess it is ok and we already do that for the Mean values, but it just loses some flexibility.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Lua.LuaScripts.json
darrenge: We need to control in this line: test: [ 'Operations.BasicOperations', 'Operations.ObjectOperations', 'Cluster.ClusterMigrate', 'Cluster.ClusterOperations', 'Lua.LuaScripts' ]   

Because if you just set the test some pre defined prefix "BDN_Tests*", then we lose the aspect where each test is ran concurrently. If you run all our BDNs in one filter, it takes 45+ mins.  

This also allows us to decide what BDNs we want to run and just don't assume if it is in the code, then it is ran in the CI.

As for variable names in the config file. It is helps to be descriptive. From other comment, we can remove the linux and win parts and just have one.  We do lose flexibility to look at other values, but I guess if we want to do that, we can just change the code again.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Operations.BasicOperations.json
darrenge: Not sure I understand why to do this?  I assume this is just an extra task near the end of the YML to do so we have it in the artifacts or are you talking about doing it instead of having separate json files?  I am not sure how the merge works but can you just continually merge them together as more tests finish? For example ... you concurrently start 10 BDN runs and #3 finishes, then #4 is finishing so at the end of that it merges, 3 and 4 together, the another test (7) finishes and so that merges 3, 4, 7 together etc until all 10 are done?  I hope it doesn't just add one and it is an actual merge. I can look into it.

test/BDNPerfTests/run_bdnperftest.ps1
```diff
@@ -0,0 +1,577 @@
...
+$allocatedColumn = "-1"   # last one is allocated, just to ensure in case other column gets added
+$paramNone = "None"
+$paramDSV = "DSV"
+$paramACL = "ACL"
+$paramAOF = "AOF"
+
+# Set the expected values based on the OS
+if ($IsLinux) {
+    # Linux expected values
+    $expectedGet_None_AllocatedValue = $object.expectedGET_None_AllocatedValue_linux
```
badrishc, line 142: this isn't maintainable. why not iterate over the retreived JSON object fields to find what we need instead of expressing each one here.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
darrenge: Done

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Lua.LuaScripts.json
darrenge: Changed it to: "expectedScript1_None": 24,   from:   "expectedScript1_None_AllocatedValue_linux": 24   so it is a closer to minimal form and still be descriptive enough

test/BDNPerfTests/run_bdnperftest.ps1
darrenge: I could do a switch statement based on what test is running and only get those values, but I was thinking it would just be easier to run through them all. Having the switch statement would clearly define what is expected from each specific test.

darrenge: Turns out the switch on config file to only read values we need for that specific tests looks better and is more concise.

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
darrenge: Since we are only pulling values we need for the specific test (based on other comment), we need to know what values we need for that specific test. Those specific values are in the config file, so I think it only makes sense to have all config files even if the values are 0.

badrishc: we could have one config file that has a section per test. for every test we add we would simply run the test, and for every result, we would comare it to either zero (by default) or an override (if specified in the JSON).

darrenge: Yeah, I guess that would be another way to implement it

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Operations.BasicOperations.json
badrishc: i was referring to the INPUT config file, not the output files. Output files have to definitely be different, not possible to merge correctly in general, so let us not even try that.

darrenge: ok - can resolve this as the Input config is already being discussed

test/BDNPerfTests/ConfigFiles/CI_CONFIG_BDN_Benchmark_Cluster.ClusterOperations.json
darrenge: I will try to implement it with one config file for all BDN test runs

darrenge: This is now implemented in a one config with a section per test.

test/BDNPerfTests/BDN_Benchmark_Config.json
```diff
@@ -0,0 +1,48 @@
+{
+  "_comment": "NOTE: If adding a new BDN perf test to this file, you will need to add the name to the test: line in ce-bdnbenchmark.yml (~line 45). This ensures all the tests are not ran sequentially when ran as github action.",
```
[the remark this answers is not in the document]
darrenge: fixed

badrishc (approved):

PR #759, opened 2024-10-30
Copy CI benchmark charts into website publishing action

badrishc:
* Grabs the latest charts from the branch `continuousbenchmark`, in the folder `website/static/charts`
* Copies the charts into the `static` folder before running `yarn`
* The pages deploy is updated to trigger on both `main` and `continuousbenchmark` branches

After merge, the benchmarking history charts should be available at https://microsoft.github.io/garnet/charts

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -28,9 +28,10 @@ jobs:
           sparse-checkout: |
             website/static/charts
           path: continuousbenchmark
-      - run: |
-          mkdir -p website/static/charts
-          cp continuousbenchmark/website/static/charts/* website/static/charts
+      - name: Copy charts
+        run: |
+          mkdir -p static/charts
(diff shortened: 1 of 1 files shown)


darrenge:
I am not sure you want to do this on every push to continuousbenchmark. Every time the BDN perf CI runs, it runs 10 BDN tests (5 tests on linux and windows). After each of those are finished, they push to continuousbenchmark.  That means for every BDN CI, this one will be ran 10 times. I do understand the timing of it though that if you do it on Main, you won't get the most up to date data on the site for that specific push. We could try it and just see how it goes.

darrenge (approved):
I think it should be ok if this is called 10 times every time the BDN perf pipeline is called.

badrishc:
Fixed to run once only when ci-bdnbenchmark completes, not on every checkin to that branch.

PR #763, opened 2024-11-01
Bugfix CustomTransaction AOF Replay Bug

hamdaankhalid:
After our command handling refactor we had an untested part that was not skipping AOF header when doing a replay. I caught it, and now skip it. This fixes the issue as seen by the UNit test.


TalZaccai (approved):
Good catch, thanks for adding the missing test!

Issue #771, opened 2024-11-05
SET EX NX does not set a string longer than expired


nightroman:
### Describe the bug

`SET EX NX` does not set a string longer than expired, e.g. if the expired
string is "short" (length 5) and the new value is "longer" (length 6) then
the result is `true` as expected ~ the string is set, but the new string is
not actually stored, the key does not exist.

### Steps to reproduce the bug

```csharp
    [Fact]
    public void ExistingExpired6379()
    {
        var redis = ConnectionMultiplexer.Connect("127.0.0.1:6379,SyncTimeout=999999");
        var db = redis.GetDatabase();

        var key = "test:lock1";
        var value1 = "short";
        var value2 = "longer";
        var ttl1 = TimeSpan.FromMilliseconds(200);
        var ttl2 = TimeSpan.FromMinutes(1);

        // set expiring string "short" and let it expire
        db.StringSet(key, value1, ttl1);
        Thread.Sleep(ttl1 * 2);

        // set expiring string "longer" over expired "short" -> true ~ expected
        bool ok = db.StringSet(key, value2, ttl2, When.NotExists);
        Assert.True(ok);

        // but the expected string does not exist, null is returned, assert fails
        string? str = db.StringGet(key);
        Assert.Equal(value2, str);
    }
```

### Expected behavior

The above xunit test is supposed to pass (new string exists) but it fails (the string does not exist, although the operation returned `true`).

Note that the test passes is we use a shorter string as `value2`, e.g. "longe" instead of "longer".

### Release version

1.0.35

### IDE

Microsoft Visual Studio Community 2022 (64-bit) - 17.11.5

### OS version

Windows 11

Vijay-Nirmal:
Raise a PR to fix it #772

PR #772, opened 2024-11-05
[Bug] Fix for SET EX NX does not set a string longer than expired

Vijay-Nirmal:
- [x] Fix for #771 
- [x] SET GET returns incorrect RESP format when the key is not found/exist
- [x] Add test cases to cover these cases

### DIFF
--- a/libs/server/Storage/Functions/MainStore/RMWMethods.cs
+++ b/libs/server/Storage/Functions/MainStore/RMWMethods.cs
@@ -558,21 +558,6 @@ namespace Garnet.server
         {
             switch (input.header.cmd)
             {
-                case RespCommand.SETEXNX:
-                    // Expired data, return false immediately
-                    // ExpireAndStop ensures that caller sees a NOTFOUND status
-                    if (oldValue.MetadataSize > 0 && input.header.CheckExpiry(oldValue.ExtraMetadata))
-                    {
-                        rmwInfo.Action = RMWAction.ExpireAndStop;
-                        return false;
-                    }
-                    // Check if SetGet flag is set
-                    if (input.header.CheckSetGetFlag())
-                    {
-                        // Copy value to output for the GET part of the command.
-                        CopyRespTo(ref oldValue, ref output);
-                    }
-                    return false;
                 default:
                     if ((ushort)input.header.cmd >= CustomCommandManager.StartOffset)
                     {
--- a/libs/server/Storage/Functions/MainStore/RMWMethods.cs
+++ b/libs/server/Storage/Functions/MainStore/RMWMethods.cs
@@ -558,6 +558,30 @@ namespace Garnet.server
         {
             switch (input.header.cmd)
             {
+                case RespCommand.SETEXNX:
+                    // Expired data, return false immediately
+                    // ExpireAndResume ensures that we set as new value, since it does not exist
+                    if (oldValue.MetadataSize > 0 && input.header.CheckExpiry(oldValue.ExtraMetadata))
+                    {
+                        rmwInfo.Action = RMWAction.ExpireAndResume;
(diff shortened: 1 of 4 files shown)


libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
```diff
@@ -206,6 +206,7 @@ public int GetRMWModifiedValueLength(ref SpanByte t, ref RawStringInput input)
 
                     case RespCommand.SET:
                     case RespCommand.SETEXXX:
+                    case RespCommand.SETEXNX:
```
badrishc, line 209: no need for this line, remove.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -558,21 +558,6 @@ public bool NeedCopyUpdate(ref SpanByte key, ref RawStringInput input, ref SpanB
         {
             switch (input.header.cmd)
             {
```
badrishc, line 560: Add this instead, before the `default` switch:

```cs
                case RespCommand.SETEXNX:
                    // Expired data, return false immediately
                    // ExpireAndResume ensures that we set as new value, since it does not exist
                    if (oldValue.MetadataSize > 0 && input.header.CheckExpiry(oldValue.ExtraMetadata))
                    {
                        rmwInfo.Action = RMWAction.ExpireAndResume;
                        return false;
                    }
                    // Check if SetGet flag is set
                    if (input.header.CheckSetGetFlag())
                    {
                        // Copy value to output for the GET part of the command.
                        CopyRespTo(ref oldValue, ref output);
                    }
                    return false;
                case RespCommand.SETEXXX:
                    // Expired data, return false immediately so we do not set, since it does not exist
                    // ExpireAndStop ensures that caller sees a NOTFOUND status
                    if (oldValue.MetadataSize > 0 && input.header.CheckExpiry(oldValue.ExtraMetadata))
                    {
                        rmwInfo.Action = RMWAction.ExpireAndStop;
                        return false;
                    }
                    return true;
```

badrishc (changes requested):

libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs
Vijay-Nirmal: Done

libs/server/Storage/Functions/MainStore/RMWMethods.cs
Vijay-Nirmal: Done

test/Garnet.test/RespTests.cs
```diff
@@ -323,15 +323,75 @@ public void SetExpiryNx()
...
+            // Test in-place update
+            ok = db.StringSet(key, newShorterValue, expireTime, When.NotExists, CommandFlags.None);
+            ClassicAssert.IsTrue(ok);
+
+            retValue = db.StringGet(key);
+            ClassicAssert.AreEqual(newShorterValue, retValue);
+        }
+
+        [Test]
+        public void SetAndGetExpiryNx()
```
badrishc, line 359: Very nice test! Can we have something similar for "Set If Exists" as well? (SetXx)

badrishc: any update on this?

Vijay-Nirmal: @badrishc Yes, I will. Bit busy with work today. I will try to do it today, if not max tomorrow morning.

Vijay-Nirmal: Done

badrishc (approved):

Issue #774, opened 2024-11-06
Create "dotnet global tool" packaging to make deployment trivial


mgravell:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

Right now, installing and running Garnet is a pain if what you want is just "give me a server"; the [getting started](https://microsoft.github.io/garnet/docs/getting-started) guide involves `git clone`, `dotnet restore` , `dotnet build`, `dotnet run` etc. Which means you need dev tools and a lot of steps - and you need to repeat all that to keep up-to-date. If you're a developer, great! But if you just want a server, this is massively overkill; you want something more like "install, (run command)"

Thought experiment; how about if, from any OS with .NET, it was as simple as:

``` txt
marc@mgx:~$ garnet-server
garnet-server: command not found
marc@mgx:~$ dotnet tool install --global garnet-server
Skipping NuGet package signature verification.
You can invoke the tool using the following command: garnet-server
Tool 'garnet-server' (version '1.0.35') was successfully installed.
marc@mgx:~$ garnet-server --port 6385
    _________
   /_||___||_\      Garnet 1.0.35 64 bit; standalone mode
   '. \   / .'      Port: 6385
     '.\ /.'        https://aka.ms/GetGarnet
       '.'

* Ready to accept connections
```

Well, it *can be* that simple, because: the above works! It also supports simple updates, etc; for example:

``` txt
marc@mgx:~$ dotnet tool update --global garnet-server
Skipping NuGet package signature verification.
Tool 'garnet-server' was reinstalled with the latest stable version (version '1.0.35').
```

### Describe the solution you'd like

So: I'm proposing that the `/main/GarnetServer` executable is packaged and uploaded to NuGet as a .NET global tool; this is not complicated; [here's how I did the above](https://github.com/StackExchange/StackExchange.Redis/tree/resp-cli/src/garnet-server). The main bits are just:

``` xml
<PackAsTool>true</PackAsTool>
<ToolCommandName>garnet-server</ToolCommandName>
```

(although other things like a readme, icon, etc are good practice)

Note:

I have uploaded [`garnet-server`](https://www.nuget.org/packages/garnet-server) for the purposes of this demonstration; I claim no ownership of this:

- if the Garnet team want the `garnet-server` package: you can have it
- if the Garnet team want to create a `Microsoft.Garnet.Server` or whatever package, and want me to nuke `garnet-server`, we can do that (I can just mark the versions as deprecated, which will also serve to prevent anyone else squatting it)

[...]

PR #779, opened 2024-11-07
package garnet-server as a .NET tool

mgravell:
fixes https://github.com/microsoft/garnet/issues/774

preview (to check package details etc) is [here](nuget.org/packages/garnet-server/1.0.36-preview1) - I've separately initiated a handover of that package to the `Garnet` account.

Usage:

image

``` txt
   15:54:40  GarnetServer net-tool  3   8.0.300   153ms 
➜ dotnet tool install --global garnet-server --version 1.0.36-preview1
You can invoke the tool using the following command: garnet-server
Tool 'garnet-server' (version '1.0.36-preview1') was successfully installed.
   15:54:45  GarnetServer net-tool  3   8.0.300   2.637s 
➜ garnet-server
    _________
   /_||___||_\      Garnet 1.0.36 64 bit; standalone mode
   '. \   / .'      Port: 6379
     '.\ /.'        https://aka.ms/GetGarnet
       '.'

* Ready to accept connections
```

### DIFF
--- a/main/GarnetServer/GarnetServer.csproj
+++ b/main/GarnetServer/GarnetServer.csproj
@@ -3,6 +3,29 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <ServerGarbageCollection>true</ServerGarbageCollection>
+
+    <!-- IMPORTANT: Keep the version in sync with .azure\pipelines\azure-pipelines-external-release.yml line ~6. -->
+    <Version>1.0.36</Version>
+    <PackageId>garnet-server</PackageId>
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>garnet-server</ToolCommandName>
+    <Title>The Microsoft.Garnet RESP server, packaged as a .NET tool</Title>
+    <GeneratePackageOnBuild Condition="'$(Configuration)'=='Release'">True</GeneratePackageOnBuild>
+    <Authors>Microsoft</Authors>
+    <Company>Microsoft</Company>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/microsoft/garnet.git</RepositoryUrl>
+    <PackageProjectUrl>https://microsoft.github.io/garnet</PackageProjectUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageRequireLicenseAcceptance>True</PackageRequireLicenseAcceptance>
+    <Description>
+      Garnet is a remote cache-store from Microsoft Research, that offers strong performance (throughput and latency),
+      scalability, storage, recovery, cluster sharding, key migration, and replication features. Garnet uses the Redis RESP wire
+      protocol and can work with existing Redis clients.
+    </Description>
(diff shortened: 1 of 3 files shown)


website/docs/welcome/releases.md
```diff
@@ -8,6 +8,16 @@ title: Releases
...
 Find releases at [https://github.com/microsoft/garnet/releases](https://github.com/microsoft/garnet/releases).
 
+## .NET Tool
+
+Garnet can be installed as a [.NET tool](https://aka.ms/global-tools):
+
+``` txt
+> dotnet tool install --global garnet-server
+>
+> garnet-server
```
Vijay-Nirmal, line 18: Just a minor request. Can [website/docs/getting-started/configuration.md](https://github.com/microsoft/garnet/blob/main/website/docs/getting-started/configuration.md) document updated to indicate `garnet-server` (global tools) can also be confirmed using either configuration file or command line arguments.

Currently, it only specifies `GarnetServer.exe`

main/GarnetServer/GarnetServer.csproj
```diff
@@ -3,6 +3,29 @@
...
+    <RepositoryUrl>https://github.com/microsoft/garnet.git</RepositoryUrl>
+    <PackageProjectUrl>https://microsoft.github.io/garnet</PackageProjectUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageRequireLicenseAcceptance>True</PackageRequireLicenseAcceptance>
+    <Description>
+      Garnet is a remote cache-store from Microsoft Research, that offers strong performance (throughput and latency),
+      scalability, storage, recovery, cluster sharding, key migration, and replication features. Garnet uses the Redis RESP wire
+      protocol and can work with existing Redis clients.
+    </Description>
+    <PackageReleaseNotes>See https://github.com/microsoft/garnet for details.</PackageReleaseNotes>
```
Vijay-Nirmal, line 26: I think we should link to `https://github.com/microsoft/garnet/releases`

badrishc: This is consistent with the current nuspec so ok to leave as it is. Users can always click through to the releases.

badrishc (approved):

website/docs/welcome/releases.md
badrishc: we can take care of these in a separate pass as well. ideally we want to rename GarnetServer.exe to garnet-server.exe so that we only have one concept overall and can just use the `garnet-server` terminology everywhere.

badrishc:
Here is where our CI pipeline does a NuGet pack of the .nuspec in order to create the nupkg, then sign it, and then push it to nuget.org --

https://github.com/microsoft/garnet/blob/main/.azure/pipelines/azure-pipelines-external-release.yml#L101

Push happens here: https://github.com/microsoft/garnet/blob/main/.azure/pipelines/azure-pipelines-external-release.yml#L191

Seems we need to add an additional step here to create this garnet-server tool. Assuming it generates a nupkg, the above pipeline should automatically sign and push to nuget.org, as it does this for all .nupkg files found in the build artifact directory.

badrishc:
Will merge this and we can make whatever changes we need to get this to push automatically, thanks!

Issue #778, opened 2024-11-07
AuthorizedAadApplicationIds input param value truncated


tarazou9:
### Describe the bug

When using auth mode ```AclWithAad``` and have config param ```--aad-authorized-app-ids``` takes multiple app ids, the value got incorrectly truncated after the first comma.

### Steps to reproduce the bug

1. Running the Garnet Server with ```--auth AclWithAad``` mode and configure the input param ```--aad-authorized-app-ids``` with value "app id 1, app id2, app id 3" (replace the app id with valid value here). 
2. Inspecting the value ```AuthorizedAadApplicationIds``` in ```Options.cs```, it is only able to parse "app id 1", everything else gets truncated.

### Expected behavior

All values correctly pass in.

### Release version

v1.0.35

PR #780, opened 2024-11-07
[Bug] Fix for AuthorizedAadApplicationIds input param value truncated

tarazou9:
- Fix for #778 
- Separator attribute should apply to IEnumerable target properties, otherwise won't parse the whole value correctly. See reference: https://github.com/commandlineparser/commandline/wiki/Option-Attribute

image


vazois (approved):
Looks good!

PR #783, opened 2024-11-07
AOF Replay using a Background Task at the Replica

vazois:
This PR separates AOF enqueue operation from replay. It contains the following

- [x] Implementation of an IBulkLogConsumer that performs the AOF replay at the replica side.
- [x] Adds throttling of enqueue operations when AOF replay lags beyond a specific amount (i.e. throttle if TailAddress - ReplicationOffset >ReplicationMaxLag)
- [x] Ensure re-initialization due to address skipping happens safely while replay consumer is reading from the log.

PS: Fixed a bug related to LRANGE which was incorrectly calling RMWObjectStoreOperationWithOutput instead of ReadObjectStoreOperationWithOutput


libs/storage/Tsavorite/cs/src/core/TsavoriteLog/TsavoriteLogIterator.cs
```diff
@@ -472,6 +470,20 @@ public unsafe bool TryBulkConsumeNext<T>(T consumer, int maxChunkSize = 0) where
             {
                 while (true)
                 {
+                    // If initializing wait for completion
+                    while (tsavoriteLog.Initializing)
+                    {
+                        Thread.Yield();
+                        try
```
badrishc, line 477: Why not just do epoch.ProtectAndDrain() instead of suspend-resume?

libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs
```diff
@@ -55,7 +55,7 @@ public void Dispose()
             cts?.Dispose();
         }
 
-        public unsafe void Consume(byte* payloadPtr, int payloadLength, long currentAddress, long nextAddress)
+        public unsafe void Consume(byte* payloadPtr, int payloadLength, long currentAddress, long nextAddress, bool isProtected)
```
badrishc, line 58: isProtected is no longer necessary here

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
```diff
@@ -48,83 +58,39 @@ public unsafe void ProcessPrimaryStream(byte* record, int recordLength, long pre
...
+                        recover: false,
+                        logger: logger);
 
-                if (ReplicationOffset != storeWrapper.appendOnlyFile.TailAddress)
-                {
-                    logger?.LogWarning("After ProcessPrimaryStream: Replication offset mismatch: ReplicaReplicationOffset {ReplicaReplicationOffset}, aof.TailAddress {tailAddress}", ReplicationOffset, storeWrapper.appendOnlyFile.TailAddress);
-                    throw new GarnetException($"After ProcessPrimaryStream: Replication offset mismatch: ReplicaReplicationOffset {ReplicationOffset}, aof.TailAddress {storeWrapper.appendOnlyFile.TailAddress}", LogLevel.Warning, clientResponse: false);
+                    Task.Run(ReplicaReplayTask);
                 }
+                //Consume(record, recordLength, currentAddress, nextAddress, isProtected: false);
```
badrishc, line 86: remove commented code

libs/host/Configuration/Options.cs
```diff
@@ -639,6 +643,7 @@ public GarnetServerOptions GetServerOptions(ILogger logger = null)
                 CheckpointThrottleFlushDelayMs = CheckpointThrottleFlushDelayMs,
                 EnableScatterGatherGet = EnableScatterGatherGet.GetValueOrDefault(),
                 ReplicaSyncDelayMs = ReplicaSyncDelayMs,
+                ReplicaMaxLag = ReplicaMaxLag,
```
badrishc, line 646: ReplicationOffsetMaxLag? Else it is not clear that the lag is expressed in terms of offset number.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -284,6 +284,11 @@ public class GarnetServerOptions : ServerOptions
         /// </summary>
         public int ReplicaSyncDelayMs = 5;
 
+        /// <summary>
+        /// Upper bound on the lag (i.e. throttle replicaAOF append if AOF.TailAddress - ReplicationOffset > ReplicaMaxLag) between primary and replica. -1 - Synchronous replay, >= 0 - background replay with specified lag 
+        /// </summary>
+        public int ReplicaMaxLag = 32768;
```
badrishc, line 290: this default looks too small and will definitely cause backup at primary every time a checkpoint is taken at the replica. why not set this to -1 as default, i.e. we do not gate the primary due to a slow replica.

vazois: Good point! This way we make async replay opt in.

libs/host/Configuration/Options.cs
```diff
@@ -342,6 +342,10 @@ internal sealed class Options
         [Option("replica-sync-delay", Required = false, HelpText = "Whether and by how much (milliseconds) should we throttle the replica sync: 0 - disable throttling")]
         public int ReplicaSyncDelayMs { get; set; }
 
+        [IntRangeValidation(-1, int.MaxValue)]
+        [Option("replica-max-lag", Required = false, HelpText = "Upper bound on the lag (i.e. throttle replicaAOF append if AOF.TailAddress - ReplicationOffset > ReplicaMaxLag) between primary and replica. -1 - Synchronous replay, >= 0 - background replay with specified lag")]
+        public int ReplicaMaxLag { get; set; }
```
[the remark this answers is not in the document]
vazois: Good idea! I added two fields
replication_offset_lag = AOF.TailAddress - ReplicationOffset ;
replication_offset_max_lag = ReplicaOffsetMaxLag

libs/host/Configuration/Options.cs
vazois: Renamed it to ReplicationOffsetMaxLag!

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
vazois: I removed the unused code.

libs/server/Servers/GarnetServerOptions.cs
badrishc: I think infinite lag should be the default. That's the right way to run the system instead of lock step.

vazois: Ok, made the following change:

0: Synchronous replay, >=1: background replay with specified lag, -1: infinite lag

libs/storage/Tsavorite/cs/src/core/TsavoriteLog/TsavoriteLogIterator.cs
vazois: Addressed this by using epoch.ProtectAndDrain()

libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs
vazois: Noting that this is the Primary AOF iteration. It does not use isProtected because it does not Acquire the epoch when consuming. The isProtected is used for the Consume part of the replica replay task.

badrishc (approved):

libs/cluster/Server/Replication/ReplicaOps/ReplicaReplayTask.cs
```diff
@@ -0,0 +1,114 @@
...
+                }
+                else if (payloadLength < 0)
+                {
+                    if (!clusterProvider.serverOptions.EnableFastCommit)
+                    {
+                        throw new GarnetException("Received FastCommit request at replica AOF processor, but FastCommit is not enabled", clientResponse: false);
+                    }
+                    TsavoriteLogRecoveryInfo info = new();
+                    info.Initialize(new ReadOnlySpan<byte>(ptr + entryLength, -payloadLength));
+                    //TODO: Verify again that this does not write into the AOF
```
badrishc, line 74: do we still need this TODO?

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
```diff
@@ -48,14 +59,23 @@ public unsafe void ProcessPrimaryStream(byte* record, int recordLength, long pre
...
+                            storeWrapper.appendOnlyFile.SafeInitialize(currentAddress, currentAddress);
                             ReplicationOffset = currentAddress;
                         }
                     }
                 }
 
-                // Address check
-                if (ReplicationOffset != storeWrapper.appendOnlyFile.TailAddress)
+                var tail = storeWrapper.appendOnlyFile.TailAddress;
+                var pageBits = storeWrapper.appendOnlyFile.UnsafeGetLogPageSizeBits();
```
badrishc, line 69: can this be done once instead of for every single call?

badrishc (changes requested):
see comments

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
```diff
@@ -48,14 +59,23 @@ public unsafe void ProcessPrimaryStream(byte* record, int recordLength, long pre
...
+                var nextPageBeginAddress = ((tail >> pageBits) + 1) << pageBits;
+                if (tail + recordLength > nextPageBeginAddress && nextPageBeginAddress != currentAddress)
```
badrishc, line 71: not sufficient, see whiteboard photo

badrishc, line 71: if tail + recordLength <= nextPageBeginAddress, we need to still verify that tail == currentAddress

badrishc: i.e. we need to first check the common case tail == currentAddress. if that is true then all good. if not, then we need the overflow based check.

libs/cluster/Server/Replication/ReplicaOps/ReplicaReplayTask.cs
vazois: no, this TODO was completed. I removed it

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
vazois: Done here: https://github.com/microsoft/garnet/pull/783/commits/26d30023c801121346eb738042fa3f507d66a264

libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs
vazois: I added a new check here https://github.com/microsoft/garnet/pull/783/commits/3fe5920d3dad3583cbe681cb129f76ccf6ad4e99.
Let me know if that looks good.

badrishc (approved):

PR #784, opened 2024-11-08
(AOF USERS BREAKING CHANGE) Provide a way to do backwards compat for AOF replay of custom proc and upserts from version 1.0.8

hamdaankhalid:
BREAKING CHANGE!

Provide a way to do backwards compat for AOF replay of custom proc and upserts from version 1.0.8

We are not guaranteed backwards compat for versions outside of 1.0.8 and are currently only adding support for upsert and custom procedure replay.

We are also not supporting backwards compat for Cluster mode.


libs/server/AOF/AofHeader.cs
```diff
@@ -5,16 +5,72 @@
...
+
+        // Used to map transactions to RespServerSessions that they came from
+        [FieldOffset(11)]
         public int sessionID;
+
+
+        // ctors are used for creating an instance when writing, having the field being set in the ctor will not affect the value when reading from disk
+        public AofHeader(AofEntryType opType, long version, int sessionID)
+        {
+            this._aofHeaderFormatVersion = ONLY_MSB_SET_MASK;
```
TalZaccai, line 59: This seems a bit confusing as the field needs to be set before the property, I'd probably move this initialization to the declaration.

libs/server/AOF/AofHeader.cs
```diff
@@ -5,16 +5,72 @@
...
         public long version;
-        [FieldOffset(10)]
+
+        // Used to map transactions to RespServerSessions that they came from
+        [FieldOffset(11)]
         public int sessionID;
+
+
+        // ctors are used for creating an instance when writing, having the field being set in the ctor will not affect the value when reading from disk
+        public AofHeader(AofEntryType opType, long version, int sessionID)
```
TalZaccai, line 57: XML comments for the ctors would be helpful here

libs/server/AOF/AofHeader.cs
```diff
@@ -5,16 +5,72 @@
...
     struct AofHeader
     {
+        // UPDATE THIS IF YOU CHANGE LAYOUT OF THIS STRUCT, this cannot exceed 127, as long as we are using MSB as heuristic to check for legacy format
+        const byte CURRENT_AOF_FORMAT_VERSION = 1;
+
+        /*
+        We use the MSB to differentiate if we are interacting with a legacy or the new AofHeader  
+        Note: When we are removing legacy support for old AOF log we can remove this and it's usage completely
+        */
+        const byte ONLY_MSB_SET_MASK = 0b1000_0000;
```
TalZaccai, line 18: 1 << 7 and 1 - (1 << 7) might be a tad more readable imo. Up to you.

libs/server/AOF/AofProcessor.cs
```diff
@@ -158,44 +167,64 @@ internal unsafe void ProcessAofRecord(IMemoryOwner<byte> entry, int length, bool
...
             entry.Dispose();
         }
 
         /// <summary>
-        /// Process AOF record
+        /// Checks if the MSB is not set of the first pointer, this indicates it is an older AOF format
         /// </summary>
+        /// <param name="ptr"></param>
+        /// <returns></returns>
+        private static unsafe bool IsLegacyFormat(byte* ptr) => (*ptr & MsbMask) == 0;
```
TalZaccai, line 189: Might make more sense to have this method in AofHeader, so that logic is contained there. Then you can reuse the constants you already set there.

libs/server/AOF/AofProcessor.cs
```diff
@@ -114,23 +121,23 @@ public void Dispose()
...
-        private unsafe void RecoverReplay(long untilAddress)
+        private unsafe bool RecoverReplay(long untilAddress)
         {
             logger?.LogInformation("Begin AOF replay");
             try
             {
                 int count = 0;
                 if (untilAddress == -1) untilAddress = storeWrapper.appendOnlyFile.TailAddress;
-                using var scan = storeWrapper.appendOnlyFile.Scan(storeWrapper.appendOnlyFile.BeginAddress, untilAddress);
+                using TsavoriteLogScanIterator scan = storeWrapper.appendOnlyFile.Scan(storeWrapper.appendOnlyFile.BeginAddress, untilAddress);
```
TalZaccai, line 138: Why change the vars here?

libs/server/AOF/AofProcessor.cs
```diff
@@ -158,44 +167,64 @@ internal unsafe void ProcessAofRecord(IMemoryOwner<byte> entry, int length, bool
         {
             fixed (byte* ptr = entry.Memory.Span)
             {
-                ProcessAofRecordInternal(ptr, length, asReplica);
+                if (replayFromLegacyAof && IsLegacyFormat(ptr))
```
TalZaccai, line 170: Should we return an error if replayFromLegacyAof is set but IsLegacyFormat is false?

libs/server/AOF/AofProcessor.cs
```diff
@@ -265,22 +310,70 @@ private unsafe bool ReplayOp(byte* entryPtr)
...
+            }
+            else
+            {
+                storeInput = new RawStringInput(RespCommand.SETEX, 0, expiration);
+            }
+
+            StoreUpsertPostParsing(basicContext, ref key, storeInput, ref value);
+        }
+
+        unsafe void RunStoredProc(byte id, CustomProcedureInput customProcInput, byte* ptr, bool isHistoricMode)
```
TalZaccai, line 364: Might want to stick to the "isLegacyFormat" naming

libs/server/InputHeader.cs
```diff
@@ -471,6 +473,29 @@ public unsafe int DeserializeFrom(byte* src)
 
             return len;
         }
+
+        public unsafe void DeserializeFromHistoricAof(byte* src)
```
TalZaccai, line 477: Rename to DeserializeFromLegacyAof, for naming consistency

libs/server/Resp/Parser/SessionParseState.cs
```diff
@@ -167,11 +168,11 @@ public void InitializeWithArguments(ArgSlice arg1, ArgSlice arg2, ArgSlice arg3,
         /// </summary>
         /// <param name="args">Set of arguments to initialize buffer with</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void InitializeWithArguments(ArgSlice[] args)
+        public void InitializeWithArguments(IList<ArgSlice> args)
```
TalZaccai, line 171: This change incurs overhead when this is called with an array. Please just call ToArray() on your list.

TalZaccai (commented):
Nice work! Added some small comments...

libs/server/StoreWrapper.cs
```diff
@@ -287,27 +304,28 @@ public void Reset()
         /// <summary>
         /// When replaying AOF we do not want to write AOF records again.
         /// </summary>
-        public long ReplayAOF(long untilAddress = -1)
+        public (long, bool) ReplayAOF(long untilAddress = -1)
```
TalZaccai, line 307: I would personally prefer ```public bool TryReplayAOF(out long offset, long untilAddress = -1)```, as .Item1, .Item2 aren't very self-explanatory...

libs/server/AOF/AofProcessor.cs
hamdaankhalid: type wasnt explicitly understood looking at the right side, figured it helps with readbility when you can tell the type from looking the right side.

libs/server/AOF/AofProcessor.cs
hamdaankhalid: this is on purpose, because this handles a server that was started in "replayFromLegacyMode" but then starts reading new aof format logs.
You dont need to worry about the log being new and having server in this backwards compat mode.

libs/server/AOF/AofProcessor.cs
hamdaankhalid: agreed

libs/server/AOF/AofHeader.cs
hamdaankhalid: keeping it as binary since one byte in binary feels more legible like this to me atm.

libs/server/Resp/Parser/SessionParseState.cs
hamdaankhalid: @badrishc  Tal and I spoke about this and would like your opinion on this.

My understanding is that this will be same cost as a lookup call on  a concrete class that we were passing before.

We dont have to go through the vtable lookup chain because of this section of the CLR book:

https://sakupinera.github.io/2022/11/04/csharp/clr-via-csharp/Chapter%2013%20Interfaces/

I may be wrong but if the type is there in my method table on the first look, is that not same cost as calling a method on a concrete class for the first time?

hamdaankhalid: Would the below version then be better?
`       [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public void InitializeWithArguments<T>(T args) where T : IList<ArgSlice>
        {
            Initialize(args.Count);

            for (var i = 0; i < args.Count; i++)
            {
                *(bufferPtr + i) = args[i];
            }
        }`

hamdaankhalid: Tal is confirmed, I decompiled to IL and I can confirm it always uses callvirt with instance

TalZaccai (approved):

libs/server/AOF/AofHeader.cs
```diff
@@ -5,16 +5,90 @@
...
+            set
+            {
+                // ignore the MSB and let the user only set the first 7 bits before the MSB
+                _aofHeaderFormatVersion = (byte)((_aofHeaderFormatVersion & ONLY_MSB_SET_MASK) | (value & MSB_IGNORED_BYTE_MASK));
+            }
+        }
+
         [FieldOffset(0)]
-        public AofEntryType opType;
+        byte _aofHeaderFormatVersion = ONLY_MSB_SET_MASK;
```
badrishc, line 38: Don't set Fields by default

TalZaccai:
This is going to be merged in the consumer's own repo.

PR #803, opened 2024-11-15
Bugfix in RespReadUtils

TalZaccai:
RespReadUtils.ReadSignedLengthHeader was throwing an overflow exception when parsing an int.MinValue, this caused BDN.benchmark to throw when running RespToInteger.

### DIFF
--- a/libs/common/RespReadUtils.cs
+++ b/libs/common/RespReadUtils.cs
@@ -378,7 +378,7 @@ namespace Garnet.common
             }
 
             // Validate length
-            if (value > int.MaxValue)
+            if (value > int.MaxValue + (ulong)(negative ? 1 : 0)) // int.MinValue = -int.MaxValue + 1
             {
                 RespParsingException.ThrowIntegerOverflow(readHead - digitsRead, (int)digitsRead);
             }


libs/common/RespReadUtils.cs
```diff
@@ -378,7 +378,7 @@ public static bool ReadSignedLengthHeader(out int length, ref byte* ptr, byte* e
             }
 
             // Validate length
-            if (value > int.MaxValue)
+            if (value > int.MaxValue + (ulong)(negative ? 1 : 0)) // int.MinValue = -int.MaxValue + 1
```
badrishc, line 381: maybe put the negative check insde the first `if` so that we do not always pay for the negative check

libs/common/RespReadUtils.cs
```diff
@@ -378,7 +378,7 @@ public static bool ReadSignedLengthHeader(out int length, ref byte* ptr, byte* e
             }
 
             // Validate length
-            if (value > int.MaxValue)
+            if (value > int.MaxValue && (!negative || value > int.MaxValue + (ulong)1)) // int.MinValue = -int.MaxValue + 1
```
badrishc, line 381: Comment should be -(int.MaxValue + 1) ?

TalZaccai: Doh! Thanks :)

badrishc (approved):

PR #807, opened 2024-11-15
Show Test FullName

vazois:
This PR updates the ci files to output the FullName of the test fixture that is being executed running Garnet tests.


darrenge (approved):
I think this will be ok. If people find it too cumbersome to look at, we can easily roll it back.

PR #824, opened 2024-11-23
Streaming snapshot checkpoint in Tsavorite

badrishc:
`StreamingSnapshot` is a new checkpoint type that yields a stream of (mostly but not all distinct) key-value pairs that correspond to a consistent snapshot (version `v`) of the database. This can be used to create a checkpoint of the user's preferred format, or directly streamed to replicas, etc. To recover, the user needs to simply iterate over this stream in order, performing a sequence of upserts, and this will restore the consistent `v` version (state) of the database.

Treating this iterator as a checkpoint allows users to pair it with AOF replay to get exactly once processing semantics (TBD).

Current implementation makes the mutable region temporarily read-only, similar to a snapshot checkpoint. However, this can be optimized to avoid the copy-on-write in future, by making the update threads write out the old version `v` before updating it to `v+1` in place.


libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/FindRecord.cs
```diff
@@ -130,6 +142,26 @@ private bool TraceBackForKeyMatch(ref TKey key, ref RecordSource<TKey, TValue, T
...
+            recSrc.LogicalAddress = recordInfo.PreviousAddress;
+            if (TraceBackForKeyMatch(ref key, recSrc.LogicalAddress, minAddress, maxAddress, out recSrc.LogicalAddress, out recSrc.PhysicalAddress))
+            {
+                recSrc.SetHasMainLogSrc();
+                return true;
+            }
+            return false;
+        }
+
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
TedHartMS, line 165: The duplication saves an "if" but makes the code more confusing. Add some begin/end comments identifying the clones so it's easier to visually parse.

TedHartMS (approved):

vazois (approved):

PR #826, opened 2024-11-24
[Compatibility] Added ZRANGESTORE command

Vijay-Nirmal:
Adding the ZRANGESTORE command to garnet

- [x] Add ZRANGESTORE command
- [x] Add documentation


libs/server/API/IGarnetApi.cs
```diff
@@ -355,6 +355,16 @@ public interface IGarnetApi : IGarnetReadApi, IGarnetAdvancedApi
...
 
+        /// <summary>
+        /// Stores a range of sorted set elements in the specified key space.
+        /// </summary>
+        /// <param name="distKey">The distribution key for the sorted set.</param>
+        /// <param name="sbKey">The sub-key for the sorted set.</param>
+        /// <param name="input">The input object containing the elements to store.</param>
+        /// <param name="result">The result of the store operation.</param>
+        /// <returns>A <see cref="GarnetStatus"/> indicating the status of the operation.</returns>
+        GarnetStatus SortedSetRangeStore(ArgSlice distKey, ArgSlice sbKey, ref ObjectInput input, out int result);
```
TalZaccai, line 366: Let's change `distKey` and `sbKey` to `dstKey` and `srcKey` for clarity (in all occurrences)

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -692,6 +692,108 @@ public GarnetStatus SortedSetAdd<TObjectContext>(byte[] key, ref ObjectInput inp
...
+            }
+
+            // SetObject
+            var objectStoreLockableContext = txnManager.ObjectStoreLockableContext;
+
+            try
+            {
+                SpanByteAndMemory rangeOutputMem = default;
+                var rangeOutput = new GarnetObjectStoreOutput() { spanByteAndMemory = rangeOutputMem };
+                var status = SortedSetRange(sbKey.ToArray(), ref input, ref rangeOutput, ref objectStoreLockableContext);
```
TalZaccai, line 731: I wonder if it would be worthwhile performance-wise to do a GET and do the range logic here instead of parsing the output. You could probably write a method in SortedSetObject that takes an ObjectInput and returns an IEnumerable<double, byte[]>, and use it in both implementations (ZRANGE & ZRANGESCORE). WDYT?

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -1086,6 +1086,41 @@ public void CheckSortedSetDifferenceStoreWithNoMatchSE()
             ClassicAssert.AreEqual(0, actualMembers.Length);
         }
 
+        [Test]
+        [TestCase("user1:obj1", "user1:objA", new[] { "Hello", "World" }, new[] { 1.0, 2.0 }, new[] { "Hello", "World" }, new[] { 1.0, 2.0 })] // Normal case
+        [TestCase("user1:emptySet", "user1:objB", new string[] { }, new double[] { }, new string[] { }, new double[] { })] // Empty set
+        [TestCase("user1:nonExistingKey", "user1:objC", new string[] { }, new double[] { }, new string[] { }, new double[] { })] // Non-existing key
+        [TestCase("user1:obj2", "user1:objD", new[] { "Alpha", "Beta", "Gamma" }, new[] { 1.0, 2.0, 3.0 }, new[] { "Beta", "Gamma" }, new[] { 2.0, 3.0 }, -2, -1)] // Negative range
+        public void CheckSortedSetRangeStoreSE(string key, string destinationKey, string[] elements, double[] scores, string[] expectedElements, double[] expectedScores, int start = 0, int stop = -1)
```
TalZaccai, line 1094: Please add a case where destKey is not empty

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -1086,6 +1086,41 @@ public void CheckSortedSetDifferenceStoreWithNoMatchSE()
...
+        {
+            int expectedCount = expectedElements.Length;
+
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+
+            var redisKey = new RedisKey(key);
```
TalZaccai, line 1101: nit: no need for these, you can call SERedis API directly with a string key

test/Garnet.test/RespSortedSetTests.cs
TalZaccai, line 1094: We should also do a sanity check for the different options ([BYSCORE | BYLEX] [REV] [LIMIT offset count])

TalZaccai (changes requested):
Very nice implementation! Please see comments.

libs/server/API/IGarnetApi.cs
Vijay-Nirmal: Done

test/Garnet.test/RespSortedSetTests.cs
Vijay-Nirmal: Done

test/Garnet.test/RespSortedSetTests.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
badrishc: Performance is of higher priority than code sharing, IMHO.

test/Garnet.test/RespSortedSetTests.cs
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: @badrishc I agree. Off-topic, sorry to bring it up here. I have a question for you here https://github.com/microsoft/garnet/pull/674#issuecomment-2515441660

TalZaccai: You're correct, we shouldn't incur any performance cost on ZRANGE. It would still be nice to at least make some portion of it common, if possible, since it's a rather hefty chunk of code, even if just the parsing / validation of options.

Vijay-Nirmal: Let me think about it, currently, I am not sure how I can do that

Vijay-Nirmal: @TalZaccai I am not sure we can do this.

> You could probably write a method in SortedSetObject that takes an ObjectInput and returns an IEnumerable<double, byte[]>

We can return `SpanByteAndMemory` from the SortedSetObject class, right? Then how can we return anything like `IEnumerable<double, byte[]>`. Maybe I didn't understand the review command.

Vijay-Nirmal:
@TalZaccai Please rereview it, I am not sure about the one open command

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
TalZaccai: Not sure I follow you... What I meant was that we could do something like:
```
-- begin transaction
srcSortedSetObj = GET(srcKey)
(items, scores) = srcSortedSetObj.GetRange(input) // new method
dstSortedSetObj = GET(dstKey)
dstSortedSetObj.Add((items, scores))
-- end transaction
```

TalZaccai: It's fine as-is, I was just wondering if we could squeeze a bit better perf here. No worries if you want to skip it for now. LMK.

Vijay-Nirmal: Ohh now only I understood your intent. Never thought we were allowed to write public methods in SortedSetObject (any Object types) and call it directly. If I have to write the new `GetRange(input)` method then most of the code with the current `SortedSetRange` method will be duplicated. I can't see a way to have a common code without allocating a new object or many dirty condition checks to check if we need to write the output to SpanByteAndMemory or a new object.

If you are fine with having lots of duplicate code then I can do it. Or if you have an idea to have a common code without affecting the performance of ZRANGE then let me know as well.

TalZaccai: I didn't mean that it would replace SortedSetRange, but that SortedSetRange would also use that helper method, and it could contain only part of the logic there (such as parsing) to avoid code duplication. I'll just let this in as-is and we can look into improving perf on this at a later stage.

TalZaccai (approved):

PR #828, opened 2024-11-25
Limit Client Connections

vazois:
This PR exposes a parameter to allow for limiting the number of  client connections

### DIFF
--- a/libs/host/Configuration/Options.cs
+++ b/libs/host/Configuration/Options.cs
@@ -339,9 +339,9 @@ namespace Garnet
         [Option("maxthreads", Required = false, HelpText = "Maximum worker and completion threads in thread pool, 0 uses the system default.")]
         public int ThreadPoolMaxThreads { get; set; }
 
-        [IntRangeValidation(0, int.MaxValue)]
-        [Option("connection-limit", Required = false, Default = 0, HelpText = "Maximum number of client connections")]
-        public int ConnectionLimit { get; set; }
+        [IntRangeValidation(-1, int.MaxValue)]
+        [Option("network-connection-limit", Required = false, Default = -1, HelpText = "Maximum number of simultaneously active network connections.")]
+        public int NetworkConnectionLimit { get; set; }
 
         [OptionValidation]
         [Option("use-azure-storage", Required = false, HelpText = "Use Azure Page Blobs for storage instead of local storage.")]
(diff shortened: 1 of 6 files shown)


libs/host/Configuration/Options.cs
```diff
@@ -339,6 +339,10 @@ internal sealed class Options
         [Option("maxthreads", Required = false, HelpText = "Maximum worker and completion threads in thread pool, 0 uses the system default.")]
         public int ThreadPoolMaxThreads { get; set; }
 
+        [IntRangeValidation(0, int.MaxValue)]
+        [Option("connection-limit", Required = false, Default = 0, HelpText = "Maximum number of client connections")]
+        public int ConnectionLimit { get; set; }
```
msft-paddy14, line 344: will this be inclusive of Gossip connections incase of cluster mode?

vazois: yes, you have to account for the internode connections, which means gossip, replication, failover and migration connections. The failover and migration are transient connections until its operation finishes. Gossip and replication and long-lasting connections.

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -134,11 +136,11 @@ private unsafe bool HandleNewConnection(SocketAsyncEventArgs e)
             string remoteEndpointName = e.AcceptSocket.RemoteEndPoint?.ToString();
             logger?.LogDebug("Accepted TCP connection from {remoteEndpoint}", remoteEndpointName);
 
-
             ServerTcpNetworkHandler handler = null;
             if (activeHandlerCount >= 0)
             {
-                if (Interlocked.Increment(ref activeHandlerCount) > 0)
+                var currentActiveHandlerCount = Interlocked.Increment(ref activeHandlerCount);
+                if (currentActiveHandlerCount is > 0 && (connectionLimit == 0 || currentActiveHandlerCount <= connectionLimit))
```
badrishc, line 143: no need for `is` for integers.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -246,6 +246,11 @@ public class GarnetServerOptions : ServerOptions
         /// </summary>
         public int ThreadPoolMaxThreads = 0;
 
+        /// <summary>
+        /// Maximum client connection limit
+        /// </summary>
+        public int ConnectionLimit = 0;
```
badrishc, line 252: We should let `0` denote "no connections allowed" which might be useful if for some reason people want to run the server without allowing any network connections. `-1` or `int.MaxValue` can be the default to allow unlimited connections.

libs/host/Configuration/Options.cs
```diff
@@ -339,6 +339,10 @@ internal sealed class Options
         [Option("maxthreads", Required = false, HelpText = "Maximum worker and completion threads in thread pool, 0 uses the system default.")]
         public int ThreadPoolMaxThreads { get; set; }
 
+        [IntRangeValidation(0, int.MaxValue)]
+        [Option("connection-limit", Required = false, Default = 0, HelpText = "Maximum number of client connections")]
```
badrishc, line 343: Reword: `Maximum number of simultaneously active client connections.` (note also: added period at the end)

badrishc, line 343: Rename to `network-connection-limit` and `NetworkConnectionLimit` to clarify this does not apply to API-based session accesses.

libs/host/defaults.conf
```diff
@@ -251,6 +251,9 @@
 	/* Maximum worker and completion threads in thread pool, 0 uses the system default. */
 	"ThreadPoolMaxThreads" : 0,
 
+	/* Maximum client connection limit. */
```
badrishc, line 254: Adjust to be consistent with the other two comment locations.

badrishc (approved):

PR #829, opened 2024-11-25
Conditionally allocate CustomCommandManager to reduce overhead

msft-paddy14:
Currently, even if CustomCommands are not required, a CustomCommandManager session is silently created with every RespSession and adds an overhead of ~1MB per connection. For ~10k connections, this overhead can be huge upto 10GB. Not trying to address the overhead amount here but disabling it for scenarios which don't even use CustomCommands.

From memory dump
image

Fixes https://github.com/microsoft/garnet/issues/834


libs/server/Resp/RespServerSession.cs
```diff
@@ -185,7 +185,7 @@ public RespServerSession(
             bool enableScripts)
             : base(networkSender)
         {
-            this.customCommandManagerSession = new CustomCommandManagerSession(storeWrapper.customCommandManager);
+            this.customCommandManagerSession = storeWrapper.serverOptions.LoadModuleCS != null ? new CustomCommandManagerSession(storeWrapper.customCommandManager) : null;
```
vazois, line 188: Good catch! However, you have to update the unit tests also to enable LoadModuleCS because the ones using modules are breaking.

vazois (changes requested):

badrishc:
The bigger issue is that this array should not have gotten so big to begin with. We are looking into this. It's a recent regression when commands moved from byte to ushort.

msft-paddy14:
@badrishc agreed regarding overhead. This PR only focuses on eliminating the allocations for the paths that don't need it (say servers that will never use these).

vazois (approved):

libs/server/Resp/RespServerSession.cs
badrishc: I dont think we want to start using Lazy in the codebase just for this. we have fixed the underlying cause of the large overhead, so the array should no longer be a bottleneck.

PR #830, opened 2024-11-25
Fix race condition in MainMonitor

msft-paddy14:
There's a race b/w the MainMonitor task which tries to reset the latency and connection dispose path which returns the latency object after setting it to null. Under high connection dispose rate, this issue becomes visible and throws below error. Adding a  lock to protect dispose. Ideally there should be handled with all public methods but based on current model only alternate code path is metrics thread, so albeit hacky, this is a solution preventing the race.
image


libs/server/Metrics/Latency/GarnetLatencyMetricsSession.cs
```diff
@@ -80,7 +84,12 @@ public void ResetAll()
         public void Reset(LatencyMetricsType cmd)
         {
             int idx = (int)cmd;
-            metrics[idx].latency[PriorVersion].Reset();
+            disposeLock.WriteLock();
+            if(metrics != null)
```
vazois, line 88: Minor issue with formatting.
```suggestion
            if (metrics != null)
```

vazois (approved):
Please correct minor formatting issue

vazois (changes requested):
Please correct minor formatting issue

libs/server/Metrics/Latency/GarnetLatencyMetricsSession.cs
```diff
@@ -80,7 +84,12 @@ public void ResetAll()
         public void Reset(LatencyMetricsType cmd)
         {
             int idx = (int)cmd;
-            metrics[idx].latency[PriorVersion].Reset();
+            disposeLock.WriteLock();
```
vazois, line 87: Add try finally here to ensure WriteUnlock is always called

libs/server/Metrics/Latency/GarnetLatencyMetricsSession.cs
```diff
@@ -30,7 +32,9 @@ public void Return()
             {
                 metrics[(int)cmd].Return();
             }
+            disposeLock.WriteLock();
```
vazois, line 35: Same here, add a try finally

vazois (changes requested):

PR #832, opened 2024-11-26
send migrate node to other primaries

vazois:
This PR fixes migrate slot walk test by adding a call to SetSlot during failed IMPORT call.
In addition, exposes a parameter that allows users to disable the epoch collision resolution algorithm.


libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -62,6 +62,11 @@ public class GarnetServerOptions : ServerOptions
         /// </summary>
         public bool EnableCluster = false;
 
+        /// <summary>
+        /// Disable epoch collision resolution algorithm.
+        /// </summary>
+        public bool DisableEpochCollision = false;
```
badrishc, line 68: this is strange to expose as option. it should be stable in main.

vazois: I reverted this.

badrishc (approved):

PR #837, opened 2024-11-27
[Compatibility] Added SINTERCARD command

Vijay-Nirmal:
Adding the SINTERCARD commands to garnet

- [x] Add SINTERCARD commands
- [x] Add documentation

### DIFF
--- a/libs/server/Resp/Objects/SetCommands.cs
+++ b/libs/server/Resp/Objects/SetCommands.cs
@@ -192,7 +192,7 @@ namespace Garnet.server
             if (parseState.Count > nKeys + 1)
             {
                 var limitArg = parseState.GetArgSliceByRef(nKeys + 1);
-                if (!limitArg.ReadOnlySpan.SequenceEqual(CmdStrings.LIMIT))
+                if (!limitArg.ReadOnlySpan.EqualsUpperCaseSpanIgnoringCase(CmdStrings.LIMIT))
                 {
                     while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
                         SendAndReset();
--- a/libs/server/Resp/Objects/SetCommands.cs
+++ b/libs/server/Resp/Objects/SetCommands.cs
@@ -151,6 +151,79 @@ namespace Garnet.server
             return true;
         }
 
+        /// <summary>
+        /// Returns the cardinality of the intersection between multiple sets.
+        /// </summary>
+        /// <param name="storageApi"></param>
+        /// <typeparam name="TGarnetApi"></typeparam>
+        /// <returns></returns>
+        private bool SetIntersectCard<TGarnetApi>(ref TGarnetApi storageApi)
+            where TGarnetApi : IGarnetApi
+        {
+            if (parseState.Count < 2) // Need at least numkeys + 1 key
+            {
+                return AbortWithWrongNumberOfArguments(nameof(RespCommand.SINTERCARD));
+            }
+
+            // Number of keys
+            if (!parseState.TryGetInt(0, out var nKeys))
+            {
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            var keys = new ArgSlice[nKeys];
+            int? limit = null;
+
+            if ((parseState.Count - 1) < nKeys)
+            {
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            for (var i = 1; i <= nKeys; i++)
+            {
+                keys[i - 1] = parseState.GetArgSliceByRef(i);
+            }
+
+            // Optional LIMIT argument
+            if (parseState.Count > nKeys + 1)
+            {
+                var limitArg = parseState.GetArgSliceByRef(nKeys + 1);
+                if (!limitArg.ReadOnlySpan.SequenceEqual(CmdStrings.LIMIT))
+                {
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+                }
+
+                if (parseState.Count != nKeys + 3 || !parseState.TryGetInt(nKeys + 2, out var limitVal))
+                {
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+                }
+
+                limit = limitVal;
+            }
+
+            var status = storageApi.SetIntersectCard(keys, limit, out var result);
+            switch (status)
+            {
+                case GarnetStatus.OK:
+                    while (!RespWriteUtils.WriteInteger(result, ref dcurr, dend))
+                        SendAndReset();
+                    break;
+                case GarnetStatus.WRONGTYPE:
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_WRONG_TYPE, ref dcurr, dend))
+                        SendAndReset();
+                    break;
+            }
+
+            return true;
+        }
 
         /// <summary>
         /// Returns the members of the set resulting from the union of all the given sets.
--- a/libs/server/Resp/Objects/SetCommands.cs
+++ b/libs/server/Resp/Objects/SetCommands.cs
@@ -177,7 +177,7 @@ namespace Garnet.server
                 return AbortWithErrorMessage(Encoding.ASCII.GetBytes(string.Format(CmdStrings.GenericErrShouldBeGreaterThanZero, "numkeys")));
             }
 
-            var keys = parseState.Parameters.Slice(1);
+            var keys = parseState.Parameters.Slice(1, nKeys);
 
             // Optional LIMIT argument
             int? limit = null;
--- a/libs/server/Resp/Objects/SetCommands.cs
+++ b/libs/server/Resp/Objects/SetCommands.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Text;
 using Garnet.common;
 using Tsavorite.core;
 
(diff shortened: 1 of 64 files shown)


libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -923,5 +923,48 @@ private GarnetStatus SetDiff<TObjectContext>(ArgSlice[] keys, ref TObjectContext
...
+                foreach (var item in keys)
+                    txnManager.SaveKeyEntryToLock(item, true, LockType.Shared);
+                _ = txnManager.Run(true);
+            }
+
+            var setObjectStoreLockableContext = txnManager.ObjectStoreLockableContext;
+
+            try
+            {
+                var status = SetIntersect(keys, ref setObjectStoreLockableContext, out var result);
```
vazois, line 956: Is it possible to break off the calculation early if the limit is reached?
The command description states the following:
"By default, the command calculates the cardinality of the intersection of all given sets. When provided with the optional LIMIT argument (which defaults to 0 and means unlimited), if the intersection cardinality reaches limit partway through the computation, the algorithm will exit and yield limit as the cardinality. Such implementation ensures a significant speedup for queries where the limit is lower than the actual intersection cardinality."

I think it is good to have this optimization

website/docs/commands/data-structures.md
```diff
@@ -674,6 +674,19 @@ If **destination** already exists, it is overwritten.
 
 ---
 
+### SINTERCARD
+
+#### Syntax
+
+```bash
+    SINTERCARD key [key ...]
```
vazois, line 682: Add LIMIT option also

website/docs/commands/data-structures.md
```diff
@@ -674,6 +674,19 @@ If **destination** already exists, it is overwritten.
...
 
+### SINTERCARD
+
+#### Syntax
+
+```bash
+    SINTERCARD key [key ...]
+```
+
+Returns the number of members in the resulting set from the intersection of all the given sets.
```
vazois, line 685: You may need to update this when implementing the above optimization with LIMIT

vazois (changes requested):

libs/server/Storage/Session/ObjectStore/SetOps.cs
Vijay-Nirmal: Not sure how efferent it will be because we have to load the entire set into memory (if on disk) and iterate over all the list as once. With this approach we will be compromising memory for CPU time. Both approach has pros and cons. To my understand (I might be wrong here), Redis doesn't have to load the entire dataset into the memory to operate on it. So limit option make scene for Redis but not sure what is the right approach for Garnet since we have to load the entire dataset on memory to operate on it.

I am open to do anything, once you config it again. Also, If we need to do this then I will create a new method to do it without modifying the SetIntersect method (but if we load every key to memory then SetIntersect can also be faster because we can avoid multiple iteration over the list).

Another way is to just respect the LIMIT option only for last list iteration, which can be done with the existing code but not sure much but benefit we will get but I am sure we will get some benefit for edge cases. If you link to do it in this manner let me know, I can do this as well.

Once you confirm the approach I will implement this review comment

Vijay-Nirmal: > It should be a simple check after each output.IntersectWith call

We can't right? Let's say we have 10 keys in the input to find the intersection then we can only exist early in the loop for the 10 key alone right? Unless you really meant that. That is what I said in my last para :)

vazois: > > It should be a simple check after each output.IntersectWith call
> 
> We can't right? Let's say we have 10 keys in the input to find the intersection then we can only exist early in the loop for the 10 key alone right? Unless you really meant that. That is what I said in my last para :)

Yes, from what I understand the semantics are as long as you found 10 members in return early. It can happen if first key is read (e.g. it has  >= 10 members) or any number of reads after that.
I think the only issue I see it if you issue SINTERCARD 3 key1 key2 key3 LIMIT 1 and you stop any time before key3 and key3 is a list object then redis expects WRONTYPE response. So you would have to read everything to ensure this, though we can be optimistic and assume until WRONTYPE is triggered, it is fine. This should be ok since SINTERSCARD is readonly.

Vijay-Nirmal: I am confused again, How can we find no of intersection among multiple sets without reading all the sets. If the input is `SINTERCARD 3 key1 key2 key3 LIMIT 1` then even if key1 has 10 members we must ready all keys and find one member that is common across all the 3 sets. So we can't return early in any case without reading key3, we can only return early after we find a common member with there prior 2 sets.

Since there is .net's IntersectWith method doesn't take a parameter to stop the intersect process early, we have to write our own IntersectWith method that access limit method. Also, we can only use the limit param for the last key, for other keys we still have to full IntersectWith.

vazois: I think I got the semantics wrong. I guess to get the best performance you have to implement a custom intersect that sorts each set does an ordered merge and breaks early. But this will increase complexity, I am not sure this w
One optimization that I recommend here is if you discover an empty set, you can break early because the result will be empty. This optimization applies for all SINTER variants.

I think the latter should be easy to add. The former is probably not part of this PR so let's leave it out of it.

Vijay-Nirmal: > One optimization that I recommend here is if you discover an empty set, you can break early because the result will be empty. This optimization applies for all SINTER variants.

Sure, will do this part

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
             return true;
         }
 
+        /// <summary>
+        /// Returns the cardinality of the intersection between multiple sets.
+        /// </summary>
+        /// <param name="storageApi"></param>
+        /// <typeparam name="TGarnetApi"></typeparam>
+        /// <returns></returns>
+        private bool SetIntersectCard<TGarnetApi>(ref TGarnetApi storageApi)
```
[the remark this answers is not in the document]
Vijay-Nirmal: Done

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
...
+            for (var i = 1; i <= nKeys; i++)
+            {
+                keys[i - 1] = parseState.GetArgSliceByRef(i);
+            }
+
+            // Optional LIMIT argument
+            if (parseState.Count > nKeys + 1)
+            {
+                var limitArg = parseState.GetArgSliceByRef(nKeys + 1);
+                if (!limitArg.ReadOnlySpan.SequenceEqual(CmdStrings.LIMIT))
```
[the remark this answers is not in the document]
Vijay-Nirmal: Done

libs/server/Storage/Session/ObjectStore/SetOps.cs
Vijay-Nirmal: There is already logic to returning early if the set is empty (I assume GET will return GarnetStatus.NOTFOUND when the set is empty or does not exist)

image

website/docs/commands/data-structures.md
Vijay-Nirmal: Done

website/docs/commands/data-structures.md
Vijay-Nirmal: Done

vazois (approved):

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
...
+            {
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            var keys = new ArgSlice[nKeys];
+            int? limit = null;
+
+            if ((parseState.Count - 1) < nKeys)
```
TalZaccai, line 179: If `nKeys <= 0` we need to return `ERR numkeys should be greater than 0`

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
...
+            {
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            for (var i = 1; i <= nKeys; i++)
```
TalZaccai, line 186: Why not use parseState.Parameters.Slice as in your previous PRs?

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
...
+            {
+                var limitArg = parseState.GetArgSliceByRef(nKeys + 1);
+                if (!limitArg.ReadOnlySpan.EqualsUpperCaseSpanIgnoringCase(CmdStrings.LIMIT))
+                {
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+                }
+
+                if (parseState.Count != nKeys + 3 || !parseState.TryGetInt(nKeys + 2, out var limitVal))
```
TalZaccai, line 202: Need to move `parseState.Count != nKeys + 3` to the previous check, as this should result in a syntax error

libs/server/Resp/Objects/SetCommands.cs
```diff
@@ -151,6 +151,79 @@ private bool SetIntersectStore<TGarnetApi>(ref TGarnetApi storageApi)
...
+                {
+                    while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+                }
+
+                limit = limitVal;
```
TalZaccai, line 209: Need to check that limit is non-negative. Error should be `ERR LIMIT can't be negative`.

TalZaccai (commented):
Couple of small comments from me, looks good otherwise

Vijay-Nirmal:
@TalZaccai Fixed review commands

PR #846, opened 2024-12-02
docs: Link defaults.conf from Configuration article

mloskot:
Let's utilise the awesome feature of hyperlinking, especially, that the [Configurable Settings](https://github.com/microsoft/garnet/blob/55f94e3ea19c9204ed9154db40bf93dfa8d4f7ce/website/docs/getting-started/configuration.md?plain=1#L68) table does not offer displaying of the default values.

### DIFF
--- a/website/docs/getting-started/configuration.md
+++ b/website/docs/getting-started/configuration.md
@@ -7,7 +7,7 @@ title: Garnet Configuration
 ## How to Configure Garnet
 
 The Garnet server (GarnetServer.exe) can be configured using a configuration file (e.g. `garnet.conf` or `redis.conf`), while command line arguments can be used to override any settings specified in the file. /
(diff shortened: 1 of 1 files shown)


TalZaccai:
> Let's utilise the awesome feature of hyperlinking, especially, that the [Configurable Settings](https://github.com/microsoft/garnet/blob/55f94e3ea19c9204ed9154db40bf93dfa8d4f7ce/website/docs/getting-started/configuration.md?plain=1#L68) table does not offer displaying of the default values.

This link would need to work from our github.io [website](https://microsoft.github.io/garnet/docs/getting-started/configuration), so we should probably opt for a full URL rather than a relative path.

mloskot:
@TalZaccai 
> we should probably opt for a full URL rather than a relative path.

Good point. Done.

TalZaccai (approved):

PR #854, opened 2024-12-04
Add Epoch Protection Documentation

hamdaankhalid:
Epoch Protection is hard to understand at first glance, and every time I look at code using it, I need to build the understanding again for it.

It has also diverged with new public methods from what is presented in Faster Whitepaper. Adding this documentation to help others and myself understand it and serve as a refresher.

I have tried my best to explain Epoch Protection after discussions with Ted and Badrish. Any misunderstandings of the topic are un-intentional, and I request a review to also confirm my understanding of Epoch Protection.


website/docs/dev/tsavorite/epochprotection.md
```diff
@@ -0,0 +1,44 @@
+---
+id: epoch
+sidebar_label: EpochProtection 
+title: EpochProtection
```
badrishc, line 4:
```suggestion
title: Epoch Protection Framework
```

website/docs/dev/tsavorite/epochprotection.md
```diff
@@ -0,0 +1,44 @@
...
+
+-	Each thread that has access to the LightEpoch object can bump the global Epoch (global to the scope of the instance of the LightEpoch class, shared across threads).
+
+-	We can add trigger actions that are executed when all threads are past a safe epoch ``` (For every thread T: SafeEpoch <= thread local Epoch <= Global Epoch) ```. Since the system holds all thread local epochs in a system accesible epoch table we can scan and find a safe epoch.
+This gives us the ability to have an exactly-once invoked function that depends on all threads logically coordinating and not having any code being executed at the time of Epoch ending for the trigger.
+
+-	If you look closely, what Epoch.Resume() essentially does is have a thread find a free entry in the epoch table, put its id there, and "claim" the current epoch (the next thread will increment the epoch by 1 and claim the next epoch). Inside Resume(), there is a loop; if the current target epoch entry is already occupied by another thread, the current thread will yield. When it wakes up, it will try to occupy the next entry. If the epoch table is full, the rest of the threads will keep yielding.
```
badrishc, line 32:
```suggestion
-	If you look closely, what `epoch.Resume()` essentially does is have a thread find a free entry in the epoch table, put its id there, and "claim" the current epoch (the next thread will increment the epoch by 1 and claim the next epoch). Inside `epoch.Resume()`, there is a loop; if the current target epoch entry is already occupied by another thread, the current thread will yield. When it wakes up, it will try to occupy the next entry. If the epoch table is full, the rest of the threads will keep yielding.
```

website/docs/dev/tsavorite/epochprotection.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Relevant Public Methods and How to use them
+
```
badrishc, line 35: put the method/variable names in  backticks so they render in correct style

hamdaankhalid: done

website/docs/dev/tsavorite/epochprotection.md
```diff
@@ -0,0 +1,44 @@
...
+-	In Ted Hart's words: LightEpoch is used because it can run certain operations after other threads that have seen the previous state are no longer executing (that is the basis of epoch control). It's not really "mutual exclusion". It's more like setting address boundaries and operating in them in such a way that there is no conflict. Really briefly: threads "protect" the current epoch by saying "I am active in this epoch" (where "epoch" is a counter). When they are done, they remove that protection. When some operation that would change shared variables (like HeadAddress) is going to run, it is done by "bumping" the epoch, which increases the counter then waits until no other thread is operating with the previous value protected. Variables are set before the bump so that any thread seeing the new counter value also sees the updated variables (again, e.g., HeadAddress) so we know that we are safe to operate up to the current HeadAddress, which is past the previous one.
```
badrishc, line 19:
```suggestion
-	LightEpoch is used because it can run certain operations after other threads that have seen the previous state are no longer executing (that is the basis of epoch control). It's not really "mutual exclusion". It's more like setting address boundaries and operating in them in such a way that there is no conflict. Really briefly: threads "protect" the current epoch by saying "I am active in this epoch" (where "epoch" is a counter). When they are done, they remove that protection. When some operation that would change shared variables (like 'HeadAddress') is going to run, it is done by "bumping" the epoch, which increases the counter then waits until no other thread is operating with the previous value protected. Variables are set before the bump so that any thread seeing the new counter value also sees the updated variables (again, e.g., 'HeadAddress') so we know that we are safe to operate up to the current 'HeadAddress', which is past the previous one.
```

website/docs/dev/tsavorite/epochprotection.md
```diff
@@ -0,0 +1,44 @@
...
+id: epoch
+sidebar_label: EpochProtection 
+title: Epoch Protection Framework
+
+---
+# Light Epoch Protection in Garnet (Latch-free lazy synchronization)
+
+## Context
+We need to ensure shared variables are not being read and mutated simultaneously without determinisitic orderings.
+Commonly used concurreny primitives such as Mutexes and sempahores provided by the language require threads to synchronize frequently with each other. This Synchronization across threads is expensive; Epoch protection **reduces the frequency of synchronization across threads**.
```
badrishc, line 11:
```suggestion
We need to ensure shared variables are not being read and mutated simultaneously without determinisitic orderings. Commonly used concurreny primitives such as Mutexes and sempahores provided by the language require threads to synchronize frequently with each other. This Synchronization across threads is expensive; Epoch protection **reduces the frequency of synchronization across threads**.
```

badrishc (approved):

PR #860, opened 2024-12-06
Fixes for SCRIPT commands and behavior

kevin-montrose:
Looking at Lua scripting integration, and fixing some issues:
 - [x] EVALSHA doesn't work in Cluster mode due to errors in slot verification
 - [x] SCRIPT implements its own sub command parsing instead of using RespCommand
 - [x] No ACL tests for SCRIPT subcommands
 - [x] SCRIPT EXIST should return an array and take multiple hashes
 - [x] SCRIPT FLUSH should take an optional argument
 - [x] Tests for SCRIPT subcommand error cases, and verify errors match Redis responses
 - [x] Remove byte[] allocations from EVAL and EVALSHA
 
In a separate PR (or set of PRs) I'll be looking into perf and correctness* on the EVAL and EVALSHA commands - I'm kinda expecting to have to tear out `NLua` and use `KeraLua` directly, current code is _very_ reflection and allocation heavy.

<sub>*In testing, it appears that keys and values that contain invalid UTF code points are mangled and thus unusable.  I haven't dug in to root case this yet, however, so I could be mistaken.  Separate PR regardless.</sub>


test/Garnet.test/Resp/ACL/RespCommandTests.cs
```diff
@@ -2493,18 +2493,68 @@ async Task DoEvalShaAsync(GarnetClient client)
...
+                "SCRIPT LOAD",
+                [DoScriptLoadAsync]
             );
 
-            async Task DoScriptAsync(GarnetClient client)
+            async Task DoScriptLoadAsync(GarnetClient client)
             {
                 string res = await client.ExecuteForStringResultAsync("SCRIPT", ["LOAD", "return 'OK'"]);
-                ClassicAssert.AreEqual("57ade87c8731f041ecac85aba56623f8af391fab", (string)res);
+                ClassicAssert.AreEqual("57ADE87C8731F041ECAC85ABA56623F8AF391FAB", (string)res);
```
badrishc, line 2506: i thought redis returns lowercase sha?

badrishc:
Can you run the BDN comparing main to this branch and post results here?

test/Garnet.test/Resp/ACL/RespCommandTests.cs
kevin-montrose: Huh, you're right.  I think I must have confused some SE.Redis behavior for plain Redis behavior.

kevin-montrose:
> Can you run the BDN comparing main to this branch and post results here?

So, best I can tell none of the existing BDN cases actually cover these changes.  I'll put one together and report back.

badrishc:
There are a few Lua and LuaRunner BDN tests.

kevin-montrose:
> There are a few Lua and LuaRunner BDN tests.

There's LuaScripts - but that's only for LuaRunner, which this change doesn't touch.  I don't see anything else.

kevin-montrose:
Alright, added [`ScriptOperations`](https://github.com/microsoft/garnet/blob/09ef77042c08680ca4e76ce78f0c9c72b324c1b1/benchmark/BDN.benchmark/Operations/ScriptOperations.cs) benchmark.

As an operations benchmark this runs AOF and ACL versions too, but I'm not including those because a) AOF isn't relevant b) `main` does not currently implement ACLs for SCRIPT operations correctly.

`main` results with `09ef77042c08680ca4e76ce78f0c9c72b324c1b1` cherry-picked:
| Method            | Params | Mean      | Error     | StdDev    | Gen0   | Allocated |
|------------------ |------- |----------:|----------:|----------:|-------:|----------:|
| ScriptLoad        | None   | 54.992 us | 0.2412 us | 0.2138 us | 0.1221 |    9600 B |
| ScriptExistsTrue  | None   | 16.857 us | 0.1802 us | 0.1686 us | 0.1221 |    9600 B |
| ScriptExistsFalse | None   | 18.472 us | 0.2467 us | 0.2308 us | 0.1221 |    9600 B |
| Eval              | None   | 65.101 us | 0.2582 us | 0.2415 us | 0.2441 |   18400 B |
| EvalSha           | None   | 31.560 us | 0.1768 us | 0.1476 us | 0.2441 |   18400 B |

`scriptFixes` results as of `09ef77042c08680ca4e76ce78f0c9c72b324c1b1`:
| Method            | Params | Mean      | Error     | StdDev    | Gen0   | Allocated |
|------------------ |------- |----------:|----------:|----------:|-------:|----------:|
| ScriptLoad        | None   | 66.476 us | 0.9060 us | 0.8475 us | 0.2441 |   16000 B |
| ScriptExistsTrue  | None   | 33.769 us | 0.3562 us | 0.3332 us |      - |         - |
| ScriptExistsFalse | None   | 25.875 us | 0.3684 us | 0.3446 us |      - |         - |
| Eval              | None   | 72.398 us | 1.0838 us | 0.8461 us | 0.2441 |   16000 B |
| EvalSha           | None   | 34.344 us | 0.1864 us | 0.1556 us | 0.1831 |   12000 B |

So allocations reduced as expected for EVAL and EVALSHA, giving up a bit of throughput.  Do note that `SCRIPT|EXISTS` is not correctly implemented in `main` today (it doesn't accept multiple hashes), so that loss comes with a grain of salt.

Given that I'm working on other perf improvements in scripting, I'm OK with this - but I can also spend more on time on this particular branch if desired.

badrishc (approved):

PR #863, opened 2024-12-08
[helm-chart] Add support for garnet.conf file via a secret object

babykart:
* Add support for garnet.conf file via a secret object
* Add example for persistence storage.


badrishc (approved):

badrishc:
This looks good, should we wait to merge this after the other open PR that is in draft?

babykart:
@badrishc  No need to wait, this PR adds a feature (use a config file) while the other is about build automation.

Issue #861, opened 2024-12-07
Garnet fails to import config if the config file is read only


benh57:
### Describe the bug

Garnet fails to import json config if the json file is read only.  Reproduced on Win11

### Steps to reproduce the bug

1. Set a json file to read-only
2. Run GarnetServer.exe --config-import-path  (path to readonly json file)

RESULT: config json fails to parse

If the file is set writable, garnet works fine.

### Expected behavior

Should parse a read only file just fine, it's only reading it. 

### OS version

Windows 11

vazois:
Can you please check the attached PR and let me know if you are still experiencing the issue?

benh57:
Looks like that fixes it! thanks!

PR #870, opened 2024-12-10
Add ReadOnly Support for IDevice

vazois:
This PR fixes issue #861

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/Device/LocalStorageDevice.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Device/LocalStorageDevice.cs
@@ -39,6 +39,7 @@ namespace Tsavorite.core
         private readonly ConcurrentQueue<SimpleAsyncResult> results;
         private static uint sectorSize = 0;
         private bool _disposed;
+        readonly bool readOnly;
 
         /// <summary>
         /// Number of pending reads on device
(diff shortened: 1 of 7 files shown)


libs/storage/Tsavorite/cs/src/core/Device/LocalStorageDevice.cs
```diff
@@ -388,14 +393,14 @@ public override long GetFileSize(int segment)
...
 
         /// <summary>
         /// Creates a SafeFileHandle for the specified segment. This can be used by derived classes to prepopulate logHandles in the constructor.
         /// </summary>
-        protected internal static SafeFileHandle CreateHandle(int segmentId, bool disableFileBuffering, bool deleteOnClose, bool preallocateFile, long segmentSize, string fileName, IntPtr ioCompletionPort, bool omitSegmentId = false)
+        protected internal static SafeFileHandle CreateHandle(int segmentId, bool disableFileBuffering, bool deleteOnClose, bool preallocateFile, long segmentSize, string fileName, IntPtr ioCompletionPort, bool omitSegmentId = false, bool readOnly = false)
         {
-            uint fileAccess = Native32.GENERIC_READ | Native32.GENERIC_WRITE;
+            uint fileAccess = readOnly ? Native32.GENERIC_READ : Native32.GENERIC_READ | Native32.GENERIC_WRITE;
             uint fileShare = unchecked(((uint)FileShare.ReadWrite & ~(uint)FileShare.Inheritable));
```
badrishc, line 404: does FileShare.ReadWrite need to change? nt sure what its semantics are, need to double check.

vazois: For windows, I tested it, and it seems to work. It was breaking before when I made the conf file read-only. That tells that fileShare does not need to be changed.

badrishc (approved):

Issue #412, opened 2024-05-23
Can I quickly import rdb data?


chunjieyaya:
### Feature request type

sample request

### Is your feature request related to a problem? Please describe

现有的redis保存成rdb以后，想导入到garnet中

### Describe the solution you'd like

可以通过命令导入

badrishc:
Workaround would be to load the rdb into redis, scan the DB, and issue set operations on Garnet.

chunjieyaya:
> Workaround would be to load the rdb into redis, scan the DB, and issue set operations on Garnet.

Can the follow-up plan do this function?

vazois:
> > Workaround would be to load the rdb into redis, scan the DB, and issue set operations on Garnet.
> 
> Can the follow-up plan do this function?

This a good idea, to have an rdb-importer for Garnet. Unfortunately, it will require constant maintenance to keep up with any updates to the rdb format.
I actually did a quick research and saw that there exist many different parsers for parsing, formatting and analyzing rdb files.
However, all of them do not seem to have support for newer redis versions.
I am adding here some of them for reference.

If we had a robust C# rdb parser then we can use the internal Garnet API to convert the data into upsert commands.
Similar to what rdb-rs is doing with the conversion to RESP protocol.
Bottomline, this is an issue that should be marked as help is needed.
If someone wants to take a stub at it, I am happy to provide some guidance/feedback.

https://github.com/SamuelFisher/CLRdb
https://github.com/badboy/rdb-rs
https://github.com/redis/librdb
https://github.com/HDT3213/rdb

https://rdb.fnordig.de/file_format.html
https://github.com/redis/redis/blob/unstable/src/rdb.h

chunjieyaya:
> > > Workaround would be to load the rdb into redis, scan the DB, and issue set operations on Garnet.
> > 
> > 
> > Can the follow-up plan do this function?
> 
> This a good idea, to have an rdb-importer for Garnet. Unfortunately, it will require constant maintenance to keep up with any updates to the rdb format. I actually did a quick research and saw that there exist many different parsers for parsing, formatting and analyzing rdb files. However, all of them do not seem to have support for newer redis versions. I am adding here some of them for reference.
> 
[...]

PR #899, opened 2025-01-02
Adding basic version of DUMP and RESTORE commands

s3w3nofficial:
Currently only work's on string type and does not support compression, but it's still a solid starting point.

### Implemented
- redis length encoding
- redis compatible crc64 implementation
- DUMP command that works on strings without compression
- RESTORE command that works on strings without compression and additional modifiers

### Not Implemented
- support for other types than strings
- support for lzf compression
- optional modifiers for RESTORE

Also about the copyright for the [crc64](https://github.com/redis/redis/blob/7.2/src/crc64.c), i only ported the pycrc generated version, but still i'm not sure if and which copyright should i add.


s3w3nofficial:

libs/common/RedisLengthEncodingUtils.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Linq;
+
+namespace Garnet.common;
+
+/// <summary>
+/// Utils for working with redis length encoding
+/// </summary>
+public static class RedisLengthEncodingUtils
```
badrishc, line 12: Rename redis -> resp, e.g., RespLengthEncodingUtils to clarify that this is related to the RESP protocol.

badrishc:
> Also about the copyright for the [crc64](https://github.com/redis/redis/blob/7.2/src/crc64.c), i only ported the pycrc generated version, but still i'm not sure if and which copyright should i add.

since you only ported the pycrc generated version, it seems we should be okay as it is ...

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -90,6 +90,7 @@ public enum RespCommand : ushort
         ZSCAN,
         ZSCORE, // Note: Last read command should immediately precede FirstWriteCommand
         ZUNION,
+        DUMP,
```
vazois, line 93: Please add a slot verification test for DUMP
E.g.: https://github.com/microsoft/garnet/blob/af9cf0eb5d1bd22036ec1fb00be71f860fadc71d/test/Garnet.test.cluster/RedirectTests/ClusterSlotVerificationTests.cs#L30

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -174,6 +175,7 @@ public enum RespCommand : ushort
         ZREMRANGEBYRANK,
         ZREMRANGEBYSCORE,
         ZUNIONSTORE,
+        RESTORE,
```
vazois, line 178: Same as above add a slot verification test.
E.g. https://github.com/microsoft/garnet/blob/af9cf0eb5d1bd22036ec1fb00be71f860fadc71d/test/Garnet.test.cluster/RedirectTests/ClusterSlotVerificationTests.cs#L30

vazois (changes requested):

libs/common/Crc64.cs
```diff
@@ -0,0 +1,82 @@
...
+        crc = Reflect64(crc) ^ 0x0000000000000000UL;
+        return crc;
+    }
+    
+    /// <summary>
+    /// Computes crc64
+    /// </summary>
+    /// <param name="data"></param>
+    /// <returns></returns>
+    public static byte[] Hash(ReadOnlySpan<byte> data)
```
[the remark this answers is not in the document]
s3w3nofficial: Yes i tried using the native one, but the output differs.

For this payload:
[0] = {byte} 3 0x3
[1] = {byte} 118 0x76
[2] = {byte} 97 0x61
[3] = {byte} 108 0x6C
[4] = {byte} 11 0xB
[5] = {byte} 0 0x0

Redis implementation:
0,219 0xDB
1,130 0x82
2,60 0x3C
3,48 0x30
4,56 0x38
5,120 0x78
6,90 0x5A
7,153 0x99

System.IO.Hashing:
0,235 0xEB
1,250 0xFA
2,243 0xF3
3,108 0x6C
4,101 0x65
5,208 0xD0
6,252 0xFC
7,175 0xAF

I did not really try to find out where exactly the implementation differs, it's probably have to do something with applying the reflection of bytes differently.

libs/server/Resp/Parser/RespCommand.cs
s3w3nofficial: done

libs/server/Resp/Parser/RespCommand.cs
s3w3nofficial: done

libs/common/RedisLengthEncodingUtils.cs
s3w3nofficial: done

vazois:
@s3w3nofficial, any updates on the open comments?

s3w3nofficial:
> @s3w3nofficial, any updates on the open comments?

Only open comment left is from @badrishc about the renaming which i already implemented, i guess i can close that one ?

vazois (approved):
LGTM!

s3w3nofficial:
Every test is passing except the RestoreACLsAsync. It throws this exception: "NOAUTH Authentication required.".

I can't figure out why. Can you guys please take a look at it ?

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -93,6 +93,7 @@ public enum RespCommand : ushort
         ZSCAN,
         ZSCORE, // Note: Last read command should immediately precede FirstWriteCommand
         ZUNION,
+        DUMP,
```
vazois, line 96: Please you position the command in alphabetical order?

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -178,6 +179,7 @@ public enum RespCommand : ushort
         ZREMRANGEBYRANK,
         ZREMRANGEBYSCORE,
         ZUNIONSTORE,
+        RESTORE,
```
vazois, line 182: Please position the command in alphabetical order?

vazois (changes requested):

libs/server/Resp/RespServerSession.cs
```diff
@@ -515,6 +515,8 @@ private bool ProcessBasicCommands<TGarnetApi>(RespCommand cmd, ref TGarnetApi st
              */
             _ = cmd switch
             {
+                RespCommand.RESTORE => NetworkRESTORE(ref storageApi),
```
vazois, line 518: Please move RESTORE and DUMP further down. We need check GET first for performance reasons.

vazois (changes requested):

libs/resources/RespCommandsDocs.json
```diff
@@ -2652,6 +2652,50 @@
       }
     ]
   },
+  {
```
TalZaccai, line 2655: Please add these commands to playground/CommandInfoUpdater/SupportedCommand.cs and run the CommandInfoUpdater tool with --force, then ensure that the produced json files (RespCommandDocs.json & RespCommandInfo.json) are the same (it might move these commands around as it's alphabetizing the commands & sub-commands). Thank you!

badrishc:
> Every test is passing except the RestoreACLsAsync. It throws this exception: "NOAUTH Authentication required.".
> 
> I can't figure out why. Can you guys please take a look at it ?

@kevin-montrose, can you help how to fix the ACL unit test here? The error is:

image

libs/server/Resp/Parser/RespCommand.cs
s3w3nofficial: done

libs/server/Resp/Parser/RespCommand.cs
s3w3nofficial: done

libs/server/Resp/RespServerSession.cs
s3w3nofficial: done

libs/resources/RespCommandsDocs.json
s3w3nofficial: done

kevin-montrose:
> > Every test is passing except the RestoreACLsAsync. It throws this exception: "NOAUTH Authentication required.".
> > I can't figure out why. Can you guys please take a look at it ?
> 
> @kevin-montrose, can you help how to fix the ACL unit test here? The error is:
> 
> image

@badrishc  @s3w3nofficial I suspect it's that try/catch not leaving the client in a nice state when an exception is encountered.  The pattern elsewhere for ACLs with commands that aren't repeatable is to use a counter to differentiate each invocation.

```charp
[Test]
public async Task RestoreACLsAsync()
{
    var count = 0;

    await CheckCommandsAsync(
        "RESTORE",
        [DoRestoreAsync]
    );

    async Task DoRestoreAsync(GarnetClient client)
    {
        var payload = new byte[]
        {
            0x00, // value type 
            0x03, // length of payload
            0x76, 0x61, 0x6C,       // 'v', 'a', 'l'
            0x0B, 0x00, // RDB version
            0xDB, 0x82, 0x3C, 0x30, 0x38, 0x78, 0x5A, 0x99 // Crc64
        };

        count++;

        var val = await client.ExecuteForStringResultAsync(
            "$7\r\nRESTORE\r\n"u8.ToArray(),
            [
                Encoding.UTF8.GetBytes($"foo-{count}"), 
                "0"u8.ToArray(), 
                payload
            ]);

        ClassicAssert.AreEqual("OK", val);
    }
}
```

Works, and all ACL tests pass.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+                while (!RespWriteUtils.WriteError(CmdStrings.RESP_ERR_TIMEOUT_NOT_VALID_FLOAT, ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            var value = parseState.GetArgSliceByRef(2);
+
+            // return error if key already exists 
+            var keyExists = storageApi.EXISTS(key);
+            switch (keyExists)
```
kevin-montrose, line 46: Why a 1 case `switch` instead of an `if`?

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            // Restore is only implemented for string type
+            if (valueSpan[0] != 0x00)
+            {
+                while (!RespWriteUtils.WriteError("ERR RESTORE currently only supports string types", ref dcurr, dend))
+                    SendAndReset();
+                return true;
+            }
+
+            // get footer (2 bytes of rdb version + 8 bytes of crc)
+            var footer = valueSpan[^10..];
```
kevin-montrose, line 65: Should an error be reported if valueSpan is too small?  Just raising an exception is kinda hostile.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            var (length, payloadStart) = RespLengthEncodingUtils.DecodeLength(ref valueSpan);
+
+            // Start from payload start and skip the value type byte
+            var val = value.ReadOnlySpan.Slice(payloadStart + 1, (int)length);
+
+            var valArgSlice = scratchBufferManager.CreateArgSlice(val);
+
+            if (ttl > 0)
+            {
+                storageApi.SETEX(key, valArgSlice, TimeSpan.FromMilliseconds(ttl));
```
kevin-montrose, line 103: What is reported if the set fails?

More generally, the Exists call on L45 won't keep the key from being created by the time we get here (or to line 107).

Perhaps a SETEX with NX (and dealing with the responses) would be a better approach?  Redis has a `REPLACE` flag to control that behavior, but that can be future work.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            }
+            else
+            {
+                storageApi.SET(key, valArgSlice);
```
kevin-montrose, line 107: Same note as above on SETEX.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            if (parseState.Count != 1)
+            {
+                return AbortWithWrongNumberOfArguments(nameof(RespCommand.DUMP));
+            }
+
+            var key = parseState.GetArgSliceByRef(0);
+
+            var status = storageApi.GET(key, out var value);
+
+            switch (status)
```
kevin-montrose, line 131: Same "why switch and not if"-note.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            var encodedLength = RespLengthEncodingUtils.EncodeLength(value.ReadOnlySpan.Length);
+
+            // Len of the dump (payload type + redis encoded payload len + payload len + rdb version + crc64)
+            var len = 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8;
+            var lengthInASCIIBytes = new Span<byte>(new byte[NumUtils.NumDigitsInLong(len)]);
+            var lengthInASCIIBytesLen = NumUtils.LongToSpanByte(len, lengthInASCIIBytes);
+
+            // Total len (% + length of ascii bytes + CR LF + payload type + redis encoded payload len + payload len + rdb version + crc64 + CR LF)
+            var totalLength = 1 + lengthInASCIIBytesLen + 2 + 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8 + 2;
+            Span<byte> buffer = stackalloc byte[totalLength];
```
kevin-montrose, line 148: This could be pretty big, right?  So unconditionally stackalloc'ing is dangerous, probably get a rented array past a certain limit (64 or 128 are probably reasonable?).

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+            lengthInASCIIBytes.CopyTo(buffer[offset..]);
+            offset += lengthInASCIIBytes.Length;
+            buffer[offset++] = 0x0D; // CR
+            buffer[offset++] = 0x0A; // LF
+
+            // value type byte
+            buffer[offset++] = 0x00;
+
+            // length of the span
+            foreach (var b in encodedLength)
```
kevin-montrose, line 162: Why copy byte-by-byte and not use a Span CopyTo method?

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -4,12 +4,189 @@
...
+                    while (!RespWriteUtils.WriteDirect(CmdStrings.RESP_ERRNOTFOUND, ref dcurr, dend))
+                        SendAndReset();
+                    return true;
+            }
+
+            var encodedLength = RespLengthEncodingUtils.EncodeLength(value.ReadOnlySpan.Length);
+
+            // Len of the dump (payload type + redis encoded payload len + payload len + rdb version + crc64)
+            var len = 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8;
+            var lengthInASCIIBytes = new Span<byte>(new byte[NumUtils.NumDigitsInLong(len)]);
```
kevin-montrose, line 143: Max length here is like 10 right?  You can just stackalloc that instead of allocating an array.

libs/common/RespLengthEncodingUtils.cs
```diff
@@ -0,0 +1,82 @@
...
+/// </summary>
+public static class RespLengthEncodingUtils
+{
+    /// <summary>
+    /// Decodes the RESP-encoded length and returns payload start
+    /// </summary>
+    /// <param name="buff"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    public static (long length, byte payloadStart) DecodeLength(ref ReadOnlySpan<byte> buff)
```
kevin-montrose, line 20: nit: C# convention is to PascalCase tuple members.  So `(long Length, byte PayloadStart)`.

libs/common/RespLengthEncodingUtils.cs
```diff
@@ -0,0 +1,82 @@
...
+        };
+    }
+
+    /// <summary>
+    /// Encoded payload length to RESP-encoded payload length
+    /// </summary>
+    /// <param name="length"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentOutOfRangeException"></exception>
+    public static byte[] EncodeLength(long length)
```
kevin-montrose, line 51: This whole method is pretty allocate-y, which worries me if it's ever extensively used.

kevin-montrose (changes requested):
Lots of little notes for improvements, and a couple questions around correctness.

s3w3nofficial:
Hey @kevin-montrose 

Thanks for the help with test and for the comments. I will try to resolve these over the weekend. 

Regarding the parsing, i'm working on a [parser](https://github.com/microsoft/garnet/issues/412) that support's the whole rdb format and i hope i will be able to implement that one in the future. For now i will look into optimizing the RespLengthEncodingUtils since the parser won't be ready any time soon.

libs/common/RespLengthEncodingUtils.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/host/defaults.conf
```diff
@@ -344,6 +344,9 @@
 	/* Fails if encounters error during AOF replay or checkpointing */
 	"FailOnRecoveryError": false,
 
+	/* Skips crc64 validation in restore command */
+	"SkipChecksumValidation": false,
```
badrishc, line 348: Rename to `SkipRestoreChecksumValidation` to clarify that this skip is specifically for the restore command.

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: Not sure if renting here is a good idea, if this method will get called, it's probably going to be multiple times, the renting and disposing might be overall less efficient than just heap allocating.

For now i changed it to heap alloc when above 128.

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: fixed

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: fixed

libs/host/defaults.conf
s3w3nofficial: I named it like this because that's how it's named in the redis source code [skip_checksum_validation](https://github.com/search?q=repo%3Aredis%2Fredis+skip_checksum_validation&type=code). Also in the redis it only can be se as a build option and not in redis.conf.

libs/common/RespLengthEncodingUtils.cs
PaulusParssinen: I agree bit @kevin-montrose here. A common approach is to have small static method to first calculate the required buffer size (it can be simple switch expression) for the caller (i.e. in our case ``NetworkDUMP`` method) to know how much buffer it needs (to slice from existing buffer, stackalloc, rent etc., doesn't matter).

Then the actual encoding method takes a ``Span<byte> output`` argument which ownership belongs to caller. Now we could do the encoding on this buffer without allocations. You can see this pattern all over .NET with ``TryWrite`` methods.

libs/host/defaults.conf
badrishc: it is fine to diverge from other resp caches

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -2,14 +2,208 @@
...
+            var len = 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8;
+            Span<byte> lengthInASCIIBytes = stackalloc byte[NumUtils.NumDigitsInLong(len)];
+            var lengthInASCIIBytesLen = NumUtils.LongToSpanByte(len, lengthInASCIIBytes);
+
+            // Total len (% + length of ascii bytes + CR LF + payload type + redis encoded payload len + payload len + rdb version + crc64 + CR LF)
+            var totalLength = 1 + lengthInASCIIBytesLen + 2 + 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8 + 2;
+
+            var buffer = totalLength <= 128
+                ? stackalloc byte[totalLength]
+                : ArrayPool<byte>.Shared.Rent(totalLength);
```
PaulusParssinen, line 164: Excellent! This rent still needs to be returned to the `ArrayPool<byte>` after we're done. I'd also recommend bumping the threshold to 256 bytes. 

This is the common pattern

```C#
byte[]? rentedBuffer = null;
var buffer = totalLength <= 256 
    ? stackalloc byte[256] // constant sized stackallocs result better codegen (We should bunch of stack left here) : 
    (rentedBuffer = ArrayPool<byte>.Shared.Rent(totalLength));
```

then after we're done:

```C#
if (rentedBuffer is not null)
    ArrayPool<byte>.Shared.Return(rentedBuffer);
```

libs/common/RespLengthEncodingUtils.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        var firstByte = input[0];
+        switch (firstByte >> 6)
+        {
+            case 0:
+                bytesRead = 1;
+                length = firstByte & 0x3F;
+                return true;
+            case 1 when input.Length > 2:
```
PaulusParssinen, line 42: I had small mistake in my snippet, sorry! 😅 
```suggestion
            case 1 when input.Length > 1:
```

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/common/RespLengthEncodingUtils.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

libs/host/defaults.conf
s3w3nofficial: done

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -2,14 +2,228 @@
...
+            Span<byte> lengthInASCIIBytes = stackalloc byte[NumUtils.NumDigitsInLong(len)];
+            var lengthInASCIIBytesLen = NumUtils.LongToSpanByte(len, lengthInASCIIBytes);
+
+            // Total len (% + length of ascii bytes + CR LF + payload type + redis encoded payload len + payload len + rdb version + crc64 + CR LF)
+            var totalLength = 1 + lengthInASCIIBytesLen + 2 + 1 + encodedLength.Length + value.ReadOnlySpan.Length + 2 + 8 + 2;
+
+            byte[] rentedBuffer = null;
+            var buffer = totalLength <= 256
+                ? stackalloc byte[256]
+                : (rentedBuffer = ArrayPool<byte>.Shared.Rent(totalLength));
```
Vijay-Nirmal, line 179: You can directly write to the output buffer for non-large objects.

```suggestion
            var buffer = totalLength <= (dend - dcurr)
                ? new Span<byte>(dcurr, (int)(dend - dcurr))
                : (rentedBuffer = ArrayPool<byte>.Shared.Rent(totalLength));
```

At the end, you have to
```csharp
if (rentedBuffer is not null)
{
    // WriteDirectLarge(buffer.Slice(0, totalLength)); Check my next command
    while (!RespWriteUtils.WriteDirect(buffer.Slice(0, totalLength), ref dcurr, dend))
        SendAndReset();

    ArrayPool<byte>.Shared.Return(rentedBuffer);
}
else
{
    dcurr += totalLength;
}
```

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -2,14 +2,228 @@
...
+            var crcBytes = Crc64.Hash(payloadToHash);
+            crcBytes.CopyTo(buffer[offset..]);
+            offset += crcBytes.Length;
+
+            // Write final CRLF
+            buffer[offset++] = 0x0D; // CR
+            buffer[offset] = 0x0A; // LF
+
+            while (!RespWriteUtils.WriteDirect(buffer.Slice(0, totalLength), ref dcurr, dend))
+                SendAndReset();
```
Vijay-Nirmal, line 218: For large objects (when the output is bigger than output buffer) `WriteDirect/SendAndReset` will fail. You can use `WriteDirectLarge` method in RespServerSession.cs file but it's currently private, you have to make it public. @TalZaccai Please provide your thoughts on making it `public` or suggest any alternative method.

Add a test case to validate dump command for large objects

libs/server/Resp/KeyAdminCommands.cs
s3w3nofficial: done

badrishc:
This PR looks ok, although a few spots for improvement remain:
* There should ideally be no need to copy over the entire value retrieved from the back-end to a temporary buffer. Instead, we would directly write the RDB and CRC to the output buffer in the back-end logic, similar to the `CopyRespTo` calls.
* There needs to be a BDN to measure and improve the performance of DUMP and RESTORE.

However, these require deeper surgery than necessary, so we can go ahead and approve/merge this PR as is, and revisit later as per needs. Thank you.

badrishc (approved):

vazois (approved):

Issue #900, opened 2025-01-02
Unresponsive Garner server upon multiple cluster meets


harshika-kashyap:
### Describe the bug

Garnet server becomes unresponsive (deadlock) upon multiple cluster meet operations, when started with very few threads (4).

### Steps to reproduce the bug

- Start 6 Garnet servers (v1.0.47) with min and max threads as 4.
```
./GarnetServer --port 6000 --checkpointdir /tmp/checkpoints/6000 --cluster True --minthreads 4 --maxthreads 4
./GarnetServer --port 6001 --checkpointdir /tmp/checkpoints/6001 --cluster True --minthreads 4 --maxthreads 4
./GarnetServer --port 6002 --checkpointdir /tmp/checkpoints/6002 --cluster True --minthreads 4 --maxthreads 4
./GarnetServer --port 6003 --checkpointdir /tmp/checkpoints/6003 --cluster True --minthreads 4 --maxthreads 4
./GarnetServer --port 6004 --checkpointdir /tmp/checkpoints/6004 --cluster True --minthreads 4 --maxthreads 4
./GarnetServer --port 6005 --checkpointdir /tmp/checkpoints/6005 --cluster True --minthreads 4 --maxthreads 4
```

- Connect to one of the servers and execute cluster meet commands.
```
redis-cli -p 6000
127.0.0.1:6000 > cluster meet 127.0.0.1 6001
OK
127.0.0.1:6000 > cluster meet 127.0.0.1 6002
OK
127.0.0.1:6000 > cluster meet 127.0.0.1 6003
OK
127.0.0.1:6000 > cluster meet 127.0.0.1 6004
OK
127.0.0.1:6000 > cluster meet 127.0.0.1 6005 <-- This one got stuck in my experiment.
```

### Expected behavior

The server should remain responsive.

### Screenshots

![Image](https://github.com/user-attachments/assets/f376760d-ca86-46aa-ae53-ddf7b6f53e20)
![Image](https://github.com/user-attachments/assets/82b2bdff-cdae-4d29-9892-3f1287387cb1)

### Release version

v1.0.47

### Additional context

I took the thread snapshot of the stuck Garnet server. 

Thread (0x4B33) has acquired the [lock in ClusterManager](https://github.com/microsoft/garnet/blob/9fb9c07ec75da058d97dda026b1b3fd836869d5b/libs/cluster/Server/ClusterManager.cs#L123C1-L123C24) and is stuck while writing to the file (nodes.conf).

[...]

PR #905, opened 2025-01-09
Skip Interlocked.Exchange and FlushConfig when no update detected

vazois:
This PR fixes an issue that caused message flooding across the cluster. This happened because the config object kept being updated even when the received config did not change the local config during merge.

The check for an updated config is happening implicitly by comparing the new and current config objects 
https://github.com/microsoft/garnet/blob/af9cf0eb5d1bd22036ec1fb00be71f860fadc71d/libs/cluster/Server/Gossip.cs#L116-L123

This meant that the call to currentCopy.Merge should return the same object if no update has happened. This occurs correctly when merging the worker info
https://github.com/microsoft/garnet/blob/af9cf0eb5d1bd22036ec1fb00be71f860fadc71d/libs/cluster/Server/ClusterConfig.cs#L917

However, an equivalent check was not there when the object was updated during the merging of the slotMap information.
The PR adds this check to fix the issue of message flooding when the cluster configuration is stable

Fixes https://github.com/microsoft/garnet/issues/900

### DIFF
--- a/Version.props
+++ b/Version.props
@@ -1,6 +1,6 @@
 <Project>
 	<!-- Versioning property for builds and packages -->
 	<PropertyGroup>
-		<VersionPrefix>1.0.49</VersionPrefix>
+		<VersionPrefix>1.0.50</VersionPrefix>
 	</PropertyGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


badrishc (approved):
LGTM, please update version for a new release as well.

PR #848, opened 2024-12-04
CustomCommandManager + CustomCommandManagerSession to use auto-expandable maps

TalZaccai:
- [x] Implementing `ExpandableMap` struct which maps a specified range of IDs to an array which doubles in size as needed.
- [x] Switching `CustomCommandManager` & `CustomCommandManagerSession` to use `ExpandableMap` instead of using set-sized arrays.
- [x] Adding tests to `RespCustomCommandTests` to test registration of multiple string commands, object commands, transactions and procedures.

Note: this relates to closed issue #834, which was addressed by temporary fix in PR #836.

Benchmark results:

main:
| Method                 | Params | Mean       | Error    | StdDev   | Gen0   | Allocated |
|----------------------- |------- |-----------:|---------:|---------:|-------:|----------:|
| CustomRawStringCommand | ACL    |   409.1 us |  7.60 us |  7.11 us |      - |         - |
| CustomObjectCommand    | ACL    | 1,061.6 us | 17.22 us | 16.11 us | 1.9531 |   24002 B |
| CustomTransaction      | ACL    | 1,738.7 us | 31.31 us | 49.66 us |      - |       2 B |
| CustomProcedure        | ACL    | 1,264.8 us | 14.86 us | 13.90 us |      - |       2 B |
| CustomRawStringCommand | AOF    |   414.0 us |  4.80 us |  4.49 us |      - |         - |
| CustomObjectCommand    | AOF    | 1,227.9 us | 16.35 us | 15.30 us | 1.9531 |   24002 B |
| CustomTransaction      | AOF    | 1,967.2 us |  6.12 us |  5.43 us |      - |       4 B |
| CustomProcedure        | AOF    | 1,678.9 us | 27.81 us | 38.99 us |      - |       2 B |
| CustomRawStringCommand | None   |   403.5 us |  7.81 us |  7.31 us |      - |         - |
| CustomObjectCommand    | None   | 1,110.4 us | 21.81 us | 31.28 us | 1.9531 |   24002 B |
| CustomTransaction      | None   | 1,732.3 us | 12.80 us |  9.99 us |      - |       2 B |
| CustomProcedure        | None   | 1,229.2 us | 24.41 us | 21.64 us |      - |       2 B |

PR #848:
| Method                 | Params | Mean       | Error    | StdDev   | Gen0   | Allocated |
|----------------------- |------- |-----------:|---------:|---------:|-------:|----------:|
[...]

PR #906, opened 2025-01-09
Returning "Friendly IDs" for Custom Commands / Custom Types

TalZaccai:
Following changes in PR #848 we now assign raw string command IDs & custom type IDs in descending order starting from last valid ID available.
To make IDs returned to the client "friendlier", this PR converts the "large" custom command / type IDs to start at index 0.

### DIFF
--- a/libs/server/Custom/ExpandableMap.cs
+++ b/libs/server/Custom/ExpandableMap.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Diagnostics;
 using System.Threading;
 using Garnet.common;
 
@@ -59,20 +60,6 @@ namespace Garnet.server
         /// <param name="id">ID if found, otherwise -1</param>
         /// <returns>True if ID found</returns>
         bool TryGetFirstId(Func<T, bool> predicate, out int id);
-
-        /// <summary>
-        /// Maps underlying map index to item ID
-        /// </summary>
-        /// <param name="index">Map index</param>
-        /// <returns>Item ID</returns>
-        int GetIdFromIndex(int index);
-
-        /// <summary>
-        /// Maps an item ID to underlying map index
-        /// </summary>
-        /// <param name="id">Item ID</param>
-        /// <returns>Map index</returns>
-        int GetIndexFromId(int id);
     }
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


libs/server/Resp/RespServerSession.cs
```diff
@@ -826,7 +826,8 @@ private bool NetworkCustomRawStringCmd<TGarnetApi>(ref TGarnetApi storageApi)
             }
 
             // Perform the operation
-            TryCustomRawStringCommand((RespCommand)currentCustomRawStringCommand.id,
+            var cmdId = customCommandManagerSession.GetRawStringCommandIdFromFriendlyId(currentCustomRawStringCommand.id);
```
badrishc, line 829: is this an interface call? solution seems a bit complex as it stands.

libs/server/Custom/ExpandableMap.cs
```diff
@@ -59,6 +59,20 @@ internal interface IExpandableMap<T>
         /// <param name="id">ID if found, otherwise -1</param>
         /// <returns>True if ID found</returns>
         bool TryGetFirstId(Func<T, bool> predicate, out int id);
+
+        /// <summary>
+        /// Maps underlying map index to item ID
+        /// </summary>
+        /// <param name="index">Map index</param>
+        /// <returns>Item ID</returns>
+        int GetIdFromIndex(int index);
```
badrishc, line 68: why can't the non-friendly ids be an internal detail of the expandable map. i.e. it always returns 0, 1, 2, etc. even though they are stored differently internally.

TalZaccai: They are not stored differently internally, they are stored by these exact indexes (0, 1, 2,...) in the underlying map, the "non-friendly" IDs are required for conversion to the RespCommand / GarnetObjectType enums which is needed by the caller to create the store input object. Maybe a slightly nicer way to do this is have the conversion methods return the enums rather than the non-friendly IDs. WDYT?

TalZaccai: Pushed this small simplification...

libs/server/Resp/RespServerSession.cs
badrishc: We can't afford an interface call at this point of the code

TalZaccai: There's no interface call here, it's all direct class method calls. If you think this should be further simplified (i.e. just K - id) I can do that, only drawback is maintainability (i.e. 2 places to change if the logic changes).

badrishc (approved):

PR #907, opened 2025-01-10
Add recompiled version for libnative_device on Ubuntu 24.04

badrishc:
Recompiled `libnative_device.so` on Ubuntu 24.04 so that it uses `libaio.so.1t64` instead of `libaio.so.1`. This lib works on Ubuntu 24.04 but it has stopped working on Ubuntu 22.04 out of the box.

Any ideas how to make it work out-of-the-box on both? We could do a symbolic link as mentioned [here](https://askubuntu.com/questions/1512196/libaio1-on-noble/1512197) but that requires sudo permission and manual intervention.

Here is the error on Ubuntu 22.04 after this PR:

```
  Error Message:
   System.DllNotFoundException : Unable to load shared library 'runtimes/linux-x64/native/libnative_device.so' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:
libaio.so.1t64: cannot open shared object file: No such file or directory
```


badrishc:
We should hold off on merging this until we figure out how to avoid breaking on Ubuntu 22.04. Thoughts?

PR #910, opened 2025-01-12
Bump Minor Version Post ETag feature Addition

hamdaankhalid:
According to -> https://semver.org/

### DIFF
--- a/Version.props
+++ b/Version.props
@@ -1,6 +1,6 @@
 <Project>
 	<!-- Versioning property for builds and packages -->
 	<PropertyGroup>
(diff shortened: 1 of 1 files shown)


Version.props
```diff
@@ -1,6 +1,6 @@
 <Project>
 	<!-- Versioning property for builds and packages -->
 	<PropertyGroup>
-		<VersionPrefix>1.0.50</VersionPrefix>
```
Vijay-Nirmal, line 4: I don't think Garnet team is following Semantic Versioning, but if we do, shouldn't this be `1.1.0`

PR #908, opened 2025-01-10
Native ETag Support

hamdaankhalid:
### Native ETag Support
Garnet provides support for ETags on raw strings. You can associate any string-based key-value pair inserted into Garnet with an automatically updated ETag.
Compatibility with non-ETag commands and the behavior of data inserted with ETags are detailed in the markdown for garnet specific commands. Please read the markdown file for API contracts, and compatibility.

### HLD:
RecordInfo for a record indicated whether or not a record contains an ETag. This bit on the record info is used to indicate that the first 8 bytes of the payload of the record include a long data type that represents the eTag associated with the value that is present after the first 8 bytes.

Reads don't modify the eTag, but any RMW operations do. Blind Upsert (SET) will remove the eTag unless sent with the WITHETAG flag.

ETags are incremented on each update, and the user can use them to handle things such as caching and Optimistic concurrency control.

PR #912, opened 2025-01-13
ETag Follow Up: Address Open Comments

hamdaankhalid:
Follow up PR addressing comments by Tal on https://github.com/microsoft/garnet/pull/908


libs/server/Resp/BasicCommands.cs
```diff
@@ -11,6 +11,16 @@
 
 namespace Garnet.server
 {
+    internal enum ExpirationOption : byte
```
TalZaccai, line 14: Not a fan of it being here... Probably should extract to a file named "RespEnums" under Resp/ (can also move EtagOptions & ExistOptions there).

libs/server/Resp/BasicCommands.cs
```diff
@@ -491,59 +491,32 @@ private bool NetworkSETEXNX<TGarnetApi>(ref TGarnetApi storageApi)
...
-                    // Validate expiry
-                    if (!parseState.TryGetInt(tokenIdx++, out expiry))
-                    {
-                        errorMessage = CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER;
-                        break;
-                    }
-
-                    if (expOption != ExpirationOption.None)
+                    // Make sure there aren't multiple expiration options in the options sent by user
+                    if (expOption != ExpirationOption.None || parsedOption == ExpirationOption.EXAT || parsedOption == ExpirationOption.PXAT)
```
TalZaccai, line 498: Probably better to have a whitelist of acceptable ExpirationOption rather than the other way around (in case that in the future this enum gets added to)

libs/server/Resp/BasicCommands.cs
```diff
@@ -491,59 +491,32 @@ private bool NetworkSETEXNX<TGarnetApi>(ref TGarnetApi storageApi)
...
-                    // Validate expiry
-                    if (!parseState.TryGetInt(tokenIdx++, out expiry))
-                    {
-                        errorMessage = CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER;
-                        break;
-                    }
-
-                    if (expOption != ExpirationOption.None)
+                    expOption = parsedOption;
+                    if (expOption != ExpirationOption.KEEPTTL)
```
TalZaccai, line 505: Same here, better to include rather than exclude

libs/server/SessionParseStateExtensions.cs
```diff
@@ -236,5 +237,37 @@ internal static bool TryGetSortedSetAggregateType(this SessionParseState parseSt
...
+        /// <param name="idx">The argument index</param>
+        /// <param name="value">Parsed expiration option value</param>
+        /// <returns>if the argument at that index is a valid expiration option set it to value and return true, else returns false</returns>
+        internal static bool TryGetExpirationOption(this SessionParseState parseState, int idx, out ExpirationOption value)
+        {
+            var sbArg = parseState.GetArgSliceByRef(idx).Span;
+            return parseState.TryGetExpirationOptionWithToken(ref sbArg, out value);
+        }
+
+        internal static bool TryGetExpirationOptionWithToken(this SessionParseState parseState, ref Span<byte> token, out ExpirationOption value)
```
TalZaccai, line 254: Add an XML comment here as well please :)

TalZaccai (changes requested):
LGTM, few small comments. Thanks for addressing all of these!

libs/server/Resp/BasicCommands.cs
hamdaankhalid: once I add the check above this, this just reduces the copmarisons in the conditional, the check before this line will make sure we are whitelisting already

TalZaccai (approved):

PR #914, opened 2025-01-14
Remove meet lock in TryMeet

msft-paddy14:
The lock is not really protecting anything in this context as we still proceed with Meet rather than returning the thread until it's available. So doesn't seem like we really need a lock. In some cases during startup if there are multiple meets (and one or more nodes are unresponsive), it can cause CAS retries during it's release so better to remove it if it's not needed.
Also note that MEET is currently Blocking and consumes a threadpool thread.

CPU stacks indicating WriteUnlock taking time
image

### DIFF
--- a/libs/cluster/Server/GarnetServerNode.cs
+++ b/libs/cluster/Server/GarnetServerNode.cs
@@ -30,11 +30,6 @@ namespace Garnet.cluster
         ClusterConfig lastConfig = null;
 
         /// <summary>
-        /// Gossip with meet command lock
-        /// </summary>
-        SingleWriterMultiReaderLock meetLock;
-
-        /// <summary>
         /// Outstanding gossip task if any
         /// </summary>
         Task gossipTask = null;
@@ -202,17 +197,9 @@ namespace Garnet.cluster
         /// <returns></returns>
         public MemoryResult<byte> TryMeet(byte[] configByteArray)
         {
-            try
-            {
-                _ = meetLock.TryWriteLock();
-                UpdateGossipSend();
-                var resp = gc.GossipWithMeet(configByteArray).WaitAsync(clusterProvider.clusterManager.clusterTimeout, cts.Token).GetAwaiter().GetResult();
-                return resp;
-            }
-            finally
-            {
-                meetLock.WriteUnlock();
-            }
+            UpdateGossipSend();
+            var resp = gc.GossipWithMeet(configByteArray).WaitAsync(clusterProvider.clusterManager.clusterTimeout, cts.Token).GetAwaiter().GetResult();
+            return resp;
         }
 
         /// <summary>


vazois:
So, the meet lock is there to throttle bursts of meet requests. The same connection is used for gossip and if that gets delayed a timeout will occur which will break and re-initialize the connection, unless gossip-delay is increases. I would make the WriteLock async if that is an issue.

vazois:
> So, the meet lock is there to throttle bursts of meet requests. The same connection is used for gossip and if that gets delayed a timeout will occur which will break and re-initialize the connection, unless gossip-delay is increases. I would make the WriteLock async if that is an issue.

Should be fine to go ahead and remove that lock

libs/cluster/Server/Gossip.cs
```diff
@@ -134,7 +134,7 @@ public bool TryMerge(ClusterConfig senderConfig, bool acquireLock = true)
         /// <param name="address"></param>
         /// <param name="port"></param>
         public void RunMeetTask(string address, int port)
-            => Task.Run(() => TryMeet(address, port));
+            => Task.Run(() => TryMeetAsync(address, port));
```
badrishc, line 137: Task.Run(async () => await TryMeetAsync(...));

libs/cluster/Server/Gossip.cs
```diff
@@ -203,15 +203,15 @@ public void TryMeet(string address, int port, bool acquireLock = true)
         /// <summary>
         /// Main gossip async task
         /// </summary>
-        async void GossipMain()
+        async Task GossipMain()
```
badrishc, line 206: Caller should be: `_ = Task.Run(async () => await GossipMain())`

libs/cluster/Server/GarnetServerNode.cs
```diff
@@ -200,19 +195,11 @@ private Task Gossip(byte[] configByteArray)
...
-                UpdateGossipSend();
-                var resp = gc.GossipWithMeet(configByteArray).WaitAsync(clusterProvider.clusterManager.clusterTimeout, cts.Token).GetAwaiter().GetResult();
-                return resp;
-            }
-            finally
-            {
-                meetLock.WriteUnlock();
-            }
+            UpdateGossipSend();
+            var resp = gc.GossipWithMeet(configByteArray).WaitAsync(clusterProvider.clusterManager.clusterTimeout, cts.Token);
```
badrishc, line 201: why not await gc.GossipMeet here?

msft-paddy14: Made consistent. thanks

libs/cluster/Server/Gossip.cs
msft-paddy14: Done

badrishc (approved):

PR #919, opened 2025-01-17
Adding Support for .Net 9

TalZaccai:


libs/common/FileUtils.cs
```diff
@@ -159,13 +159,15 @@ public static bool TryLoadAssemblies(IEnumerable<string> assemblyPaths, out IEnu
                     Assembly assembly;
                     try
                     {
-                        assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+                        // NET9 seems to hold the assembly locked if it's opened from the path directly, causing tests to fail.
+                        //assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+                        assembly = AssemblyLoadContext.Default.LoadFromStream(new FileStream(path, FileMode.Open, FileAccess.Read));
```
PaulusParssinen, line 164: If I'm reading this right; we should `Dispose` or at least call `Close` for this new instantiated `FileStream`. (Avoids relying on finalizer queue to clean it up)

TalZaccai: Very true, thanks for noticing!

badrishc (approved):
Great work, thank you all!

PR #923, opened 2025-01-18
Adding ObjectStoreOutputFlags to GarnetObjectStoreOutput

TalZaccai:
This change is to avoid checking for `GarnetObjectStoreOutput.spanByteAndMemory.Length == 0` in order to determine if object operation was attempted on a mismatched type (and by that to allow object operations to return an empty output and avoid unnecessary allocations)

Benchmarks for module operations after the change:
Note: # of allocated bytes decreased for ObjRmw, ObjRead & JsonSet.
| Method                         | Params | Mean      | Error    | StdDev    | Median    | Gen0   | Allocated |
|------------------------------- |------- |----------:|---------:|----------:|----------:|-------:|----------:|
| ModuleNoOpRawStringReadCommand | None   |  55.37 us | 1.969 us |  5.422 us |  53.53 us |      - |         - |
| ModuleNoOpRawStringRmwCommand  | None   |  63.87 us | 1.275 us |  3.031 us |  63.01 us |      - |         - |
| ModuleNoOpObjRmwCommand        | None   |  89.03 us | 1.994 us |  5.786 us |  87.00 us |      - |    3200 B |
| ModuleNoOpObjReadCommand       | None   |  73.87 us | 1.469 us |  2.794 us |  73.74 us |      - |    3200 B |
| ModuleNoOpProc                 | None   |  45.70 us | 0.889 us |  1.092 us |  45.64 us |      - |         - |
| ModuleNoOpTxn                  | None   |  38.86 us | 0.760 us |  1.519 us |  38.79 us |      - |         - |
| ModuleJsonGetCommand           | None   | 154.14 us | 2.852 us |  4.765 us | 153.92 us |      - |   72800 B |
| ModuleJsonSetCommand           | None   | 269.16 us | 5.345 us | 13.312 us | 268.49 us | 0.4883 |  223200 B |


badrishc:
Why isn't allocated at zero after this change? Need to check what else gets allocated on critical path. There doesn't logically seem to be anything else needed.

libs/server/Objects/Types/IGarnetObject.cs
```diff
@@ -35,8 +35,9 @@ public interface IGarnetObject : IDisposable
         /// <param name="output"></param>
         /// <param name="sizeChange"></param>
         /// <param name="removeKey"></param>
+        /// <param name="wrongType"></param>
         /// <returns></returns>
-        bool Operate(ref ObjectInput input, ref SpanByteAndMemory output, out long sizeChange, out bool removeKey);
+        bool Operate(ref ObjectInput input, ref SpanByteAndMemory output, out long sizeChange, out bool removeKey, out bool wrongType);
```
badrishc, line 40: Please clean this up if possible, remove size change and remove key. Put into enum perhaps?

libs/server/Objects/Types/GarnetObjectStoreOutput.cs
```diff
@@ -20,6 +20,11 @@ public struct GarnetObjectStoreOutput
         /// </summary>
         public IGarnetObject garnetObject;
 
+        /// <summary>
+        /// True if an operation was attempted on the wrong type of object
+        /// </summary>
+        public bool wrongType;
```
badrishc, line 26: We discussed making this an enum, please check that...

test/BDNPerfTests/BDN_Benchmark_Config.json
```diff
@@ -261,28 +261,28 @@
...
     "expected_ModuleNoOpProc_ACL": 0,
     "expected_ModuleNoOpTxn_ACL": 0,
     "expected_ModuleJsonGetCommand_ACL": 72800,
-    "expected_ModuleJsonSetCommand_ACL": 228800,
+    "expected_ModuleJsonSetCommand_ACL": 223200,
     "expected_ModuleNoOpRawStringReadCommand_AOF": 0,
     "expected_ModuleNoOpRawStringRmwCommand_AOF": 0,
-    "expected_ModuleNoOpObjRmwCommand_AOF": 8800,
-    "expected_ModuleNoOpObjReadCommand_AOF": 5600,
+    "expected_ModuleNoOpObjRmwCommand_AOF": 3200,
```
badrishc, line 272: Scope of this PR may be a bit narrow, IMHO. Better if one PR handles all the technical debt related to custom ops perf, since each individual item is rather straightforward, technically.

TalZaccai: I think this one is self-contained and unrelated to the changes which have to do with the CustomCommandManager, makes it easier to review and pinpoint to a certain commit if something needs to be debugged / reverted.

libs/server/Objects/Types/GarnetObjectStoreOutput.cs
TalZaccai: Oh I missed that comment :) Can do!

libs/server/Objects/Types/IGarnetObject.cs
badrishc: They go into GOSO as enum entries only if they are needed by the back end.

libs/server/Objects/Types/IGarnetObject.cs
```diff
@@ -33,10 +33,8 @@ public interface IGarnetObject : IDisposable
         /// </summary>
         /// <param name="input"></param>
         /// <param name="output"></param>
-        /// <param name="sizeChange"></param>
-        /// <param name="removeKey"></param>
         /// <returns></returns>
-        bool Operate(ref ObjectInput input, ref SpanByteAndMemory output, out long sizeChange, out bool removeKey);
```
badrishc, line 39: Reinstate sizeChange as discussed

TalZaccai: SizeChange was added to GarnetObjectStoreOutput

badrishc: GOSO fields should be limited to what information needs to flow from the back-end to the front-end. Let us make sure we verify this holds for all additions to GOSO.

libs/server/Objects/Types/GarnetObjectStoreOutput.cs
```diff
@@ -1,24 +1,52 @@
...
+    }
+
     /// <summary>
     /// Output type used by Garnet object store.
     /// </summary>
     public struct GarnetObjectStoreOutput
     {
         /// <summary>
-        /// span byte and memory
+        /// Span byte and memory
```
badrishc, line 37: Add a comment to GOSO that we should only add fields that need to flow from backend (IFunctions) to frontend (caller of GarnetApi) in that specific direction.

badrishc (approved):

PR #925, opened 2025-01-18
Make parsing and formatting utility method naming more accurate and consistent

PaulusParssinen:
I have been going back and forth for long time when would be best time to try introduce these changes, so here goes.

This PR makes major naming changes in the "low-level" utility classes: `NumUtils` and `RespRead/WriteUtils`. The naming changes are largely mechanical:
1. Adding `Try`-prefix to all parsing/formatting methods that return `bool` indicating whether reading/writing was successful or not. 
  i. This matches the .NET API design convention I _think_ majority .NET developers are/should be familiar with. I'm talking about the distinction between `int.Parse` and `int.TryParse`. It makes it _clear_ to the developer they need to think about the case where the method can fail "silently" if the return value is ignored. Generally these ``Try`` methods are also the more performant ones, they handle the majorirty that they don't __usually__ throw, and as such incur any exception handling costs.
2. .NET primitive numeric naming in method names, i.e. `(U)Long` to `(U)Int64`. This one is very much preference thing, but after I saw method named as `Read64Int`, I went ahead with this naming change proposal. (This is also the convention in BCL)
3. PR also attempts to improve the xmldoc situation in `NumUtils`. 

Method naming is largely very subjective and hard, but I think these changes are warranted as _I believe_ they improve readability and reduce the mental overhead when working with the parsing/formatting parts of the code-base. I also think that more accurate method naming conventions especially improve code-review process.

I'm very happy and prefer to wait with this PR to minimize any all the conflicts this will have with older PRs still in progress.


badrishc:
I think this is a good and sensible code clean-up, we should review and get this merged.

TalZaccai (approved):
Big fan of this PR! :) Thanks for taking the time to do this!

Issue #379, opened 2024-05-13
integrate OpenTelemetry to enhance  observability 


springmin:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

integrate OpenTelemetry or other trance component  to enhance  observability。 

### Describe the solution you'd like

integrate OpenTelemetry or other trance component  to enhance  observability。

PaulusParssinen:
I've been thinking about this for a while (and implementing the server using Generic Host in general).

By using the new System.Diagnostics.Metrics APIs (`Counter<T>`, `UpDownCounter<T>`, `Histogram<T>` etc. ..)  would allow us to delete HdrHistogram fork altogether from the repo and simplify our metrics/monitoring logic a lot. The places where we pull the metrics inside the server (`LATENCY` etc?) we should be able to use simple `MeterListener` to aggregate them). I'm not sure if `Histogram<T>` covers everything that `HdrHistogram` offers and someone who knows statistics better than me should weigh in on that (https://github.com/dotnet/runtime/issues/63650 seems to be useful addition).

We should follow the existing [OTel semantic conventions for Redis](https://opentelemetry.io/docs/specs/semconv/database/redis/) but of course modify/extend it if needed. Also configurations knob whether to record [db.statement](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/) or not, as I believe there is currently no redaction capabilities in Garnet for potentially sensitive parts of the command.

Measuring anything should become a lot easier, but we should follow how ASP.NET guards the metrics logic to keep the overhead next to none IF the metrics are completely disabled ([`Instrument.Enabled`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.instrument.enabled?view=net-8.0#system-diagnostics-metrics-instrument-enabled)). Should the metrics codepaths should be entirely eliminated when Garnet is built with `<MetricsSupport>false</MetricsSupport>` given that would also affect the behavior of the RESP commands that rely on it, if they are implemented using `MeterListener`?

We should also measure if any extra contention (IIRC currently Garnet's counters are done by atomic `Interlocked.Increment`s) for normal use-cases. We should keep eye on how big difference there is in RPS with, for example, ``dotnet-counters`` hooked in 
. Some existing tracking issues for `S.D.Metrics` performance concerns https://github.com/dotnet/aspnetcore/issues/50412, https://github.com/dotnet/runtime/issues/71563 to keep eye on too.

badrishc:
[...]

PR #934, opened 2025-01-21
Implement generic host

s3w3nofficial:
Integrate GarnetServer with [generic host](https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host?tabs=appbuilder).

- simplify configuration of Garnet 
- allow for overriding config with GARNET_ environment variables 
- simplify passing of dependencies using DI
- allow for future integration with [open telemetry](https://opentelemetry.io/docs/languages/net/getting-started/), #379 
- allow for future integration with [IOptions and IOptionsMonitor](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-9.0)


libs/host/GarnetApplication.cs
```diff
@@ -0,0 +1,106 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Garnet.host;
+
+/// <summary>
+/// 
+/// </summary>
+public class GarnetApplication : IHost 
+{
+    /// <summary>
```
Meir017, line 20: should these be properties?

s3w3nofficial:
Hi @badrishc, could you please take a look at this ? There is still work to be done but i would appreciate some initial review.

PR #938, opened 2025-01-21
Add separate active connection tracker

vazois:
This PR adds a separate field to track active number of connections separately from received number of connections.


libs/server/Metrics/GarnetServerMonitor.cs
```diff
@@ -258,6 +258,7 @@ private async void MainMonitorTask(CancellationToken token)
                     var garnetServer = ((GarnetServerBase)server);
                     globalMetrics.total_connections_received = garnetServer.get_conn_recv();
                     globalMetrics.total_connections_disposed = garnetServer.get_conn_disp();
+                    globalMetrics.total_connections_active = garnetServer.get_conn_active();
```
badrishc, line 261: use `received - disposed` to compute this instead of `activeHandlers.Count` so that the three reported numbers are self-consistent.

badrishc (approved):

PR #940, opened 2025-01-21
Perf Improvements to Custom Commands' Execution Path

TalZaccai:
This PR introduces some perf improvements to the execution path of custom commands:
- [x] Unlatched access to custom commands' data structures in CustomCommandManager
- [x] Eliminating expensive string-based RespCommandInfo access for arity check by adding arity property to commands classes.

**Benchmarking results:**

_**Current branch:**_

**CustomOperations:**
| Method                 | Params | Mean      | Error    | StdDev   | Median    | Allocated |
|----------------------- |------- |----------:|---------:|---------:|----------:|----------:|
| CustomRawStringCommand | None   |  43.23 us | 0.949 us | 2.783 us |  42.49 us |         - |
| CustomObjectCommand    | None   | 146.70 us | 2.880 us | 4.568 us | 145.93 us |   18400 B |
| CustomTransaction      | None   |  92.77 us | 1.545 us | 2.706 us |  92.05 us |         - |
| CustomProcedure        | None   |  72.35 us | 0.595 us | 0.557 us |  72.15 us |         - |

**ModuleOperations:**

| Method                         | Params | Mean      | Error    | StdDev   | Median    | Gen0   | Allocated |
|------------------------------- |------- |----------:|---------:|---------:|----------:|-------:|----------:|
| ModuleNoOpRawStringReadCommand | None   |  39.59 us | 0.784 us | 1.863 us |  38.96 us |      - |         - |
| ModuleNoOpRawStringRmwCommand  | None   |  48.62 us | 0.963 us | 2.604 us |  47.87 us |      - |         - |
| ModuleNoOpObjRmwCommand        | None   |  64.37 us | 1.277 us | 2.491 us |  64.53 us |      - |    3200 B |
| ModuleNoOpObjReadCommand       | None   |  55.30 us | 1.096 us | 2.770 us |  54.64 us |      - |    3200 B |
| ModuleNoOpProc                 | None   |  31.71 us | 0.368 us | 0.344 us |  31.70 us |      - |         - |
| ModuleNoOpTxn                  | None   |  30.33 us | 0.601 us | 1.186 us |  30.14 us |      - |         - |
| ModuleJsonGetCommand           | None   | 125.22 us | 2.487 us | 5.354 us | 124.29 us |      - |   72800 B |
| ModuleJsonSetCommand           | None   | 216.43 us | 2.273 us | 2.015 us | 216.43 us | 0.4883 |  223200 B |

_**main:**_

**CustomOperations:**
| Method                 | Params | Mean      | Error    | StdDev    | Median    | Allocated |
|----------------------- |------- |----------:|---------:|----------:|----------:|----------:|
| CustomRawStringCommand | None   |  46.06 us | 0.692 us |  0.824 us |  45.70 us |         - |
| CustomObjectCommand    | None   | 175.01 us | 4.435 us | 12.216 us | 170.85 us |   24000 B |
| CustomTransaction      | None   |  98.49 us | 1.959 us |  2.932 us |  98.16 us |         - |
| CustomProcedure        | None   |  83.08 us | 1.629 us |  2.938 us |  82.53 us |         - |

**ModuleOperations:**
| Method                         | Params | Mean      | Error    | StdDev   | Gen0   | Allocated |
|------------------------------- |------- |----------:|---------:|---------:|-------:|----------:|
| ModuleNoOpRawStringReadCommand | None   |  46.36 us | 0.927 us | 1.624 us |      - |         - |
| ModuleNoOpRawStringRmwCommand  | None   |  54.67 us | 1.083 us | 0.960 us |      - |         - |
| ModuleNoOpObjRmwCommand        | None   |  84.71 us | 1.627 us | 4.400 us |      - |    8800 B |
| ModuleNoOpObjReadCommand       | None   |  69.37 us | 1.373 us | 1.527 us |      - |    5600 B |
| ModuleNoOpProc                 | None   |  39.89 us | 0.794 us | 1.259 us |      - |         - |
| ModuleNoOpTxn                  | None   |  34.98 us | 0.649 us | 1.136 us |      - |         - |
| ModuleJsonGetCommand           | None   | 131.93 us | 2.474 us | 4.884 us |      - |   72800 B |
| ModuleJsonSetCommand           | None   | 235.78 us | 3.061 us | 2.863 us | 0.4883 |  228801 B |


libs/server/Custom/ExpandableMap.cs
```diff
@@ -386,39 +386,39 @@ public bool TryGetFirstId(Func<T, bool> predicate, out int id)
     }
 
     /// <summary>
-    /// Extension methods for ConcurrentExpandableMap
+    /// Extension methods for ExpandableMap
```
badrishc, line 389: As discussed:
* We need one global CCM (at the storeWrapper level)
* CCMS is a session specific wrapper for the global CCM, that caches when necessary/possible
* All access to CCM from a session (both front-end and back-end) would happen via CCMS and not CCM
* We can unify the map into a single latch-free expandable map
* We can share this latch-free ExpandableMap between CCM and CCMS for scalable and efficient access

libs/server/Custom/ExpandableMap.cs
```diff
@@ -189,31 +164,37 @@ public bool TryGetNextIdSafe(out int id)
...
         /// Try to set item by ID
+        /// This method should only be called from a thread-safe context
         /// </summary>
         /// <param name="id">Item ID</param>
         /// <param name="value">Item value</param>
         /// <param name="noExpansion">True if should not attempt to expand the underlying array</param>
-        /// <param name="updateSize">True if should update actual size of the map</param>
         /// <returns>True if assignment succeeded</returns>
-        internal bool TrySetValue(int id, ref T value, bool noExpansion, bool updateSize)
+        private bool TrySetValueUnsafe(int id, ref T value, bool noExpansion)
```
badrishc, line 197: The invariant here should be that we are setting a entry that is currently null. We cannot _change_ an existing entry that is non-null. This is because in `TryGetFirstId`, there would be a race where we might apply the predicate on the old entry and succeed, but by the time we return that id, it is overwritten by a new entry. It might be better to debug assert that this invariant is true.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -1,100 +1,218 @@
...
     {
+        // The instance of the CustomCommandManager that this session is associated with
         readonly CustomCommandManager customCommandManager;
 
-        // These session specific arrays are indexed by the same ID as the arrays in CustomCommandManager
-        ExpandableMap<CustomTransactionProcedureWithArity> sessionTransactionProcMap;
-        ExpandableMap<CustomProcedure> sessionCustomProcMap;
+        // These session specific maps are indexed by the same ID as the arrays in CustomCommandManager
+        // Maps between the transaction ID and a tuple of the per-session transaction procedure and its arity
+        readonly Dictionary<int, (CustomTransactionProcedure, int)> sessionTransactionProcMap;
```
badrishc, line 22: Revert back to maps for now, until we figure out how important the memory savings versus higher CPU tradeoff would be.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -1,100 +1,218 @@
...
+        // Maps between the custom procedure ID and the per-session custom procedure instance
+        readonly Dictionary<int, CustomProcedure> sessionCustomProcMap;
+
+        // Cache of custom command info by command name
+        readonly Dictionary<string, RespCommandsInfo> sessionCustomCommandsInfo;
```
badrishc, line 27: do we need these two dictionaries? seems like we do not need to access them on critical path as the required info should be within the cmd instance?

badrishc (changes requested):
looking good, thanks!

badrishc (approved):

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -1,100 +1,190 @@
...
+        // Maps between the custom procedure ID and the per-session custom procedure instance
         ExpandableMap<CustomProcedure> sessionCustomProcMap;
 
         public CustomCommandManagerSession(CustomCommandManager customCommandManager)
         {
             this.customCommandManager = customCommandManager;
-            sessionTransactionProcMap = new ExpandableMap<CustomTransactionProcedureWithArity>(CustomCommandManager.MinMapSize, 0, byte.MaxValue);
-            sessionCustomProcMap = new ExpandableMap<CustomProcedure>(CustomCommandManager.MinMapSize, 0, byte.MaxValue);
+
+            sessionTransactionProcMap = new ExpandableMap<(CustomTransactionProcedure, int)>(MinMapSize, 0, byte.MaxValue);
```
badrishc, line 33: why not start the expandable map at 0 so that we dont pay the cost of memory in the common case that there are no custom commands? then then grow as 1,2,4, etc.

libs/server/Custom/CustomCommandManagerSession.cs
```diff
@@ -1,100 +1,190 @@
...
     /// <summary>
-    /// Server session for RESP protocol - basic commands are in this file
+    /// Session-specific access to custom command data held by CustomCommandManager
+    /// For both custom procedures and custom transactions, this class maintains a cached instance of the transaction / procedure that can be called directly per-session
+    /// This class also maintains a cache of custom command info and docs to avoid the cost of synchronization when calling CustomCommandManager
     /// </summary>
     internal sealed class CustomCommandManagerSession
     {
+        // Initial size of expandable maps
+        private static readonly int MinMapSize = 8;
```
badrishc, line 18: why not start the expandable map at 0 so that we dont pay the cost of memory in the common case that there are no custom commands? then then grow as 1,2,4, etc.

badrishc (changes requested):
fix memory use by default.

libs/server/Custom/CustomCommandManager.cs
```diff
@@ -13,26 +14,35 @@ namespace Garnet.server
     /// </summary>
     public class CustomCommandManager
     {
+        // Maximum number of custom raw string commands
         private static readonly int MaxCustomRawStringCommands = 256;
 
+        // Initial size of expandable maps
+        private static readonly int MinMapSize = 8;
```
badrishc, line 21: why not start the expandable map at 0 so that we dont pay the cost of memory in the common case that there are no custom commands? then then grow as 1,2,4, etc. (for CCM and CCMS)

libs/server/Custom/CustomCommandManagerSession.cs
TalZaccai: The expandable map does start at 0, MinMapSize is the initial size once you set the first element. I can change MinMapSize to 1.

badrishc (approved):

PR #942, opened 2025-01-22
Support and accelerate bitmap operations on more platforms

PaulusParssinen:
This PR refactors the SIMD logic used in bitmap operations to reduce duplication and simplify control flow. The goal is to improve maintainability and set the stage for consistent performance across architectures.

- Unifies vectorized paths across x86 and ARM64 with the .NET cross-architecture intrinsics.
  - Add `Vector512` path. This like others only lights up and is not DCE'd during JIT import if it is supported _and_ hardware accelerated.
  - These can be consolidated further with [`ISimdVector<TSelf, T>`](https://github.com/dotnet/runtime/issues/76244) in the _future_
- Adds/updates tests to be parameterized
- Implement new `DIFF` operation
- `StackExchange.Redis` to 2.9.25 for the new bitops in tests
- No functional behavior changes intended

<details>
<summary>ARM64 demo on Google Pixel 8a</summary>
Now we can do BITOPs on when running Garnet inside Android VMs!

https://github.com/user-attachments/assets/419262c2-2197-422d-a7ae-36a5ac1e0592
</details>

### Follow-up work:
* ~~Refactor BITCOUNT logic similarly to be cross-platform~~ (https://github.com/microsoft/garnet/pull/1034)  and to follow "modern" .NET SIMD practices


libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
```diff
@@ -44,607 +43,286 @@ public static bool BitOpMainUnsafeMultiKey(byte* dstPtr, int dstLen, byte** srcS
...
-        /// <param name="minSize">Minimum size of source bitmaps.</param>
-        private static void __bitop_multikey_simdX256_or(byte* dstPtr, int dstLen, byte** srcStartPtrs, byte** srcEndPtrs, int srcKeyCount, int minSize)
-        {
-            int batchSize = 8 * 32;
-            long slen = minSize;
-            long stail = slen & (batchSize - 1);
 
-            byte* dstCurr = dstPtr;
-            byte* dstEnd = dstCurr + (slen - stail);
+            // Scalar: 8 bytes x 4
```
prvyk, line 160: Shouldn't there be an 'else' here?

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
```diff
@@ -44,607 +43,286 @@ public static bool BitOpMainUnsafeMultiKey(byte* dstPtr, int dstLen, byte** srcS
...
+                Vectorized256(ref firstKeyPtr, srcStartPtrs, srcKeyCount, ref dstPtr, dstBatchEndPtr);
             }
-        #endregion
-
-        fillTail:
-            #region scalar_1x1    
-            byte* dstMaxEnd = dstPtr + dstLen;
-            int offset = 0;
-            while (dstCurr < dstMaxEnd)
+            else if (Vector128.IsHardwareAccelerated && Vector128<byte>.IsSupported)
```
prvyk, line 150: It would be nice if all three paths were tested. In practice, all of Github's x86 based runners will have Vector256.IsHardwareAcceleated as true, not testing the other paths by default.

(If Garnet had an arm64 runner - which it should anyway - then I strongly suspect it would have Vector128.IsHardwareAccelerated as true.)

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
PaulusParssinen: This explicitly follows original processing, i.e. there is unrolled scalar loop even after vectorized main workhorses, if there is enough left in the shared lengths of the source keys.

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
PaulusParssinen: I agree, it would be nice. I would prefer to have "outerloop" test runs that either manually and/or scheduled with `DOTNET_EnableAVX2=0` or `DOTNET_EnableHWIntrinsic=0` environment variables for all of the tests. 

`ISimdVector` in future .NET versions is gonna push this testing effort back to .NET runtime team hopefully, as the 128 & 256 & other paths are just Vector### templates and if there was non-typo bug in them, it would be bug in the .NET BCL.

prvyk: We can do the same TestProcess solution #1041 ended up with for these 2-3 specific tests (it can be generalized a bit further, but we just need to test the paths once or twice, so adding the same using var p = TestProcess and env.Add() isn't much).

It's also possible to do what the Vector documentation suggests to do with BDN and allow an option to run the entire suit with acceleration off (I suspect these hardware accelerations affect other parts of the program, via library code or JIT).

It would also make sense to have a regular ARM runner, as ARM has other differences like memory model which may be important (the BITCOUNT portable code issue wasn't detected since no runner - despite Garnet offering binaries for these platforms...).

PaulusParssinen: Added out-of-proc tests to exercise these paths (https://github.com/microsoft/garnet/pull/942/commits/1d5672114c662635a796c231d9b689ebb1bcd706)

PaulusParssinen:
Some mixed results for benchmarks on different microarchitectures on some inputs. Secondary goal of this PR is to not regress performance, so still need to investigate that.

PaulusParssinen:
Extremely hard to measure, especially wrt. binary input `TensorPrimitives` fast-path non-temporal vs. temporal store difference in context of micro-benchmarks (i.e. very different from the saturated server CPU state) and with all the micro arch differences. Could probably use some more massaging in many places to find a sweet spot for most of them.

If some performance numbers seem now completely off from before on some important hardware you target, do report!

test/Garnet.test/TestProcess.cs
```diff
@@ -13,29 +13,29 @@ namespace Garnet.test
...
+        private readonly Process process = default;
         private readonly Stopwatch stopWatch = default;
         private readonly LightClientRequest lightClientRequest = default;
 
-        internal GarnetServerTestProcess(out ConfigurationOptions opts,
-                                         Dictionary<string, string> env = default,
-                                         int port = 7000)
+        public ConfigurationOptions Options { get; }
+
+        internal GarnetServerTestProcess(Dictionary<string, string> env = default, int port = 7000)
```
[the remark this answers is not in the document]
PaulusParssinen: It is a internal class for in the test project.

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
```diff
@@ -1,650 +1,356 @@
...
+                    if (srcPtrs[i] < srcEndPtrs[i])
+                    {
+                        d00 = TBinaryOperator.Invoke(d00, *srcPtrs[i]);
+                        srcPtrs[i]++;
+                    }
+                    else if (typeof(TBinaryOperator) == typeof(BitwiseAndOperator))
                     {
-                        d00 |= srcStartPtrs[i][offset];
-                        if (d00 == 0xff) goto writeBack;
+                        d00 = 0;
```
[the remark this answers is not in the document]
PaulusParssinen: That's not a runtime comparison..

libs/common/Numerics/IBinaryOperator.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Garnet.common.Numerics
+{
+    /// <summary>Operator that takes two input values and returns a single value.</summary>
+    public interface IBinaryOperator
+    {
+        /// <summary>
+        /// Computes the binary operation of two scalar values.
+        /// </summary>
+        static abstract T Invoke<T>(T x, T y) where T : IBinaryInteger<T>;
```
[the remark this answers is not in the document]
PaulusParssinen: Not a thing in this case.

PaulusParssinen:
Last CI test failure is from cluster flakiness

libs/common/Numerics/IBinaryOperator.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Numerics;
+using System.Runtime.Intrinsics;
+
+namespace Garnet.common.Numerics
+{
+    /// <summary>Operator that takes two input values and returns a single value.</summary>
+    public interface IBinaryOperator
```
vazois, line 10: Really nice abstraction. I just checked that Redis 8.2 adds additional operations (i.e. DIFF,ANDOR etc.).
Is it possible to implement those also?

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
```diff
@@ -1,650 +1,356 @@
...
-            byte* srcCurr = srcBitmap;
-            byte* srcEnd = srcCurr + (slen - stail);
-            byte* dstCurr = dstPtr;
+                if (op == BitmapOperation.NOT)
+                {
+                    TensorPrimitives.OnesComplement(srcBitmap, dstBitmap);
+                }
+                else
+                {
+                    srcBitmap.CopyTo(dstBitmap);
```
vazois, line 46: If we have only one source vector, isn't it always guaranteed that the operation we are performing is NOT? In that case, it should be a Debug assertion or exception here.

vazois (approved):
Great work! Please add some missing operators that have been recently added with Redis 8.2

PaulusParssinen:
Did some thinking on the new operators. `DIFF` is trivial to add as just `x & ~y`  compute kernel. I'll file feature request for the remaining as they are more involved and which I have to try work out with some pen and paper first. I'm not even sure if there's any big opportunities over naive scalar implementation on those yet.

<sub>Otherwise I fear it'll soon be like 2026 dec and non-AVX CPUs will still be crashing from the unsupported instructions </sub>

libs/server/Resp/Bitmap/BitmapManagerBitOp.cs
PaulusParssinen: I recall testing against RESP reference implementations such as Valkey and Redis early on and they behave exactly like this. Just confirmed and for the currently implemented binary BITOPs it indeed works like copy. 

**BUT** the new additions do require two source keys, so I have to return the error for the `DIFF` that is trivial to add in this PR.

libs/common/Numerics/IBinaryOperator.cs
PaulusParssinen: It's excellent design and is taken many steps further in the `S.N.TensorPrimitives` by none other than Tanner Gooding et al. from who I ~~stole~~ learned it!  

TIL new BITOPs! Will add `DIFF` and file issue for the rest as enchancement.

vazois:
@PaulusParssinen, I think this PR has introduced some braking tests. Can you take a look what is the issue?
It seems to be breaking at :

Failed Garnet.test.GarnetBitmapTests.BitmapSimpleBitCountTest("DOTNET_EnableAVX2=0")

Might have to do with the fast that the testing machines do not have support for AVX2?!

https://github.com/microsoft/garnet/actions/runs/19355003613/job/55374605852#step:6:52

PR #943, opened 2025-01-22
Fix issue with unassigned worker initialization when restoring ClusterConfig

Mathos1432:
We saw issues when running the forget command that would cause the different nodes to return invalid information where it looked like a different node was responding. This was caused due to the Role of worker[0] in the cluster config being set to primary (by default) since it was not initialized in this constructor and we skipped initialization of worker 0 here:
https://github.com/microsoft/garnet/blob/27ab5a8227700a3170303d0f939c4f81f35b32e7/libs/cluster/Server/ClusterConfigSerializer.cs#L106

This then caused the check to see if we knew about the node in the forget command to always be true and we would delete the node that was at index 0 in the array, which would cause all kinds of issue down the line.
https://github.com/microsoft/garnet/blob/c85e281acede27498f239dab41c3f28684abfa57/libs/cluster/Server/ClusterManagerWorkerState.cs#L64C1-L65C1

This was validated with a new test and copying the binaries on a broken cluster and verifying that it no longer showed the issue after restarting it.


kcheekuri (approved):

test/Garnet.test.cluster/ClusterConfigTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            Assert.That(address == "unassigned");
+            Assert.That(port == 0);
+            Assert.That(NodeRole.UNASSIGNED == restoredConfig.GetNodeRoleFromNodeId("asdasdqwe"));
+        }
+
+        [Test, Order(2)]
+        [Category("CLUSTER-CONFIG"), CancelAfter(1000)]
+        public void ClusterForgetAfterNodeRestartTest()
+        {
+            int nbInstances = 4;
```
vazois, line 67: please fix this warning

vazois (approved):

PR #949, opened 2025-01-23
Fix HEXPIRE test case failure in pipeline

### LINKED ISSUE #948: Pipeline failing on HEXPIREAT Test
### Describe the bug

The pipeline is failing on the following test:
https://github.com/microsoft/garnet/actions/runs/12934431504/job/36075533943

Run the pipeline multiple time ideally using a small number of cores.
It seems this might be a timing issue due to the nature of the command being tested.

### Expected behavior

Test should not fail

### Screenshots

Image

### Release version

_No response_

### IDE

_No response_

### OS version

_No response_

### Additional context

_No response_

Vijay-Nirmal:
I am not sure about the cause of the failure as the test runtime is not more than 1 sec to trigger this issue. Increased the expiration time to avoid the possibility of it getting expired.

fix #948

### DIFF
--- a/test/Garnet.test/RespHashTests.cs
+++ b/test/Garnet.test/RespHashTests.cs
@@ -1243,10 +1243,10 @@ namespace Garnet.test
 
             (var expireTimeField1, var expireTimeField3, var newExpireTimeField) = command switch
             {
-                "HEXPIRE" => ("2", "4", "3"),
-                "HPEXPIRE" => ("2000", "4000", "3000"),
-                "HEXPIREAT" => (DateTimeOffset.UtcNow.AddSeconds(2).ToUnixTimeSeconds().ToString(), DateTimeOffset.UtcNow.AddSeconds(4).ToUnixTimeSeconds().ToString(), DateTimeOffset.UtcNow.AddSeconds(3).ToUnixTimeSeconds().ToString()),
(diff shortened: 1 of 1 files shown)


badrishc:
> I am not sure about the cause of the failure as the test runtime is not more than 1 sec to trigger this issue. Increased the expiration time to avoid the possibility of it getting expired.
> 
> fix #948

Did you see the failure log to verify the failed test did not take more than a second? Then we need to diagnose this further as this fix isn't the culprit

Vijay-Nirmal:
@badrishc Yes, I have mentioned the same thing in the attached issue. At the time of raising this PR, it was just a quick fix to unblock the pipeline, also I was pretty confident that it's not an issue from the actual expiration logic.

Now I have diagnosed this issue further. The issue is with the code in the test file only. The culprit is the `DateTimeOffset.UtcNow.AddSeconds(1).ToUnixTimeSeconds().ToString()` line in unit testing. This line can return a time that is less than 1 sec from the current time, even it can be almost the same as the current time. Below is an example of when that can happen.

Current Time: 01:01:00.9980000
Current Time + 1 sec: 01:01:01.9980000
Current Time + 1 sec as Seconds: 01:01:01 (assume this timestamp as seconds, this value is almost the same as current time)

This happens because `ToUnixTimeSeconds` method counts the number of full seconds in provided time, basically it works like `Math.Floor` for seconds.

Provided the able reasoning, I can say the fix provided in the PR is good enough to solve this issue

badrishc (approved):

PR #950, opened 2025-01-23
Fix linsert before handling (microsoft#945)

prvyk:
ListObjectImpl made a case-sensitive comparison that ignored non-uppercase 'before' in LINSERT command. Do a proper comparison as advised by @badrishc . Also add a testcase to ensure this doesn't break in the future.


prvyk:

TalZaccai (approved):
Thanks for finding and fixing this issue!

PR #951, opened 2025-01-24
SETNX command should return integer value

prvyk:
https://redis.io/docs/latest/commands/setnx/

Per redis docs, SETNX command should return 1 or 0. However, garnet translates it internally to SET with NX flag and therefore incorrectly returns SET return values ("OK" or nil). Fix by inlining the correct code path into NetworkSETNX function and then returning the correct response. Also add test and fix existing test.

### DIFF
--- a/libs/server/Resp/BasicCommands.cs
+++ b/libs/server/Resp/BasicCommands.cs
@@ -444,10 +444,8 @@ namespace Garnet.server
             var input = new RawStringInput(RespCommand.SETEXNX, ref parseState, startIdx: 1, arg1: 0);
             var status = storageApi.SET_Conditional(ref sbKey, ref input);
 
-            // the status returned for SETNX as NOTFOUND is the expected status in the happy path, so flip the ok flag
-            bool ok = status == GarnetStatus.NOTFOUND;
-
-            if (ok)
+            // The status returned for SETNX as NOTFOUND is the expected status in the happy path
+            if (status == GarnetStatus.NOTFOUND)
             {
                 while (!RespWriteUtils.TryWriteInt32(1, ref dcurr, dend))
                     SendAndReset();
--- a/libs/server/Resp/BasicCommands.cs
+++ b/libs/server/Resp/BasicCommands.cs
@@ -440,16 +440,9 @@ namespace Garnet.server
             var status = storageApi.SET_Conditional(ref sbKey, ref input);
 
             // The status returned for SETNX as NOTFOUND is the expected status in the happy path
-            if (status == GarnetStatus.NOTFOUND)
-            {
-                while (!RespWriteUtils.TryWriteInt32(1, ref dcurr, dend))
-                    SendAndReset();
-            }
-            else
-            {
-                while (!RespWriteUtils.TryWriteInt32(0, ref dcurr, dend))
-                    SendAndReset();
-            }
+            var retVal = status == GarnetStatus.NOTFOUND ? 1 : 0;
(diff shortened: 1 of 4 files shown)


Vijay-Nirmal (approved):
Nice catch

test/Garnet.test/RespSetTest.cs
```diff
@@ -1390,6 +1390,26 @@ public void CanDoSinterCardLC(string values1, string values2, string expectedCou
             ClassicAssert.AreEqual(expectedResponse, response.AsSpan().Slice(0, expectedResponse.Length).ToArray());
         }
 
+        [Test]
```
TalZaccai, line 1393: RespSetTest.cs is meant for tests of the set object, please move this test to RespTests.cs

libs/server/Resp/BasicCommands.cs
```diff
@@ -438,7 +438,25 @@ private bool NetworkSETNX<TGarnetApi>(bool highPrecision, ref TGarnetApi storage
             var getOption = ArgSlice.FromPinnedSpan(CmdStrings.NX);
             parseState.InitializeWithArguments(key, value, getOption);
```
TalZaccai, line 439: If you're calling SET_Conditional directly there's no need to initialize a new parseState, as it only uses the set value, so esentially you can get rid of lines 437-439.

libs/server/Resp/BasicCommands.cs
```diff
@@ -438,7 +438,25 @@ private bool NetworkSETNX<TGarnetApi>(bool highPrecision, ref TGarnetApi storage
...
 
-            return NetworkSETEXNX(ref storageApi);
+            var nkey = parseState.GetArgSliceByRef(0);
```
TalZaccai, line 441: No need for this nKey, you can use key

libs/server/Resp/BasicCommands.cs
```diff
@@ -438,7 +438,25 @@ private bool NetworkSETNX<TGarnetApi>(bool highPrecision, ref TGarnetApi storage
...
+            var sbKey = nkey.SpanByte;
+
+            var input = new RawStringInput(RespCommand.SETEXNX, ref parseState, startIdx: 1, arg1: 0);
```
TalZaccai, line 444: no need to set arg1 here, it's optional

libs/server/Resp/BasicCommands.cs
```diff
@@ -438,7 +438,25 @@ private bool NetworkSETNX<TGarnetApi>(bool highPrecision, ref TGarnetApi storage
...
+            var status = storageApi.SET_Conditional(ref sbKey, ref input);
+
+            // The status returned for SETNX as NOTFOUND is the expected status in the happy path
+            if (status == GarnetStatus.NOTFOUND)
```
TalZaccai, line 448: nit: slightly cleaner so you don't have to duplicate the write code:
```cs
var retVal = status == GarnetStatus.NOTFOUND ? 1 : 0;
while (!RespWriteUtils.TryWriteInt32(retVal, ref dcurr, dend))
    SendAndReset();
```

TalZaccai (changes requested):
Good catch! A few comments...

libs/server/Resp/BasicCommands.cs
prvyk: A previous version already tried this but failed the new test in the GET stage - it returned a different serialization - and was force pushed out. Well, now it works. I must have had some mistake elsewhere then...

TalZaccai (approved):

PR #953, opened 2025-01-24
Switching Enum.HasFlag calls with bitwise operations

TalZaccai:
Switching a few areas in the code from using Enum.HasFlag (which apparently uses reflection and thus incur overhead) to using more efficient bitwise operations.

### DIFF
--- a/libs/server/Objects/Types/GarnetObjectStoreOutput.cs
+++ b/libs/server/Objects/Types/GarnetObjectStoreOutput.cs
@@ -50,6 +50,16 @@ namespace Garnet.server
         /// </summary>
         public ObjectStoreOutputFlags OutputFlags;
 
+        /// <summary>
+        /// True if output flag WrongType is set
+        /// </summary>
+        public bool IsWrongType => (OutputFlags & ObjectStoreOutputFlags.WrongType) == ObjectStoreOutputFlags.WrongType;
+
+        /// <summary>
+        /// True if output flag RemoveKey is set
+        /// </summary>
+        public bool RemoveKey => (OutputFlags & ObjectStoreOutputFlags.RemoveKey) == ObjectStoreOutputFlags.RemoveKey;
+
         public void ConvertToHeap()
         {
             // Does not convert to heap when going pending, because we immediately complete pending operations for object store.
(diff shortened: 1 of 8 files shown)


libs/server/Objects/Types/GarnetObjectStoreOutput.cs
```diff
@@ -50,6 +50,16 @@ public struct GarnetObjectStoreOutput
...
 
+        /// <summary>
+        /// True if output flag WrongType is set
+        /// </summary>
+        public bool IsWrongType => (OutputFlags & ObjectStoreOutputFlags.WrongType) == ObjectStoreOutputFlags.WrongType;
+
+        /// <summary>
+        /// True if output flag RemoveKey is set
+        /// </summary>
+        public bool RemoveKey => (OutputFlags & ObjectStoreOutputFlags.RemoveKey) == ObjectStoreOutputFlags.RemoveKey;
```
TedHartMS, line 61: C;hange name to IsRemoveKey (or HasRemoveKey)

TedHartMS (changes requested):

TedHartMS (approved):

PaulusParssinen:
`Enum.HasFlag` has been intrinsic [since 2017](https://github.com/dotnet/coreclr/pull/13748). The codegen _should_ be same as bitwise check. It doesn't have to do reflection if the call `HasFlag` call targets the generic overload.

TalZaccai:
> `Enum.HasFlag` has been intrinsic [since 2017](https://github.com/dotnet/coreclr/pull/13748). The codegen _should_ be same as bitwise check. It doesn't have to do reflection if the call `HasFlag` call targets the generic overload. I suggest reverting this PR.

Interesting... I've definitely seen unnecessary allocations when running a profiler, which I don't see any longer, so I'm not sure why that would be the case.

PaulusParssinen:
> Interesting... I've definitely seen unnecessary allocations when running a profiler, which I don't see any longer, so I'm not sure why that would be the case.

It shouldn't be boxing in even in debug mode or Tier-0 codegen anymore (I could be wrong here). But if there clearly is difference with profiler runs I think the changes should then stay. It definitely gets eliminated in Tier-1 (i.e. fully optimized hot code) 😅

TalZaccai:
That's good to know! Thanks for looking into it!

PR #956, opened 2025-01-25
BUGFIX: Fix setting srcRecordInfo during interleaving of checkpointing and RMW

hamdaankhalid:
During checkpointing and RMW interleaving the goto CreateNewRecord from CPR shift detection would be called without any srcRecordInfo, this would further cause an issue in CreateNewRecordRMW method since an in-memory source from something that was tombstoned in mutable region would cause us a null reference exception becuause it would be sent down CopyUpdate instead of initialUpdate...

Super rare case that we hit at scale...


TedHartMS (approved):
Nice catch!

PR #959, opened 2025-01-26
Add ACL GETUSER Command Initial Implementation

### LINKED ISSUE #958: Add Support for ACL GETUSER Command
### Feature request type

sample request

Currently Garnet Cache does not support the `ACL GETUSER` command. This command is used to obtain the ACL rules for a specified user. This command is useful for inspecting individual identities authorized to access the cluster.

I would like to be able to issue a command against Garnet in the form of `ACL GET USER {USERNAME}`. The command would return details about the specified users ACL in a RESP protocol compliant array with this output:

```
1) "flags"
2) 1) "on"
3) "passwords"
4) 1) "#**REDACTED**"
5) "commands"
6) "+get +incr"
```

In the case where the user does not exist the command should return null.

Currently the `ACL LIST` command is supported. A client interested in this information could parse the response of the list command for similar information about a user's ACL. However, with `ACL GETUSER` this parsing is not required.

### Additional context

I have a forthcoming contribution/PR that implements this command for the team's consideration for inclusion into the project.

kevinmichaelbowersox:
This PR provides the implementation of the `ACL GETUSER` command which is used to obtain ACL rules for a specified user. The change is accompanied by tests for the command.

The changeset includes:
- [x] ACL GETUSER command implementation
- [x] Unit Tests and ACL Test
- [x] Documentation

This resolves #958 where further details can be found. 

Additionally a few instances of unnecessary whitespace were cleaned up.


libs/server/ACL/User.cs
```diff
@@ -420,6 +420,33 @@ public string DescribeUser()
             return stringBuilder.ToString();
         }
 
+        /// <summary>
+        /// Returns flags for the <see cref="User"/>.
+        /// </summary>
+        /// <returns>A <see cref="List{T}"/> of <see cref="string"/> representing flags for the user.</returns>
+        public List<string> GetFlags()
```
kevin-montrose, line 427: Why allocate here?  It's just switching between two constants.

libs/server/ACL/User.cs
```diff
@@ -420,6 +420,33 @@ public string DescribeUser()
...
+            return new() { IsEnabled ? "on" : "off" };
+        }
+
+        /// <summary>
+        /// Returns password hashes for the <see cref="User"/>.
+        /// </summary>
+        /// <returns>A <see cref="List{T}"/> of <see cref="string"/> representing password hashes for the user.</returns>
+        public List<string> GetPasswordHashes()
+        {
+            return _passwordHashes.Select(hash => $"#{hash}").ToList();
```
kevin-montrose, line 438: This is not thread safe, _passwordHashes is mutable.

Also, there's no need to allocate here, we're just prepending a `#` so that can be done while writing the response out.

libs/server/Resp/ACLCommands.cs
```diff
@@ -366,5 +366,90 @@ private bool NetworkAclSave()
...
+                {
+                    while (!RespWriteUtils.TryWriteNull(ref dcurr, dend))
+                        SendAndReset();
+                }
+                else
+                {
+                    while (!RespWriteUtils.TryWriteArrayLength(6, ref dcurr, dend))
+                        SendAndReset();
+
+                    var flags = user.GetFlags();
```
kevin-montrose, line 417: I think broadly this will run into issues if the user is modified while the command is running.

GetFlags(), GetPasswordHashes(), and GetEnabledCommandsDescription() won't necessarily return consistent results.

kevin-montrose (changes requested):

libs/server/ACL/User.cs
kevinmichaelbowersox: Are you onboard with making this field accessible to clients to avoid the allocations?

libs/server/Resp/ACLCommands.cs
kevinmichaelbowersox: I see that `enabledCommands` has a copy method. What are your thoughts on returning a copy of the commands back to the client to avoid any issues with modifications?

libs/server/ACL/User.cs
kevin-montrose: Given the thread safety concerns, I don't think exposing the field is appropriate.

libs/server/Resp/ACLCommands.cs
kevin-montrose: I think some variant of "one method (that internally takes appropriate) locks, and copies the relevant data into the return"-is the only path forward.

I'd be tempted to try and fit things into Span's created by the caller to avoid allocations (or rent some arrays, or something), so the there's some flexibility in the implementation.

kevinmichaelbowersox:
Closing this PR in favor of its successor found here: https://github.com/microsoft/garnet/pull/1038

PR #964, opened 2025-01-27
Adding ROLE command

prvyk:
An initial attempt to add the ROLE command + docs + tests. Likely this will need some revisions.

https://redis.io/docs/latest/commands/role/

1. After #961 the new ACL test blows up (it passed previously). I suspect that introduced an unrelated regression affecting this test, since we just use normal serialization methods.
2. I'm not sure the replication state check is correct at all. Note we check the new IsPrimary() since the recovering case should go to the replica branch unlike normal IsReplica().
3. Didn't mention 'sentinel' possibility in docs, as Garnet does not support this mode.
4. Far from sure this is the best code organization, it seemed cleaner than doing the logic in one method.
5. Minor effect on Metrics code path (see first commit), but with same result.


badrishc:
I don't think the unit test error is related to #961. It might be related to #925. The issue is that `GarnetClient` cannot handle an array response where one of the array elements is itself another array. It is possible that prior to #925, we were swallowing this somehow?

```
RECV: [*1|$4|ROLE|]
SEND: [*3|$6|master|:0|*0|]
Exception thrown: 'Garnet.common.Parsing.RespParsingException' in Garnet.common.dll
Unexpected character '*'.
```

One solution would be to add basic support for an array field in RespReadResponseUtils.cs line 192, something like this:

```cs
                else if (*ptr == ':')
                {
                    if (!TryReadIntegerAsString(out result[i], ref ptr, end))
                        return false;
                }
                else if (*ptr == '*')
                {
                    if (!TryReadStringArrayWithLengthHeader(out var subArray, ref ptr, end))
                        return false;
                    result[i] = string.Join(", ", subArray);
                }
                else
                {
                    RespParsingException.ThrowUnexpectedToken(*ptr);
                }
```

prvyk:
Thanks. Added this to unblock review for now. Perhaps this should be fixed separately, if it is it would be easy to revert/force push/ to get that out from the PR.

badrishc:
Feel free to split this change out to its own PR. It is not perfect, but we don't really plan to make GarnetClient perfect for all cases anyway. it just needs to handle these nestings gracefully.

libs/common/Metrics/ReplicaInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Garnet.common
+{
+    public struct ReplicaInfo
```
vazois, line 6: Why don't we consolidate ReplicaInfo and RoleReplicaInfo, under one class RoleInfo? It seems most of the fields are shared.

vazois: Also, we need to add this under libs/cluster/server/

libs/common/Metrics/ReplicaInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public long offset;
```
vazois, line 8: Add comment describing each public parameter.

test/Garnet.test.cluster/ClusterRoleTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test.cluster
+{
+    [TestFixture, NonParallelizable]
+    public class ClusterRoleTests
+    {
+        ClusterTestContext context;
+        readonly int defaultShards = 3;
```
vazois, line 20: Please fix warning

test/Garnet.test.cluster/ClusterRoleTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test.cluster
```
vazois, line 14: Please add this under ClusterManagementTests.

test/Garnet.test.cluster/ClusterRoleTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            context.TearDown();
+        }
+
+        [Test]
+        public void ClusterTestRoleCommand()
```
vazois, line 38: Please add a test for CLUSTER NODES also.

vazois (changes requested):
Great work! Some minor comments

libs/common/Metrics/ReplicaInfo.cs
prvyk: Good idea, consolidated and added comments. However, under libs/cluster/server IClusterProvider can't compile. Moved it to libs/server/cluster and the build passes. Is that fine?

test/Garnet.test.cluster/ClusterRoleTests.cs
prvyk: This is unrelated. Added it anyway. 

I'm uncertain whether CLUSTER NODES output should end with a newline (Either redis doesn't do it, or their doc page omits it), but that's for a separate PR.

prvyk:
The new CLUSTER NODES test has an interesting Linux-only failure... CLUSTER NODES is unrelated to this PR, I think it should get its own?

test/Garnet.test.cluster/ClusterRoleTests.cs
prvyk: The new CLUSTER NODES test has an interesting Linux-only failure... CLUSTER NODES is unrelated to this PR, I think it should get its own?

libs/common/Metrics/ReplicaInfo.cs
vazois: yes, it fine to put it there.

test/Garnet.test.cluster/ClusterRoleTests.cs
vazois: yes, you are right. I forgot that CLUSTER NODES uses the ConnectionStore and not the AofTaskStore. You can revert this.

vazois:
> The new CLUSTER NODES test has an interesting Linux-only failure... CLUSTER NODES is unrelated to this PR, I think it should get its own?

Yes, revert the change it is not related to this PR.

vazois (approved):

libs/cluster/Server/Replication/PrimaryOps/AofTaskStore.cs
```diff
@@ -63,27 +64,30 @@ internal void SafeTailShiftCallback(long oldTailAddress, long newTailAddress)
             }
         }
 
-        public List<(string, string)> GetReplicaInfo(long PrimaryReplicationOffset)
+        public List<(int count, RoleInfo roleInfo)> GetReplicaInfo(long PrimaryReplicationOffset)
         {
             // secondary0: ip=127.0.0.1,port=7001,state=online,offset=56,lag=0
-            List<(string, string)> replicaInfo = new List<(string, string)>();
+            List<(int, RoleInfo)> replicaInfo = new(numTasks);
```
vazois, line 70: I just noticed this, why do you generate the offset here? Can't you just do it at the ClusterProvider on demand? Since we generate the response later we don't have to track the offset here.

libs/cluster/Server/ClusterProvider.cs
```diff
@@ -249,12 +253,49 @@ public MetricsItem[] GetReplicationInfo()
...
 
+        /// <inheritdoc />
+        public (long replication_offset, RoleInfo[] replicaInfo) GetPrimaryInfo()
+        {
+            if (!serverOptions.EnableCluster)
+            {
+                return (0, default);
+            };
+
+            return (replicationManager.ReplicationOffset, replicationManager.GetReplicaInfo().Select(w => w.roleInfo).ToArray());
```
vazois, line 270: It is best to avoid allocations where necessary. ToArray performs an unecessary allocation. Also, you iterate through the list to extract the Role information. If you change the GetReplicaInfo to return List<RoleInfo> then you will not need either and can just return the list to the caller and iterate to produce the response.

vazois (changes requested):
I just noticed some more things. Please address.

libs/cluster/Server/ClusterProvider.cs
prvyk: Both have the same underlying reason: ReplicationManager.GetReplicaInfo was also used by Metrics code and I used the same function for both uses instead of splitting the code paths.

I split it up (restoring the previous Metrics code, only the function name is slightly different), and now we only use the List .

libs/cluster/Server/Replication/PrimaryOps/AofTaskStore.cs
prvyk: Changed to get the offset directly from the replicationmanager.

vazois (approved):

PR #970, opened 2025-01-28
Separate device factory into a creator layer

badrishc:
Separate device factory into a creator layer so that expensive connection costs can be incurred once, for device types such as cloud storage. Workflow now is:

(1) Instantiate a factory creator
(2) Ask creator for a factory, given a base path
(2) Ask factory for devices instances

Fixes https://github.com/microsoft/garnet/issues/739.


badrishc:
Note: It seems that the AzureStreamProvider currently creates its own "factory creator" using the provided connection string (even before GarnetServerOptions are created), so it is possible we instantiate the connection twice in case user were reading the config file from Azure. But this seems to be a rare case, so did not optimize for it.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -277,9 +277,9 @@ public class GarnetServerOptions : ServerOptions
         public int NetworkConnectionLimit = -1;
 
         /// <summary>
-        /// Creator of device factories
+        /// Func to help create device factories
```
TalZaccai, line 280: nit: looks like the comment wasn't updated properly?

libs/storage/Tsavorite/cs/src/core/Index/CheckpointManagement/LocalStorageNamedDeviceFactoryCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        readonly bool preallocateFile;
+        readonly bool deleteOnClose;
+        readonly int? throttleLimit;
+        readonly bool disableFileBuffering;
+        readonly bool useNativeDeviceLinux;
+        readonly bool readOnly;
+        readonly ILogger logger;
+
+        public LocalStorageNamedDeviceFactoryCreator(bool preallocateFile = false, bool deleteOnClose = false, bool disableFileBuffering = true, int? throttleLimit = null, bool useNativeDeviceLinux = false, bool readOnly = false, ILogger logger = null)
```
TalZaccai, line 18: Would be nice to have an XML comment that describes each parameter

TalZaccai (approved):
LGTM!

PR #971, opened 2025-01-29
Add support for slowlog

badrishc:
* Added support for `SLOWLOG [HELP|GET|LEN|RESET]` commands.
* Support options of `SlowLogThreshold` and `SlowLogMaxEntries`.


libs/host/Configuration/Options.cs
```diff
@@ -312,6 +312,14 @@ internal sealed class Options
         [Option("latency-monitor", Required = false, HelpText = "Track latency of various events.")]
         public bool? LatencyMonitor { get; set; }
 
+        [IntRangeValidation(0, int.MaxValue)]
```
TalZaccai, line 315: Since these parameters also exist in `redis.conf`, you can consider adding them to our `RedisOptions` class that converts from Redis to Garnet parameters. If you want to test that the conversion was done correctly, you can uncomment these parameters in `Garnet.test/redis.conf` and add an assert to `GarnetServerConfigTests.ImportExportRedisConfigLocal` that captures the values of these.

libs/host/Configuration/Options.cs
```diff
@@ -630,6 +638,11 @@ public GarnetServerOptions GetServerOptions(ILogger logger = null)
                 CompactionForceDelete = true;
             }
 
+            if (SlowLogThreshold > 0)
+            {
+                if (SlowLogThreshold < 100)
```
TalZaccai, line 643: You can add the min range in the `IntRangeValidation` attribute of this property

libs/server/Metrics/Slowlog/RespSlowlogCommands.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// <summary>
+        /// Processes SLOWLOG HELP subcommand.
+        /// </summary>
+        /// <returns>true</returns>
+        private bool NetworkSlowLogHelp()
+        {
+            // No additional arguments
+            if (parseState.Count != 0)
+            {
+                while (!RespWriteUtils.TryWriteError($"ERR Unknown subcommand or wrong number of arguments for SLOWLOG HELP.", ref dcurr, dend))
```
TalZaccai, line 23: Call AbortWithWrongNumberOfArguments instead (same comment for all methods in this file)

libs/server/Metrics/Slowlog/RespSlowlogHelp.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Info on what slow log commands are supported
+    /// </summary>
+    static class RespSlowLogHelp
+    {
+        /// <summary>
+        /// Get supported latency commands and a short description
+        /// </summary>
+        /// <returns></returns>
+        public static List<string> GetSlowLogCommands()
```
TalZaccai, line 17: This can probably be a lazily instantiated static array of strings?

libs/server/Metrics/Slowlog/SlowLogContainer.cs
```diff
@@ -0,0 +1,67 @@
...
+                logEntries.TryDequeue(out _);
+            }
+        }
+
+        /// <summary>
+        /// Clear the slow log buffer
+        /// </summary>
+        public void Clear()
+        {
+            while (logEntries.TryDequeue(out _)) { }
```
TalZaccai, line 53: Why not call logEntries.Clear()?

TalZaccai (approved):
Some minor comments, generally looks good to me

libs/host/Configuration/Options.cs
badrishc: But it can also be zero, so 100 cannot be the min of the range, right?

libs/host/Configuration/Options.cs
badrishc: Done, please check my commit on this.

libs/server/Metrics/Slowlog/RespSlowlogHelp.cs
badrishc: leaving as is, because other HELP messages use this paradigm, and it is wasteful to leave the static help strings in memory after the call.

libs/server/Metrics/Slowlog/SlowLogContainer.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Get a snapshot of the latest count entries in the slow log (-1 returns all entries)
+        /// </summary>
+        public List<SlowLogEntry> GetEntries(int count)
+        {
+            List<SlowLogEntry> entries = [.. logEntries];
+            if (count < 0 || count >= entries.Count)
+                return entries;
+            return entries.GetRange(entries.Count - count, count);
```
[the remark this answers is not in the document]
badrishc: GetRange does exactly this. The initial extraction (`entries = [.. logEntries]`) is necessary because we need the latest count entries, which queue does not support directly without enumerating all the items.

PR #978, opened 2025-01-30
Add SortedSetCount to IGarnetApi

badrishc:
We already have `SortedSetCount` exposed via network as `ZCOUNT` but there was no way to access it programmatically from the custom stored procedure.

This PR adds the `SortedSetCount` API to IGarnetApi, and uses it in a sample test custom transaction (stored proc) [test/Garnet.test/Extensions/SortedSetCountTxn.cs](https://github.com/microsoft/garnet/blob/main/test/Garnet.test/Extensions/SortedSetCountTxn.cs).

We can also use this PR as a template for how to add API calls to IGarnetApi in future:
* Add method signature to IGarnetApi or IGarnetReadApi interface
* Add implementation of API to StorageSession (e.g. SortedSetOps.cs)
* Add forwarder methods to GarnetApi and GarnetWatchApi (if it is a read operation) struct wrappers

Fixes https://github.com/microsoft/garnet/issues/979


test/Garnet.test/Extensions/SortedSetCountTxn.cs
```diff
@@ -0,0 +1,31 @@
...
+using Tsavorite.core;
+
+namespace Garnet
+{
+    sealed class SortedSetCountTxn : CustomTransactionProcedure
+    {
+        public override bool Prepare<TGarnetReadApi>(TGarnetReadApi api, ref CustomProcedureInput input)
+        {
+            int offset = 0;
+            AddKey(GetNextArg(ref input, ref offset), LockType.Exclusive, true);
```
TalZaccai, line 15: Can't this be a shared lock since it's a read op?

TalZaccai (approved):

PR #985, opened 2025-01-31
Add ZRANGEBYLEX command synonym.

prvyk:
ZRANGEBYLEX was listed as supported by the website docs but actually was not. Link to ZRANGE BYLEX in the same way ZRANGEBYSCORE does with BYSCORE.


test/Garnet.test/Resp/ACL/RespCommandTests.cs
```diff
@@ -6403,6 +6403,27 @@ static async Task DoZRangeStoreAsync(GarnetClient client)
             }
         }
 
+        [Test]
```
TalZaccai, line 6406: We probably should have a test or two for this command (even if it's deprecated), please see `RespSortedSetTests.CanDoZRangeByScoreLC` & `RespSortedSetTests.CanDoZRangeByScoreWithLimitLC` as reference

TalZaccai (changes requested):
Thanks your contribution! Added a comment about needing tests...

test/Garnet.test/Resp/ACL/RespCommandTests.cs
prvyk: Done. 

Didn't do it earlier since it's a synonym to ZRANGE BYLEX which was already tested. It's better to test though. Also added bylex testing to AddWithOptions().

TalZaccai (approved):

PR #986, opened 2025-02-01
Regression Fix: RMW for checking record location on hlog, and remove assignment in immutable region in Upsert during CPR

hamdaankhalid:
As a part of my last PR making a change in RMW to set srcRecordInfo correctly, I also added a change in Upsert route.
However, the RMW was missing the fuzzy region, and the Upsert was erroneously trying to set SrcRecordInfo even when record in readonly region of hybrid log, when it should only do so for mutable region of hybrid log.
 
HeadAddress may change value while we are still holding the Epoch during InternalRMW.
Although the page won't be evicted, we test the record's poisition being in the inmemory region using HeadAddress, during checkpointing, this bumping of headAddress can produce the below scenario:

SafeHeadAddress <= record.Logicaladdress < HeadAddres

As for the change in upsert, we only scan from tail to READONLYOFFSET, but our checking was also checking for immutable region, this means if record was in immutable region we dont set physical address and then we try to assign it a value using unsafeRefCasting in GetInfo method, resulting in setting srcRecordInfo to null. This then produces error in calling Seal!


TedHartMS (commented):
Per discussion, it seems possible that the HeadAddress test at line 145 of InternalUpsert (and corresponding in InternalRMW) may fail if HeadAddress is incremented after HasMainLogSrc is set. So the test should be of HasMainLogSrc instead.

badrishc:
Closing as this was not updated for a while, in favor of https://github.com/microsoft/garnet/pull/1071

PR #987, opened 2025-02-01
Accept lowercase SCRIPT subcommands

prvyk:
script load return -> ERR unknown command
SCRIPT LOAD return -> "63143b6f8007b98c53ca2149822777b3566f9241"
SCRIPT load return -> ERR unknown command
script LOAD return -> "63143b6f8007b98c53ca2149822777b3566f9241"

What happens is that the second pass uppercases SCRIPT, but can't uppercase the subcommand. Later the 'fast' check only checks the uppercase FLUSH/EXISTS/LOAD.

Solve this by adding the script command also to the 'slow' parsing step. This handles all the mixed-cases and matches the MEMORY command handling, while retaining 'fast' parsing for uppercase subcommands. Garnet already does the 'you get slightly slower parsing if you do lowercase' tradeoff .

The second commit standardizes subcommand 'slow' handling a bit: always the correct error message when no subcommand and subcommand is required, always the correct error message when non-supported subcommand is used. Add and use an error message not mentioning HELP subcommand when that subcommand is not implemented.


badrishc (approved):

badrishc:
could you resolve conflicts with main?

kevin-montrose (approved):

PR #989, opened 2025-02-01
Improve ACL SETUSER Thread Handling

### LINKED ISSUE #988: Improve ACL SETUSER Thread Handling
### Describe the bug

While analyzing thread safety/handling for [another pull request](https://github.com/microsoft/garnet/pull/959), I discovered improvements for thread handling in the `ACL SETUSER` command implementation.

Currently, the implementation of this command performs in place modifications to the `User` properties and modifies commands within the `CommandPermissionSet` as the ACL input to the command is parsed. In place modification of these objects/properties works well in a single threaded context, however it can produce inconsistent results in a multithreaded context due to modifications to shared state. ([see code described here](https://github.com/microsoft/garnet/blob/a160787b307420100f9e2e1cfd5cbf2161a83067/libs/server/Resp/ACLCommands.cs#L148)).

```csharp
        // Please Note: Omitted non-essential code / comments
        private bool NetworkAclSetUser()
        {
                // User object retrieved from ACL shared by all sessions (aka shared state).
                var user = aclAuthenticator.GetAccessControlList().GetUser(username);

                try
                {
                    // Remaining parameters are ACL operations
                    for (var i = 1; i < parseState.Count; i++)
                    {
                        var op = parseState.GetString(i);
                        ACLParser.ApplyACLOpToUser(ref user, op); // Modifications to shared state in multithreaded context.
                    }
                }
                catch (ACLException exception)
                // Rest of method...
```

In this scenario, the `User` (aka shared state) is stored in memory within a `ConcurrentDictionary` located within the `AccessControlList`. I believe the `AccessControlList` is shared across all instances of `RespServerSession` ([see code reference here](https://github.com/microsoft/garnet/blob/a160787b307420100f9e2e1cfd5cbf2161a83067/libs/server/ACL/AccessControlList.cs#L24)). If multiple clients concurrently call `ACL SETUSER` for the same user it can produce inconsistent results that combine input arguments from multiple invocations of `ACL SETUSER`.

Additionally, there is a race condition that can occur if multiple clients attempt to issue the initial `ACL SETUSER` command for a user due to contention on the `ConcurrentDictionary` in the `AccessControlList`.

The following unit tests can be used to recreate the issue. I have included these in an accompanying PR that addresses the issue.

```csharp
        /// <summary>
        /// Tests that ACL SETUSER works in parallel without corrupting the user's ACL.
        ///
        /// Test launches multiple clients that apply two simple ACL changes to the same user many times in parallel.
        /// Validates that ACL result after each execution is one of the possible valid responses.
        ///
        /// Race conditions are not deterministic so test uses repeat.
        ///
        /// </summary>
        [TestCase(128, 2048)]
        [Repeat(2)]
        public async Task ParallelAclSetUserTest(int degreeOfParallelism, int iterationsPerSession)
        {
            string activeUserWithGetCommand = $"ACL SETUSER {TestUserA} on >{DummyPassword} +get";
            string inactiveUserWithoutGetCommand = $"ACL SETUSER {TestUserA} off >{DummyPassword} -get";

            string activeUserWithGet = $"user {TestUserA} on #{DummyPasswordHash} +get";
            string inactiveUserWithoutGet = $"user {TestUserA} off #{DummyPasswordHash} -get";
            string inactiveUserWithNoCommands = $"user {TestUserA} off #{DummyPasswordHash}";

            var c = TestUtils.GetGarnetClientSession();
            c.Connect();
            _ = await c.ExecuteAsync(activeUserWithGetCommand.Split(" "));

            // Run multiple sessions that stress AUTH
            await Parallel.ForAsync(0, degreeOfParallelism, async (t, state) =>
            {
                using var c = TestUtils.GetGarnetClientSession();
                c.Connect();

                for (uint i = 0; i < iterationsPerSession; i++)
                {
                    await Task.WhenAll(
                        c.ExecuteAsync(activeUserWithGetCommand.Split(" ")),
                        c.ExecuteAsync(inactiveUserWithoutGetCommand.Split(" ")));

                    var aclListResponse = await c.ExecuteForArrayAsync("ACL", "LIST");

                    if (!aclListResponse.Contains(activeUserWithGet) &&
                        !aclListResponse.Contains(inactiveUserWithoutGet) &&
                        !aclListResponse.Contains(inactiveUserWithNoCommands))
                    {
                        string corruptedAcl = aclListResponse.First(line => line.Contains(TestUserA));
                        throw new AssertionException($"Invalid ACL: {corruptedAcl}");
                    }
                }
            });
        }

        /// <summary>
        /// Tests that ACL SETUSER works in parallel without fatal contention on user in ACL map.
        ///
        /// Test launches multiple clients that apply the same ACL change to the same user. Creates race to become the
        /// the first client to add the user to the ACL. Throws after initial insert into ACL if threading issues exist.
        ///
        /// Race conditions are not deterministic so test uses repeat.
        ///
        /// </summary>
        [TestCase(128, 2048)]
        [Repeat(10)]
        public async Task ParallelAclSetUserAvoidsMapContentionTest(int degreeOfParallelism, int iterationsPerSession)
        {
            string command1 = $"ACL SETUSER {TestUserA} on >{DummyPassword}";

            await Parallel.ForAsync(0, degreeOfParallelism, async (t, state) =>
            {
                using var c = TestUtils.GetGarnetClientSession();
                c.Connect();

                List<Task> tasks = new();
                for (uint i = 0; i < iterationsPerSession; i++)
                {
                    // Creates race between threads contending for first insert into ACL. Throws after first ACL insert.
                    tasks.Add(c.ExecuteAsync(command1.Split(" ")));
                }

                await Task.WhenAll(tasks);
            });

            ClassicAssert.Pass();
        }
    }
```

### Expected behavior

I would expect the following behavior from the `ACL SETUSER` command:
- It would not throw an exception when two threads or clients race to insert the first user. Instead, one thread/client would insert the user and the other would perform an update.
- It would maintain integrity of the user ACL, avoiding the combination of input arguments across client invocations of the command.
- It would handle contention among clients attempting to modify the same user, producing linearizable results.
- Subsequent calls to `ACL LIST` or in the future `ACL GETUSER` would return with deterministic results for the ACL operations performed via `ACL SETUSER`.

### Screenshots

_No response_

### Release version

I am building and testing against the source code in the repo as of February 1st, 2025.

### IDE

_No response_

### OS version

_No response_

### Additional context

I have submitted this accompanying PR to address this issue: https://github.com/microsoft/garnet/pull/989

kevinmichaelbowersox:
This pull requests fixes #988. Please see the detailed discussion provided in the issue for a deeper discussion of the problem this PR addresses.

**Quick Problem Overview**

The `ACL SETUSER` command may produce inconsistent results when multiple clients/threads attempt to concurrently modify the same user. Please find test cases in this PR that demonstrate the concurrency issues this PR addresses.

**Solution Overview**

This PR addresses the issue by implementing the following changes:
- Establishing a table of lock objects in the `AccessControlList` that can be used for taking granular level locks on a per user basis. This locking strategy should be used sparingly, possibly only for modifications to users via `ACL SETUSER`.
- In the `ACL SETUSER` implementation, a user level lock is taken prior to a read, copy, write sequence performed in the critical section of the lock. This is necessary to prevent one client from reading the user object from the `AccessControlList` while another client is in the process of modifying it during this sequence. A copy constructor was added to `User` to protect against external modification while performing this sequence. 
- An observer pattern is implemented for subscribers interested in being notified of when ACL modifications are performed on the `AccessControlList`. Sessions subscribe to these notifications at creation time, giving them awareness of changes to the ACL for the  user in the current session and the ability to refresh the ACL for the respective user. With this capability, ACL modifications are still applied in real time when users are modified.
- A flag was added to the session to signal staleness of the ACL. When the flag is raised, the session will refresh the ACL for the user before the next command execution.

**Testing and Verification**
- [x] Added unit tests that repro the situation prior to making changes and verified existence of the issue.
- [x] Performed local testing manually to verify expected results used in test cases.
- [x] Verified new and existing tests passed once fix was implemented.


libs/server/ACL/AccessControlList.cs
```diff
@@ -282,5 +331,26 @@ void Import(StreamReader input, string configurationFile = "<undefined>")
                 }
             }
         }
+
+        /// <summary>
+        /// Registers a subscriber to receive notifications when modifications are performed to the <see cref="AccessControlList"/>.
+        /// </summary>
+        /// <param name="subscriber">The <see cref="IAccessControlListSubscriber"/> to register.</param>
+        internal void Subscribe(IAccessControlListSubscriber subscriber)
```
kevin-montrose, line 339: I don't see any way subscriptions are removed - won't this extend the lifetime of all `RespServerSession`s indefinitely?

libs/server/ACL/AccessControlList.cs
```diff
@@ -282,5 +331,26 @@ void Import(StreamReader input, string configurationFile = "<undefined>")
...
+            _subscribers[subscriber.AclSubscriberKey] = subscriber;
+        }
+
+        /// <summary>
+        /// Notify the registered subscribers when modifications are performed to the <see cref="AccessControlList"/>.
+        /// </summary>
+        /// <param name="user">The created or updated <see cref="User"/> that triggered the notification.</param>
+        private void NotifySubscribers(User user)
+        {
+            foreach (IAccessControlListSubscriber subscriber in _subscribers.Values)
```
kevin-montrose, line 350: I believe there is a race here:
 1. A new `RespServerSession` is created, it runs through `this.AuthenticateUser` (L236) and the thread is paused before subscription runs
 2. Another thread executes a `ACL SETUSER`, triggering this `NotifySubscribers` method
 3. The thread running in #1 resumes, and registers itself

Net effect, that new `RespServerSession` ends up with the old user.

libs/server/ACL/IAccessControlListSubscriber.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Garnet.server.ACL
+{
+    /// <summary>
+    /// An interface for types that subscribe to <see cref="AccessControlList"/> changes.
+    /// </summary>
+    internal interface IAccessControlListSubscriber
```
kevin-montrose, line 9: pseudo-nit: There's a single implementer of this, I'd just take `RespServerSession`s directly and delete the interface.  More recent .NETs will try to devirtualize these calls, but it's best not to rely on them.

libs/server/Resp/AdminCommands.cs
```diff
@@ -114,6 +114,10 @@ internal bool CheckScriptPermissions(RespCommand cmd)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal bool CheckACLPermissions(RespCommand cmd)
         {
+            if (!(!_authenticator.IsAuthenticated || (_user != null)))
```
kevin-montrose, line 117: I don't quite follow what is happening here.

Generally changes here make me nervous, it's very hot code.  I'd want to see perf numbers.

libs/server/Resp/RespServerSession.cs
```diff
@@ -33,7 +33,7 @@ namespace Garnet.server
     /// <summary>
     /// RESP server session
     /// </summary>
-    internal sealed unsafe partial class RespServerSession : ServerSessionBase
+    internal sealed unsafe partial class RespServerSession : ServerSessionBase, IAccessControlListSubscriber
```
kevin-montrose, line 36: See nit above, I'd ditch the interface - there's only one implementer.

libs/server/Resp/RespServerSession.cs
```diff
@@ -438,6 +469,11 @@ private void ProcessMessages()
                 // Check ACL permissions for the command
                 if (cmd != RespCommand.INVALID)
                 {
+                    if (_isUserAclStale && cmd != RespCommand.AUTH)
```
kevin-montrose, line 472: This is very hot code, we'll want to see benchmarks to measure the impact.

I'd be more comfortable if this was in a failure path or something, rather than at the very start of the command dispatch.

libs/server/ACL/AccessControlList.cs
```diff
@@ -18,16 +18,31 @@ public class AccessControlList
...
+
         /// <summary>
         /// Dictionary containing all users defined in the ACL
         /// </summary>
         ConcurrentDictionary<string, User> _users = new();
 
+        /// <summary>
+        /// Dictionary containing stable lock objects for each user in the ACL.
+        /// </summary>
+        ConcurrentDictionary<string, object> _userLockObjects = new();
```
kevin-montrose, line 34: I don't quite get why we're introducing lock objects - we can just lock on the `User` if it's already exists?

For new users things could be phrased in terms of `GetOrAdd()`.

kevin-montrose (changes requested):

libs/server/ACL/AccessControlList.cs
kevinmichaelbowersox: I considered using the user as the lock object, however we are replacing the user entry in the map. This makes it a bad candidate for a lock object. We need a stable object for each user to lock on.

libs/server/Resp/AdminCommands.cs
kevinmichaelbowersox: Thank you for catching this. It was debug code I accidentally left in my PR.

libs/server/Resp/RespServerSession.cs
kevinmichaelbowersox: This part is essential to the solution because it forces the refresh when the user's acl has been updated. We'll need to check the benchmarks to see if the approach is viable. I was hoping the conditional evaluation of the simple boolean would not be a drain on performance.

If the benchmarks do not perform well, the entire design will need reconsidered.

libs/server/ACL/AccessControlList.cs
kevinmichaelbowersox: Pushed a fix that unsubscribes the session from the ACL notifications when the session is killed.

libs/server/Resp/RespServerSession.cs
kevinmichaelbowersox: I removed this interface from the solution.

libs/server/Resp/AdminCommands.cs
kevinmichaelbowersox: This was removed.

libs/server/ACL/IAccessControlListSubscriber.cs
kevinmichaelbowersox: Removed this interface.

libs/server/ACL/AccessControlList.cs
kevinmichaelbowersox: Pushed a fix that checks the authenticated user is current prior to finishing instantiation of the session.

```
            // Ensure user was not updated during initialization.
            if (_authenticator is GarnetACLAuthenticator aclAuthenticator && _user != aclAuthenticator.GetUser())
            {
                this.RefreshUser();
            }
```

libs/server/Resp/RespServerSession.cs
kevinmichaelbowersox: Main Commit Sha(a63ad801ceb155179084b6c728c3989c4e9a424c) BDN Results:
 Method      | Params | Mean     | Error    | StdDev   | Allocated |
------------ |------- |---------:|---------:|---------:|----------:|
 **Set**         | **ACL**    | **14.20 μs** | **0.052 μs** | **0.049 μs** |         **-** |
 SetEx       | ACL    | 19.98 μs | 0.035 μs | 0.031 μs |         - |
 SetNx       | ACL    | 20.45 μs | 0.055 μs | 0.052 μs |         - |
 SetXx       | ACL    | 23.73 μs | 0.101 μs | 0.085 μs |         - |
 GetFound    | ACL    | 15.55 μs | 0.029 μs | 0.026 μs |         - |
 GetNotFound | ACL    | 10.95 μs | 0.016 μs | 0.015 μs |         - |
 Increment   | ACL    | 21.16 μs | 0.044 μs | 0.039 μs |         - |
 Decrement   | ACL    | 21.12 μs | 0.064 μs | 0.053 μs |         - |
 IncrementBy | ACL    | 25.56 μs | 0.032 μs | 0.028 μs |         - |
 DecrementBy | ACL    | 27.57 μs | 0.096 μs | 0.085 μs |         - |
 **Set**         | **AOF**    | **19.96 μs** | **0.064 μs** | **0.057 μs** |         **-** |
 SetEx       | AOF    | 25.55 μs | 0.089 μs | 0.079 μs |         - |
 SetNx       | AOF    | 27.04 μs | 0.138 μs | 0.129 μs |         - |
 SetXx       | AOF    | 28.60 μs | 0.292 μs | 0.273 μs |         - |
 GetFound    | AOF    | 15.62 μs | 0.028 μs | 0.025 μs |         - |
 GetNotFound | AOF    | 10.80 μs | 0.024 μs | 0.022 μs |         - |
 Increment   | AOF    | 27.29 μs | 0.120 μs | 0.107 μs |         - |
 Decrement   | AOF    | 27.98 μs | 0.053 μs | 0.049 μs |         - |
 IncrementBy | AOF    | 31.82 μs | 0.121 μs | 0.107 μs |         - |
 DecrementBy | AOF    | 32.33 μs | 0.150 μs | 0.133 μs |         - |
 **Set**         | **None**   | **14.11 μs** | **0.018 μs** | **0.015 μs** |         **-** |
 SetEx       | None   | 19.96 μs | 0.118 μs | 0.110 μs |         - |
 SetNx       | None   | 20.70 μs | 0.038 μs | 0.035 μs |         - |
 SetXx       | None   | 23.47 μs | 0.039 μs | 0.033 μs |         - |
 GetFound    | None   | 15.64 μs | 0.041 μs | 0.034 μs |         - |
 GetNotFound | None   | 11.03 μs | 0.023 μs | 0.019 μs |         - |
 Increment   | None   | 20.76 μs | 0.027 μs | 0.023 μs |         - |
 Decrement   | None   | 21.87 μs | 0.037 μs | 0.033 μs |         - |
 IncrementBy | None   | 27.58 μs | 0.180 μs | 0.160 μs |         - |
 DecrementBy | None   | 27.71 μs | 0.075 μs | 0.066 μs |         - |

kevinmichaelbowersox: This PR Commit SHA(4ca6db52373090c3b124adaec30b230695b9b23c) BDN Results:
| Method      | Params | Mean     | Error    | StdDev   | Allocated |
|------------ |------- |---------:|---------:|---------:|----------:|
| Set         | ACL    | 15.13 us | 0.103 us | 0.091 us |         - |
| SetEx       | ACL    | 20.49 us | 0.064 us | 0.060 us |         - |
| SetNx       | ACL    | 21.70 us | 0.072 us | 0.060 us |         - |
| SetXx       | ACL    | 23.91 us | 0.058 us | 0.052 us |         - |
| GetFound    | ACL    | 15.58 us | 0.105 us | 0.087 us |         - |
| GetNotFound | ACL    | 10.92 us | 0.016 us | 0.014 us |         - |
| Increment   | ACL    | 21.11 us | 0.049 us | 0.043 us |         - |
| Decrement   | ACL    | 21.81 us | 0.057 us | 0.050 us |         - |
| IncrementBy | ACL    | 26.21 us | 0.114 us | 0.101 us |         - |
| DecrementBy | ACL    | 26.17 us | 0.063 us | 0.059 us |         - |
| Set         | AOF    | 19.36 us | 0.077 us | 0.068 us |         - |
| SetEx       | AOF    | 25.41 us | 0.063 us | 0.056 us |         - |
| SetNx       | AOF    | 27.82 us | 0.081 us | 0.072 us |         - |
| SetXx       | AOF    | 27.71 us | 0.086 us | 0.076 us |         - |
| GetFound    | AOF    | 15.55 us | 0.029 us | 0.026 us |         - |
| GetNotFound | AOF    | 11.46 us | 0.010 us | 0.009 us |         - |
| Increment   | AOF    | 26.35 us | 0.036 us | 0.032 us |         - |
| Decrement   | AOF    | 27.86 us | 0.043 us | 0.040 us |         - |
| IncrementBy | AOF    | 32.92 us | 0.109 us | 0.097 us |         - |
| DecrementBy | AOF    | 33.06 us | 0.090 us | 0.085 us |         - |
| Set         | None   | 14.89 us | 0.019 us | 0.017 us |         - |
| SetEx       | None   | 20.96 us | 0.080 us | 0.071 us |         - |
| SetNx       | None   | 20.12 us | 0.039 us | 0.037 us |         - |
| SetXx       | None   | 23.88 us | 0.045 us | 0.040 us |         - |
| GetFound    | None   | 15.77 us | 0.026 us | 0.022 us |         - |
| GetNotFound | None   | 10.59 us | 0.020 us | 0.018 us |         - |
| Increment   | None   | 22.16 us | 0.206 us | 0.183 us |         - |
| Decrement   | None   | 22.51 us | 0.115 us | 0.096 us |         - |
| IncrementBy | None   | 26.93 us | 0.138 us | 0.122 us |         - |

badrishc:
Minor nit: can you fix the title of this PR to be a bit descriptive? Thanks.

libs/server/ACL/ACLParser.cs
```diff
@@ -110,6 +109,11 @@ public static User ParseACLRule(string input, AccessControlList acl = null)
                 ApplyACLOpToUser(ref user, tokens[i]);
             }
 
+            if (acl != null)
```
kevin-montrose, line 112: nit: `acl?.AddOrReplaceUser(user);`

libs/server/ACL/AccessControlList.cs
```diff
@@ -98,7 +139,15 @@ public bool DeleteUser(string username)
             {
                 throw new ACLException("The special 'default' user cannot be removed from the system");
             }
-            return _users.TryRemove(username, out _);
+
+            bool userDeleted = _users.TryRemove(username, out _);
```
kevin-montrose, line 143: There's a race here.

If Session A deletes a user and Session B adds the user it is possible for this to happen:

|Sessions A|Session B|
|--|--|
| `_users.TryRemove("foo", ...)` => true| |
| Thread Suspends | AddUser called with "foo", populates `_users`, does not modify `_userLockObjects` |
| Thread Resumes | |
| Because `userDeleted == true`, ` _userLockObjects.TryRemove("foo", out _)` called | |

At the end `_users` is populated, but `_userLockObject` is not.

libs/server/ACL/AccessControlList.cs
kevin-montrose: Fair.

The dance with ConcurrentDictionaries isn't quite right though, which makes me think there might be a simpler approach that does enable User as a lock.

libs/server/ACL/AccessControlList.cs
```diff
@@ -18,16 +18,31 @@ public class AccessControlList
...
+
         /// <summary>
         /// The currently configured default user (for fast default lookups)
         /// </summary>
         User _defaultUser;
 
+        /// <summary>
+        /// The <see cref="RespServerSession"/>s that will receive access control list change notifications.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, RespServerSession> _subscribedSessions = new();
```
kevin-montrose, line 44: nit: it's not clear to me why this is keyed by a string?  You could just use the `RespServerSesssion` directly, object identity should be sufficient.

libs/server/Resp/RespServerSession.cs
```diff
@@ -249,6 +263,12 @@ public RespServerSession(
                 if (this.networkSender.GetMaxSizeSettings?.MaxOutputSize < sizeof(int))
                     this.networkSender.GetMaxSizeSettings.MaxOutputSize = sizeof(int);
             }
+
+            // Ensure user was not updated during initialization.
+            if (_authenticator is GarnetACLAuthenticator aclAuthenticator && _user != aclAuthenticator.GetUser())
+            {
+                this.RefreshUser();
```
kevin-montrose, line 270: I believe you need `force: true` here.

kevin-montrose (changes requested):

kevin-montrose:
The more I look at this, the more I'm tempted to try an approach that skips all the ConcurrentDictionary, lock, and notify stuff.

Something akin to:
```csharp
class User
{
   // If User has been modified, the resulting user will be stored here
   private User successorUser;

   public User GetEffectiveUser()
   => successUser?.GetEffectiveUser() ?? this;
}

// --- and then in RespServerSession, somewhere
_user = _user.GetEffectiveUser();

// --- and then in ACL updates
// 1. still clone the user and make changes to the copy
// 2. use an Interlocked.CompareExchange(ref successorUser, newUser, null)
// 3. if that returns non-null, restart the command
```

If I wasn't on-call I'd sketch it out properly...

kevinmichaelbowersox:
The approach without the concurrent dictionary would be simpler. We'll hold on this implementation and shift efforts towards this alternative design in a separate PR.

kevinmichaelbowersox:
Closing this pull request in favor of the alternative solution discussed in the thread above and implemented in this new pull request: https://github.com/microsoft/garnet/pull/1007

PR #990, opened 2025-02-02
Clean up and speed up pub/sub logic

badrishc:
* Remove unsafe code and unnecessary generics
* No clobbering of network buffers, use ArgSlice correctly
* Use a new `ReadOptimizedConcurrentSet` structure for efficient sub matching
* Publish path is now zero copy (except for the single copy from publisher network buffer to subscriber network buffer)
* Improve the async publish (via TsavoriteLog) to eliminate all memory allocations
* Fix various small bugs in pub/sub responses and implementation


libs/server/PubSub/ConcurrentList.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Garnet.common;
+
+namespace Garnet.server
+{
+    class ConcurrentList<T>
```
TalZaccai, line 9: This could be useful outside the context of PubSub (might want to move to Garnet.common and maybe add some comments to address why it's better to use than, say BlockingCollection?)

libs/server/PubSub/ConcurrentList.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public void Add(T item)
+        {
+            rwLock.WriteLock();
+            list.Add(item);
+            rwLock.WriteUnlock();
+        }
+
+        public bool RemoveAll(T item)
```
TalZaccai, line 38: Would be good to add an XML comment that specifies under what condition this returns true, might not be obvious.

libs/server/ByteArrayWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System;
+using System.Numerics;
+
+namespace Garnet.server
+{
+    /// <summary>
+    /// Specialized key type for storing byte arrays (pinned and unpinned).
+    /// </summary>
+    public readonly struct ByteArrayWrapper
```
TalZaccai, line 12: Might be useful in other places, consider moving to Garnet.common

libs/server/PubSub/SubscribeBroker.cs
```diff
@@ -131,166 +112,129 @@ private unsafe int Broadcast(byte[] key, byte* valPtr, int valLength, bool ascii
...
-                done.Reset();
-                subscriptions = new ConcurrentDictionary<byte[], ConcurrentDictionary<int, ServerSessionBase>>(ByteArrayComparer.Instance);
-                prefixSubscriptions = new ConcurrentDictionary<byte[], (bool, ConcurrentDictionary<int, ServerSessionBase>)>(ByteArrayComparer.Instance);
-                Task.Run(() => Start(cts.Token));
+                Initialize();
             }
             else
             {
-                while (prefixSubscriptions == null) Thread.Yield();
+                while (patternSubscriptions == null) Thread.Yield();
```
TalZaccai, line 192: I would suggest having something dedicated to signaling an initialized state, in case in the future someone adds more logic to the initialization stage and forgets to update this condition.

TalZaccai (approved):
OMG this is SO MUCH CLEANER! Love it! Few suggestions.

libs/server/ByteArrayWrapper.cs
badrishc: It depends on ArgSlice which lives in Garnet.server. Hence, leaving this in server.

PR #994, opened 2025-02-04
Replace String concat with StringBuilder in Cluster Nodes

msft-paddy14:
StackExchange.Redis does periodic CLUSTER NODES and for large clusters a lot of allocations can happen with such calls.
We recently saw an issue where lots of allocation increased our .NET committed memory and also GC activity when new clients would be deployed/control plane would probe nodes where cluster size was ~100 nodes.

image

Using StringBuilder in an adhoc BDN test shows we can get significant improvements using this

**| TestMultiStringConcat        | None   | 125.13 us | 1.894 us | 2.395 us | 93.5059 | 4.3945 | 1567640 B |
| TestMultiStringConcatBuilder | None   |  41.59 us | 0.627 us | 0.586 us |  9.5215 | 0.9155 |  160152 B |**


badrishc:
There is no need to create a string or use stringbuilder here, if this is a common operation. We should directly write to the output network buffer using a sequence of the usual style of writing:

```cs
while (!RespWriteUtils.TryWriteXxx(...))
   SendAndReset();
```

msft-paddy14:
> Ideally, there would be no need to create strings at all, or use stringbuilder, if this is a very common operation. We could directly write to a session-local scratch buffer that is reusable for that session, and write to it a sequence of the usual style of writing:
> 
> ```cs
> while (!RespWriteUtils.TryWriteXxx(...))
>    ...
> ```
> 
> Here is a rough (non-working) prototype of this for `CLUSTER NODE`: https://github.com/microsoft/garnet/compare/badrishc/cluster-node?expand=1
> 
> However, the `StringBuilder` approach is reasonable and simple to begin with, if you are seeing it is sufficient for your use case.

Your approach seems more performant. However, since it's relatively more code and needs few interface changes, maybe string builder is okay to start with for simplicity and more gains compared to main code. We can optimize further with scratchbuffer if needed?

badrishc:
> > Ideally, there would be no need to create strings at all, or use stringbuilder, if this is a very common operation. We could directly write to a session-local scratch buffer that is reusable for that session, and write to it a sequence of the usual style of writing:
> > ```cs
> > while (!RespWriteUtils.TryWriteXxx(...))
> >    ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here is a rough (non-working) prototype of this for `CLUSTER NODE`: https://github.com/microsoft/garnet/compare/badrishc/cluster-node?expand=1
> > However, the `StringBuilder` approach is reasonable and simple to begin with, if you are seeing it is sufficient for your use case.
> 
> Your approach seems more performant. However, since it's relatively more code and needs few interface changes, maybe string builder is okay to start with for simplicity and more gains compared to main code. We can optimize further with scratchbuffer if needed?

Yeah this is at least better than main.

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -516,17 +517,22 @@ public string GetNodeInfo(ushort workerId, ConnectionInfo info)
...
-                $"{info.pong} " +
-                $"{workers[workerId].ConfigEpoch} " +
-                $"{(info.connected || workerId == 1 ? "connected" : "disconnected")}" +
-                $"{GetSlotRange(workerId)}" +
-                $"{GetSpecialStates(workerId)}\n";
+            var nodeInfoStringBuilder = new StringBuilder();
+            return nodeInfoStringBuilder
+                .Append(workers[workerId].Nodeid).Append(" ")
+                .Append(workers[workerId].Address).Append(":").Append(workers[workerId].Port)
+                .Append("@").Append(workers[workerId].Port + 10000).Append(",").Append(workers[workerId].hostname).Append(" ")
```
badrishc, line 524: use `.Append(char c)` for all single characters such as ' ', ':', etc.

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -516,17 +517,22 @@ public string GetNodeInfo(ushort workerId, ConnectionInfo info)
...
+                .Append(workerId == 1 ? "myself," : "")
+                .Append(workers[workerId].Role == NodeRole.PRIMARY ? "master" : "slave").Append(" ")
+                .Append(workers[workerId].Role == NodeRole.REPLICA ? workers[workerId].ReplicaOfNodeId : "-").Append(" ")
+                .Append(info.ping).Append(" ")
+                .Append(info.pong).Append(" ")
+                .Append(workers[workerId].ConfigEpoch).Append(" ")
+                .Append(info.connected || workerId == 1 ? "connected" : "disconnected")
+                .Append(GetSlotRange(workerId))
```
badrishc, line 532: `GetSlotRange` and `GetSpecialStates` can return `StringBuilder` instances to avoid intermediate string allocation.

badrishc: i.e. we can call `GetSlotRange(workerId, nodeInfoStringBuilder)` here.

badrishc:
@msft-paddy14 - have left a few small comments to address here, thanks.

badrishc (approved):

vazois (approved):

PR #996, opened 2025-02-04
MSETNX atomic fix #2

prvyk:
A pull request for issue #980, this time doing a transaction the same way other commands do.


badrishc:
Nicely done. I was worried we do not have sufficient documentation for how to structure this as a transaction, but this is great!

badrishc (approved):

PR #999, opened 2025-02-05
Add a DEBUG command.

prvyk:
The idea is to allow special commands during testing in order to simulate conditions that are difficult to achieve normally - for example, seeing whether a cluster can survive a couple of its nodes crashing.

Redis happens to already have a DEBUG command for similar uses, might as well use the keyword. Note the subcommands have no official documentation and there are no compatibility issues even if Garnet behaviour differs. Also implement the redis safeguards, they were very likely borne from experience.

Since this is intended for testing, disable the command in normal use (like Redis does) and enable it by default for the test suits.


prvyk:
It's not very clear what's the difference between ServerOptions and GarnetServerOptions. Both seem to serve the same purpose (and comment), yet one inherits from the other?

libs/server/Auth/Settings/ConnectionProtectionOption.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Garnet.server.Auth.Settings
+{
+    public enum ConnectionProtectionOption
```
TalZaccai, line 6: Why is this not called `DebugCommandOption` or something similar? This might be a bit confusing as to the purpose of this configuration. Could also use some XML comments to describe its function.

libs/server/Resp/AdminCommands.cs
```diff
@@ -651,6 +653,47 @@ private bool NetworkProcessClusterCommand(RespCommand command)
...
+            if (parseState.Count == 0)
+            {
+                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_WRONG_NUMBER_OF_ARGUMENTS, ref dcurr, dend))
+                    SendAndReset();
+
+                return true;
+            }
+
+            if (storeWrapper.serverOptions.EnableDebugCommand == ConnectionProtectionOption.Block)
+            {
```
TalZaccai, line 667: nit: Can merge these two conditions with an `||` as they return the same error

libs/server/Resp/AdminCommands.cs
```diff
@@ -651,6 +653,47 @@ private bool NetworkProcessClusterCommand(RespCommand command)
...
+            var command = parseState.GetArgSliceByRef(0).ReadOnlySpan;
+
+            if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.SEGFAULT))
+                throw new GarnetException(Microsoft.Extensions.Logging.LogLevel.Debug, panic: true);
+
+            if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.VERSION))
+            {
+                while (!RespWriteUtils.TryWriteAsciiBulkString(storeWrapper.version, ref dcurr, dend))
+                    SendAndReset();
+            }
```
TalZaccai, line 692: Would be helpful to have a `DEBUG HELP` command for discoverability of these subcommands

libs/server/Resp/AdminCommands.cs
```diff
@@ -651,6 +653,47 @@ private bool NetworkProcessClusterCommand(RespCommand command)
...
+            {
+                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_DEUBG_DISALLOWED, ref dcurr, dend))
+                    SendAndReset();
+
+                return true;
+            }
+
+            var command = parseState.GetArgSliceByRef(0).ReadOnlySpan;
+
+            if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.SEGFAULT))
```
TalZaccai, line 685: Shouldn't this be PANIC instead of SEGFAULT?

libs/server/Resp/CmdStrings.cs
```diff
@@ -250,6 +253,7 @@ static partial class CmdStrings
         public static ReadOnlySpan<byte> RESP_COMMAND_HAS_NO_KEY_ARGS => "The command has no key arguments"u8;
         public static ReadOnlySpan<byte> RESP_ERR_INVALID_CLIENT_UNBLOCK_REASON => "ERR CLIENT UNBLOCK reason should be TIMEOUT or ERROR"u8;
         public static ReadOnlySpan<byte> RESP_UNBLOCKED_CLIENT_VIA_CLIENT_UNBLOCK => "UNBLOCKED client unblocked via CLIENT UNBLOCK"u8;
+        public static ReadOnlySpan<byte> RESP_ERR_DEUBG_DISALLOWED => "DEBUG command not allowed."u8;
```
TalZaccai, line 256: Would be helpful to have the full error message, as it also helps the user with how to enable this option - `ERR DEBUG command not allowed. If the EnableDebugCommand option is set to "local", you can run it from a local connection, otherwise you need to set this option in the configuration file, and then restart the server.`

TalZaccai (changes requested):
Cool addition! Added a few comments :)

website/docs/commands/api-compatibility.md
```diff
@@ -277,7 +277,7 @@ Note that this list is subject to change as we continue to expand our API comman
 |  | CONFIG REWRITE | ➖ |  |
 |  | [CONFIG SET](server.md#config-set) | ➕ |  |
 |  | [DBSIZE](server.md#dbsize) | ➕ |  |
-|  | DEBUG | ➖ | Internal command |
+|  | DEBUG | ➕ | Internal command |
```
badrishc, line 280: this command and its subcommands need to be documented, perhaps in garnet-specific commands if they are specific, or in a different section if that makes sense.

badrishc:
> It's not very clear what's the difference between ServerOptions and GarnetServerOptions. Both seem to serve the same purpose (and comment), yet one inherits from the other?

It's just a historical split of functionality. Can be eliminated if necessary, but we left it as is because GarnetServerOptions is already quite large.

libs/host/Configuration/Options.cs
```diff
@@ -498,6 +498,10 @@ internal sealed class Options
         [Option("object-scan-count-limit", Required = false, HelpText = "Limit of items to return in one iteration of *SCAN command")]
         public int ObjectScanCountLimit { get; set; }
 
+        [OptionValidation]
+        [Option("enable-debug-command", Required = false, HelpText = "Enable DEBUG command for 'no', 'local' or 'all' connections", Hidden = true)]
+        public string EnableDebugCommand { get; set; }
```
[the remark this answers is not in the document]
prvyk: Done.

libs/resources/RespCommandsInfo.json
```diff
@@ -1078,6 +1078,14 @@
       "response_policy:agg_sum"
     ]
   },
+  {
```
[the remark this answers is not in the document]
prvyk: Ah, there's a tool? Done.

libs/server/Auth/Settings/ConnectionProtectionOption.cs
prvyk: Because this method is used also for other things in Redis:

> # no    - Block for any connection (remain immutable)
> # yes   - Allow for any connection (no protection)
> # local - Allow only for local connections. Ones originating from the
> #         IPv4 address (127.0.0.1), IPv6 address (::1) or Unix domain sockets.
> #
> # enable-protected-configs no
> enable-debug-command local
> # enable-module-command no

enable-module-command is likely not relevant, but I can see enable-debug-command being used.

libs/server/Resp/AdminCommands.cs
prvyk: Done.

libs/server/Resp/AdminCommands.cs
prvyk: Done.

libs/server/Resp/AdminCommands.cs
prvyk: Done.

libs/server/Resp/CmdStrings.cs
prvyk: Done.

website/docs/commands/api-compatibility.md
prvyk: Added command documentation.

The subcommands are internal and used for testing, not a public API, and therefore IMHO do not need documentation (redis doesn't document them either).

libs/host/Configuration/Redis/RedisOptions.cs
```diff
@@ -27,6 +27,9 @@ internal class RedisOptions
         [RedisOption("bind", nameof(Options.Address), BindWarning, typeof(ArrayToFirstItemTransformer<string>))]
         public Option<string[]> Bind { get; set; }
 
+        [RedisOption("enable-debug-command", nameof(Options.EnableDebugCommand))]
+        public Option<string> EnableDebugCommand { get; set; }
```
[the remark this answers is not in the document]
prvyk: Done. 

Every attempt at adding a new type converter blew up at option parsing, even though the code matched RedisTlsAuthClients and other Redis enum types. Eventually I just used the string values as keys for the new enum type and that worked without a new converter.

prvyk:
> > It's not very clear what's the difference between ServerOptions and GarnetServerOptions. Both seem to serve the same purpose (and comment), yet one inherits from the other?
> 
> It's just a historical split of functionality. Can be eliminated if necessary, but we left it as is because GarnetServerOptions is already quite large.

Ah. I asked since I wasn't sure where to add the option. Well, if the idea is to split Options, one could put Redis options and inherit Garnet from that. Or do something like General options. : Network options : (base) DB options .

badrishc:
> > > It's not very clear what's the difference between ServerOptions and GarnetServerOptions. Both seem to serve the same purpose (and comment), yet one inherits from the other?
> > 
> > 
> > It's just a historical split of functionality. Can be eliminated if necessary, but we left it as is because GarnetServerOptions is already quite large.
> 
> Ah. I asked since I wasn't sure where to add the option. Well, if the idea is to split Options, one could put Redis options and inherit Garnet from that. Or do something like General options. : Network options : (base) DB options .

Nowadays we've been adding all the options in GarnetServerOptions. We can probably folder ServerOptions into it, to keep it simple.

libs/host/Configuration/Redis/RedisOptions.cs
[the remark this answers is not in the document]
TalZaccai: This is actually important to us to have the separation between the two types, we want to have our Garnet type and support the Redis type only for compatibility with redis.conf, we don't necessarily want to tie ourselves to their syntax. I've added a type converter to this branch - [talzacc/type_converter](https://github.com/microsoft/garnet/tree/talzacc/type_converter), I actually can't push to yours so if you can, please merge it and it should work.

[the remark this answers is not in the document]
prvyk: Ah, I see the issue. I wasn't clear earlier.

Redis allows enabling debug command either through configuration or command line. The commandline option also accepts the same 'no/local/yes' syntax (i.e. I'd like to accept "--enable-debug-command yes"). But these aren't the names that are most clear internally, it would be nice to use different names. So I tried to use the typeconverter for commandline option parsing and that was what blew up. 

First pass was to accept string on the commandline and convert from that.
Second pass tried and failed to use typeconverter on commandline, so instead it accepted the Redis style options on the commandline and converted from that.
But if we must use the same enum for commandline and internally, I can just add no/yes directly. (Another option would be to use a third enum for commandline parsing, that's a bit much?)

I pulled your diff and added no/yes so the converter will accept --enable-debug-command no/local/yes.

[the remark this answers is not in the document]
TalZaccai: I think it's confusing to have this overload of enum values, why can't it just be Yes, No & Local?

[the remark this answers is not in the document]
prvyk: Ok, standardized on No/Local/Yes and added comments to compensate.

TalZaccai (approved):
Great! Thanks for all the prompt fixes! :)

PR #1002, opened 2025-02-05
Added Mirrored networking mode detail for WSL

Vijay-Nirmal:
Fix: #844


badrishc:
This is not necessary, and further, this mirrored networking mode works only on Windows 11.

Here is how we make WSL clients work with a Garnet Windows server in general:
* Start GarnetServer in Windows, listening to 0.0.0.0 (`IPAddress.Any`)
* Add `GarnetServer.exe` to firewall exceptions (Windows Defender Firewall -> Allowed apps)
* On WSL, get the host IP address using `ip route show | grep -i default | awk '{ print $3}'`
* Connect from WSL using `redis-cli -h <address>`

Vijay-Nirmal:
Will do it

badrishc (approved):

badrishc (approved):

Issue #993, opened 2025-02-04
In release 1.0.54 container image bind Garnet on 127.0.0.1 instead 0.0.0.0


rnz:
### Describe the bug

In release 1.0.54 container image bind Garnet on 127.0.0.1 instead 0.0.0.0 

### Steps to reproduce the bug

docker/podman pull 
docker/podman run 

### Expected behavior

garnet in container bind on 0.0.0.0

### Release version

v1.0.54

md-owes:
I was wondering since last 4 hours as to why it was happening on my system. As i have done a full upgrade of my Arch Linux, which also updated docker desktop resulting in updating of all docker images including garnet, hence i was confused as to which upgrade might have caused it. I already searched through all issues in this repo but could not find anything relatable. After scratching my head with continuous prompts on chatgpt, deepseek etc which still could not help finding the root cause. Now accidentally refreshed issues page of this repo and your message flashes. Thank you for creating this issue as it would have taken me another few hours after which i would have evetually created this issue.

badrishc:
This seems to be a regression from https://github.com/microsoft/garnet/pull/944 - @PaulusParssinen

badrishc:
It seems to be an issue only in the latest docker, not the released version. You can access 1.0.54 here: https://github.com/microsoft/garnet/pkgs/container/garnet/346713800?tag=1.0.54

md-owes:
for now, i fixed it by adding args property to my garnet-deployment.yaml file as shown below

```
command: ["/app/GarnetServer"]
args: ["--bind", "0.0.0.0"]
```

PR #1003, opened 2025-02-06
Fix bind address

badrishc:
Set bind address to IPAddress.Any when address is null (default) to restore older semantics. Fixes https://github.com/microsoft/garnet/issues/993


PaulusParssinen:
Ah, right.. 🤦‍♂️ Thanks for fixing!

PR #1005, opened 2025-02-07
Multi-Database Support

TalZaccai:
This PR enables multiple logical databases in a single Garnet server instance.

This PR includes:
- [x] Adding MaxDatabases configuration to garnet.conf + compatibility with redis.conf
- [x] Multi-DB support in StoreWrapper, including checkpointing & AOF
- [x] Current-DB context switching support in RespServerSession
- [x] Update GarnetInfoMetrics to reflect current database data
- [x] Support for the following commands:
  - [x] SELECT - switch current DB context
  - [x] SWAPDB - swap two DB instances
  - [x] FLUSHDB - flush all keys from current DB in context
  - [x] FLUSHALL - flush all keys from all active databases
- [x] Add support for specifying a DB ID in the following commands:
  - [x] SAVE - synchronous checkpointing database by ID
  - [x] BGSAVE - asynchronous checkpointing database by ID
  - [x] LASTSAVE - number of seconds since last checkpoint of database by ID
  - [x] COMMITAOF - synchronous database AOF commit by ID
- [x] Add DB ID to client info
- [x] Testing for supported commands
- [x] Add website docs for new commands
- [x] Add website dev docs 

**Note:** This PR does not include support for SELECT & SWAPDB commands within a transaction, this gap will be addressed in a separate PR.


prvyk:
Some tiny nitpicking if it's not too early:

A) Some more bounds checking can be done on the lower bound.
e.g. 'LASTSAVE -1'  or starting with --max-databases=0 (both quickly crash).

B) SWAPDB doesn't update CLIENT LIST db field if the current connection's db is swapped. Also doesn't update the current db when called from lua inside the script context.

C) Redis also adds a line for each db under 'Keyspace' in INFO command.

TalZaccai:
> Some tiny nitpicking if it's not too early:
> 
> A) Some more bounds checking can be done on the lower bound. e.g. 'LASTSAVE -1' or starting with --max-databases=0 (both trigger an assertion).
> 
> B) SWAPDB doesn't update CLIENT LIST db field if the current connection's database is swapped. Also doesn't update the current database when called from lua inside the script context.
> 
> C) Redis also adds a line for each db under 'Keyspace' in INFO command.

Thanks @prvyk! Super helpful!
A - fixed.
B - not sure what you mean here, say the active DB is 1 and we do a SWAPDB 0 1, the active db is still 1 (and the CLIENT LIST sould show so). 
C - we currently don't support keyspace, even when we had just one logical database, I'm assuming it's because it's a fairly slow operation (since it requires scanning through all keys in each store), so I left it as-is.

prvyk:
> Thanks @prvyk! Super helpful!  B - not sure what you mean here

I think I got confused with CLIENT LIST. The lua thing is this:

```
SET a 0
SELECT 2
SET a 2
SELECT 0
EVAL "return { redis.call('SWAPDB', 0, 2); redis.call('GET', 'a') }" 0
```

Redis -> 
1) "OK"
2) "2"

Garnet ->
1) "OK"
2) "0"

(Afterwards, 'get a' returns "2" with both)

Redis seems to switch during the script context.

TalZaccai:
> > Thanks @prvyk! Super helpful!
> 
> > A - fixed.
> 
> In some cases the check is "db < -1" (and not db < 0). What is -1 supposed to mean when it's 'legal'?
> 

-1 is the same as not specifying an ID (i.e. "save all etc."), I suppose I can just disable it from the input as to not cause confusion.

> > B - not sure what you mean here
> 
> I think I got confused with CLIENT LIST. The lua thing is this:
> 
> ```
> SET a 0
> SELECT 2
> SET a 2
> SELECT 0
> EVAL "return { redis.call('SWAPDB', 0, 2); redis.call('GET', 'a') }" 0
> ```
> 
> Redis ->
> 
> 1. OK
> 2. "2"
> 
> Garnet ->
> 
> 1. "OK"
> 2. "0"
> 
> (Afterwards, 'GET a' returns "2" with both)
> 
> Redis seems to switch during the script context.

Good to know, I haven't tested SWAPDB from a LUA context, I'll add that to the tests as well!

TalZaccai:
@prvyk fixed the Lua swap issue, let me know if you see anything else that needs attention :) Thanks again!

prvyk:
It's interesting to note interaction between WATCH/MULTI and FLUSH commands.

```
SET KEY 1
WATCH KEY
MULTI
EXISTS KEY
(another client does DEL KEY meanwhile).
EXEC
```
return nil (correctly).

However, 

```
SET KEY 1
WATCH KEY
MULTI
EXISTS KEY
(another client does FLUSHALL or FLUSHDB meanwhile).
EXEC
```

returns 0 (it should have been nil too).

---

Another minor observation is that observers are not separated by databases. That is, the CollectionUpdatedEvent will fire if an update is done on the same key name in a different db number, even if the key does not exist in the watched db. However, the check itself is the origin session and as far as I can tell checks the right db. (So it's just a get and a lock in another db).

Unless I missed something creative, this seems to have merely performance implications. The scenario isn't all that different from multiple clients connecting to a single db and doing stuff. This will need to be tested when MOVE is implemented (what happens if someone moves a list into an observed key?).

badrishc:
> However,
> 
> ```
> SET KEY 1
> WATCH KEY
> MULTI
> EXISTS KEY
> (another client does FLUSHALL or FLUSHDB meanwhile).
> EXEC
> ```
> 
> returns 0 (it should have been nil too).
> 

It seems `FLUSHALL` and `FLUSHDB` need to increment the version for all entries in `functionsState.watchVersionMap` so that ongoing transactions see them as an update to every key. Something like `functionsState.watchVersionMap.IncrementAllVersions()` ...

darrenge (approved):
The changes to the BDN test framework all look good.

libs/server/Resp/ArrayCommands.cs
```diff
@@ -231,10 +231,68 @@ private bool NetworkSELECT()
...
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// SWAPDB
+        /// </summary>
+        /// <returns></returns>
+        private bool NetworkSWAPDB()
```
vazois, line 269: nit: can you put this under AdminCommands.cs

libs/server/Resp/ArrayCommands.cs
```diff
@@ -231,10 +231,68 @@ private bool NetworkSELECT()
...
+
+            // Validate index
+            if (!parseState.TryGetInt(0, out var index1) || !parseState.TryGetInt(1, out var index2))
+            {
+                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
                     SendAndReset();
+                return true;
+            }
+
+            if (index1 < storeWrapper.serverOptions.MaxDatabases && index2 < storeWrapper.serverOptions.MaxDatabases)
```
vazois, line 284: I am getting an error:server closed the connection when issuing SWAPDB 0 1 and --cluster=true.
I think we need to check options.EnableCluster earlier.
Maybe you can a test for all related DB management commands when cluster is enabled.

PS: SELECT 0 command also might need to be corrected.

vazois (changes requested):
Reviewing in stages because this PR is pretty big.

libs/server/AOF/AofProcessor.cs
```diff
@@ -108,42 +89,51 @@ public AofProcessor(
...
         {
-            logger?.LogInformation("Begin AOF recovery");
-            RecoverReplay(untilAddress);
+            logger?.LogInformation("Begin AOF recovery for DB ID: {id}", db.Id);
+            return RecoverReplay(ref db, untilAddress);
         }
 
-        MemoryResult<byte> output = default;
-        private unsafe void RecoverReplay(long untilAddress)
+        private long RecoverReplay(ref GarnetDatabase db, long untilAddress)
```
vazois, line 114: could you add some comments here since this is recovery and we know what each new addition does?

libs/host/GarnetServer.cs
```diff
@@ -294,82 +310,77 @@ private void LoadModules(CustomCommandManager customCommandManager)
...
-                return;
+                if (opts.CommitFrequencyMs != 0 || opts.WaitForCommit)
+                    throw new Exception("Cannot use CommitFrequencyMs or CommitWait without EnableAOF");
+                return (null, null);
             }
 
-            if (opts.CommitFrequencyMs != 0 || opts.WaitForCommit)
-                throw new Exception("Cannot use CommitFrequencyMs or CommitWait without EnableAOF");
+            if (opts.FastAofTruncate && opts.CommitFrequencyMs != -1)
+                throw new Exception("Need to set CommitFrequencyMs to -1 (manual commits) with MainMemoryReplication");
```
vazois, line 376: nit: " with FastAofTruncate"

libs/server/AOF/AofProcessor.cs
```diff
@@ -108,42 +89,51 @@ public AofProcessor(
         /// </summary>
         public void Dispose()
         {
-            basicContext.Session?.Dispose();
-            objectStoreBasicContext.Session?.Dispose();
+            var databaseSessionsSnapshot = respServerSession.GetDatabaseSessionsSnapshot();
+            foreach (var dbSession in databaseSessionsSnapshot)
+            {
+                dbSession.StorageSession.basicContext.Session?.Dispose();
+                dbSession.StorageSession.objectStoreBasicContext.Session?.Dispose();
```
vazois, line 96: Should we be checking if objectStoreBasicContext == null here?

libs/server/Databases/DatabaseManagerFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+                return false;
+
+            // If there are multiple databases to recover, create a multi database manager, otherwise create a single database manager.
+            using (createDatabaseDelegate(0))
+            {
+                // Check if there are multiple databases to recover from checkpoint
+                var checkpointParentDir = serverOptions.MainStoreCheckpointBaseDirectory;
+                var checkpointDirBaseName = serverOptions.GetCheckpointDirectoryName(0);
+
+                if (MultiDatabaseManager.TryGetSavedDatabaseIds(checkpointParentDir, checkpointDirBaseName,
```
vazois, line 43: I think we need to handle the case where the data db data/folders conflict with the value of AllowMultiDb.
If AllowMultiDb = false and we discover many databases on disk what happens?

test/Garnet.test/MultiDatabaseTests.cs
```diff
@@ -0,0 +1,1188 @@
...
+using Garnet.common;
+using Garnet.server;
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test
+{
+    [TestFixture]
+    public class MultiDatabaseTests
```
vazois, line 15: If you have not done so, can you please add a test for DBSize for multiple databases

libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -0,0 +1,984 @@
...
+            try
+            {
+                if (dbId >= 0 && dbId < databasesMapSize && !databasesMapSnapshot[dbId].IsDefault())
+                {
+                    success = true;
+                    return ref databasesMapSnapshot[dbId];
+                }
+
+                var db = CreateDatabaseDelegate(dbId);
+                if (!databases.TrySetValueUnsafe(dbId, ref db, false))
```
vazois, line 650: Can you please add a note here explaining why the order matters due to the underlying limitations of ExpandableMap

libs/server/Resp/ArrayCommands.cs
TalZaccai: SWAPDB is not an admin command (weird, I know)

libs/server/Resp/ArrayCommands.cs
TalZaccai: Ah that's a good find, it's because I didn't place it in the appropriate spot in the RespCommand enum, `IsDataCommand()` returned true and it failed in `CanServeSlot()`. We may want to have a test that iterates through the commands and checks that each command that doesn't have a KeySpec is not a data command.

TalZaccai: Also, SELECT 0 should be allowed in cluster mode (it just doesn't do anything)... But the error messages for SELECT and SWAP DB need to be updated.

libs/server/AOF/AofProcessor.cs
TalZaccai: Not sure if I understand what you mean here, comments regarding what?

vazois: If you can add more comments that will be very helpful.

TalZaccai: Ok I added some (I'll push them with the rest of the fixes, LMK if there's anything specific you think needs extra-attention)

libs/server/AOF/AofProcessor.cs
TalZaccai: It's a struct so it's never null, only Session will be null if object store is turned off

libs/server/Databases/DatabaseManagerFactory.cs
TalZaccai: You can see in the beginning of this method:
```cs
if (!serverOptions.AllowMultiDb || !serverOptions.Recover)
    return false;
```
i.e. if AllowMultiDb is disabled we create a `SingleDatabaseManager`, regardless of how many saved databases are on disk. When you call `SingleDatabaseManager.RecoverCheckpoint` it'll only recover the default database.

libs/server/Databases/DatabaseManagerFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+                        out var dbIds) && dbIds.Any(id => id != 0))
```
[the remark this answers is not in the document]
TalZaccai: That's a good question... It can happen if someone messes with the checkpoint files before running recovery. I can demand of `TryGetSavedDatabaseIds` to only return true if the default DB checkpoint directory exists.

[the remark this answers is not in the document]
TalZaccai: On second thought, if the data is not there, we should just not recover the database, right? It is possible to checkpoint one database and not others, so it is possible that DB 0 was not checkpointed.

libs/server/Resp/ArrayCommands.cs
vazois: Where we put it in the code is orthogonal to how it is classified. It will be better if it was part of the AdminCommands. Also are you saying from the ACL perspective it is not classified as an admin command?
This is potentially dangerous, if users can SWAPDB like that. It may lead to a privacy leak.
Can we check if SWAPDB can be disabled through ACL? It might be ok because I remember that ACL should be able to restrict execution of individual commands, but I am not sure if we have implemented/enabled it yet.

badrishc: SWAPDB is no more powerful than the combination of SELECT, DEL, SET so it is probably fine as defined. Also, as you said, it can be restricted on a per-command basis using ACLs.

badrishc (commented):
It seems we checkpoint Tsavorite even for databases where nothing was added:

127.0.0.1:6379[1]> select 15
OK
127.0.0.1:6379[15]> select 0
OK
127.0.0.1:6379> save
OK


Results in a checkpoint for 0 and 15.

badrishc (commented):
select 2

called after recovery of a store with 0 and 1 save.



gave this crash:


```
03::07::01 crit: Session[0] [127.0.0.1:63805] [02F3E4CC] ProcessMessages threw an exception: System.NullReferenceException: Object reference not set to an instance of an object.    at Garnet.server.MultiDatabaseManager.HandleDatabaseAdded(Int32 dbId) in D:\Git\garnet\libs\server\Databases\MultiDatabaseManager.cs:line 860    at Garnet.server.MultiDatabaseManager.TryGetOrAddDatabase(Int32 dbId, Boolean& success, Boolean& added) in D:\Git\garnet\libs\server\Databases\MultiDatabaseManager.cs:line 689    at Garnet.server.MultiDatabaseManager.CreateFunctionsState(Int32 dbId) in D:\Git\garnet\libs\server\Databases\MultiDatabaseManager.cs:line 635    at Garnet.server.StoreWrapper.CreateFunctionsState(Int32 dbId) in D:\Git\garnet\libs\server\StoreWrapper.cs:line 475    at Garnet.server.StorageSession..ctor(StoreWrapper storeWrapper, ScratchBufferManager scratchBufferManager, GarnetSessionMetrics sessionMetrics, GarnetLatencyMetricsSession LatencyMetrics, ILogger logger, Int32 dbId) in D:\Git\garnet\libs\server\Storage\Session\StorageSession.cs:line 72    at Garnet.server.RespServerSession.CreateDatabaseSession(Int32 dbId) in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 1463    at Garnet.server.RespServerSession.TryGetOrSetDatabaseSession(Int32 dbId, Boolean& success, Int32 dbIdForSessionCreation) in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 1441    at Garnet.server.RespServerSession.TrySwitchActiveDatabaseSession(Int32 dbId) in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 1355    at Garnet.server.RespServerSession.NetworkSELECT() in D:\Git\garnet\libs\server\Resp\ArrayCommands.cs:line 249    at Garnet.server.RespServerSession.ProcessArrayCommands[TGarnetApi](RespCommand cmd, TGarnetApi& storageApi) in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 719    at Garnet.server.RespServerSession.ProcessMessages() in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 530    at Garnet.server.RespServerSession.TryConsumeMessages(Byte* reqBuffer, Int32 bytesReceived) in D:\Git\garnet\libs\server\Resp\RespServerSession.cs:line 379
03::07::01 trce: GarnetServer[0] Accept socket was disposed at RecvEventArg_Completed
```

badrishc (commented):
max databases should be logger.Logged at startup

libs/server/Databases/SingleDatabaseManager.cs
```diff
@@ -0,0 +1,419 @@
...
+    /// </summary>
+    internal class SingleDatabaseManager : DatabaseManagerBase
+    {
+        /// <inheritdoc/>
+        public override ref GarnetDatabase DefaultDatabase => ref defaultDatabase;
+
+        /// <inheritdoc/>
+        public override int DatabaseCount => 1;
+
+        GarnetDatabase defaultDatabase;
```
badrishc, line 24: Can this be made readonly, and the callers/constructors adjusted, as the invariant is that this manager is associated with exactly one pre-defined database.

TalZaccai:
> It seems we checkpoint Tsavorite even for databases where nothing was added:
> 
> 127.0.0.1:6379[1]> select 15 OK 127.0.0.1:6379[15]> select 0 OK 127.0.0.1:6379> save OK
> 
> Results in a checkpoint for 0 and 15.

Correct, when you do a first SELECT on a dbId (that isn't the default DB) it creates a new database instance and the database is marked as "active". When you save, all active databases are checkpointed. Is there a different logic that you had in mind?

libs/server/Databases/SingleDatabaseManager.cs
TalZaccai: I understand the reasoning behind what you're asking, the reason for having it this way is to be able to pass it back by ref and avoid creating copies of this struct as much as possible. LMK if this is ok by you or if it still needs to be altered.

badrishc: Is there a good reason for GarnetDatabase to be a struct and not a class? If yes, then no point making readonly.

libs/server/GarnetDatabase.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// </summary>
+        /// <returns>True if default struct</returns>
+        public bool IsDefault() => MainStore == null;
+
+        /// <summary>
+        /// Dispose method
+        /// </summary>
+        public void Dispose()
+        {
+            if (disposed) return;
```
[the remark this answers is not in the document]
TalZaccai: Not critically important, it can be called more than once, but ideally shouldn't.

badrishc (approved):

badrishc (dismissed):

vazois (approved):
Looks great! Thanks!

TedHartMS (approved):
Approved with a few "fit and finish" comments

libs/server/Databases/IDatabaseManager.cs
```diff
@@ -0,0 +1,251 @@
...
+
+    /// <summary>
+    /// Interface for logical database management
+    /// </summary>
+    public interface IDatabaseManager : IDisposable
+    {
+        /// <summary>
+        /// Default database (DB 0)
+        /// </summary>
+        public GarnetDatabase DefaultDatabase { get; }
```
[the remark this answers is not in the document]
TalZaccai: Yes, this is mainly to avoid code churn in the Garnet cluster code, we're still supporting access to the default stores etc. through StoreWrapper via this DefaultDatabase.

badrishc (approved):

PR #1006, opened 2025-02-09
BITPOS bit search on the last byte should account for the byte's position.

prvyk:
The code In question is clearly wrong. The return value is defined relative to start of string. However, the called function returns bit offset relative to start of byte - it should add the byte's offset in the result like it already does for the starting byte. Fix this by adding the offset. Add a test.

Existing faulty code:
```
            // Search prefix and terminate if found position of bit
            var _ppos = BitPosIndexBitSearch(value, setVal, startOffset);
            if (_ppos != -1) return startOffset + _ppos;

            ....

            // Search suffix
            var _spos = BitPosIndexBitSearch(value, setVal, endOffset); // should add here too like it does for start!
            return _spos;
```

(I have some doubts regarding over other parts of the bit path, but these require more careful consideration. It's best to push a clearly right fix now and consider from there),

### DIFF
--- a/libs/server/Resp/Bitmap/BitmapManagerBitPos.cs
+++ b/libs/server/Resp/Bitmap/BitmapManagerBitPos.cs
@@ -126,7 +126,14 @@ namespace Garnet.server
 
             // Search suffix
             var _spos = BitPosIndexBitSearch(value, setVal, endOffset);
-            return _spos;
+            if (_spos != -1)
+            {
+                var retpos = (endByte * 8) + _spos;
+                if (retpos <= endOffset)
+                    return retpos;
+            }
+
+            return -1;
         }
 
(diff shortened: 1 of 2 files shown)


vazois:
There is certainly a bug for BIT variant. However, this PR only fixes a single case. See example below
Garnet Output:
image
Expected Output:
image

You will probably have to re-write the search for BIT option to make it work correctly.

prvyk:
Yep, there are other suspicious cases, this only handles a single cause. The BYTE variant probably isn't clean either (running the same testcase against redis shows a divergence in one case). I tried fixing the code but got lost a bit... Maybe a rewrite would do better.

prvyk:
Obsoleted by #1016 .

PR #1016, opened 2025-02-12
Refactor BITPOS Implementation

vazois:
This PR fixes the BITPOS to support BIT search correctly.

### DIFF
--- a/libs/server/Resp/Bitmap/BitmapManagerBitPos.cs
+++ b/libs/server/Resp/Bitmap/BitmapManagerBitPos.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Buffers.Binary;
 using System.Runtime.Intrinsics.X86;
 
 namespace Garnet.server
@@ -53,46 +54,6 @@ namespace Garnet.server
         /// <param name="endBitOffset"></param>
         /// <param name="searchFor"></param>
         /// <returns></returns>
-        private static long BitPosBitSearch2(byte* input, long inputLen, long startBitOffset, long endBitOffset, byte searchFor)
-        {
-            // Initialize variables
-            var currentBitOffset = startBitOffset;
-            var searchBit = searchFor == 1;
-            var mask = searchBit ? 0x00 : 0xff;
-
-            // Iterate through the byte array
-            while (currentBitOffset <= endBitOffset)
-            {
-                // Calculate byte and bit positions
-                var byteIndex = currentBitOffset / 8;
-                var bitIndex = (int)(currentBitOffset & 7);
-
-                var payload = input[byteIndex] & (1 << 8 - (int)bitIndex) - 1;
-                if (mask != payload)
-                {
-                    // Create a payload with the current byte shifted to the most significant byte position
-                    payload = payload << 56;
-
-                    // Trim leading bits
-                    payload <<= bitIndex;
-
-                    // Transform to count leading zeros
-                    payload = searchBit ? payload : ~payload;
-
-                    // Check if the payload matches the mask
-                    // Calculate the position of the first matching bit
-                    var lzcnt = (long)Lzcnt.X64.LeadingZeroCount((ulong)payload);
-                    return currentBitOffset + lzcnt;
-                }
-
-                // Move to the next byte
-                currentBitOffset += 8 - bitIndex;
-            }
-
-            // Return -1 if no matching bit is found
-            return -1;
-        }
-
         private static long BitPosBitSearch(byte* input, long inputLen, long startBitOffset, long endBitOffset, byte searchFor)
         {
             var searchBit = searchFor == 1;
@@ -125,8 +86,6 @@ namespace Garnet.server
             return -1;
         }
 
-
-
         /// <summary>
         /// Search for position of bit set in byte array using byte offset for start and end range
         /// </summary>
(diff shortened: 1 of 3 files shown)


prvyk:
This is great; I couldn't find time for this. But one thing, there was a test which I couldn't pass last try, and this one doesn't seem to pass it either (maybe I got the test wrong). Try applying this diff manually (or via git apply). It adds a few lines to BitmapBitPosOffsetsTest test.

```
diff --git a/test/Garnet.test/GarnetBitmapTests.cs b/test/Garnet.test/GarnetBitmapTests.cs
index 44c94f45..55ece807 100644
--- a/test/Garnet.test/GarnetBitmapTests.cs
+++ b/test/Garnet.test/GarnetBitmapTests.cs
@@ -620,6 +620,9 @@ namespace Garnet.test
                 pos = db.StringBitPosition(key, set, startOffset, endOffset);
 
                 ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset}");
+
+                pos = db.StringBitPosition(key, set, startOffset * 8, endOffset * 8, StringIndexType.Bit);
+                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset * 8} {endOffset * 8} bit");
             }
 
             // check negative offsets in range
@@ -635,6 +638,9 @@ namespace Garnet.test
                 expectedPos = Bitpos(buf, startOffset, endOffset, set);
                 pos = db.StringBitPosition(key, set, startOffset, endOffset);
                 ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset}");
+
+                pos = db.StringBitPosition(key, set, startOffset * 8, endOffset * 8, StringIndexType.Bit);
+                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset * 8} {endOffset * 8} bit");
             }
         }
 

```

You'd expect referring to the same positions (in different syntax) lead to the same result? Yet the test fails:

>   Failed BitmapBitPosOffsetsTest [288 ms]
>   Error Message:
>      1 False 4032 4040 bit
> Assert.That(actual, Is.EqualTo(expected))
>   Expected: 4037
>   But was:  4040
> 
>   Stack Trace:
>      at NUnit.Framework.Legacy.ClassicAssert.AreEqual(Object expected, Object actual, String message, Object[] args)
>    at Garnet.test.GarnetBitmapTests.BitmapBitPosOffsetsTest() in garnet/test/Garnet.test/GarnetBitmapTests.cs:line 625
> 
> 1)    at NUnit.Framework.Legacy.ClassicAssert.AreEqual(Object expected, Object actual, String message, Object[] args)
>    at Garnet.test.GarnetBitmapTests.BitmapBitPosOffsetsTest() in garnet/test/Garnet.test/GarnetBitmapTests.cs:line 625
> 

(Last time I also tried with a randomized seed a couple of times just to be sure it always passes).

libs/server/Resp/Bitmap/BitmapManagerBitPos.cs
```diff
@@ -1,185 +1,163 @@
...
+                {
+                    var payload = *(long*)(input + currentStartOffset);
+                    payload = BinaryPrimitives.ReverseEndianness(payload);
+
+                    // Process only if payload is valid (i.e. not all bits are set or clear based on searchFor parameter)
+                    if (payload != invalidMask64)
+                    {
+                        // Transform to count leading zeros
+                        payload = searchBit ? payload : ~payload;
+                        var lzcnt = (long)Lzcnt.X64.LeadingZeroCount((ulong)payload);
```
PaulusParssinen, line 123: While we're here, all CPU arch specific intrinsic calls should use portable ones, i.e. `BitOperations.LeadingZeroCount` or should either be guarded by `.IsSupported` calls.

vazois: Good point I keep forgetting. Added portable ones!

prvyk:
It turns out the distribution of the default seed is somehow very specific. I tried a few, and they pass consistently except for one case: when startOffset equals endOffset. Didn't find problems when that was excluded.

```
diff --git a/test/Garnet.test/GarnetBitmapTests.cs b/test/Garnet.test/GarnetBitmapTests.cs
index 827f0e25..b953a005 100644
--- a/test/Garnet.test/GarnetBitmapTests.cs
+++ b/test/Garnet.test/GarnetBitmapTests.cs
@@ -607,6 +607,11 @@ namespace Garnet.test
             long expectedPos;
             long pos;
 
+	     //var r2 = new Random();
+	     //int x = r2.Next();
+	     int x = 1756844280; // 243876438 also fails with same case.
+	     r = new Random(x);
+
             for (var j = 0; j < iter; j++)
             {
                 r.NextBytes(buf);
@@ -619,12 +624,12 @@ namespace Garnet.test
                 expectedPos = Bitpos(buf, startOffset, endOffset, set);
                 pos = db.StringBitPosition(key, set, startOffset, endOffset);
 
-                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset}");
+                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset} {x}");
 
                 var startBitOffset = startOffset << 3;
                 var endBitOffset = endOffset << 3;
                 pos = db.StringBitPosition(key, set, startBitOffset, endBitOffset, StringIndexType.Bit);
-                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startBitOffset} {endBitOffset} bit");
+                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startBitOffset} {endBitOffset} bit {x}");
             }
 
             // check negative offsets in range

```

Result is:

>   Failed BitmapBitPosOffsetsTest [294 ms]
>   Error Message:
>      9 False 3264 3264 bit 1756844280
> Assert.That(actual, Is.EqualTo(expected))
>   Expected: 3265
>   But was:  -1
> 

(bit output actually can differ in redis, but only when end argument isn't applied and the result is extended past end of string. This isn't the case in this test, so likely a bug.)

vazois:
> It turns out the distribution of the default seed is somehow very specific. I tried a few, and they pass consistently except for one case: when startOffset equals endOffset. Didn't find problems when that was excluded.
> 
> ```
> diff --git a/test/Garnet.test/GarnetBitmapTests.cs b/test/Garnet.test/GarnetBitmapTests.cs
> index 827f0e25..b953a005 100644
> --- a/test/Garnet.test/GarnetBitmapTests.cs
> +++ b/test/Garnet.test/GarnetBitmapTests.cs
> @@ -607,6 +607,11 @@ namespace Garnet.test
>              long expectedPos;
>              long pos;
>  
> +	     //var r2 = new Random();
> +	     //int x = r2.Next();
> +	     int x = 1756844280; // 243876438 also fails with same case.
> +	     r = new Random(x);
> +
>              for (var j = 0; j < iter; j++)
>              {
>                  r.NextBytes(buf);
> @@ -619,12 +624,12 @@ namespace Garnet.test
>                  expectedPos = Bitpos(buf, startOffset, endOffset, set);
>                  pos = db.StringBitPosition(key, set, startOffset, endOffset);
>  
> -                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset}");
> +                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startOffset} {endOffset} {x}");
>  
>                  var startBitOffset = startOffset << 3;
>                  var endBitOffset = endOffset << 3;
>                  pos = db.StringBitPosition(key, set, startBitOffset, endBitOffset, StringIndexType.Bit);
> -                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startBitOffset} {endBitOffset} bit");
> +                ClassicAssert.AreEqual(expectedPos, pos, $"{j} {set} {startBitOffset} {endBitOffset} bit {x}");
>              }
>  
>              // check negative offsets in range
> ```
> 
> Result is:
> 
> > Failed BitmapBitPosOffsetsTest [294 ms]
> > Error Message:
> > 9 False 3264 3264 bit 1756844280
> > Assert.That(actual, Is.EqualTo(expected))
> > Expected: 3265
> > But was:  -1
> 
> (bit output actually can differ in redis, but only when end argument isn't applied and the result is extended past end of string. This isn't the case in this test, so likely a bug.)

The seed needs to be specific because the unit tests need to be deterministic.
Overall, the problem with this test is that it compares against a naive implementation of bitpos which is not ideal if that naive implementation has a bug.
In this case, the BitPos implementation that we are comparing against does not work with the BIT option only the BYTE option which explains the inconsistency for a single bit search.
You might want to try another test that sets or clears a bit at a specific position, so you know the expected pos return,
or maybe have a trace of the Redis output.
I added a simple test to ensure correctness when range is a single bit.

image

prvyk:
I'm aware of the seed requirement (it would be a bad idea to change it), I'm just using this to verify.

Now, the bitops implementation is naive, yet the exception is on the bit test. So the byte test 'passed' with the same result as the naive implementation - which suggests a BYTE range and an equivalent BIT range returned a different result. I don't think this should be happening when the end offset is specified?

vazois:
> Edit: Ah, but the redis BYTE range includes the entire byte when the naive bit range is limited to a bit? So it's not the same range exactly. I was wrong here, sorry.

Yes, I was about to explain that. Cool, if you find something else let me know. It has been tricky to test it extensively.

badrishc (approved):

prvyk:
> > Edit: Ah, but the redis BYTE range includes the entire byte when the naive bit range is limited to a bit? So it's not the same range exactly. I was wrong here, sorry.
> 
> Yes, I was about to explain that. Cool, if you find something else let me know. It has been tricky to test it extensively and you did a great job providing feedback.

I compared with a Redis implementation and couldn't find differences when start end arguments are given and identical regardless of bit/byte.

PR #1000, opened 2025-02-05
Added support to expire specfic member of a sorted set

Vijay-Nirmal:
Adding support for member-specific expiration in a sorted set. This PR adds ZEXPIRE, ZPEXPIRE, ZEXPIREAT, ZPEXPIREAT, ZTTL, ZPTTL, ZEXPIRETIME, ZPEXPIRETIME, ZPERSIST and ZCOLLECT commands to garnet to support this feature.

- [x] Implementation of ZEXPIRE, ZPEXPIRE, ZEXPIREAT, ZPEXPIREAT, ZTTL, ZPTTL, ZEXPIRETIME, ZPEXPIRETIME and ZPERSIST commands
- [x] Implementation of ZCOLLECT to delete expired fields on demand
- [x] Add Integration Test cases to get all the new commands
- [x] Add Integration to test all the existing Hash set commands with expiring fields
- [x] ACL Test
- [x] Slot Verification Test
- [x] Add documentation
- [x] Add a background task to call ZCOLLECT to delete expired items
- [x] Benchmarking and making sure there is no performance regression

Since this is a garnet specific command, I used hashset expire API format as the base line. Below is the syntax for each command
```
ZEXPIRE => ZEXPIRE key seconds [NX | XX | GT | LT] MEMBERS nummembers member [member ...]
ZPEXPIRE => ZPEXPIRE key milliseconds [NX | XX | GT | LT] MEMBERS nummembers member [member ...]
ZEXPIREAT => ZEXPIREAT key unix-time-seconds [NX | XX | GT | LT] MEMBERS nummembers member [member ...]
ZPEXPIREAT => ZPEXPIREAT key unix-time-milliseconds [NX | XX | GT | LT] MEMBERS nummembers member [member ...]
ZTTL => ZTTL key MEMBERS nummembers member [member ...]
ZPTTL => ZPTTL key MEMBERS nummembers member [member ...]
ZEXPIRETIME => ZEXPIRETIME key MEMBERS nummembers member [member ...]
ZPEXPIRETIME => ZPEXPIRETIME key MEMBERS nummembers member [member ...]
ZPERSIST => ZPERSIST key MEMBERS nummembers member [member ...]
ZCOLLECT => ZCOLLECT key [key ...]
```

PR #1021, opened 2025-02-14
Combining ZRange options in SortedSetObject

Vijay-Nirmal:
This PR combines multiple ZRange options into one to decrease the number of items in the enum to make sure it's less than 32

Details: https://github.com/microsoft/garnet/pull/1000#issuecomment-2644510985


libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -494,7 +495,7 @@ public unsafe GarnetStatus SortedSetRange<TObjectContext>(ArgSlice key, ArgSlice
             }
 
             // Reverse
-            if (sortedOperation != SortedSetOperation.ZREVRANGE && reverse)
+            if (reverse && sortedSetOrderOperation != SortedSetOrderOperation.ByRank)
```
TalZaccai, line 498: Shouldn't this be `if (reverse && ((sortedSetOrderOperation  & SortedSetOrderOperation.Reverse) == 0))`?

TalZaccai (changes requested):
Looks good! Thanks for doing this!

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -494,7 +495,7 @@ public unsafe GarnetStatus SortedSetRange<TObjectContext>(ArgSlice key, ArgSlice
...
             {
                 arguments.Add(scratchBufferManager.CreateArgSlice("REV"u8));
```
TalZaccai, line 500: Won't it be cheaper to do `rangeOpts |= SortedSetRangeOpts.Reverse`?

Vijay-Nirmal: Yes, Don't want to make many changes to this as there is no proper unit test for this method to validate all path

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: There is no `SortedSetOrderOperation.Reverse` also, it's not flagged enum. Current logic should be the same as before. Anyway I have changed the logic to use the reverse flag directly which should solve both this and the other command. Let me know if it has issues.

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
Vijay-Nirmal: Fixed

TalZaccai (approved):

PR #1023, opened 2025-02-14
BITFIELD and BITFIELD_RO improvements

prvyk:
It's very easy to get a parsing exception with either:

BITFIELD a SET i8
BITFIELD a GET i8 ""

(I managed to get crashes but it didn't replicate later. I guess it depends on build settings).

Error messages can be confusing:

bitfield_ro a OVERFLOW fail GET i8 1
ERR Invalid bitfield type. Use something like i16 u8. Note that u64 is not supported but i64 is

Return type is wrong when key was not found:

BITFIELD nx GET i8 0
1) "0"
(should be an integer).

----

First commit cleans up the parsing and separates the '_RO' version so both sides are simplified and give decent messages. Adds a test.

Next commit continues the _RO separation to the storage layer.
It's very mechanical (we know only the GET operation can activate and therefore overflow is false and then edit accordingly). I suspect the _RO part can be reduced a bit more but didn't want to risk it.

### DIFF
--- a/libs/server/Resp/Bitmap/BitmapCommands.cs
+++ b/libs/server/Resp/Bitmap/BitmapCommands.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -10,6 +9,8 @@ using Tsavorite.core;
 
 namespace Garnet.server
 {
+    using SecondaryCommandList = List<(RespCommand, ArgSlice[])>;
+
     ///  (1) , (2) , (3) 
     /// overflow check, ptr protection, and status not found implemented for below
     /// GETBIT, SETBIT, BITCOUNT, BITPOS (1),(2)
@@ -359,7 +360,7 @@ namespace Garnet.server
         /// <summary>
         /// Performs arbitrary bitfield integer operations on strings.
         /// </summary>
-        private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly = false)
+        private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi)
             where TGarnetApi : IGarnetApi
         {
             if (parseState.Count < 1)
@@ -371,12 +372,11 @@ namespace Garnet.server
             // Extract Key
             var sbKey = parseState.GetArgSliceByRef(0).SpanByte;
 
-            var currTokenIdx = 1;
-
             var isOverflowTypeSet = false;
             ArgSlice overflowTypeSlice = default;
-            var secondaryCommandArgs = new List<(RespCommand, ArgSlice[])>();
+            var secondaryCommandArgs = new SecondaryCommandList();
 
+            var currTokenIdx = 1;
             while (currTokenIdx < parseState.Count)
             {
                 // Get subcommand
@@ -384,80 +384,56 @@ namespace Garnet.server
                 var command = commandSlice.ReadOnlySpan;
 
                 // Process overflow command
-                if (!readOnly && command.EqualsUpperCaseSpanIgnoringCase("OVERFLOW"u8))
+                if (command.EqualsUpperCaseSpanIgnoringCase("OVERFLOW"u8))
                 {
-                    // Get overflow parameter
-                    overflowTypeSlice = parseState.GetArgSliceByRef(currTokenIdx);
-                    isOverflowTypeSet = true;
-
                     // Validate overflow type
-                    if (!parseState.TryGetBitFieldOverflow(currTokenIdx, out _))
+                    if (currTokenIdx >= parseState.Count || !parseState.TryGetBitFieldOverflow(currTokenIdx, out _))
                     {
-                        while (!RespWriteUtils.TryWriteError(
-                                   $"ERR Overflow type {parseState.GetString(currTokenIdx)} not supported",
-                                   ref dcurr, dend))
+                        while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_INVALID_OVERFLOW_TYPE, ref dcurr, dend))
                             SendAndReset();
                         return true;
                     }
 
-                    currTokenIdx++;
+                    // Get overflow parameter
+                    overflowTypeSlice = parseState.GetArgSliceByRef(currTokenIdx++);
+                    isOverflowTypeSet = true;
 
                     continue;
                 }
 
                 // [GET <encoding> <offset>] [SET <encoding> <offset> <value>] [INCRBY <encoding> <offset> <increment>]
                 // Process encoding argument
-                var encodingSlice = parseState.GetArgSliceByRef(currTokenIdx);
-                var offsetSlice = parseState.GetArgSliceByRef(currTokenIdx + 1);
-                var encodingArg = parseState.GetString(currTokenIdx);
-                var offsetArg = parseState.GetString(currTokenIdx + 1);
-                currTokenIdx += 2;
-
-                // Validate encoding
-                if (encodingArg.Length < 2 ||
-                    (encodingArg[0] != 'i' && encodingArg[0] != 'u') ||
-                    !int.TryParse(encodingArg.AsSpan(1), out var bitCount) ||
-                    bitCount > 64 ||
-                    (bitCount == 64 && encodingArg[0] == 'u'))
+                if ((currTokenIdx >= parseState.Count) || !parseState.TryGetBitfieldEncoding(currTokenIdx, out _, out _))
                 {
                     while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_INVALID_BITFIELD_TYPE, ref dcurr,
                                dend))
                         SendAndReset();
                     return true;
                 }
+                var encodingSlice = parseState.GetArgSliceByRef(currTokenIdx++);
 
-                // Validate offset
-                var isOffsetValid = offsetArg[0] == '#'
-                    ? long.TryParse(offsetArg.AsSpan(1), out _)
-                    : long.TryParse(offsetArg, out _);
-
-                if (!isOffsetValid)
+                // Process offset argument
+                if ((currTokenIdx >= parseState.Count) || !parseState.TryGetBitfieldOffset(currTokenIdx, out _, out _))
                 {
                     while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_GENERIC_BITOFFSET_IS_NOT_INTEGER, ref dcurr,
                                dend))
                         SendAndReset();
                     return true;
                 }
+                var offsetSlice = parseState.GetArgSliceByRef(currTokenIdx++);
 
-                // Subcommand takes 2 args, encoding and offset
-                if (command.EqualsUpperCaseSpanIgnoringCase("GET"u8))
+                // GET Subcommand takes 2 args, encoding and offset
+                if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.GET))
                 {
                     secondaryCommandArgs.Add((RespCommand.GET, [commandSlice, encodingSlice, offsetSlice]));
                 }
                 else
                 {
-                    if (readOnly)
-                    {
-                        while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_SYNTAX_ERROR, ref dcurr, dend))
-                            SendAndReset();
-                        return true;
-                    }
-
                     RespCommand op;
                     // SET and INCRBY take 3 args, encoding, offset, and valueArg
-                    if (command.EqualsUpperCaseSpanIgnoringCase("SET"u8))
+                    if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.SET))
                         op = RespCommand.SET;
-                    else if (command.EqualsUpperCaseSpanIgnoringCase("INCRBY"u8))
+                    else if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.INCRBY))
                         op = RespCommand.INCRBY;
                     else
                     {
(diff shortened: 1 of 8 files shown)


libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -527,14 +571,50 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly
...
+
+            if (offsetSlice.Length == 0)
+                return false;
+
+            var offsetArg = offsetSlice.ToString();
+
+            return offsetArg[0] == '#'
+                ? long.TryParse(offsetArg.AsSpan(1), out _)
+                : long.TryParse(offsetArg, out _);
         }
```
vazois, line 618: Is it possible to unify this so as not to pollute the namespace?

```suggestion
        bool GetEncodingAndOffset(SessionParseState parseState,
                                 ref int currTokenIdx,
                                 out ArgSlice encodingSlice,
                                 out ArgSlice offsetSlice)
        {
            encodingSlice = default;
            offsetSlice = default;
            // [GET <encoding> <offset>] [SET <encoding> <offset> <value>] [INCRBY <encoding> <offset> <increment>]
            // Process encoding argument
            if (!ValidateEncoding(parseState, currTokenIdx++, out encodingSlice))
            {
                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_INVALID_BITFIELD_TYPE, ref dcurr,
                           dend))
                    SendAndReset();
                return false;
            }

            // Process offset argument
            if (!ValidateOffset(parseState, currTokenIdx++, out offsetSlice))
            {
                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_GENERIC_BITOFFSET_IS_NOT_INTEGER, ref dcurr,
                           dend))
                    SendAndReset();
                return false;
            }
            return true;


            bool ValidateEncoding(SessionParseState parseState,
                                     int currTokenIdx,
                                     out ArgSlice encodingSlice)
            {
                if (currTokenIdx >= parseState.Count)
                {
                    encodingSlice = default;
                    return false;
                }

                encodingSlice = parseState.GetArgSliceByRef(currTokenIdx);

                var encodingArg = encodingSlice.ToString();

                return
                    encodingArg.Length >= 2 &&
                    (encodingArg[0] == 'i' || encodingArg[0] == 'u') &&
                    int.TryParse(encodingArg.AsSpan(1), out var bitCount) &&
                    (bitCount < 64 ||
                        (bitCount == 64 && encodingArg[0] == 'i')
                    )
                ;
            }

            bool ValidateOffset(SessionParseState parseState, int currTokenIdx, out ArgSlice offsetSlice)
            {
                if (currTokenIdx >= parseState.Count)
                {
                    offsetSlice = default;
                    return false;
                }

                offsetSlice = parseState.GetArgSliceByRef(currTokenIdx);

                if (offsetSlice.Length == 0)
                    return false;

                var offsetArg = offsetSlice.ToString();

                return offsetArg[0] == '#'
                    ? long.TryParse(offsetArg.AsSpan(1), out _)
                    : long.TryParse(offsetArg, out _);
            }
        }
```

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -527,14 +571,50 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly
...
+                offsetSlice = default;
+                return false;
+            }
+
+            offsetSlice = parseState.GetArgSliceByRef(currTokenIdx);
+
+            if (offsetSlice.Length == 0)
+                return false;
+
+            var offsetArg = offsetSlice.ToString();
```
vazois, line 613: better to avoid doing ToString(), it results in allocation and is bad for performance.
Try offsetArg.ptr[0] = '#'

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -527,14 +571,50 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly
...
-            return StringBitField(ref storageApi, true);
+            if (currTokenIdx >= parseState.Count)
+            {
+                encodingSlice = default;
+                return false;
+            }
+
+            encodingSlice = parseState.GetArgSliceByRef(currTokenIdx);
+
+            var encodingArg = encodingSlice.ToString();
```
vazois, line 587: Same here, don't use ToString. encodingArg.Length should work.

vazois (commented):
Some minor initial comments

libs/server/Resp/Bitmap/BitmapCommands.cs
prvyk: How about putting them in parseState? Overflow is already there.

libs/server/Resp/Bitmap/BitmapCommands.cs
prvyk: Done.

libs/server/Resp/Bitmap/BitmapCommands.cs
prvyk: Done.

libs/server/Resp/Bitmap/BitmapCommands.cs
vazois: @TalZaccai what do you think?

prvyk: I suspect the most efficient solution would be to be able to pass a pointer in RawStringInput and then do the parsing only once - instead of at this layer and later again in MainSessionFunctions.GetBitFieldArguments() - but changing RawStringInput was too much for this PR.

vazois: Just to clarify, you mean something like the following. I think it might be a good idea.
Let the other chime in also. @TalZaccai, @badrishc 

// Bitfield method
                var encodingSlice = default(ArgSlice);
                if (!parseState.TryGetBitfieldEncoding(currTokenIdx++, ref encodingSlice))
                {
                    while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_INVALID_BITFIELD_TYPE, ref dcurr,
                               dend))
                        SendAndReset();
                    return true;
                }       
// ParseState
        public bool TryGetBitfieldEncoding(int i, ref ArgSlice encodingSlice)
        {
            encodingSlice = default;
            if (i >= Count)
                return false;

            encodingSlice = ref Unsafe.AsRef<ArgSlice>(bufferPtr + i);

            return
                encodingSlice.Length >= 2 &&
                (encodingSlice.ptr[0] == 'i' || encodingSlice.ptr[0] == 'u') &&
                int.TryParse(encodingSlice.Span.Slice(1), out var bitCount) &&
                (bitCount < 64 ||
                    (bitCount == 64 && encodingSlice.ptr[0] == 'i')
                );
        }

libs/server/SessionParseStateExtensions.cs
```diff
@@ -139,6 +139,60 @@ internal static bool TryGetBitFieldOverflow(this SessionParseState parseState, i
...
+        }
+
+        /// <summary>
+        /// Parse bit field OFFSET slice from parse state at specified index
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="offsetSlice">Parsed slice</param>
+        /// <returns></returns>
+        internal static unsafe bool TryGetOFfsetSlice(this SessionParseState parseState, int idx, out ArgSlice offsetSlice)
```
vazois, line 175: nit: TryGetOffsetSlice()

libs/server/Resp/Parser/ParseUtils.cs
```diff
@@ -53,6 +53,24 @@ public static bool TryReadInt(ref ArgSlice slice, out int number)
...
         }
 
+        /// <summary>
+        /// Try to read a signed 32-bit long from a given ArgSlice and offset.
+        /// </summary>
+        /// <returns>
+        /// True if integer parsed successfully
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryReadInt(ref ArgSlice slice, int offset, out int number)
```
vazois, line 63: I think it is better to do this in the caller. In fact, you need to add ArgSlice.Slice() method and use it with the already implemented TryReadInt

libs/server/Resp/Parser/ParseUtils.cs
```diff
@@ -92,6 +110,23 @@ public static bool TryReadLong(ref ArgSlice slice, out long number)
...
         }
 
+        /// <summary>
+        /// Try to read a signed 64-bit long from a given ArgSlic and offset..
+        /// </summary>
+        /// <returns>
+        /// True if long parsed successfully
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryReadLong(ref ArgSlice slice, int offset, out long number)
```
vazois, line 120: Same as above. We should add Slice to ArgSlice and use it with existing TryReadLong

libs/server/SessionParseStateExtensions.cs
```diff
@@ -139,6 +139,60 @@ internal static bool TryGetBitFieldOverflow(this SessionParseState parseState, i
...
         }
 
+        /// <summary>
+        /// Parse bit field ENCODING slice from parse state at specified index
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="encodingSlice">Parsed slice</param>
+        /// <returns></returns>
+        internal static unsafe bool TryGetEncodingSlice(this SessionParseState parseState, int idx, out ArgSlice encodingSlice)
```
vazois, line 149: nit: TryGetBitfieldEncoding

libs/server/SessionParseStateExtensions.cs
vazois: Sorry, I should have written; TryGetBitfieldOffset

vazois (changes requested):
Great work! Some minor comments

libs/server/SessionParseStateExtensions.cs
prvyk: Done.

libs/server/Resp/Parser/ParseUtils.cs
prvyk: >you need to add ArgSlice.Slice()

Wouldn't that be an extra allocation?

(in ArgSlice.cs):
```
        public ArgSlice Slice(int offset)
        {
            // Should we check length here since others may call the API? What would failure look like?
            // Adding an extra TryReadLong with offset seems clearer than this...
            return new ArgSlice(ptr + 1, length - 1);
        }
```

Or am I misunderstanding the idea?

libs/server/SessionParseStateExtensions.cs
prvyk: Done.

libs/server/Resp/Parser/ParseUtils.cs
vazois: I was thinking that it might not be a big issue given that struct are allocated in the stack (though ArgSlice is not a ref struct).@badrishc, @talzaccai any comments here? You can always check/add a BDN benchmark and see the allocations.
It is not necessary to add Slice for ArgSlice but I would put the corresponding ptr logic inside the TryGetBitfieldOffset and TryGetBitfieldEncoding instead of introducing new methods in ParseUtils.

prvyk: A struct version must be very slightly worse, perhaps we can gain more clarity by other means. How about ditching the int version and using the long version with encoding too? It doesn't really matter for encoding (the allowed values are much smaller; Both functions call TryReadUInt64 internally). The long function is now called from two locations, and it looks neat:

```
            var ret = (*offsetSlice.ptr == '#')
                ? ParseUtils.TryReadLong(ref offsetSlice, byteOffset: 1, out bitFieldOffset)
                : ParseUtils.TryReadLong(ref offsetSlice, out bitFieldOffset);
```

prvyk: Done.

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -469,24 +442,93 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly
...
+                    return true;
+                }
+
+                secondaryCommandArgs.Add((RespCommand.GET, [commandSlice, encodingSlice, offsetSlice]));
+            }
+
+            return StringFieldAction(ref storageApi, ref sbKey, RespCommand.BITFIELD_RO, secondaryCommandArgs);
+        }
+
+        private bool StringFieldAction<TGarnetApi>(ref TGarnetApi storageApi,
```
TalZaccai, line 520: Should this be `StringBitFieldAction`?

TalZaccai (changes requested):
Thanks! Added some comments...

libs/server/Resp/Bitmap/BitmapCommands.cs
prvyk: Done.

libs/server/SessionParseStateExtensions.cs
```diff
@@ -139,6 +139,78 @@ internal static bool TryGetBitFieldOverflow(this SessionParseState parseState, i
...
+        /// <summary>
+        /// Parse bit field OFFSET slice from parse state at specified index
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="offsetSlice">Parsed slice</param>
+        /// <returns></returns>
+        internal static unsafe bool TryGetBitfieldOffset(this SessionParseState parseState, int idx, out ArgSlice offsetSlice)
+        {
+            if (idx >= parseState.Count)
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/SessionParseStateExtensions.cs
```diff
@@ -139,6 +139,78 @@ internal static bool TryGetBitFieldOverflow(this SessionParseState parseState, i
...
+        /// <summary>
+        /// Parse bit field ENCODING slice from parse state at specified index
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="encodingSlice">Parsed slice</param>
+        /// <returns></returns>
+        internal static unsafe bool TryGetBitfieldEncoding(this SessionParseState parseState, int idx, out ArgSlice encodingSlice)
+        {
+            if (idx >= parseState.Count)
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Resp/Parser/ParseUtils.cs
```diff
@@ -92,6 +92,23 @@ public static bool TryReadLong(ref ArgSlice slice, out long number)
...
         }
 
+        /// <summary>
+        /// Try to read a signed 64-bit long from a given ArgSlic and byteOffset.
+        /// </summary>
+        /// <returns>
+        /// True if long parsed successfully
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryReadLong(ref ArgSlice slice, int byteOffset, out long number)
```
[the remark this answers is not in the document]
prvyk: Leftover from previous discussion. Removed.

libs/server/SessionParseStateExtensions.cs
```diff
@@ -139,6 +139,78 @@ internal static bool TryGetBitFieldOverflow(this SessionParseState parseState, i
...
         }
 
+        /// <summary>
+        /// Parse bit field ENCODING slice from parse state at specified index
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="encodingSlice">Parsed slice</param>
+        /// <returns></returns>
+        internal static unsafe bool TryGetBitfieldEncoding(this SessionParseState parseState, int idx, out ArgSlice encodingSlice)
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -371,93 +372,65 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi, bool readOnly
...
 
-            var currTokenIdx = 1;
-
             var isOverflowTypeSet = false;
             ArgSlice overflowTypeSlice = default;
-            var secondaryCommandArgs = new List<(RespCommand, ArgSlice[])>();
+            var secondaryCommandArgs = new SecondaryCommandList();
 
-            while (currTokenIdx < parseState.Count)
+            for (var currTokenIdx = 1; currTokenIdx < parseState.Count;)
```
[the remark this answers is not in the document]
prvyk: Done.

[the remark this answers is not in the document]
TalZaccai: Why not leave the `while (currTokenIdx < parseState.Count)`? Less prone to issues and then you don't have to have that extra check after.

TalZaccai (approved):
Nice! Looks much better now. One minor comment.

libs/server/Resp/Bitmap/BitmapCommands.cs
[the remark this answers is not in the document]
prvyk: Yeah, that would be better. Done.

vazois (approved):

PR #1034, opened 2025-02-18
Platform independent bitcount operation

prvyk:
Tiny change to make the scalar bitcount operation platform independent. Use that as software fallback when Avx2 and other x86-based operations are not supported. Tested on ARM VM, BitmapSimpleBitCountTest didn't pass before and it passes after.


prvyk:
Obviously ARM Advsimd acceleration could be done in a same manner as SSE/AVX acceleration. Running tests on the VM was slow enough to not even try. The scalar operation is needed anyway and it works.

vazois (approved):

PR #1040, opened 2025-02-20
Always create barrier at checkpoint version switch

badrishc:
Simplifying side effects:
* A v thread will never see a v+1 entry
* No need to cpr_shift_detected
* There is no longer a _fuzzy region_ in the log, i.e., a region with both v and v+1 records.
* Recovery simplified, as it does not need to skip v+1 versions


badrishc:
Closing in favor of https://github.com/microsoft/garnet/pull/1059

PR #1041, opened 2025-02-22
Add ability to disable intrinsics for tests

prvyk:
This adds a DEBUG INTRINSICS command so the fallback can be tested. The command allows turning off their use even when the runner supports the intrinsics. Use this with some BITCOUNT tests.

### DIFF
--- a/test/Garnet.test/GarnetBitmapTests.cs
+++ b/test/Garnet.test/GarnetBitmapTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.Intrinsics.X86;
 using Garnet.common;
 using Garnet.server;
 using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


prvyk:
@PaulusParssinen Do you think this is the right approach? 

* Used a static as this is how the typical Intrinsics are tested.
* Ignored the non-constant warning since in ordinary use this is a constant set at startup, only in single-threaded tests this is used.
* Another method of doing this would have been to add a GarnetServerOption, start another instance for the overrides,  that would have been much more complicated in the testsuit...

prvyk:
RemoteExecutor has a nice API. It's not trivial to install**, and it fails with marshalling errors (a la https://github.com/dotnet/arcade/issues/14634 ). I suspect it's not compatible with NUnit but rather with xUnit - which the dotnet team uses. I do like xUnit more, but rewriting the suit is a bit much...

Another idea was to use [Tmds.ExecFunction](https://github.com/tmds/Tmds.ExecFunction) - which is apparently incompatible with the v3 test platform ("unable to determine parent 'dotnet' process").

** From arcade's package repo, so one has to either vendor or get the action runners to import from there too.

prvyk:
The idea here is to include GarnetServer itself as a dependency, and then execute that from the build. It's a bit of an hack (getting the .exe extension on Windows or hardcoding the port) - let's see if it will pass the builders here.

test/Garnet.test/GarnetBitmapTests.cs
```diff
@@ -259,25 +264,53 @@ public void BitmapSetGetBitTest_LTM(bool preSet)
...
+                var port = 7000;
 
-            long count = db.StringBitCount(key);
-            ClassicAssert.AreEqual(expectedCount, count);
+                Dictionary<string, string> env = new()
+                {
+                    { "DOTNET_EnableHWIntrinsic", "0" }
+                };
+
+                using var p = TestUtils.StartProcess(typeof(Garnet.Program), port, env);
```
badrishc, line 286: instead of using Garnet.Program, can you create your own test-only GarnetServer process that does not load all the extensions. We ideally don't want to pollute the main/ codebase with these InternalsVisibleTo. Or have to change GarnetServer.csproj for this.

prvyk: Done.

Trivia: using Unix domain sockets for this is consistently slower. Possibly setup costs, but possible something along the stack is not optimized.

badrishc: > Done.
> 
> Trivia: using Unix domain sockets for this is consistently slower. Possibly setup costs, or perhaps it's not the most optimized path in .NET.

Thanks! 

@PaulusParssinen - any idea why unix domain sockets would be slower? We should add a BDN for this at some point ...

badrishc: I didn't see any InternalsVisibleTo in your test server, could we have just used main/GarnetServer without any modifications to it's contents?

prvyk: Done, with only csproj changes to GarnetServer.

What happens is as follows:

* We want to ensure an exe file is built when building the testsuit to run it as a process.

* We need to figure out its location too. 
The testsuit could be built from a number of runners. VSTest, dotnet test, etc.
dotnet test could be run from the solution or from the project, and it could have the --output argument to build at an entirely different filesystem.
 
* Most obvious method to handle both is including GarnetServer as a dependency of the testsuit. At which point it's likely in the same directory (just in case we do Assembly.Location instead of assuming the same directory and constant filename). This seems reasonable given the new usage, as now it is a dependency of sorts, but it leads to csproj changes:

* * We need to strong sign GarnetServer when using it as a dependency otherwise the build fails.
  * More significantly, Some files are included in both projects and so conflict with themselves. The InternalsVisibleTo attribute was a way to remove the file includes but without changing their visibility. That's why a separate test server project didn't need that - it didn't include the custom transaction code again.
     
So we need to achieve the same effect without changing the contents. It turns out .NET 5 added a way to do InternalsVisibleTo in the csproj. Since we need to change the csproj anyway, this seems reasonable.

badrishc (approved):

vazois (approved):

PR #1047, opened 2025-02-24
LMOVing a list element to same list should not remove the element.

prvyk:
RPUSH list a b c
LMOVE list test-lmove1 RIGHT RIGHT
LRANGE list 0 -1

Redis:
"a"
"b"
"c"

Garnet:
"a"
"b"

We just need to remove an apparently unnecessary check (all tests pass) to fix this. Add a test too.


libs/server/Storage/Session/ObjectStore/ListOps.cs
```diff
@@ -299,9 +299,9 @@ public GarnetStatus ListMove(ArgSlice sourceKey, ArgSlice destinationKey, Operat
                     else
                     {
                         // When the source and the destination key is the same the operation is done only in the sourceList
-                        if (sourceDirection == OperationDirection.Right && destinationDirection == OperationDirection.Left)
+                        if (destinationDirection == OperationDirection.Left)
```
TalZaccai, line 302: Nice catch! You can even break earlier before the `RemoveFirst`/`RemoveLast` if the directions are the same.

TalZaccai (changes requested):
Nice catch again here! Small change request...

libs/server/Storage/Session/ObjectStore/ListOps.cs
prvyk: Done.

TalZaccai (approved):

PR #1051, opened 2025-02-26
Gracefully terminating TCP connections during server.Dispose

Xizt:
When a k8s pod hosting garnet service is gracefully shutting down, socket.Dispose() terminates the tcp with a RST packet which is forced termination of the tcp connection. This causes the client to keep connecting with SYN packets with stale IP resulting in SYN flooding from SERedis client. 
Using socket.Shutdown() and socket.Close() to send a FIN closes the socket connection gracefully. SERedis client handles FIN by having exponential backoff retries.


Xizt:
If PR is okay, would also be incrementing the Garnet version in same PR.

libs/common/Networking/TcpNetworkHandlerBase.cs
```diff
@@ -140,6 +140,12 @@ void Start(bool useTLS)
         /// <inheritdoc />
         public override void Dispose()
         {
+            if (socket.Connected)
```
badrishc, line 143: Try this:

```cs
        try
        {
            if (socket.Connected)
            {
                // Gracefully shutdown the socket connection
                socket.Shutdown(SocketShutdown.Both);
            }
        }
        catch (Exception ex)
        {
            logger?.LogError(ex, "Error shutting down socket");
        }
        finally
        {
            // Always close the socket to release the resources
            socket.Close();
            // Dispose of the socket to free up unmanaged resources
            socket.Dispose();
        }
    }
```

Xizt: Thanks, updated.

badrishc (approved):

TalZaccai (approved):

PR #1057, opened 2025-02-27
Fix Connection Dispose when Object Store Disabled Exception is thrown

vazois:


test/Garnet.test/DisableObjectStoreTests.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT license.
+
+using System;
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test
+{
+    public class DisableObjectStoreTests
```
TalZaccai, line 11: Is it necessary to have a new test class just for this one test? Might want to consider moving it to `RespTests`, there are some other disabled-object-store-related tests there (e.g. `SingleDeleteWithObjectStoreDisabled`).

test/Garnet.test/DisableObjectStoreTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            var mykey = "mykey";
+            try
+            {
+                _ = db.ListLength(mykey);
+            }
+            catch (Exception ex)
+            {
+                var msg = ex.Message;
+                ClassicAssert.AreEqual("ERR Garnet Exception: Object store is disabled", msg);
```
TalZaccai, line 45: This assert won't be reached if an exception was not thrown (which we should require), consider something like this instead: 
```cs
var ex = Assert.Throws<RedisServerException>(() => db.ListLength(mykey));
ClassicAssert.AreEqual("ERR Garnet Exception: Object store is disabled", ex!.Message);
```

TalZaccai (changes requested):
Couple comments...

libs/common/GarnetException.cs
```diff
@@ -18,44 +18,62 @@ public class GarnetException : Exception
         public LogLevel LogLevel { get; } = LogLevel.Trace;
         public bool ClientResponse { get; } = true;
         public bool Panic { get; } = false;
+        public bool DisposeSession { get; } = false;
```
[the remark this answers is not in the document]
vazois: yes, it needs to be true by default. I will adjust

test/Garnet.test/DisableObjectStoreTests.cs
vazois: I will move this to RespTests, though it might be easier to have a class that contains all the ObjectStore negative tests, so people can add more if they want.

test/Garnet.test/DisableObjectStoreTests.cs
vazois: Done!

TalZaccai (approved):

PR #1058, opened 2025-02-28
Implement remaining GEOSEARCH functionality.

prvyk:
Implement GEOSEARCH BYRADIUS. 
Implement also ASC, DESC orderings, ANY, WITHHASH and FROMLONLAT.
Implement GEORADIUS_RO and GEORADIUSBYMEMBER_RO.
Implement GEORADIUS and GEORADIUSBYMEMBER.
Fix GEOSEARCHSTORE without STOREDIST and other bugs.
Add tests.


libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -149,24 +190,81 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
         /// <typeparam name="TGarnetApi"></typeparam>
+        /// <param name="cmd"></param>
         /// <param name="storageApi"></param>
         /// <returns></returns>
-        private unsafe bool GeoSearchStore<TGarnetApi>(ref TGarnetApi storageApi)
+        private unsafe bool GeoSearchStore<TGarnetApi>(RespCommand cmd, ref TGarnetApi storageApi)
             where TGarnetApi : IGarnetApi
         {
-            if (parseState.Count < 4)
+            int GetDestIdx()
```
TalZaccai, line 199: It would be more consistent with the rest of the code to validate all the arguments here and return the appropriate error message if necessary, rather than defer that to the storage layer. Refer to, for example, `SortedSetDifference` in `SortedSetCommands`.

libs/server/Resp/Parser/RespCommand.cs
```diff
@@ -1756,6 +1774,10 @@ private RespCommand SlowParseCommand(ref int count, ref ReadOnlySpan<byte> speci
             {
                 return RespCommand.ECHO;
             }
+            else if (command.SequenceEqual(CmdStrings.GEORADIUSBYMEMBER_RO))
```
TalZaccai, line 1777: Why is `GEORADIUSBYMEMBER_RO` in slow parse and `GEORADIUS_RO` etc. in fast parse?
I would probably put all of these commands in the slow parse path, as they are all essentially deprecated and can be replaced by `GEOSEARCH` / `GEOSEARCHSTORE`...

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -196,9 +201,71 @@ private static int AdminCommands(RespCommand command)
             };
         }
 
-        private int SortedSetObjectKeys(SortedSetOperation command, int inputCount)
+        private int GeoCommands(RespCommand command, int inputCount)
         {
+            var idx = 0;
+            ArgSlice key;
 
+            if (command == RespCommand.GEOSEARCHSTORE)
```
TalZaccai, line 209: This method could probably use some comments :)

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -196,9 +201,71 @@ private static int AdminCommands(RespCommand command)
...
+            {
+                key = respSession.parseState.GetArgSliceByRef(0);
+                SaveKeyEntryToLock(key, true, LockType.Exclusive);
+                SaveKeyArgSlice(key);
+                idx = 1;
+            }
+
+            key = respSession.parseState.GetArgSliceByRef(++idx);
```
TalZaccai, line 217: I don't think this is right, it would mean that the first key in commands that are not `GEOSEARCHSTORE` is in index 1? Shouldn't this be `idx++`?

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -196,9 +201,71 @@ private static int AdminCommands(RespCommand command)
...
+            SaveKeyEntryToLock(key, true, LockType.Shared);
+            SaveKeyArgSlice(key);
+
+            if (command == RespCommand.GEORADIUS_RO)
+                return 1;
+
+            if (command == RespCommand.GEORADIUSBYMEMBER_RO || command == RespCommand.GEORADIUSBYMEMBER)
+            {
+                key = respSession.parseState.GetArgSliceByRef(++idx);
```
TalZaccai, line 226: This would be a member, not a key, right?

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -196,9 +201,71 @@ private static int AdminCommands(RespCommand command)
...
+                        span.EqualsUpperCaseSpanIgnoringCase(CmdStrings.STOREDIST))
+                    {
+                        key = respSession.parseState.GetArgSliceByRef(i + 1);
+                        SaveKeyEntryToLock(key, true, LockType.Exclusive);
+                        SaveKeyArgSlice(key);
+                        break;
+                    }
+                }
+
+                return (command == RespCommand.GEORADIUS) ? 1 : 2;
```
TalZaccai, line 249: This should be 1 regardless

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -196,9 +201,71 @@ private static int AdminCommands(RespCommand command)
...
+            }
+
+            // GEOSEARCH, GEOSEARCHSTORE
+            for (var i = idx; i < inputCount - 1; ++i)
+            {
+                if (respSession.parseState.GetArgSliceByRef(i).ReadOnlySpan.EqualsUpperCaseSpanIgnoringCase(CmdStrings.FROMMEMBER))
```
TalZaccai, line 255: Again, these are members, not keys

test/Garnet.test/RespSortedSetGeoTests.cs
```diff
@@ -423,15 +465,37 @@ public void CanUseGeoSearchWithCities(int bytesSent)
...
+            ClassicAssert.IsTrue(actualValue.IndexOf("Washington") != -1);
+
+            responseBuf = lightClientRequest2.SendCommandChunks($"GEOSEARCH cities FROMLONLAT {lon} {lat} BYRADIUS 500 km COUNT 3 ANY WITHCOORD WITHDIST ASC", bytesSent, 16);
             expectedResponse = "*3\r\n*3\r\n$10\r\nWashington\r\n$1\r\n0\r\n*2\r\n$17\r\n-77.0368704199791\r\n$17\r\n38.90719190239906\r\n*3\r\n$12\r\nPhiladelphia\r\n$17\r\n198.4242996738795\r\n*2\r\n$18\r\n-75.16521960496902\r\n$18\r\n39.952582865953445\r\n*3\r\n$8\r\nNew York\r\n$18\r\n327.67645879712575\r\n*2\r\n$17\r\n-74.0059420466423\r\n$18\r\n40.712782591581345\r\n+PONG\r\n";
-            actualValue = Encoding.ASCII.GetString(responseBuf).Substring(0, expectedResponse.Length);
+            actualValue = Encoding.ASCII.GetString(responseBuf, 0, expectedResponse.Length);
+            ClassicAssert.IsTrue(actualValue.IndexOf("Washington") != -1);
+
+            using var lightClientRequest3 = TestUtils.CreateRequest();
+            responseBuf = lightClientRequest3.SendCommands("GEORADIUSBYMEMBER_RO cities Washington 500 km WITHCOORD WITHDIST ASC", "PING");
```
TalZaccai, line 491: Please also add tests for `GEORADIUS` and `GEORADIUSBYMEMBER`

TalZaccai (changes requested):
Thanks for this contribution, this is great! 
Added a few initial comments, haven't looked at `GeoSearch` in `SortedSetGeoObjectImpl` thoroughly yet...

libs/server/Objects/SortedSetGeo/GeoHash.cs
```diff
@@ -263,6 +263,20 @@ public static double Distance(double sourceLat, double sourceLon, double targetL
             return 2 * Math.Asin(Math.Sqrt(latHaversine + (tmp * lonHaversine))) * EarthRadiusInMeters;
         }
 
+        /// <summary>
+        /// Find if a point is in the circle.
+        /// </summary>
+        public static bool GetDistanceWhenInCircle(double radius, double latCenterPoint, double lonCenterPoint, double lat2, double lon2, ref double distance)
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Objects/SortedSetGeo/SortedSetGeoObjectImpl.cs
```diff
@@ -23,26 +24,104 @@ private struct GeoSearchData
...
             public double Distance;
+            public long GeoHash;
             public string GeoHashCode;
             public (double Latitude, double Longitude) Coordinates;
         }
 
+        /// <summary>
+        /// Type of GeoSearch
+        /// </summary>
+        internal enum GeoSearchType
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Resp/Parser/RespCommand.cs
prvyk: GEORADIUSBYMEMBER_RO is 20 characters, and so gets chopped up into 10 char parts and a fast parsing becomes more complicated than usual.

Moved everything to slow-parse path to keep it consistent.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Right. Fixed.

test/Garnet.test/RespSortedSetGeoTests.cs
prvyk: Done. TODO: check why radius search in CanUseGeoSearchStore() has an higher error level. There's a loss of accuracy here somewhere...

prvyk: What happens is that Washington is defined as (38.9071923, -77.0368707), but using a member leads to storing it and retrieving it (from long) as (38.907191902399063, -77.0368704199791). @PaulusParssinen, I guess that's normal?

I can emulate the accuracy loss by storing and retrieving a supplied lon/lat pair, and so get same result, but I doubt that's the intent here....

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -61,7 +61,7 @@ private unsafe bool GeoAdd<TGarnetApi>(ref TGarnetApi storageApi)
         /// <param name="command"></param>
         /// <param name="storageApi"></param>
         /// <returns></returns>
-        private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi storageApi)
+        private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi storageApi, bool @readonly = true)
```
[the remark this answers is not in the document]
prvyk: Parameter removed. We'll do it by changing the command instead.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -149,24 +190,81 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
         /// GEOSEARCHSTORE: Store the the members of a sorted set populated with geospatial data, which are within the borders of the area specified by a given shape.
         /// </summary>
         /// <typeparam name="TGarnetApi"></typeparam>
+        /// <param name="cmd"></param>
         /// <param name="storageApi"></param>
         /// <returns></returns>
-        private unsafe bool GeoSearchStore<TGarnetApi>(ref TGarnetApi storageApi)
+        private unsafe bool GeoSearchStore<TGarnetApi>(RespCommand cmd, ref TGarnetApi storageApi)
```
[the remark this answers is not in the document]
prvyk: I unified GeoCommands and GeoSearchStore so this doesn't apply anymore I guess. I'll put the general flow talk in the other conversation.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -92,20 +100,49 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                    break;
+                case RespCommand.GEODIST:
+                    op = SortedSetOperation.GEODIST;
+                    break;
+                case RespCommand.GEOPOS:
+                    op = SortedSetOperation.GEOPOS;
+                    break;
+                case RespCommand.GEORADIUS:
+                    op = SortedSetOperation.GEOSEARCH;
+                    opts = SortedSetGeoOpts.ByRadius | SortedSetGeoOpts.Store;
```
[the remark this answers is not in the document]
prvyk: Changed this to not set by default but examine the command first.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Ok, I unified the parsing later, and split out GeoSearchCommandParse() in the storage layer.

1) Call GeoSearchCommandParse in parsing layer too. This gives us early validation at the cost of tiny bit of speed.

2) Move GeoSearchCommandParse() code into GeoCommands() in the actual parsing layer, change signatures/API to accept GeoSearchOptions struct, and then somehow pass it to private void GeoSearch(). This is called from Operate(). The ref struct is too much for input or output, so I'm not sure how to go about this.

3) Move GeoSearchCommandParse() code into GeoCommands() in the actual parsing layer, change signatures/API to accept GeoSearchOptions struct, and then do it in a a bit higher layer above the Operate() thingy so we can pass the options there. So the actual RO operation is something like a SortedSetGet(), and then we call the equivalent of GeoSearch() there. The RW operation is similar (SET/EXPIRE etc.), maybe or maybe not that involves the ParseState.Read() hack. This sounds doable I think.

test/Garnet.test/RespSortedSetGeoTests.cs
prvyk: I see this happens also on regular Redis, though the offset between a member and itself is smaller there. In both servers, it's far less than the 0.5% error threshold GEOADD command page allows.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Done. The code flow is more idiomatic now, and there's no need for SortedSetGeoOpts flag at all.

test/Garnet.test/RespSortedSetGeoTests.cs
PaulusParssinen: Yes, that's sound like normal and within the expected GeoHash error bounds. When you go into the Geospatial rabbit hole, all sort horrors await you as it turns out "give me distance between two coordinates thanks" is quite difficult and inaccurate. 

Something I think in future we should do is remove all the instances from tests where we look for _exact_ matches of coordinates in the RESP responses as there's inherent inaccuracies/confusion involved that come with how the coordinates are stored in Garnet (and Redis etc) as 52-bit GeoHash integer. The string 11-character representation of this geohash string is then of course also still 52-bit precision, not 55-bit which may be assumed from a 11-character GeoHash "standard"<sub>there is no such thing, except one attempt just last year</sub> representation (base32 = 5-bits ber char)). So this and _other stuff_ makes me think tests and the Garnet/Redis can in general promise that the results are within certain error bounds and tests should also then maybe reflect that contract? Anyways, whenever we have extra time to do some test housekeeping that is something to maybe consider.

prvyk: There isn't even a Geohash standard??? !!!!

I think the GEODIST/GEOADD help pages** can be interpreted to give us some bounds. If base 0.5% distance error is acceptable, than additional error of 0.05% surely isn't much (then instead of 0.5% error, we have worst case 0.6% error), and maybe we can even say an 0.1% is fine (so worst case is 0.7%).

Note that Redis 8 already returns different results than Redis 7 (it rounds to different precisions), so even with Redis itself the only way for tests to properly work is use an error bound...

** https://microsoft.github.io/garnet/docs/commands/data-structures#geodist
Redis has the same comment in GEOADD page.

libs/server/Objects/SortedSetGeo/GeoHash.cs
```diff
@@ -263,6 +263,20 @@ public static double Distance(double sourceLat, double sourceLon, double targetL
...
+        {
+            distance = Distance(latCenterPoint, lonCenterPoint, lat2, lon2);
+            if (distance > radius)
```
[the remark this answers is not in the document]
TalZaccai: Oh I just meant to switch the `if` with a conditional `return`. I'm not sure if the radius should be included or not?

libs/server/Transaction/TxnKeyManager.cs
TalZaccai: There is no member-level locking, you just lock the source & dest object keys and that's it...

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -197,9 +202,76 @@ private static int AdminCommands(RespCommand command)
...
+            SaveKeyEntryToLock(key, true, LockType.Shared);
+            SaveKeyArgSlice(key);
+
+            if (command == RespCommand.GEORADIUS_RO)
+                return 1;
+
+            // Member commands have COMMAND key member syntax.
+            if (command == RespCommand.GEORADIUSBYMEMBER_RO || command == RespCommand.GEORADIUSBYMEMBER)
+            {
+                var member = respSession.parseState.GetArgSliceByRef(++idx);
```
TalZaccai, line 230: There's no member-level locking, you should lock the key that comes after STOREDIST (if exists)

TalZaccai (changes requested):
Added a couple of comments, still need to do a final pass here...

libs/server/Transaction/TxnKeyManager.cs
prvyk: Ah, got it. Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                        {
+                            argNumError = true;
+                            break;
+                        }
+
+                        // Read width, height
+                        if (!input.parseState.TryGetDouble(currTokenIdx++, out searchOpts.boxWidth) ||
+                            !input.parseState.TryGetDouble(currTokenIdx++, out var height))
+                        {
+                            errorMessage = CmdStrings.RESP_ERR_NOT_VALID_FLOAT;
```
TalZaccai, line 383: Error here should be `ERR need numeric width` / `ERR need numeric height` respectively

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                            break;
+                        }
+                        searchOpts.boxHeight = height;
+
+                        if (searchOpts.boxWidth < 0 || searchOpts.boxHeight < 0)
+                        {
```
TalZaccai, line 389: Error here should be: `ERR height or width cannot be negative`

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                }
+
+                if (tokenBytes.EqualsUpperCaseSpanIgnoringCase("ASC"u8))
+                {
+                    searchOpts.sort = GeoOrder.Ascending;
+                    continue;
+                }
+
+                if (tokenBytes.EqualsUpperCaseSpanIgnoringCase("DESC"u8))
+                {
```
TalZaccai, line 408: Should probably disallow the other if one is already set

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                    }
+
+                    if (!input.parseState.TryGetInt(currTokenIdx++, out var countValue))
+                    {
+                        errorMessage = CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER;
+                        break;
+                    }
+
+                    if (countValue <= 0)
+                    {
```
TalZaccai, line 428: Error here should be: `ERR COUNT must be > 0`

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                        errorMessage = CmdStrings.RESP_ERR_GENERIC_VALUE_IS_OUT_OF_RANGE;
+                        break;
+                    }
+
+                    searchOpts.countValue = countValue;
+                    continue;
+                }
+
+                if (tokenBytes.EqualsUpperCaseSpanIgnoringCase("ANY"u8))
```
TalZaccai, line 437: This should only be allowed immediately after parsing the count value

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                {
+                    searchOpts.withCountAny = true;
+                    continue;
+                }
+
+                if (!searchOpts.readOnly)
+                {
+                    if (byRadius && tokenBytes.EqualsUpperCaseSpanIgnoringCase(CmdStrings.STORE))
```
TalZaccai, line 445: using `byRadius` here (and in the `CmdStrings.STOREDIST` case) is a bit confusing, should probably filter by command

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -44,17 +43,72 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
           where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
             => ReadObjectStoreOperationWithOutput(key, ref input, ref objectContext, ref outputFooter);
 
+        /// <summary>
+        /// GEOSEARCH: Returns the members of a sorted set populated with geospatial data, which are within the borders of the area specified by a given shape.
```
TalZaccai, line 47: Should add the additional commands that call this method in the docs.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+            // Check if we have a wrong number of arguments
+            if (argNumError)
+            {
+                errorMessage = Encoding.ASCII.GetBytes(string.Format(CmdStrings.GenericErrWrongNumArgs,
+                                                       command.ToString()));
+            }
+
+            if (destIdx != -1)
+            {
+                searchOpts.readOnly = false;
```
TalZaccai, line 510: Wouldn't this already be false at this point? Otherwise we wouldn't have read destIdx...

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
```diff
@@ -44,17 +43,72 @@ public GarnetStatus GeoCommands<TObjectContext>(byte[] key, ref ObjectInput inpu
...
+        /// </summary>
+        /// <typeparam name="TObjectContext"></typeparam>
+        /// <param name="key"></param>
+        /// <param name="input"></param>
+        /// <param name="output"></param>
+        /// <param name="objectContext"></param>
+        /// <returns></returns>
+        public GarnetStatus GeoSearch<TObjectContext>(ArgSlice key, GeoSearchOptions opts,
```
TalZaccai, line 55: I would probably just reparse GeoSearchOptions in the backend using some common helper method instead of having to manage the locks here, seems error-prone to me.

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -197,9 +202,59 @@ private static int AdminCommands(RespCommand command)
...
+            var idx = 0;
 
+            // GEOSEARCHSTORE dest key....
+            // While all other commands here start with GEOsomething key...
+            if (command == RespCommand.GEOSEARCHSTORE)
+            {
+                var destinationKey = respSession.parseState.GetArgSliceByRef(0);
+                SaveKeyEntryToLock(destinationKey, true, LockType.Exclusive);
+                SaveKeyArgSlice(destinationKey);
+                idx = 1;
```
TalZaccai, line 216: nit: can just do `respSession.parseState.GetArgSliceByRef(idx++)`

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -197,9 +202,59 @@ private static int AdminCommands(RespCommand command)
...
+            SaveKeyArgSlice(key);
+
+            switch (command)
+            {
+                case RespCommand.GEOSEARCH:
+                case RespCommand.GEORADIUS_RO:
+                case RespCommand.GEORADIUSBYMEMBER_RO:
+                    return 1;
+                case RespCommand.GEOSEARCHSTORE:
+                    return 2;
```
TalZaccai, line 231: Shouldn't this be 1?

TalZaccai (changes requested):
Added some more comments... Thanks for bearing through these, there's a lot of nuance here :)

libs/server/Transaction/TxnKeyManager.cs
prvyk: We skip the destination key too, right? Like XSTOREKeys() method in the same file.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+                {
+                    // From Member
+                    searchOpts.fromMember = input.parseState.GetArgSliceByRef(currTokenIdx++).SpanByte.ToByteArray();
+                    searchOpts.origin = GeoOriginType.FromMember;
+                }
+                else
+                {
+                    if (!input.parseState.TryGetDouble(currTokenIdx++, out searchOpts.lon) ||
+                        !input.parseState.TryGetDouble(currTokenIdx++, out searchOpts.lat) ||
+                        (searchOpts.lon < GeoHash.LongitudeMin) ||
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -145,40 +147,424 @@ private unsafe bool GeoCommands<TGarnetApi>(RespCommand command, ref TGarnetApi
...
+            {
+                searchOpts.withHash = true;
+            }
+            #endregion
+
+            GarnetStatus status;
+            if (destIdx != -1)
+            {
+                var destinationKey = parseState.GetArgSliceByRef(destIdx);
+                status = storageApi.GeoSearchStore(sourceKey, destinationKey, searchOpts, ref input, ref output);
```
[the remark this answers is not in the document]
prvyk: Great call. Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Right. It should be the other way around, of setting it to true (so GEORADIUS WITHHASH without STORE prints the result as integer). Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: And here I thought I could get away with being more lax/lazy and keep the previous code... ;-)

Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: I checked, Redis allows both and takes the last one given.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
prvyk: We still have to lock the keys here, no? How does reparsing help avoid that?

Reparsing has its own set of possible bugs, e.g. if backend and frontend get out of sync. Since we have to save it into variables anyway (GeoSearch is a bit too complex to be done inline I think?), it won't even save memory.

libs/server/Objects/SortedSetGeo/SortedSetGeoObjectImpl.cs
```diff
@@ -198,9 +181,13 @@ private void GeoDistance(ref ObjectInput input, ref SpanByteAndMemory output)
...
                 // Read units
-                var units = input.parseState.Count > 2
-                    ? input.parseState.GetArgSliceByRef(2).ReadOnlySpan
-                    : "M"u8;
+                var units = GeoDistanceUnitType.M;
+
+                if (input.parseState.Count > 2)
+                {
+                    var validUnit = input.parseState.TryGetGeoDistanceUnit(2, out units);
+                    //Debug.Assert(validUnit);
```
prvyk, line 189: @TalZaccai 
We should be parsing GEODIST so the unit error message comes out here too. But this is a bit out of scope I think? I'd rather save this for a future PR.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Done.

libs/server/Transaction/TxnKeyManager.cs
TalZaccai: Yes, you're right, my bad.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: 1. You need to slow down on fixing things, I can't keep up ;)
2. You will directly call `ReadObjectStoreOperationWithOutput`, locking will be handled by the backend. My worry is that multiple commands call this same method, so it could cause confusion with it's original read-only purpose. 
3. I agree that reparsing is not ideal, but currently that's the way many commands are handled - parse for validation in the frontend, parse with assumption that it should succeed in the backend. If possible, we encode data from the validation phase in `ObjectInput` (in `arg1` and `arg2`) so that we don't have to reparse, problem here is that there's a lot of data, too much to encode.

prvyk: >You need to slow down on fixing things, I can't keep up ;)

It's very late here really. ;) I'll just post this reply.

>My worry is that multiple commands call this same method, so it could cause confusion with it's original read-only purpose.

Maybe the best solution is to inline into GeoSearchStore() applicable contents from GeoSearch()? That way there's a clean separation between the readonly and readwrite methods, and there really isn't much at all in GeoSearch(). I think that's a good idea anyway.

(Aside, that confusion predates the diff, the current code just does it with GeoCommands() function).

>I agree that reparsing is not ideal, but currently that's the way many commands are handled -

Saw it during the BITFIELD issues... I now see how to do it, it's just a bit ugly and the main effect on the user is being a bit slower. At least if there was a way to tell storage "there's COUNT 3 ANY, you can use that to read only 3 items" there would be some performance optimization here, but I don't think it knows to partially read a sorted set.

prvyk: Done. Come to think of it now, GeoSearch() was only called for GEOSEARCH style methods (it was separated from GeoCommands() ), so I dunno if there really was a conflict earlier. 

Now that it calls ReadObjectStoreOperationWithOutput again, there's no need to keep this apart from GeoCommands().

libs/server/Resp/Objects/SortedSetGeoCommands.cs
TalZaccai: As long as you're messing around in there might as well make it better :)

libs/server/Objects/SortedSetGeo/SortedSetGeoObjectImpl.cs
TalZaccai: Sure, we can skip it for now.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
TalZaccai: > Maybe the best solution is to inline into GeoSearchStore() applicable contents from GeoSearch()? That way there's a clean separation between the readonly and readwrite methods, and there really isn't much at all in GeoSearch(). I think that's a good idea anyway.

Not entirely sure what you mean here. LMK if you want to explore this further. At the very least, we should rename this method (and the one in `SortedSetGeoOps`) to something like `GeoSearchReadOnly` so that it's clear that it can only do read-only operations, and we can come up with a better solution later.

TalZaccai (changes requested):
Few extra comments

libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
```diff
@@ -0,0 +1,461 @@
...
+                        }
+
+                        // Read radius and units
+                        if (!input.parseState.TryGetDouble(currTokenIdx++, out searchOpts.radius))
+                        {
+                            return CmdStrings.RESP_ERR_NOT_VALID_RADIUS;
+                        }
+
+                        if (searchOpts.radius < 0)
+                        {
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
```diff
@@ -0,0 +1,461 @@
...
+                            break;
+                        }
+
+                        // Read width, height
+                        if (!input.parseState.TryGetDouble(currTokenIdx++, out searchOpts.boxWidth))
+                        {
+                            return CmdStrings.RESP_ERR_NOT_VALID_WIDTH;
+                        }
+
+                        if (!input.parseState.TryGetDouble(currTokenIdx++, out var height))
```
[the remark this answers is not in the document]
prvyk: It's a property (mapping to Radius), it doesn't allow use in out clauses.

libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
```diff
@@ -0,0 +1,461 @@
...
+        internal byte[] fromMember;
+        internal double lon, lat;
+
+        internal bool withCoord;
+        internal bool withHash;
+        internal bool withCountAny;
+        internal bool withDist;
+        internal GeoOrder sort;
+
+        public static unsafe ReadOnlySpan<byte> Parse(RespCommand command,
```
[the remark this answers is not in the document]
prvyk: Done.

libs/server/Storage/Session/ObjectStore/SortedSetGeoOps.cs
prvyk: >Not entirely sure what you mean here. LMK if you want to explore this further.

We have two possible flows here. 

The first is parse to GeoSearchOptions struct and pass the ref struct inwards. It's like SortedSetDifference()
The second one is parse, call GeoCommands() and parse again at Operate(), pass the ref struct to the sorted set operation.

Both of these flows internally have a readonly path and a readwrite path.

The difference is basically this:
https://github.com/microsoft/garnet/pull/1058/commits/325fb3b243b36fcc4c4574cab9159d7ed2556778

The first doesn't need to reparse, but does locking with transaction manager.
The second method parses again, but does locking with ReadObjectStoreOperationWithOutput in the read paths.

First we talked about the first method (like SortedSetDifference), and now we're talking about reparsing. The problem being that the first method's locking might be confusing.

Now what I did was to completely separate the readonly and write paths, so I think I can avoid confusion without reparsing. The readwrite path will not call the readonly path to do the GET. The readonly path is only for GEOSEARCH/GEORADIUS*_RO style commands (GEORADIUS* without STORE* also applies) and not for GEODIST etc.

If you still prefer the reparsing flow, I'll just revert the latest commit.

prvyk:
Both recent test runs have odd different random failure in net8.0 Debug Garnet.test. I don't think it's related to this PR at all.

Failure on memory usage:
https://github.com/microsoft/garnet/actions/runs/14003801195/job/39214826304?pr=1058#step:7:1681

Inexplicable custom command failure:
https://github.com/microsoft/garnet/actions/runs/14003929070/job/39215149669#step:7:1579

libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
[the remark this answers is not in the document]
TalZaccai: Ah, makes sense.

TalZaccai:
> Both recent test runs have odd different random failure in net8.0 Debug Garnet.test (and only there). I don't think it's related to this PR at all.
> 
> Failure on memory usage: https://github.com/microsoft/garnet/actions/runs/14003801195/job/39214826304?pr=1058#step:7:1681
> 
> Inexplicable custom command failure: https://github.com/microsoft/garnet/actions/runs/14003929070/job/39215149669#step:7:1579

Seems odd to me, I don't think we've seen any test failures in main... I'll double-check with the team.

TalZaccai:
> > Both recent test runs have odd different random failure in net8.0 Debug Garnet.test (and only there). I don't think it's related to this PR at all.
> > Failure on memory usage: https://github.com/microsoft/garnet/actions/runs/14003801195/job/39214826304?pr=1058#step:7:1681
> > Inexplicable custom command failure: https://github.com/microsoft/garnet/actions/runs/14003929070/job/39215149669#step:7:1579
> 
> Seems odd to me, I don't think we've seen any test failures in main... I'll double-check with the team.

@Vijay-Nirmal Could you take a look at the failure of `CanDoHashCollect`?

Vijay-Nirmal:
@TalZaccai I have raised a PR to fix it.

badrishc:
> @TalZaccai I have raised a PR to fix it.

More important is what changed that might cause this to occur more frequently in recent times? Hopefully there isn't some other slowdown that is exposing the timing limit.

Vijay-Nirmal:
@badrishc Not sure how often this is happening in the pipeline but to reproduce it locally, I had to run the test ~150 times to get the failure (Using the "Run Until Failure" feature)

badrishc:
> @badrishc Not sure how often this is happening in the pipeline but to reproduce it locally, I had to run the test ~150 times to get the failure (Using the "Run Until Failure" feature)

Can you try with the commit where the test initially merged to see if it used to fail then as well?

Vijay-Nirmal:
> Inexplicable custom command failure:
https://github.com/microsoft/garnet/actions/runs/14003929070/job/39215149669#step:7:1579

Raise a separate #1127 PR to fix this failure as well

Vijay-Nirmal:
> Can you try with the commit where the test initially merged to see if it used to fail then as well?

@badrishc Checked this, still failing in the old commits

prvyk:
@badrishc
I don't think this PR is done yet?

badrishc:
> @badrishc I don't think this PR is done yet? [Edit: nevermind, I'll resubmit. It would also be nice to let Copilot give his opnion].

Oh looks like GH closed this accidentally when 1126 closed, based on the text Vijay had in the PR description of https://github.com/microsoft/garnet/pull/1126 ... :-D

PR #1063, opened 2025-02-28
Less annoying lightClient request builder

prvyk:
Lightclient does very naive parsing, which is reasonable except it does things like 'interpret extra spaces as an empty argument'. For example, accidentally trying to send 

```
GET  a
``` 
 sends '*3$3GET$0$1a' which is unexpected.

This PR teaches lightClientRequest to do slightly less naive parsing, so extra spaces don't usually turn into empty arguments. While at it, teach it a bit of quotes and escaping. Quotes are useful to be able to do 'get "A B"' (sends '*2$3GET$3A B').

There was a test actually depending on the empty argument behaviour so I left an option to enable it.

### DIFF
--- a/libs/common/LightClientRequest.cs
+++ b/libs/common/LightClientRequest.cs
@@ -202,7 +202,20 @@ namespace Garnet.common
                 {
                     sb = sb.Append(to - tokenStart - hadEscapedChar);
                     sb = sb.Append("\r\n");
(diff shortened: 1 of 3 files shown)


prvyk:
>It is experimental and in-fact we started removing it from the codebase as much as possible... Can you use GarnetClientSession instead?

Ah, I didn't know that (just wanted to remove a footgun). Sure I can use GarnetClientSession instead.

PR #1064, opened 2025-03-01
Fix Single Threaded Client Usage Issues in Parallel ACL SETUSER Tests

kevinmichaelbowersox:
# Summary
The `ParallelTests#ParallelAclSetUserTest` and `ParallelTests#ParallelAclSetUserAvoidsMapContentionTest` have issues surrounding the client selected for the test and how the client is used within the test. This misuse can lead to sporadic thread safety and memory access violation issues when executing tests.

## ParallelAclSetUserTest

The `ParallelAclSetUserTest` uses the single threaded `GarnetClientSession`, however it launches concurrent tasks using `Task.WhenAll`. This violates the contract of the single threaded client causing sporadic issues with the test. To fix this problem unnecessary usage of `Task.WhenAll` was removed and each individual async command executed by the client was awaited.

## ParallelAclSetUserAvoidsMapContentionTest

The `ParallelAclSetUserAvoidsMapContentionTest` uses the single threaded `GarnetClientSession` to start many tasks concurrently using `Task.WhenAll`. The goal is to create a close race between the first successful task and the runner up, causing contention on the ACL map. However, it launches the concurrent tasks using `Task.WhenAll`, violating the contract of the ingle threaded client. To fix this problem, the test was updated to use the `GarnetClient`, which has support for multiple threads.

### DIFF
--- a/test/Garnet.test/Resp/ACL/ParallelTests.cs
+++ b/test/Garnet.test/Resp/ACL/ParallelTests.cs
@@ -119,31 +119,24 @@ namespace Garnet.test.Resp.ACL
         /// <summary>
         /// Tests that ACL SETUSER works in parallel without fatal contention on user in ACL map.
         ///
-        /// Test uses a multi-threaded client that applies the same ACL change to the same user. Creates race to
-        /// become the first command invocation to add the user to the ACL. Throws after initial insert into ACL if
-        /// threading issues exist.
+        /// Test launches multiple clients that apply the same ACL change to the same user. Creates race to become the
+        /// the first client to add the user to the ACL. Throws after initial insert into ACL if threading issues exist.
         ///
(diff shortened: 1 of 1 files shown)


badrishc:
If we want to create server side races, shouldn't we be using multiple GarnetClientSession (one per parallel thread/task) instead of a shared GarnetClient? Because a single GarnetClient will linearize all the requests into a fifo stream.

kevinmichaelbowersox:
> If we want to create server side races, shouldn't we be using multiple GarnetClientSession (one GCS per parallel thread/task) instead of a shared GarnetClient? Because a single GarnetClient will linearize all the requests into a fifo stream on a single TCP connection at the server.

Agree. Thanks for pointing this out. I switched back to using multiple clients and was able to repro (testing with our fixes removed).

vazois (approved):

PR #1076, opened 2025-03-07
[TestSuit] Use consistent assertion for lightClient responses.

prvyk:
LightClient is used for a lot of the testsuit. It uses a large fixed array, which is trimmed to compare a fixed expected responsed. There are at least 4 separate ways this is done in the testsuit**. Also the comparison is sometimes wrong in the argument order (doesn't use expected response as expected) or very rarely just wrong (CanUseZDiffSTORE test). 

This PR introduces a standard TestUtils methods for this assertion, changes almost all relevant locations to use it, and fixes the few cases when it's done incorrectly.

**

1. There's this:
Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length).

(Runs GetString + allocation over the entire fixed array).

2. There's also:
Encoding.ASCII.GetString(response, 0, expectedResponse.Length)

(Reasonable, but allocates. Maybe could be improved)

3. And also:
  ClassicAssert.AreEqual(response.AsSpan().Slice(0, expectedResponse.Length).ToArray(), expectedResponse);
  
(Order of arguments is wrong. Otherwise, it's interesting to note NUnit can handle the different types, not clear if that is better or worse).

4. Lastly there are calls to lightclient's Execute() method which just does the second under the hood. These aren't modified for now.

This diff uses the Encoding.ASCII.GetString(response, 0, expectedResponse.Length) method. It would be easy the try the latter method (in the correct argument order) if it's useful.

(Re #1063 comments: I don't view this as 'improving lightClient' but as improving the testsuit. If we are stuck with some use for now, might as well test the assertion in a consistent and safe way.)

### DIFF
--- a/test/Garnet.test/RespAdminCommandsTests.cs
+++ b/test/Garnet.test/RespAdminCommandsTests.cs
@@ -42,8 +42,7 @@ namespace Garnet.test
             using var lightClientRequest = TestUtils.CreateRequest();
             var expectedResponse = "+PONG\r\n";
             var response = lightClientRequest.SendCommand("PING");
-            var actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
         }
 
         [Test]
@@ -52,8 +51,7 @@ namespace Garnet.test
             using var lightClientRequest = TestUtils.CreateRequest();
             var expectedResponse = "$5\r\nHELLO\r\n";
             var response = lightClientRequest.SendCommand("PING HELLO");
-            var actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
         }
 
         [Test]
@@ -62,8 +60,7 @@ namespace Garnet.test
             using var lightClientRequest = TestUtils.CreateRequest();
             var expectedResponse = $"-{string.Format(CmdStrings.GenericErrWrongNumArgs, $"{nameof(RespCommand.PING)}")}\r\n";
             var response = lightClientRequest.SendCommand("PING HELLO WORLD", 1);
-            var actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
         }
 
         [Test]
@@ -72,10 +69,8 @@ namespace Garnet.test
             using var lightClientRequest = TestUtils.CreateRequest();
             var expectedResponse = $"-{string.Format(CmdStrings.GenericErrWrongNumArgs, $"{nameof(RespCommand.ECHO)}")}\r\n";
             var response = lightClientRequest.SendCommand("ECHO", 1);
-            var actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
         }
-
 
         [Test]
         public void EchoWithMessagesReturnErrorTest()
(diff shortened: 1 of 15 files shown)


badrishc:
> (Re https://github.com/microsoft/garnet/pull/1063 comments: I don't view this as 'improving lightClient' but as improving the testsuit. If we are stuck with some use for now, might as well test the assertion in a consistent and safe way.)

Makes sense. It is worth examining the usage sites of `LightClient` to make sure it was specifically necessary to use it, versus GCS (GarnetClientSession) or SE.Redis. For example. some tests care about specific network splits which GCS and SE.Redis cannot directly control.

test/Garnet.test/RespSetTest.cs
```diff
@@ -929,7 +908,7 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
 
             // Get three random members
             response = lightClientRequest.SendCommand("SRANDMEMBER myset 3", 3);
-            strResponse = Encoding.ASCII.GetString(response);
+            var strResponse = Encoding.ASCII.GetString(response);
             ClassicAssert.AreEqual('*', strResponse[0]);
```
TalZaccai, line 912: Why not use `AssertEqualUpToExpectedLength("*", response)`?

TalZaccai (changes requested):
Very nice cleanup! Thanks! Few small comments...

test/Garnet.test/RespAdminCommandsTests.cs
```diff
@@ -107,31 +100,27 @@ public void EchoTwoCommandsTest()
...
         [Test]
         public void TimeCommandTest()
         {
             // this is an example, we just compare the length of the response with the expected one.
             using var lightClientRequest = TestUtils.CreateRequest();
             var expectedResponse = "*2\r\n$10\r\n1626282789\r\n$6\r\n621362\r\n";
             var response = lightClientRequest.SendCommand("TIME", 3);
-            var actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
+            var actualValue = Encoding.ASCII.GetString(response, 0, expectedResponse.Length);
             ClassicAssert.AreEqual(expectedResponse.Length, actualValue.Length);
```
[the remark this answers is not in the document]
prvyk: Indeed. I went over all spots again to fix places where I missed it.

test/Garnet.test/RespHashTests.cs
```diff
@@ -1405,45 +1389,38 @@ public void CanDeleteOnepairLC()
...
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
 
             response = lightClientRequest.SendCommand("MEMORY USAGE myhash");
             expectedResponse = ":408\r\n";
-            actualValue = Encoding.ASCII.GetString(response).Substring(0, expectedResponse.Length);
-            ClassicAssert.AreEqual(expectedResponse, actualValue);
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
 
             var result = lightClientRequest.SendCommand("HDEL myhash field1");
```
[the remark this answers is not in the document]
prvyk: Ouch. Renamed responses to 'response' when possible.

test/Garnet.test/RespSetTest.cs
prvyk: Done.

test/Garnet.test/RespAdminCommandsTests.cs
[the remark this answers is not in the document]
prvyk: Ah, here is a special spot where it just checks the length (see comment). But there were indeed a few other places where I missed replacing the comparison.

test/Garnet.test/RespSetTest.cs
TalZaccai: Missed a commit? I don't see a change here :)

TalZaccai: Oh I think you fixed it in `CanDoSPOPWithCountCommandLC` but not here :)

TalZaccai (commented):
LGTM, one comment remaining

test/Garnet.test/RespSetTest.cs
prvyk: Right. Actually done.

TalZaccai (approved):

PaulusParssinen:
Noticed this PR now, big fan of it!

PR #1077, opened 2025-03-08
Single state machine across main and object stores

badrishc:
Changes:
- [x] Share Epoch and StateMachineDriver (SMD) across main and object stores
- [x] Garnet checkpointing uses this shared driver to checkpoint both stores with a single state machine
- [x] Transactions use SMD to register transactions, so that a single transaction version is maintained across the two stores.
- [x] Transactions no longer barrier during the `(v) -> (v+1)` switchover. We ensure correctness by verifying the transaction version after all locks are acquired.
- [x] Move `NumActiveLockingSessions` (renamed to `NumActiveTransactions`) to SMD and ensure correct `(v) -> (v+1)` switchover with active transactions.
- [x] Removed `targetVersion` concept so that checkpoint versions always increment in steps of 1.
- [x] Refactored `CheckpointVersionShift` to a pair of `CheckpointVersionShiftStart` and `CheckpointVersionShiftEnd` to help demarcate start and end of fuzzy region in the AOF ((region that contains mix of `v` and `v+1` log entries).
- [x] Fix AOF replay to correctly process single-key operations and transactions in fuzzy region. Specifically,
   * `checkpoint_start` marker: Track `ReplicationCheckpointStartOffset` in AOF
   *  Process all `(v)` ops and txns in the fuzzy region of AOF, store away the `(v+1)` entries in a fuzzy buffer
   * `checkpoint_end` marker: Take a checkpoint, potentially truncating until `ReplicationCheckpointStartOffset`, replay all the stored `(v+1)` ops and txns from the fuzzy buffer.

Notes:
- Recovery logic is left unmodified, it assumes checkpoints are separate across main and object stores. This allows the PR to remain backwards compatible.


libs/storage/Tsavorite/cs/src/core/Index/Checkpointing/StateMachineDriver.cs
```diff
@@ -23,28 +23,119 @@ public class StateMachineDriver
...
+        public long AcquireTransactionVersion()
+        {
+            var isProtected = epoch.ThisInstanceProtected();
+            if (!isProtected)
+                epoch.Resume();
+            try
+            {
+                // We create a barrier preventing new transactions from starting in the PREPARE_GROW phase
+                // since the lock table needs to be drained and transferred to the larger hash index.
+                while (systemState.Phase == Phase.PREPARE_GROW)
```
TedHartMS, line 79: I believe this is also what ensures there are only 2 "active" txnVersions?

TedHartMS, line 79: Should this include Phase.PREPARE?

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/TsavoriteThread.cs
```diff
@@ -17,27 +17,26 @@ internal void InternalRefresh<TInput, TOutput, TContext, TSessionFunctionsWrappe
...
+                // Adjust session's effective state if there is an ongoing transaction
+                if (sessionFunctions.Ctx.txnVersion > 0 &&
+                    sessionFunctions.Ctx.SessionState.Phase == Phase.IN_PROGRESS &&
+                    sessionFunctions.Ctx.txnVersion == sessionFunctions.Ctx.SessionState.Version - 1)
+                {
+                    sessionFunctions.Ctx.SessionState = SystemState.Make(Phase.PREPARE, sessionFunctions.Ctx.txnVersion);
+                }
+
+                // Session threads need to wait in PREPARE_GROW phase unless they are in a transaction
+                if (sessionFunctions.Ctx.phase == Phase.PREPARE_GROW && !sessionFunctions.Ctx.isAcquiredLockable)
```
TedHartMS, line 39: Should this still include Phase.PREPARE? Maybe like IsInV1 we need IsInPREPARE

libs/storage/Tsavorite/cs/src/core/Index/Checkpointing/StateMachineDriver.cs
badrishc: PREPARE_GROW is specifically for index growth. That is the only case we need to barrier transactions in this PR.

badrishc: PREPARE_GROW is specifically for index growth. That is the only case we need to barrier transactions in this PR.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/TsavoriteThread.cs
badrishc: We do not need to barrier in case of PREPARE.

libs/storage/Tsavorite/cs/src/core/Index/Checkpointing/StateMachineDriver.cs
badrishc: There are 2 active txnVersions for normal checkpoints. Transactions in (v) drain out at the same time as non-conflicting transactions in (v+1). For index growth, version numbers do not increment.

libs/storage/Tsavorite/cs/test/StateMachineDriverTests.cs
```diff
@@ -58,111 +58,225 @@ public async ValueTask DoCheckpointVersionSwitchEquivalenceCheck(
...
-                    output = completedOutputs.Current.Output;
-                    result = completedOutputs.Next();
-                    ClassicAssert.IsFalse(result);
+                    var status = await task;
+                    ClassicAssert.IsTrue(status);
                 }
-                ClassicAssert.IsTrue(status.Found, $"status = {status}");
+                else
+                {
+                    var status = task.GetAwaiter().GetResult();
```
TedHartMS, line 214: Is there a need to support this sync-over-async pattern?

TedHartMS (approved):
approved with one comment

TedHartMS (approved):

vazois (approved):

Issue #1052, opened 2025-02-26
Blocking pops block the connection when the origin already has enough items


prvyk:
### Describe the bug

Example:

ZADD a 1 one
ZADD a 2 two
ZADD b 3 three
BZMPOP 0 2 a b MIN COUNT 2 
-> redis-cli never returns.

The order in InitializeObserver() is likely wrong (what if a key is inserted after the loop checking it but before its observer being set up?), but that's not the cause here. The lock is actually freed?

This discussion may be related:
https://github.com/microsoft/garnet/discussions/1046

### Steps to reproduce the bug

ZADD a 1 one
ZADD a 2 two
ZADD b 3 three
BZMPOP 0 2 a b MIN COUNT 2

### Expected behavior

1) "a"
2) 1) 1) "one"
      2) "1"
   2) 1) "two"
      2) "2"

### Release version

v1.0.57

prvyk:
Ok, I get it now. The 'block' isn't from a Garnet lock. It's because Garnet sends out a mismatched array and redis-cli waits on read. I kept thinking this is blocking API but it's just not from there. That's easy to fix **.  

The race condition is a bit more complex (and maybe related to the discussion?). I think the simplest (but slowish) solution would be to create the observer unconditionally (better make KeysToObservers a ConcurrentDictionary to be safest) and raise CollectionUpdatedEvent at the end. CollectionUpdatedEvent will call TryAssignItem which will use ObserverStatusLock to synchronize and line 295 will check for the item. At most another CollectionUpdatedEvent comes up earlier or later and locks serialize it. The same idea of creating the observer first and locking so a concurrent CollectionUpdatedEvent is processed right can be done without the extra event, that's slightly more complex.

**
```
diff --git a/libs/server/Resp/Objects/SortedSetCommands.cs b/libs/server/Resp/Objects/SortedSetCommands.cs
index d0c3626..63d2b2f 100644
--- a/libs/server/Resp/Objects/SortedSetCommands.cs
+++ b/libs/server/Resp/Objects/SortedSetCommands.cs
@@ -1685,7 +1685,7 @@ namespace Garnet.server
             while (!RespWriteUtils.TryWriteArrayLength(result.Items.Length, ref dcurr, dend))
                 SendAndReset();

-            for (var i = 0; i < result.Items.Length; i += 2)
+            for (var i = 0; i < result.Items.Length; ++i)
             {
[...]

PR #1078, opened 2025-03-08
Fix BZMPOP return and add more argument checking.

prvyk:
This PR fixes one issue of #1052 , where the mistaken array length return causes clients to hang waiting for items that will never arrive. Also adds more argument checking and a test.

Locking issues, from #1052 and maybe #1046 are not included here, these require writing good tests first.

### DIFF
--- a/libs/server/Resp/Objects/SortedSetCommands.cs
+++ b/libs/server/Resp/Objects/SortedSetCommands.cs
@@ -1605,9 +1605,14 @@ namespace Garnet.server
             var currTokenId = 0;
 
             // Read timeout
-            if (!parseState.TryGetDouble(currTokenId++, out var timeout) || (timeout < 0))
+            if (!parseState.TryGetDouble(currTokenId++, out var timeout))
             {
                 return AbortWithErrorMessage(CmdStrings.RESP_ERR_TIMEOUT_NOT_VALID_FLOAT);
+            }
+
+            if (timeout < 0)
+            {
+                return AbortWithErrorMessage(CmdStrings.RESP_ERR_TIMEOUT_IS_NEGATIVE);
             }
 
             // Read count of keys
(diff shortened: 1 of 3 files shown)


libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -1552,7 +1552,7 @@ private unsafe bool SortedSetBlockingPop(RespCommand command)
                 return AbortWithWrongNumberOfArguments(command.ToString());
             }
 
-            if (!parseState.TryGetDouble(parseState.Count - 1, out var timeout))
+            if (!parseState.TryGetDouble(parseState.Count - 1, out var timeout) || (timeout < 0))
```
TalZaccai, line 1555: The error message for a negative timeout should be: `ERR timeout is negative`

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -1605,13 +1605,13 @@ private unsafe bool SortedSetBlockingMPop()
...
             // Read timeout
-            if (!parseState.TryGetDouble(currTokenId++, out var timeout))
+            if (!parseState.TryGetDouble(currTokenId++, out var timeout) || (timeout < 0))
             {
                 return AbortWithErrorMessage(CmdStrings.RESP_ERR_TIMEOUT_NOT_VALID_FLOAT);
             }
 
             // Read count of keys
-            if (!parseState.TryGetInt(currTokenId++, out var numKeys))
+            if (!parseState.TryGetInt(currTokenId++, out var numKeys) || (numKeys <= 0))
```
TalZaccai, line 1614: `numKeys` can't be 0 either

TalZaccai (changes requested):
Couple of minor comments.

libs/server/Resp/Objects/SortedSetCommands.cs
prvyk: It can't, so the error check is <= 0, right?

TalZaccai: Doh! sorry :)

TalZaccai (approved):

PR #1083, opened 2025-03-10
Support for Multiple Endpoints

vazois:
This PR adds support for multiple endpoints through --bind option.


prvyk:
You can use several different loopback addresses together for the test and skip dealing with Dns/network interfaces. IPv4 has the entire 127.x.x.x range for this.

vazois:
> You can use several different loopback addresses together for the test and skip dealing with Dns/network interfaces. IPv4 has the entire 127.x.x.x range for this.

Thanks for the suggestion. I want to test also the non-loopback addresses. So, added test for loopback ipv4 and ipv6

prvyk:
This all works, just a few notes (IMHO it would be reasonable to save them for a future PR):

* Error message when failing to bind to an address is IMHO a bit worse now. Better to not print out the stacktrace by not passing the exception to LogError.
* RedisConf import currently uses only first bind address and does not yet support multiple endpoints.
* Redis allows creating a Unix socket concurrently. ('--bind 127.0.0.1 --unixsocket ./socket'. In Redis both are defined, here only the socket).

vazois:
> This all works, just a few notes (IMHO it would be reasonable to save them for a future PR):
> 
> * Error message when failing to bind to an address is IMHO a bit worse now. Better to not print out the stacktrace by not passing the exception to LogError.
> * RedisConf import currently uses only first bind address and does not yet support multiple endpoints.
> * Redis allows creating a Unix socket concurrently. ('--bind 127.0.0.1 --unixsocket ./socket'. In Redis both are defined, here only the socket).

I made a change to log the message instead of the stack trace for the first one.
Not 100% positive on this but the other two suggestions seem to require a little bit more work that might be better done at another PR. Especially the last one. I will check for the second one.

badrishc:
Unix socket and TCP socket at the same time was, I thought, the use case that first drove this feature. Would be useful to see what makes that harder than two TCP sockets.

vazois:
> Unix socket and TCP socket at the same time was, I thought, the use case that first drove this feature. Would be useful to see what makes that harder than two TCP sockets.

The original purpose of this PR was to fully support hostnames that could map to multiple interfaces when provided through the bind option.
I am wondering if supporting unix sockets is as a simple as creating another instance of GarnetServerTcp.
If so, why we did not do this in the original PR? @PaulusParssinen

badrishc:
> I am wondering if supporting unix sockets is as a simple as creating another instance of GarnetServerTcp. If so, why we did not do this in the original PR? @PaulusParssinen

I _think_ it should be as simple as creating another instance of `GarnetServerTcp` with that endpoint (@PaulusParssinen?) .. We did not support multiple `GarnetServerTcp` instances until this PR, so it was not done earlier, IIRC.

PaulusParssinen:
> > I am wondering if supporting unix sockets is as a simple as creating another instance of GarnetServerTcp. If so, why we did not do this in the original PR? @PaulusParssinen
> 
> I _think_ it should be as simple as creating another instance of `GarnetServerTcp` with that endpoint (@PaulusParssinen?) .. We did not support multiple `GarnetServerTcp` instances until this PR, so it was not done earlier, IIRC.

The reason I didn't try to do this all in one PR (the one that added support for unix socket) is that I don't have confident mental model for what sort of changes that could possibly entail and hoped that just like this PR does, someone who has much clearer picture of the cluster mode (and the current networking stack) would have stab at it instead (Thanks!).

Glad to see it was as simple as making many `GarnetServerTcp` instances 🎉

libs/host/Configuration/Options.cs
```diff
@@ -637,17 +633,11 @@ public GarnetServerOptions GetServerOptions(ILogger logger = null)
             var checkpointDir = CheckpointDir;
             if (!useAzureStorage) checkpointDir = new DirectoryInfo(string.IsNullOrEmpty(checkpointDir) ? (string.IsNullOrEmpty(logDir) ? "." : logDir) : checkpointDir).FullName;
 
-            EndPoint endpoint;
+            if (!Format.TryParseAddressList(Address, Port, out var endpoints, out _) || endpoints.Length == 0)
```
TalZaccai, line 657: `Address` should probably be renamed to `Addresses` and the `HelpText` should also be updated to reflect that multiple addresses could be supplied.

libs/common/Format.cs
```diff
@@ -31,68 +33,105 @@ internal static bool IsNullOrWhiteSpace([NotNullWhen(false)] this string s) =>
...
+            endpoints = null;
+            errorHostnameOrAddress = null;
+            // Check if input null or empty
+            if (string.IsNullOrEmpty(addressList) || string.IsNullOrWhiteSpace(addressList))
+            {
+                endpoints = [new IPEndPoint(IPAddress.Any, port)];
+                return true;
+            }
+
+            var addresses = addressList.Split(' ');
```
TalZaccai, line 57: We can support more than one separator, we can have a constant `char[]` defined here that can be referenced by the parser docs.

TalZaccai, line 57: Might be helpful to strip extra whitespaces / ignore empty addresses after `Split`

prvyk: What happens if there are multiple spaces? Could StringSplitOptions.RemoveEmptyEntries be added here?

test/Garnet.test/GarnetServerConfigTests.cs
```diff
@@ -631,5 +633,30 @@ public void UnixSocketPermission_InvalidPermissionFails()
             var parseSuccessful = ServerSettingsManager.TryParseCommandLineArguments(args, out _, out _, out _, silentMode: true);
             ClassicAssert.IsFalse(parseSuccessful);
         }
+
+        [Test]
```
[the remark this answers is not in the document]
prvyk: Do we have a test for the situation where Garnet is start with both Unix socket and TCP address endpoints together? I see a multiple TCP endpoint test and Unixsocket tests, but not combined?

libs/common/Format.cs
vazois: > We can support more than one separator, we can have a constant `char[]` defined here that can be referenced by the parser docs.

It is simpler to support only space separated addresses for now. If in the future there is strong reason for it we can revisit it.

vazois: > What happens if there are multiple consecutive spaces? Could StringSplitOptions.RemoveEmptyEntries be added here?
Sure, I added that option.

libs/common/Format.cs
```diff
@@ -85,6 +85,9 @@ public static async Task<EndPoint[]> TryCreateEndpoint(string singleAddressOrHos
             if (string.IsNullOrEmpty(singleAddressOrHostname) || string.IsNullOrWhiteSpace(singleAddressOrHostname))
                 return [new IPEndPoint(IPAddress.Any, port)];
 
+            if (singleAddressOrHostname[0] == '-')
```
prvyk, line 88: In redis, the dash means that the address is optional, so a bind failure won't stop start up. I don't see that here?

vazois: Yes, that is a conscious choice. I really don't see a strong reason for this feature.
You might argue that people should be aware of the IPs the server runs on, so they can avoid a security incident.
However, if you feel this a needed feature, please feel free to provide some examples and maybe a PR.

badrishc:
After update the branch to the latest in main, you can mirror your branch to garnet GH so that we can manually trigger the CI BDN benchmark (via workflow dispatch) and the comparison to main will appear as comments on the latest commit of the branch.

libs/common/Format.cs
```diff
@@ -31,116 +33,109 @@ internal static bool IsNullOrWhiteSpace([NotNullWhen(false)] this string s) =>
...
+            endpoints = null;
+            errorHostnameOrAddress = null;
+            // Check if input null or empty
+            if (string.IsNullOrEmpty(addressList) || string.IsNullOrWhiteSpace(addressList))
+            {
+                endpoints = [new IPEndPoint(IPAddress.Any, port)];
+                return true;
+            }
+
+            var addresses = addressList.Trim().Split(' ', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
```
badrishc, line 56: `addressList.Split([',', ' '], StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);`

libs/host/GarnetServer.cs
```diff
@@ -184,7 +184,7 @@ private void InitializeServer()
             var clusterFactory = opts.EnableCluster ? new ClusterFactory() : null;
 
             this.logger = this.loggerFactory?.CreateLogger("GarnetServer");
-            logger?.LogInformation("Garnet {version} {bits} bit; {clusterMode} mode; Endpoint: {endpoint}", version, IntPtr.Size == 8 ? "64" : "32", opts.EnableCluster ? "cluster" : "standalone", opts.EndPoint);
+            logger?.LogInformation("Garnet {version} {bits} bit; {clusterMode} mode; Endpoint: {endpoint}", version, IntPtr.Size == 8 ? "64" : "32", opts.EnableCluster ? "cluster" : "standalone", opts.EndPoints[0]);
```
badrishc, line 187: the list of endpoints should be logged here, not just the first one.

libs/cluster/Server/ClusterManager.cs
```diff
@@ -50,8 +50,18 @@ public unsafe ClusterManager(ClusterProvider clusterProvider, ILogger logger = n
             clusterConfigDevice = deviceFactory.Get(new FileDescriptor(directoryName: "", fileName: "nodes.conf"));
             pool = new(1, (int)clusterConfigDevice.SectorSize);
 
-            if (opts.EndPoint is not IPEndPoint endpoint)
-                throw new NotImplementedException("Cluster mode for unix domain sockets has not been implemented.");
+            IPEndPoint endpoint = null;
+            foreach (var endPoint in opts.EndPoints)
```
badrishc, line 54: Add `cluster-announce-ip` and `cluster-announce-port` options as optional. If not specified, using the first IPEndpoint is fine. If specified, we use that.

badrishc (changes requested):
LGTM overall, just a couple small requests.

libs/host/GarnetServer.cs
vazois: I am not sure we should be doing this, logging the endpoints due to privacy concerns.
https://github.com/microsoft/garnet/security/code-scanning/53

libs/cluster/Server/ClusterManager.cs
vazois: I added these options.

libs/common/Format.cs
vazois: I added this

badrishc: Why do you still have that Trim() I think it's not necessary

vazois: Sure, I removed the redundant Trim() call.

badrishc (approved):

TalZaccai (approved):

davidglassborow:
Is it worth making the default being to listen on v4 and v6 any ?  The documentation suggests that is what's happening, and with 'localhost' often resolving to the v6 endpoint on modern Windows 

image

vazois:
> Is it worth making the default being to listen on v4 and v6 any ? The documentation suggests that is what's happening, and with 'localhost' often resolving to the v6 endpoint on modern Windows
> 
> image

I am not sure we should put much effort on the default values, unless it is a parameter that has non-obvious consequences like setting memory size, compaction etc.
Most users can make informed decisions on how to setup the correct IP through the bind option or .conf file.
Personally, I am not a fun of Any option. We should probably default to Loopback because listening to any address is generally considered unsafe, especially if you are spinning up an instance for testing purposes.
Maybe the easier solution is to update the docs to be explicit that we are using IPV4.Any as default.
@badrishc, @talzaccai, @PaulusParssinen what is your take here?

davidglassborow:
Or maybe go for a safe default binding of V4 + V6 Loopback only ?

Redis has a 'protected mode' which will only allow an non-loopback bind if authentication has been turned on - https://redis.io/docs/latest/operate/oss_and_stack/management/security/

vazois:
> Or maybe go for a safe default binding of V4 + V6 Loopback only ?
> 
> Redis has a 'protected mode' which will only allow an non-loopback bind if authentication has been turned on - https://redis.io/docs/latest/operate/oss_and_stack/management/security/

Makes sense. You can open a PR for this if you think it is worthwhile.

davidglassborow:
> > Or maybe go for a safe default binding of V4 + V6 Loopback only ?
> > Redis has a 'protected mode' which will only allow an non-loopback bind if authentication has been turned on - https://redis.io/docs/latest/operate/oss_and_stack/management/security/
> 
> Makes sense. You can open a PR for this if you think it is worthwhile.

https://github.com/microsoft/garnet/pull/1156

PR #1093, opened 2025-03-12
Ignoring Lua RedisCallErrors test on .net9

TalZaccai:
also removing env var setting from ci.yml


badrishc (approved):

badrishc:
@kevin-montrose - fyi, we are ignoring the Lua tests that cause issues on .NET 9. Once the `longjmp` issue is fixed, these tests can be re-enabled. Thanks!

PR #1100, opened 2025-03-13
Reduce allocations and hash calculations in SortedSet operations

PaulusParssinen:
todo: PR description & check if visible in any benchmarks

PR #1105, opened 2025-03-16
Reduce allocations in HashObject operations on .NET 9

PaulusParssinen:
Follows from #1100. 

Trying to use alternative lookup in `HashObject` gets _much uglier_ and error-prone because we have to do expiration dictionary lookups in addition to the hash lookups. This means we have to be extremely careful how to layer utility methods and make sure that the overload resolution does not accidentally regress either .NET 9 or .NET 8 by binding `byte[] -> ReadOnlySpan<byte>` causing double key allocation where we try to only make one.

I'm not really too happy how this reads currently and I'm down to leave `HashObject` be until .NET 8 falls out of support in far future.


PaulusParssinen:
Yea this looks bad. Will look again how to de-dupe and unify expiration for both upcoming SortedSet expiration (#1000) and HashObjects some day in the far future.

badrishc:
>  expiration dictionary lookups

expiration dictionary is most likely always null/empty and never used except for the extremely rare case where users are goin per-element expiration. so that code path incurring extra cost is acceptable.

PR #1112, opened 2025-03-19
experiment: Make Tsavorite allocators unmanaged

PaulusParssinen:
wip. Testing if we can stop using POH for the various Tsavorite buffers and making the aligned allocations faster and simpler by using the new ``NativeMemory.Aligned*`` APIs. (i.e. delegate to same allocation logic and APIs as original FASTER C++ impl)

### Some notes about the progress
* Figure out why GenericAllocator logic started reading garbage for unblittable keys and values.
  * Was it relying on the buffer pool's over-allocation during alignment? Seems so at the moment.
  * I forgot that the original pool allocations were zero-init arrays, makes sense. This fixed the corruption of the first read record. The tail record keeps reading unitialized memory, if we don't over-allocate!
  * Maybe something up in the log alignment arithmetics..? Need figure out where exactly we are reading outside the rounded sector size that we ask for from the pool.
    * Seems like there was more logic relying on this over-allocation judging from the AVs.
  * Making Allocators disposal free the its owned buffers has started to break test that reused session from already disposed log (https://github.com/microsoft/garnet/pull/1112/commits/3cc1c0a78e19447d4ee91b845330b8ab80eee4ab) or where log was disposed before the device with its IO threads still operating on it (https://github.com/microsoft/garnet/pull/1112/commits/3ef93aad2453ea543261226ba7c687c2bbc0c956).
* The properties ``ValidOffset``, ``RequiredBytes`` and ``AvailableBytes`` seem to be record allocation concepts and probably don't belong to the aligned memory class. Consider pulling them out while we're here.


PaulusParssinen:
Trying to get tests pass so I can start to properly diagnose what parts of the allocator's record (de)serialization is messing around beyond the bounds what the code asks for. 

The changes made to the way allocator handles the memory buffers (that were GC tracked arrays before this PR) has "exposed" that the allocators pages were never _explicitly_ unrooted by the usual disposal procedures. By changing to unmanaged buffers, we must of course explicitly free them and this seems to have upsetted asynchronous/multithreaded log tests in couple cases that I could test and diagnose properly.

Little bit out of the depth in certain parts of the storage logic, so thought to ask first: Is there existing mechanism in the storage to either make the tests wait for all log device writes to finish before starting the clean-up logic (in-memory device had blocking logic in its disposal, do we want to build IAsyncDisposable one?), or existing logic/approach to telling the device to "abort everything the buffers you have from log are bad". cc @badrishc or anyone

badrishc:
We _may_ (dont recall 100% at the moment) be simply letting the GC handle the clean up of pages, for the reason that it was overhead to track and ensure the actual completion of IO-in-flight, and this was thought to be unnecessary as it was a test-only problem. So we just let the GC handle it. With explicit native memory, things get more complicated. We will need to drain the IO. There is a CompletePending API for the KV, but it is possible other parts of the code do not have similar mechanisms yet? TsavoriteLog uses an `_disposed` variable, but we need to make sure to wait for races where the callback thread might have just gone past the `_disposed ` check.

cc @TedHartMS

PaulusParssinen:
> @PaulusParssinen - can you show an example where this is occurring?

I never managed to track this down but I *_think_* it was must be happenign somewhere in the log record (de)serialization. I'll have re-do and take a look at this again after the new storage changes.

PaulusParssinen:
I'll do this again when `storage-v2` gets merged. This also contains unrelated changes that should be extracted to other PRs.

PR #1115, opened 2025-03-20
Add Overflow Bucket Statistics to Info Command

hamdaankhalid:
Adding information we use internally to debug large over flow table sizes in Garnet. These statistics help us get a better understanding of the hash index memory usage and any degradation that may happen in the future.


libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Tsavorite.cs
```diff
@@ -653,32 +653,87 @@ private unsafe string DumpDistributionInternal(int version)
...
                         if (((!x.ReadCache) && (x.Address >= beginAddress)) || (x.ReadCache && (x.AbsoluteAddress >= readCacheBase.HeadAddress)))
                         {
                             if (tags.Contains(x.Tag) && !x.Tentative)
                                 throw new TsavoriteException("Duplicate tag found in index");
                             tags.Add(x.Tag);
-                            ++cnt;
+                            ++total_valid_records_in_this_bucket_cnt;
                             ++total_record_count;
                         }
+                        else if (x.word != default && x.Address < beginAddress)
```
TedHartMS, line 691: The && part should be unneeded bc we verified beginAddress above; also it doesn't check the readCache case. If there was some reason an additional check was needed it should be inside the "if (!default)" case, to avoid mistakenly incrementing the zeroed_out_slots below

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Tsavorite.cs
```diff
@@ -653,32 +653,87 @@ private unsafe string DumpDistributionInternal(int version)
...
+                        break;
+
                     b = *(HashBucket*)overflowBucketsAllocator.GetPhysicalAddress(b.bucket_entries[Constants.kOverflowBucketIndex] & Constants.kAddressMask);
+                    is_bucket_in_ofb_table = true;
                 }
 
-                if (!histogram.ContainsKey(cnt)) histogram[cnt] = 0;
-                histogram[cnt]++;
+                if (!histogram.ContainsKey(total_valid_records_in_this_bucket_cnt)) histogram[total_valid_records_in_this_bucket_cnt] = 0;
+                histogram[total_valid_records_in_this_bucket_cnt]++;
```
TedHartMS, line 730: it doesn't matter so much here, but generally this should be "set to 1 else increment" to avoid an additional lookup

TedHartMS (approved):

PR #1120, opened 2025-03-21
Adding a producer/consumer test for BLPOP/BRPOP

TalZaccai:
This is to test the validity of the scenario suggested in this [discussion](https://github.com/microsoft/garnet/discussions/1046)


prvyk:
Does Garnet know the original consumer session is not alive? There was another issue I've been meaning to report where it doesn't clean up enough on dispose...  Maybe that should have been cleaned up too?

TalZaccai:
> Does Garnet know the original consumer session is not alive? There was another issue I've been meaning to report where it doesn't clean up enough on dispose... Maybe that should have been cleaned up too?

Garnet doesn't know at that point that the remaining session is a "zombie" session, as far as Garnet's concerned that session is still alive until it receives that item and dies (at which point it gets disposed).

badrishc (approved):

badrishc:
> There was another issue I've been meaning to report where it doesn't clean up enough on dispose

if you see an issue with clean up on dispose, we may need to look into that. Let us know what is your finding.

badrishc:
> Garnet doesn't know at that point that the remaining session is a "zombie" session

What is a zombie session? If the session disconnected, then that socket will be broken and we dispose RespServerSession in the `catch` of the socket read loop. That dispose should correctly be handled to notify any blocking writers?

PR #1121, opened 2025-03-21
Add more test coverage for Tsavorite primitives

PaulusParssinen:
Add little more `SpanByte` and its associated comparison logic unit test coverage. Also contains some Tsavorite test code housekeeping (mostly doing switching to simpler UTF-8 logic instead of UTF-16).

## Follow up work
- There could always be more `SpanByte` specific unit tests in `Tsavorite.test`


badrishc:
This may still be needed, as the MetadataSize check is different from Length check. A SpanByte may have an 8 byte optional header (a bit tells us whether there is such a header). The MetadataSize check ensures that both left and right have such a metadata header. MetadataSize will be either 8 or 0, depending on this bit.

PaulusParssinen:
Ah, thanks for explanation. I should have looked where that getter leads to before jumping to conclusion. 🤦‍♂️ 

I probably should pivot this PR to add PR to catch this detail.

libs/storage/Tsavorite/cs/test/SpanByteComparerTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        [Test]
+        public void Equals_ReturnsTrue_ForEmptyAndInvalid()
+        {
+            var invalid = new SpanByte();
+            invalid.Invalid = true;
+
+            var valid = new SpanByte();
+
+            ClassicAssert.IsTrue(SpanByteComparer.Instance.Equals(ref invalid, ref valid));
+        }
```
PaulusParssinen, line 81: Thought to just show this here and ask whether it is intentional or leave it as "undefined" behaviour and we should not try make any contract in the tests?

PaulusParssinen:
Seems like unrelated flaky test failure. There is probably more to do in Tsavorite tests but this will do as a start. Suggestions for any interesting or important test cases to still implement in this PR are welcome 👍

badrishc:
Ted's large PR at https://github.com/microsoft/garnet/pull/1186 might clash with this one. We were hoping to get that in without too many more clashes. Any chance this could be revised to work off of that branch?

PaulusParssinen:
> Ted's large PR at #1186 might clash with this one. We were hoping to get that in without too many more clashes. Any chance this could be revised to work off of that branch?

Sounds good, I can revisit this once that merges 👍

badrishc:
Thanks @PaulusParssinen , would also appreciate if you can review that PR https://github.com/microsoft/garnet/pull/1186 (fine to review the draft, as the pending changes right now are focused on larger-than-memory modifications) -- as it makes a lot of the improvements that have been suggested by you and other .NET experts over the last year.

PaulusParssinen:
I'll do this again when `storage-v2` gets merged.

PR #1140, opened 2025-03-27
Added try catch in object store object serialization

pradeepyadavmsft:
We occasionally encounter an exception indicating that the collection is being modified during serialization. This suggests that another record still holds a reference to the same object.

To address this, we have added a try-catch block around the serialization process. If serialization fails, we now set the object to null to prevent further issues.


libs/storage/Tsavorite/cs/src/core/Allocator/GenericAllocatorImpl.cs
```diff
@@ -433,13 +434,24 @@ private void WriteAsync<TContext>(long flushPage, ulong alignedDestinationAddres
...
-                                valueSerializer.Serialize(ref src[i].value);
-
-                                // Store the value address into the 'buffer' AddressInfo image as an offset into 'ms'.
-                                value_address->Address = pos;
-                                value_address->Size = (int)(ms.Position - pos);
-                                addr.Add((long)value_address);
-                                endPosition = pos + value_address->Size;
+                                try
+                                {
+                                    valueSerializer.Serialize(ref src[i].value);
```
badrishc, line 439: Try/catch around only this, if failed then reset pos and retry after setting src[i].value to default. No need to set invalid etc.

TedHartMS (approved):

PR #1143, opened 2025-03-29
[TestSuit] Fix CanDoSRANDMEMBERWithCountCommandLC test

prvyk:
The test didn't really test for duplicate members, because it didn't update arrLenEndIdx, so the next loop run just compared against the previous - passing regardless of actual behaviour (which happened to be per spec, but the test should still work).

Also fix remaining Encoding.ASCII.GetString usages.

### DIFF
--- a/test/Garnet.test/GarnetServerConfigTests.cs
+++ b/test/Garnet.test/GarnetServerConfigTests.cs
@@ -866,6 +866,7 @@ namespace Garnet.test
             var hostname = TestUtils.GetHostName();
             var addresses = Dns.GetHostAddresses(hostname);
             addresses = [.. addresses, IPAddress.IPv6Loopback, IPAddress.Loopback];
+            addresses = [.. addresses.Distinct()];
 
             var endpoints = addresses.Select(address => new IPEndPoint(address, TestUtils.TestPort)).ToArray();
             var server = TestUtils.CreateGarnetServer(TestUtils.MethodTestDir, endpoints: endpoints);
--- a/test/Garnet.test/RespSetTest.cs
+++ b/test/Garnet.test/RespSetTest.cs
@@ -1007,37 +1007,39 @@ namespace Garnet.test
         }
 
         [Test]
-        public async Task CanDoSRANDMEMBERWithCountCommandLC()
+        public unsafe void CanDoSRANDMEMBERWithCountCommandLC()
         {
             var myset = new HashSet<string> { "one", "two", "three", "four", "five" };
 
-            using var c = TestUtils.GetGarnetClientSession(raw: true);
-            c.Connect();
-
             // Check SRANDMEMBER with non-existing key
-            var response = await c.ExecuteAsync("SRANDMEMBER", "myset");
-            ClassicAssert.AreEqual("$-1\r\n", response);
+            using var lightClientRequest = TestUtils.CreateRequest();
+            var response = lightClientRequest.SendCommand("SRANDMEMBER myset");
+            var expectedResponse = "$-1\r\n";
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
 
             // Check SRANDMEMBER with non-existing key and count
-            response = await c.ExecuteAsync("SRANDMEMBER", "myset", "3");
-            ClassicAssert.AreEqual("*0\r\n", response);
+            response = lightClientRequest.SendCommand("SRANDMEMBER myset 3");
+            expectedResponse = "*0\r\n";
+            TestUtils.AssertEqualUpToExpectedLength(expectedResponse, response);
 
             CreateLongSet();
 
-            c.RawResult = false;
-            var item = await c.ExecuteAsync("SRANDMEMBER", "myset");
+            response = lightClientRequest.SendCommand("SRANDMEMBER myset", 1);
+            var strLen = Encoding.ASCII.GetString(response, 1, 1);
+            var item = Encoding.ASCII.GetString(response, 4, int.Parse(strLen));
             ClassicAssert.IsTrue(myset.Contains(item));
 
             // Get three random members
-            var results = await c.ExecuteForArrayAsync("SRANDMEMBER", "myset", "3");
-            ClassicAssert.AreEqual(3, results.Length);
-            ClassicAssert.IsTrue(results.All(myset.Contains));
+            response = lightClientRequest.SendCommand("SRANDMEMBER myset 3", 3);
+            TestUtils.AssertEqualUpToExpectedLength("*3\r\n", response);
 
             // Get 6 random members and verify that at least two elements are the same
-            results = await c.ExecuteForArrayAsync("SRANDMEMBER", "myset", "-6");
-            ClassicAssert.AreEqual(6, results.Length);
-            ClassicAssert.IsTrue(results.All(myset.Contains));
+            response = lightClientRequest.SendCommand("SRANDMEMBER myset -6", 6);
+            TestUtils.AssertEqualUpToExpectedLength("*6\r\n", response);
 
+            var results = TestSimpleReadRESP.ReadRESP(response);
+
+            ClassicAssert.IsTrue(results.All(a => myset.Contains((string)a)));
             ClassicAssert.IsTrue(results.Distinct().Count() != results.Length,
                                  "At least two members are repeated.");
         }
--- a/test/Garnet.test/RespSetTest.cs
+++ b/test/Garnet.test/RespSetTest.cs
@@ -1007,7 +1007,7 @@ namespace Garnet.test
         }
 
         [Test]
-        public unsafe void CanDoSRANDMEMBERWithCountCommandLC()
+        public void CanDoSRANDMEMBERWithCountCommandLC()
         {
             var myset = new HashSet<string> { "one", "two", "three", "four", "five" };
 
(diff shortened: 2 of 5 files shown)


TalZaccai (changes requested):

test/Garnet.test/RespSetTest.cs
```diff
@@ -902,15 +902,15 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
...
+            var strLen = Encoding.ASCII.GetString(response, 1, 1);
+            var item = Encoding.ASCII.GetString(response, 4, int.Parse(strLen));
             ClassicAssert.IsTrue(myset.Contains(item));
 
             // Get three random members
             response = lightClientRequest.SendCommand("SRANDMEMBER myset 3", 3);
             TestUtils.AssertEqualUpToExpectedLength("*", response);
 
-            var strResponse = Encoding.ASCII.GetString(response);
+            var strResponse = Encoding.ASCII.GetString(response, 0, 20);
```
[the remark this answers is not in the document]
prvyk: Done.

test/Garnet.test/RespSetTest.cs
```diff
@@ -920,25 +920,16 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
 
             // Get 6 random members and verify that at least two elements are the same
             response = lightClientRequest.SendCommand("SRANDMEMBER myset -6", 6);
-            var strReponse = Encoding.ASCII.GetString(response);
-            arrLenEndIdx = strReponse.IndexOf("\r\n", StringComparison.InvariantCultureIgnoreCase);
-            strArrLen = strReponse.AsSpan().Slice(1, arrLenEndIdx - 1);
+            strResponse = Encoding.ASCII.GetString(response, 0, 200);
```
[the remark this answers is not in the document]
prvyk: This continues #1076, converting the whole buffer added meaningful time to test run***, using an assert was both more readable and more efficient. Here we know how many tokens we want, we can just test for *{number}\r\n directly, without adding any complications... Done without constants here. 

In other locations where there was no choice ("CLIENT ID" etc.) add a constant and used that. Doubled to let some room for control characters.

*** Which has been getting slower recently despite that, with all the new tests.

test/Garnet.test/RespSetTest.cs
```diff
@@ -920,25 +920,16 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
...
-            {
-                var member = strReponse.Substring(arrLenEndIdx + 2, response.Length - arrLenEndIdx - 5);
-                if (members.Contains(member))
-                {
-                    repeatedMembers = true;
-                    break;
-                }
-                members.Add(member);
-            }
+            var results = strResponse.Split('\n').Where(w => w[0] != '$' && w[0] != '\0' && w[0] != '*').ToArray();
```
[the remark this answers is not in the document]
prvyk: Done. Used an existing method from RespReadUtils for that, it returns false or throws if the RESP response is malformed.

test/Garnet.test/GarnetServerConfigTests.cs
```diff
@@ -784,6 +784,7 @@ public async Task MultiTcpSocketTest()
             var hostname = TestUtils.GetHostName();
             var addresses = Dns.GetHostAddresses(hostname);
             addresses = [.. addresses, IPAddress.IPv6Loopback, IPAddress.Loopback];
+            addresses = [.. addresses.Distinct()];
```
prvyk, line 787: This is because the test failed here:

https://github.com/microsoft/garnet/actions/runs/14272307974/job/40008210658#step:7:189

test/Garnet.test/RespSetTest.cs
```diff
@@ -1025,43 +1026,31 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
             CreateLongSet();
 
             response = lightClientRequest.SendCommand("SRANDMEMBER myset", 1);
-            var strLen = Encoding.ASCII.GetString(response).Substring(1, 1);
-            var item = Encoding.ASCII.GetString(response).Substring(4, Int32.Parse(strLen));
+            var strLen = Encoding.ASCII.GetString(response, 1, 1);
```
TalZaccai, line 1029: I don't think that we necessarily need these changes, if the data changes it's not clear that these magic numbers need to be updated, the previous solution was more generic.

test/Garnet.test/RespSetTest.cs
```diff
@@ -1025,43 +1026,31 @@ public void CanDoSRANDMEMBERWithCountCommandLC()
...
-            ClassicAssert.IsTrue(int.TryParse(strArrLen, out arrLen));
-
-            var members = new HashSet<string>();
-            var repeatedMembers = false;
-            for (var i = 0; i < arrLen; i++)
+            TestUtils.AssertEqualUpToExpectedLength("*6\r\n", response);
+
+            string[] results;
+
+            fixed (byte* p = &response[0])
```
TalZaccai, line 1043: These changes might be a bit hard to read and also make the innerworkings of LightClient not well hidden.
We're also not validating that the members returned are in the set. 
Since we don't have to worry about perf here, I'd say we should just read the members one by one and validate their set membership.

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -4208,15 +4208,16 @@ public void CanUseZRandMember(int bytesSent)
             _ = lightClientRequest.SendCommand("ZADD dadi 1 uno 2 due 3 tre 4 quattro 5 cinque 6 six 7 sept 8 huit 9 nough 10 dis");
 
             // ZRANDMEMBER
-            var s = Encoding.ASCII.GetString(lightClientRequest.SendCommandChunks("ZRANDMEMBER dadi", bytesSent));
-            int startIndexField = s.IndexOf('\n') + 1;
-            int endIndexField = s.IndexOf('\n', startIndexField) - 1;
-            string memberValue = s.Substring(startIndexField, endIndexField - startIndexField);
+            var response = lightClientRequest.SendCommandChunks("ZRANDMEMBER dadi", bytesSent);
```
TalZaccai, line 4211: Why are these changes necessary?

test/Garnet.test/RespTests.cs
```diff
@@ -4958,17 +4958,19 @@ public async Task ClientUnblockBasicTest(string clientType, string mode, bool ex
 
             // Start blocking client
             using var blockingClient = TestUtils.CreateRequest();
-            var clientIdResponse = Encoding.ASCII.GetString(blockingClient.SendCommand("CLIENT ID"));
+            var clientIdResponse = Encoding.ASCII.GetString(blockingClient.SendCommand("CLIENT ID"), 0,
```
TalZaccai, line 4961: Same here, why are these changes necessary?

TalZaccai (changes requested):

test/Garnet.test/RespSortedSetTests.cs
prvyk: They're not critical, I just wanted to remove some uses of Encoding.ASCII.GetString(). Removed from PR.

test/Garnet.test/RespTests.cs
prvyk: Removed unnecessary changes. I did leave a check making sure the clientId is 64 bit per spec, and moving the Substring() subscript inwards down below (it's equivalent except for not needing to decode the entire array).

test/Garnet.test/RespSetTest.cs
prvyk: I would rather avoid yet another copy of RESP parsing code.

Rewrote the test to use GarnetClientSession and it can parse the array for us. Also, added testing that all members returned are in the set.

test/Garnet.test/RespSetTest.cs
prvyk: Rewrote the test to use GarnetClientSession.

test/Garnet.test/RespSetTest.cs
TalZaccai: Well the original test was testing the command with LightClient, so using GarnetClient (while valuable) is inconsistent with other tests in this context. I understand your concern about duplicated parsing logic, but it's how it's been done throughout the test code, so I think that for consistency's sake we should stick to it until we decide to do some wide overhaul of the LC-related code.

badrishc:
Where are we on the pending requests for this PR? It's been open quite a while, would be good to close on it, thanks!

prvyk:
I need to return the test to lightClient (reverting the last commit or so) and then copy some RespReadUtils code to Garnet.test and adjust to read out of a normal string (we don't care about being ultra fast there). It's not a problem, I was just prioritizing other PRs...

TalZaccai:
> I need to return the test to lightClient (reverting the last commit or so) and then copy some RespReadUtils code to Garnet.test and adjust to read out of a normal string (we don't care about being ultra fast there). It's not a problem, I was just prioritizing other PRs...

Please re-request a review when this is ready so I can take a look :) Thanks!

TalZaccai:
Since we want future contributors to follow a similar testing pattern, we think that the changes in this PR should be a part of a greater overhaul of the Garnet.test project. Since most of the code is written in the original form, it is the current reference and no one will know to follow this new pattern. So closing this for now and we'll get working on a larger change in the future.

PR #1147, opened 2025-03-31
Consolidate Standalone and Cluster Snapshot Metadata

vazois:
This PR consolidates the standalone and cluster snapshot metadata format by making the user cookie part of HybridLogRecoveryInfo. To achieve the following features were added

- [x] Cookie is integrated into the HybridLogRecoveryInfo
- [x] An interface was exposed to enable injecting a function that is used to build a custom cookie
- [x] The HistoryId parameters were moved inside the DeviceLogManager to be used across standalone and cluster.
- [x] Added unit test to validate conversion from cluster to standalonde and vice-versa.

NOTE: Added backwards compatibility code to ensure that old format is supported. This code should be deprecated as soon as we release the next version of Garnet.


libs/storage/Tsavorite/cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
```diff
@@ -232,13 +227,15 @@ public byte[] GetIndexCheckpointMetadata(Guid indexToken)
         }
 
         /// <inheritdoc />
-        public virtual unsafe void CommitLogCheckpoint(Guid logToken, byte[] commitMetadata)
+        public virtual unsafe void CommitLogCheckpoint(Guid logToken, HybridLogRecoveryInfo hlri)
```
badrishc, line 230: HybridLogRecoveryInfo is an internal detail of Tsavorite that ideally should not be visible to plugin API. Instead Tsavorite interacts with the checkpoint plugin with two APIs:

* `byte[] CheckpointCookieState()`
* `void RecoverCookieState(byte[] cookie)`

During checkpoint, it uses #1 to get the cookie to add to HLRI. During recovery, it uses #2 to tell the plugin what the recovered cookie is.

vazois: resolved this

badrishc (approved):

PR #1154, opened 2025-04-01
Fix GEOADD default and add GEODIST unit checking

prvyk:
GEOADD currently does add OR update, when default should be add AND update.
Also fix argument parsing (it should accept 'XX XX XX') and add GEODIST unit checking.
Add test and remove invalid test (reference doesn't behave that way).


libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
```diff
@@ -1,5 +1,16 @@
...
+﻿using System;
+
+namespace Garnet.server
 {
+    [Flags]
+    internal enum GeoAddOptions : byte
+    {
+        None = 0,
+        NX = 1,
+        XX = 2,
```
TalZaccai, line 10: For consistency, values should be `0, 1 << 0, 1 << 1, 1 << 2`

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -25,17 +25,69 @@ private unsafe bool GeoAdd<TGarnetApi>(ref TGarnetApi storageApi)
...
             // Get the key for SortedSet
-            var sbKey = parseState.GetArgSliceByRef(0).SpanByte;
-            var keyBytes = sbKey.ToByteArray();
+            var sbKey = parseState.GetArgSliceByRef(currTokenIdx++).SpanByte;
+
+            while (currTokenIdx < parseState.Count)
+            {
+                var byteOptions = parseState.GetArgSliceByRef(currTokenIdx).ReadOnlySpan;
+
+                if (byteOptions.EqualsUpperCaseSpanIgnoringCase("CH"u8))
```
TalZaccai, line 38: We can add this to the `CmdStrings` as well

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -25,17 +25,69 @@ private unsafe bool GeoAdd<TGarnetApi>(ref TGarnetApi storageApi)
...
+
+            var currTokenIdx = 0;
             // Get the key for SortedSet
-            var sbKey = parseState.GetArgSliceByRef(0).SpanByte;
-            var keyBytes = sbKey.ToByteArray();
+            var sbKey = parseState.GetArgSliceByRef(currTokenIdx++).SpanByte;
+
+            while (currTokenIdx < parseState.Count)
+            {
+                var byteOptions = parseState.GetArgSliceByRef(currTokenIdx).ReadOnlySpan;
```
TalZaccai, line 36: naming: Should be `addOptionSpan` or something similar

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -25,17 +25,69 @@ private unsafe bool GeoAdd<TGarnetApi>(ref TGarnetApi storageApi)
...
 
+            GeoAddOptions ex = 0;
+
+            var currTokenIdx = 0;
             // Get the key for SortedSet
-            var sbKey = parseState.GetArgSliceByRef(0).SpanByte;
-            var keyBytes = sbKey.ToByteArray();
+            var sbKey = parseState.GetArgSliceByRef(currTokenIdx++).SpanByte;
+
+            while (currTokenIdx < parseState.Count)
```
TalZaccai, line 34: You can't parse an unbound number of these options - there can be a maximal number of 2

libs/server/Resp/Objects/SortedSetGeoCommands.cs
```diff
@@ -25,17 +25,69 @@ private unsafe bool GeoAdd<TGarnetApi>(ref TGarnetApi storageApi)
                 return AbortWithWrongNumberOfArguments("GEOADD");
             }
 
+            GeoAddOptions ex = 0;
```
TalZaccai, line 28: naming: Should be `addOption`

libs/server/SessionParseStateExtensions.cs
```diff
@@ -743,6 +719,42 @@ internal static bool TryGetGeoDistanceUnit(this SessionParseState parseState, in
...
+        /// Parse geo longitude and latitude from parse state at specified index.
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="currIdx">The argument index</param>
+        /// <param name="lon"></param>
+        /// <param name="lat"></param>
+        /// <param name="idx">Outgoing argument index</param>
+        /// <param name="error"></param>
+        /// <returns>True if value parsed successfully</returns>
+        internal static bool TryGetGeoLonLat(this SessionParseState parseState, int currIdx, out double lon, out double lat,
```
TalZaccai, line 732: You can pass `currIdx` by ref instead of having both a parameter and an out parameter

libs/server/SessionParseStateExtensions.cs
```diff
@@ -743,6 +719,42 @@ internal static bool TryGetGeoDistanceUnit(this SessionParseState parseState, in
             return true;
         }
 
+        /// <summary>
+        /// Parse geo longitude and latitude from parse state at specified index.
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="currIdx">The argument index</param>
```
TalZaccai, line 726: nit: The *first* argument index
also, please add XML comments for the rest of the parameters.

TalZaccai (changes requested):
Nice changes, please see a few of my comments...

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Surprisingly, redis accepts this. GEOADD key CH XX XX XX XX XX CH (lon lat member etc.) will work.

libs/server/SessionParseStateExtensions.cs
prvyk: I tried that first, and got errors like "This returns local by reference but it is not a ref local" (since int currTokenIdx is not a ref in the caller), and a different ref scoping error which doesn't even appear in Visual Studio docs (when passed a reference to currTokenIdx).

I also tried passing a & reference directly (could make that work, but why go unsafe when I can avoid it?). Ultimately, using the out parameter was the simplest solution I could do...

TalZaccai: Ah ok I see, in that case maybe a clearer approach is to have an `out int tokensRead`, and then `currIdx += tokensRead` outside of the method.

prvyk: If we do that, maybe it's clearest to avoid the parameter entirely, and just do += 2 outside?

libs/server/SessionParseStateExtensions.cs
prvyk: Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Done.

libs/server/Resp/Objects/SortedSetGeoCommands.cs
prvyk: Done.

libs/server/Objects/SortedSetGeo/GeoSearchOptions.cs
prvyk: Done.

libs/server/SessionParseStateExtensions.cs
prvyk: Did += 2 outside instead.

TalZaccai (approved):

PR #1151, opened 2025-03-31
Enforce Cluster Announce Parameters on Recovery

vazois:
This PR fixes issue #640 because now we support cluster-announce-ip and cluster-announce-port options

PR #1156, opened 2025-04-02
Use Loopback (IPv4+IPv6) as the default end points, rather than IPv4 Any

davidglassborow:
- Render all endpoints during startup
- If localhost is specified as the address, explicitly add v4 and v6, modern Windows uses IPv6 for 'localhost', macos and linux use IPv4 currently
- Makes the defaults on the Options objects consistent with what is assigned during parsing of addresses

Discussed in https://github.com/microsoft/garnet/pull/1083#issuecomment-2766761513

Main drivers for change:
- Increased security for defaults, prevent uses unknowingly exposing their instances to the network accidentally
- Bind to both IPv4 and IPv6 stacks to localhost works consistently between OSs

### DIFF
--- a/libs/common/Format.cs
+++ b/libs/common/Format.cs
@@ -49,7 +49,7 @@ namespace Garnet.common
             // Check if input null or empty
             if (string.IsNullOrEmpty(addressList) || string.IsNullOrWhiteSpace(addressList))
             {
-                endpoints = [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)];
+                endpoints = Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];
                 return true;
             }
 
@@ -83,13 +83,13 @@ namespace Garnet.common
         public static async Task<EndPoint[]> TryCreateEndpoint(string singleAddressOrHostname, int port, bool tryConnect = false, ILogger logger = null)
         {
             if (string.IsNullOrEmpty(singleAddressOrHostname) || string.IsNullOrWhiteSpace(singleAddressOrHostname))
-                return [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)];
+                return Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];
 
             if (singleAddressOrHostname[0] == '-')
                 singleAddressOrHostname = singleAddressOrHostname.Substring(1);
 
             if (singleAddressOrHostname.Equals("localhost", StringComparison.CurrentCultureIgnoreCase))
-                return [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)];
+                return Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];
 
             if (IPAddress.TryParse(singleAddressOrHostname, out var ipAddress))
                 return [new IPEndPoint(ipAddress, port)];
--- a/libs/common/Format.cs
+++ b/libs/common/Format.cs
@@ -49,7 +49,7 @@ namespace Garnet.common
             // Check if input null or empty
             if (string.IsNullOrEmpty(addressList) || string.IsNullOrWhiteSpace(addressList))
             {
-                endpoints = [new IPEndPoint(IPAddress.Any, port)];
+                endpoints = [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)];
                 return true;
             }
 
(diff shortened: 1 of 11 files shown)


davidglassborow:
Thanks for the feedback @prvyk 

Latest commit will only bind to IPv6 if it's supported in the underlying OS

As for containers, I could change the entrypoint to explicitly listen on Any if considered appropriate ?

prvyk:
I think it is appropriate to open up containers. Docker does not automatically expose ports, but requires using -p etc. It has its own ways to firewall, and we shouldn't add extra friction on top of that.

davidglassborow:
Ok, docker version will listen to Ipv4 Any as before.

prvyk:
Looks nice now.  👍

In #1083 GarnetServer only prints Endpoint[0], while this does string.Join etc. 

This is a reasonable change, I just wonder whether @vazois had a particular reason he chose to limit output then.

badrishc:
The solution for Docker presented here won't work because user might have a Garnet.conf that uses a specific address, and this approach will always override it to 0.0.0.0.

Also, as this this a breaking change for current users and we had issues in the past, we have made the default Any for v4 and v6 in https://github.com/microsoft/garnet/pull/1151. Please review that.

prvyk:
I recall redis.conf uses a setting called 'protected mode'. Normally it defaults to 'yes' and limits connections to loopback by default. But when running a container, the setting is switched to 'no', and default is any. That way, --bind overrides the address in both cases.

badrishc:
> redis.conf uses a setting called 'protected mode'. Normally it defaults to 'yes' and limits connections to loopback by default. But > when running a container, the setting is switched to 'no', and default is any. That way, --bind overrides the address in both cases.

Agreed, we would need a separate contribution that adds protected mode and container override as you mention.

> Are you happy to keep [localhost being bound to both IPv4 and IPv6](https://github.com/microsoft/garnet/pull/1156/files#diff-86fb5fafb9a5a37420e5fdb5e79db65366e6bede8401ee8db06740e78d57c55aR92) or shall I just close this PR ?

I think the open PR at https://github.com/microsoft/garnet/pull/1151 already adds localhost being bound to IPv4 and IPv6, so this can be closed. We need to make sure the comment made [above](https://github.com/microsoft/garnet/pull/1156#issuecomment-2771759985) re: disabled IPv6 is handled in that PR.

badrishc:
> I just wonder whether @vazois had a particular reason he chose to limit output then. I think printing more than one is a very reasonable change.

Agreed, this part of the current PR can potentially be retained, IMHO.

libs/common/Format.cs
```diff
@@ -83,13 +83,13 @@ public static bool TryParseAddressList(string addressList, int port, out EndPoin
         public static async Task<EndPoint[]> TryCreateEndpoint(string singleAddressOrHostname, int port, bool tryConnect = false, ILogger logger = null)
         {
             if (string.IsNullOrEmpty(singleAddressOrHostname) || string.IsNullOrWhiteSpace(singleAddressOrHostname))
-                return [new IPEndPoint(IPAddress.Any, port)];
+                return Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];
```
vazois, line 86: Can you make this function call?
For example, defaultAny() => Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];

libs/common/Format.cs
```diff
@@ -83,13 +83,13 @@ public static bool TryParseAddressList(string addressList, int port, out EndPoin
...
 
             if (singleAddressOrHostname[0] == '-')
                 singleAddressOrHostname = singleAddressOrHostname.Substring(1);
 
             if (singleAddressOrHostname.Equals("localhost", StringComparison.CurrentCultureIgnoreCase))
-                return [new IPEndPoint(IPAddress.Loopback, port)];
+                return Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];
```
vazois, line 92: Same here, make it a function call.

vazois:
> Looks nice now. 👍
> 
> In #1083 GarnetServer only prints Endpoint[0], while this does string.Join etc.
> 
> I just wonder whether @vazois had a particular reason he chose to limit output then. I think printing more than one is a very reasonable change.

The main reason was aesthetics. If you a long list of IPs because of the hostname then it will not look nice with the watermark.
Actually, Redis prints only the port not the address in their watermark we might do that. 
Of course, with the assumption that in the future we will not support Endpoints with different ports.

libs/common/Format.cs
vazois: static EndPoint[] defaultBindLoopBack(int port)
            => Socket.OSSupportsIPv6 ? [new IPEndPoint(IPAddress.Loopback, port), new IPEndPoint(IPAddress.IPv6Loopback, port)] : [new IPEndPoint(IPAddress.Loopback, port)];

davidglassborow:
I'm glad the address is shown, I've had problems in production because I developed on Linux, where localhost is v4, and deployed to Windows where localhost is v6, took me ages to figure out the problem, would have been even worse without the address being shown on startup - Maybe render under the logo if concerned ?

Any timescales on where you might release a new version on nuget ?  I'm blocked in prod until the multi-address binding feature is release.

Given the discussions, I'll close this PR as most of it is covered by #1151

PR #1157, opened 2025-04-02
UnixDomainSocketEndPoint - improve error message if the user doesn't also specify the UnixSocketPath

davidglassborow:
The current error is confusing because the parameter name is actually the underlying parameter name of File.Delete.
I came across this when testing with UnixDomainSocketEndPoint, as to why I needed to specify the path twice, but can see that UnixDomainSocketEndPoint doesn't expose the path it's created with.

`System.ArgumentNullException : Value cannot be null. (Parameter 'path')` => `System.ArgumentNullException : Value cannot be null. (Parameter 'UnixSocketPath')`

### DIFF
--- a/libs/host/GarnetServer.cs
+++ b/libs/host/GarnetServer.cs
@@ -259,6 +259,8 @@ namespace Garnet
                 {
                     if (opts.EndPoints[i] is UnixDomainSocketEndPoint)
                     {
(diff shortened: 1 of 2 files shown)


badrishc (approved):

badrishc:
Looks like the formatting failed, could you run `dotnet format` to get past the formatting error.

badrishc (approved):

PR #1159, opened 2025-04-03
Update BDN tests base to only run framework that is current instead of both 8.0 and 9.0 each time

darrenge:
The current BDN tests always run .net80 and .net90 BDN tests no matter what the current framework is. This caused problems in the data as it put both .net80 results and .net90 results in the same chart. If the BDN tests just run one at a time, then the BDN scripts can put the proper data in the proper chart.


benchmark/BDN.benchmark/Program.cs
```diff
@@ -32,14 +33,21 @@ public BaseConfig()
...
-            .WithEnvironmentVariables(new EnvironmentVariable("DOTNET_TieredPGO", "0"));
-        Net9BaseJob = baseJob.WithRuntime(CoreRuntime.Core90)
-            .WithEnvironmentVariables(new EnvironmentVariable("DOTNET_TieredPGO", "0"));
-
-        AddJob(
-            Net8BaseJob.WithId(".NET 8"),
-            Net9BaseJob.WithId(".NET 9")
-            );
+        var runtime = System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription;
+        switch (runtime)
```
badrishc, line 37: Let the user specify this as command line parameter. If they didn't specify anything then run both frameworks so we are backward compatible with main.

darrenge:
It turns out to be an issue / Bug in BDN: When run with 'dotnet run', all optional BDN parameters are counted as args to program.cs. It also overwrites the last BDN arg with the '--' arg sent to program.cs.

Example:
dotnet run -c Release -f net9.0 --project C:/GarnetGitHub/benchmark/BDN.benchmark --filter BDN.benchmark.Operations.BasicOperations.* --exporters json -- net9.0

BDN fails with "net9.0 is invalid value for '--exporters'."

Workaround: Use the last parameter to determine the BDN framework (net8.0 or net9.0).  Then set the last arg to the second-to-last arg value (e.g., "json").  This ensures BDN runs correctly by using the last arg as the value for the second-to-last arg.

TedHartMS (approved):

Issue #1145, opened 2025-03-31
ZUNION returns empty result when the first input set is empty


zard-abutton:
### Describe the bug

ZUNION command produces an empty result when the first input set is empty, even if other input sets contain elements. This behavior differs from Redis, where ZUNION correctly returns the union of all provided sets regardless of the first set being empty.

Quote from https://microsoft.github.io/garnet/docs/commands/data-structures#zunion
> Keys that do not exist are considered to be empty sets.

### Steps to reproduce the bug

```
> zadd set1 1 a 2 b
(integer) 2

> zunion 2 set1 empty_set
1) "a"
2) "b"

> zunion 2 empty_set set1
(empty list or set)
```

### Expected behavior

```
> zadd set1 1 a 2 b
(integer) 2

> zunion 2 set1 empty_set
1) "a"
2) "b"

> zunion 2 empty_set set1
1) "a"
2) "b"

```

### Release version

1.0.61

PR #1162, opened 2025-04-05
Fix wrong type and non existing key issues in S/Z UNION, DIFF, INTER commands

Vijay-Nirmal:
Fix: #1145

This pull request includes updates to the handling of set and sorted set operations for UNION, DIFF and INTER when the first key does not exist and when there is a wrong type issue. Added new test cases to cover scenarios where the first key does not exist.

### DIFF
--- a/libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
+++ b/libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
@@ -1126,57 +1126,72 @@ namespace Garnet.server
 
             // Get the first sorted set
             var status = GET(keys[0].ToArray(), out var firstObj, ref objectContext);
+
+            if (status == GarnetStatus.WRONGTYPE)
+            {
+                return GarnetStatus.WRONGTYPE;
+            }
+
+            Dictionary<byte[], double> sortedSetDictionary = null;
+            if (status == GarnetStatus.NOTFOUND)
+            {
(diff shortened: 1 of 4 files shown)


libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -1126,57 +1126,72 @@ private GarnetStatus SortedSetUnion<TObjectContext>(ReadOnlySpan<ArgSlice> keys,
...
+
+            if (status == GarnetStatus.WRONGTYPE)
+            {
+                return GarnetStatus.WRONGTYPE;
+            }
+
+            Dictionary<byte[], double> sortedSetDictionary = null;
+            if (status == GarnetStatus.NOTFOUND)
+            {
+                sortedSetDictionary = new Dictionary<byte[], double>(ByteArrayComparer.Instance);
```
badrishc, line 1138: Instead of creating an empty dictionary here and immediately cloning it later, can we directly create `pairs` as empty later on if `sortedSetDictionary` is null. That way we avoid the extra object creation here.

Vijay-Nirmal: Done

Vijay-Nirmal:
The failed test is unrelated to my changes cc: @vazois 
image

badrishc (approved):

PR #1177, opened 2025-04-20
Bugfix Incorrect Etag increment on SETIFGREATER

hamdaankhalid:
SETIFGREATER has incorrect etag setting semantics

### DIFF
--- a/libs/server/Storage/Functions/MainStore/RMWMethods.cs
+++ b/libs/server/Storage/Functions/MainStore/RMWMethods.cs
@@ -587,7 +587,7 @@ namespace Garnet.server
                         return false;
 
                     // doesn't update etag, since it's only the metadata that was updated
-                    return true; ;
+                    return true;
                 case RespCommand.PEXPIREAT:
                 case RespCommand.EXPIREAT:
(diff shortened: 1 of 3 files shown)


libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -1049,10 +1050,12 @@ public bool CopyUpdater(ref SpanByte key, ref RawStringInput input, ref SpanByte
 
                     long etagFromClient = input.parseState.GetLong(1);
 
-                    // change the current etag to the the etag sent from client since rest remains same
+
+                    // outside the switch statment we always increment the Etag, by setting it one below we keep logic consistent with rest of SETIFMATCH
                     if (cmd == RespCommand.SETIFGREATER)
-                        functionsState.etagState.etag = etagFromClient;
+                        functionsState.etagState.etag = etagFromClient - 1;
```
badrishc, line 1056: This seems a bit hacky, can't we set the etag correctly in both cases directly

badrishc (approved):

PR #1179, opened 2025-04-23
Add DELIFGREATER ETag Command

hamdaankhalid:
Adds DELIFGREATER RMW Operation to Garnet.
This can be used to synchronize deletes without relying on transactions.

Also minor bugfix of GETDEL overallocating in RCU route

### DIFF
--- a/libs/server/Storage/Functions/MainStore/RMWMethods.cs
+++ b/libs/server/Storage/Functions/MainStore/RMWMethods.cs
@@ -908,6 +908,17 @@ namespace Garnet.server
         {
             switch (input.header.cmd)
             {
+                case RespCommand.DELIFGREATER:
+                    if (rmwInfo.RecordInfo.ETag)
+                        EtagState.SetValsForRecordWithEtag(ref functionsState.etagState, ref oldValue);
+                    long etagFromClient = input.parseState.GetLong(0);
+                    if (etagFromClient <= functionsState.etagState.etag)
+                    {
+                        EtagState.ResetState(ref functionsState.etagState);
+                        return false;
+                    }
+
+                    return true;
                 case RespCommand.SETIFGREATER:
                 case RespCommand.SETIFMATCH:
                     if (rmwInfo.RecordInfo.ETag)
@@ -1021,6 +1032,7 @@ namespace Garnet.server
                 return false;
             }
 
+            // threw an exception here
             rmwInfo.ClearExtraValueLength(ref recordInfo, ref newValue, newValue.TotalSize);
 
             RespCommand cmd = input.header.cmd;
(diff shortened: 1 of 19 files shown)


libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -903,6 +910,9 @@ public bool NeedCopyUpdate(ref SpanByte key, ref RawStringInput input, ref SpanB
             {
                 case RespCommand.SETIFGREATER:
                 case RespCommand.SETIFMATCH:
+                    if (rmwInfo.RecordInfo.ETag)
```
hamdaankhalid, line 913: This is to catch anytime we skip IPU -> NCU -> RCU pipeline and instead end up doing NCU -> RCU due to ongoing checkpointing.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -27,6 +27,7 @@ public bool NeedInitialUpdate(ref SpanByte key, ref RawStringInput input, ref Sp
                 case RespCommand.PEXPIREAT:
                 case RespCommand.GETDEL:
                 case RespCommand.GETEX:
+                case RespCommand.DELIFGREATER:
```
hamdaankhalid, line 30: Init update does not apply, so early exit

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -355,9 +356,15 @@ private bool InPlaceUpdaterWorker(ref SpanByte key, ref RawStringInput input, re
                     EtagState.ResetState(ref functionsState.etagState);
                     // Nothing is set because being in this block means NX was already violated
                     return true;
+                case RespCommand.DELIFGREATER:
+                    long etagFromClient = input.parseState.GetLong(0);
```
hamdaankhalid, line 360: expire and stop will cause tombstoning in SessionFunctions, and cancelling will let us just early exit the from internalRMW without expiring.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -1026,10 +1035,17 @@ public bool CopyUpdater(ref SpanByte key, ref RawStringInput input, ref SpanByte
 
             switch (cmd)
             {
+                case RespCommand.DELIFGREATER:
+                    // We are doing this in copy updater instead of NCU because we need to tombstone the record in InternalRMW.
+                    // NCU does not tombstone the record.
+                    long etagFromClient = input.parseState.GetLong(0);
+                    rmwInfo.Action = etagFromClient > functionsState.etagState.etag ? RMWAction.ExpireAndStop : RMWAction.CancelOperation;
```
hamdaankhalid, line 1042: We either expire and add a tombstone record, or we had one extra allocation on the hybrid log that can then be put in the free record pool because of cancelling.

badrishc (approved):

PR #1180, opened 2025-04-25
Implement active expiration for mutable region

hamdaankhalid:


wessisaac:
@hamdaankhalid To enable active expiration and free list revivification, which settings need to be configured in the Garnet.conf file?

PR #1188, opened 2025-05-07
Improve CollectionItemBroker scalability

badrishc:
Improve scalability of `CollectionItemBroker` logic at high scale, misc improvements to sorted set expiration that regressed performance. Add correct finally logic to scan iterator.


TedHartMS (approved):
The "Conversation" tab still has some comments about opt-in which is no longer part of this

PR #1189, opened 2025-05-07
[Issue 1191] Handle free list for revivification in RMW for expiration/deletes

hamdaankhalid:
RMW does not allow for Free list Revivification when the code decides to either actively expire a key or delete (via expiration).

RMW can work off of the same semantics of when to perform elision of record from hybrid log as DELETE flow does. This PR refactors the code and moves logic into RMW as DELETE

New places of revivification:
When going via IPU and we find expired record we can tombstone and free list it.
When going via NCU we can potentially do complete elision, no new record, or do tombstone + old record elision.


TedHartMS (approved):

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -1046,13 +1048,6 @@ public bool CopyUpdater(ref SpanByte key, ref RawStringInput input, ref SpanByte
 
             switch (cmd)
             {
-                case RespCommand.DELIFGREATER:
```
hamdaankhalid, line 1049: moved this code to NCU since it is more efficient to handle not creating a new record at all when working with deletion.

hamdaankhalid: TODO check if NCU is tombstoning correctly in 2 cases:
1. CanElide and we can either elide from hei OR elide from HEI and free list.
2. Cannot Elide, then this should drop into RCU route except this would just create a tombstone record.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/Helpers.cs
```diff
@@ -253,5 +253,55 @@ private bool FindTagAndTryTransientSLock<TInput, TOutput, TContext, TSessionFunc
             stackCtx.SetRecordSourceToHashEntry(hlogBase);
             return true;
         }
+
+
+        // Note: We do not currently consider this reuse for mid-chain records (records past the HashBucket), because TracebackForKeyMatch would need
+        //  to return the next-higher record whose .PreviousAddress points to this one, *and* we'd need to make sure that record was not revivified out.
+        //  Also, we do not consider this in-chain reuse for records with different keys, because we don't get here if the keys don't match.
+        private void HandleRecordElision<TInput, TOutput, TContext, TSessionFunctionsWrapper>(
```
hamdaankhalid, line 261: Lift and shift code from Delete.

This function assumes you have already checked the record can be elided and now you do approporiate ellision and further steps based on config of running server.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalDelete.cs
```diff
@@ -119,51 +119,10 @@ internal OperationStatus InternalDelete<TInput, TOutput, TContext, TSessionFunct
...
-                                    // HashBucketEntry.word to 0, it means we do not expect any records for this key's tag to exist after the elision. Therefore,
-                                    // we can re-insert the record iff the HashBucketEntry's address is <= kTempInvalidAddress.
-                                    stackCtx.hei = new(stackCtx.hei.hash);
-                                    FindOrCreateTag(ref stackCtx.hei, hlogBase.BeginAddress);
-
-                                    if (stackCtx.hei.entry.Address <= Constants.kTempInvalidAddress && stackCtx.hei.TryCAS(stackCtx.recSrc.LogicalAddress))
-                                        srcRecordInfo.UnsealAndValidate();
-                                }
-                            }
+                            HandleRecordElision<TInput, TOutput, TContext, TSessionFunctionsWrapper>(
```
hamdaankhalid, line 124: Lift and shift into helpers.
Code is validated by running existing delete unit tests.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -138,16 +138,34 @@ internal OperationStatus InternalRMW<TInput, TOutput, TContext, TSessionFunction
                     }
 
                     // rmwInfo's lengths are filled in and GetValueLengths and SetLength are called inside InPlaceUpdater.
-                    if (sessionFunctions.InPlaceUpdater(stackCtx.recSrc.PhysicalAddress, ref key, ref input, ref recordValue, ref output, ref rmwInfo, out status, ref srcRecordInfo)
```
hamdaankhalid, line 141: expire and stop is moved out of IPU hotpath. Into it's own block

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -138,16 +138,34 @@ internal OperationStatus InternalRMW<TInput, TOutput, TContext, TSessionFunction
...
                     {
                         MarkPage(stackCtx.recSrc.LogicalAddress, sessionFunctions.Ctx);
 
+                        pendingContext.recordInfo = srcRecordInfo;
+                        pendingContext.logicalAddress = stackCtx.recSrc.LogicalAddress;
+                        // status has been set by InPlaceUpdater
+                        goto LatchRelease;
+                    }
+
+                    if (rmwInfo.Action == RMWAction.ExpireAndStop)
```
hamdaankhalid, line 151: Mutable region deletion/expiration where we just mark an existing record as tombstoned. Now the code should be able to pick up and use perform ellision if applicable to the tombstoned record.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -388,7 +406,26 @@ private OperationStatus CreateNewRecordRMW<TInput, TOutput, TContext, TSessionFu
                         forExpiration = true;
                     }
                     else if (rmwInfo.Action == RMWAction.ExpireAndStop)
+                    {
+                        MarkPage(stackCtx.recSrc.LogicalAddress, sessionFunctions.Ctx);
+
+                        // Handle potential record revivification for records in in-memory region. Handle readonly region and potentially Mutable region undergoing CPR. 
+                        if (stackCtx.recSrc.HasMainLogSrc)
```
hamdaankhalid, line 413: If NCU wants to expire and stop at this point, we are not adding a tombstone. The caller will see the record as deleted from the interface pov. This does not incur the cost of a new record either.

Not sure if there is a potential edge case I am missing where this would cause an issue.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -388,7 +406,26 @@ private OperationStatus CreateNewRecordRMW<TInput, TOutput, TContext, TSessionFu
...
+                        {
+                            // HK TODO: Ask @TedHart if record should be marked dirty and modified here?
```
hamdaankhalid, line 415: @TedHartMS

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
hamdaankhalid: I think this should still create an entry in hybrid log the same way we do for RCU tombstoning.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -911,14 +911,16 @@ public bool NeedCopyUpdate(ref SpanByte key, ref RawStringInput input, ref SpanB
...
                     if (rmwInfo.RecordInfo.ETag)
                         EtagState.SetValsForRecordWithEtag(ref functionsState.etagState, ref oldValue);
+
                     long etagFromClient = input.parseState.GetLong(0);
-                    if (etagFromClient <= functionsState.etagState.etag)
+                    if (etagFromClient > functionsState.etagState.etag)
                     {
-                        EtagState.ResetState(ref functionsState.etagState);
-                        return false;
+                        rmwInfo.Action = RMWAction.ExpireAndStop;
```
TedHartMS, line 918: Add a comment here explaining why there's no "true" return: Either we ExpireAndStop (in which case we must return false) or there is no CopyUpdate to be done (so again we must return false).

libs/storage/Tsavorite/cs/src/core/Allocator/SpanByteAllocator.cs
```diff
@@ -80,6 +80,9 @@ public readonly (int actualSize, int allocatedSize, int keySize) GetRMWCopyDesti
             where TVariableLengthInput : IVariableLengthInput<SpanByte, TInput>
              => _this.GetRMWCopyDestinationRecordSize(ref key, ref input, ref value, ref recordInfo, varlenInput);
 
+        /// <inheritdoc/>
+        public (int actualSize, int allocatedSize, int keySize) GetTombtoneRecordSize(ref SpanByte key) => _this.GetTombstoneRecordSize(ref key);
```
TedHartMS, line 84: Tombtone == sinister background music...?

hamdaankhalid: exactly what I wanted hahaha.

JK I will fix this.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
hamdaankhalid: done

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -138,16 +138,35 @@ internal OperationStatus InternalRMW<TInput, TOutput, TContext, TSessionFunction
...
+                        pendingContext.recordInfo = srcRecordInfo;
+                        pendingContext.logicalAddress = stackCtx.recSrc.LogicalAddress;
+                        // status has been set by InPlaceUpdater
+                        goto LatchRelease;
+                    }
+
+                    if (rmwInfo.Action == RMWAction.ExpireAndStop)
+                    {
+                        MarkPage(stackCtx.recSrc.LogicalAddress, sessionFunctions.Ctx);
+                        // HK TODO: Ask @TedHart if record should be marked dirty and modified here? By now it is already tombstoned
```
hamdaankhalid, line 154: @TedHartMS can you confirm this please.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -138,16 +138,35 @@ internal OperationStatus InternalRMW<TInput, TOutput, TContext, TSessionFunction
...
+                        srcRecordInfo.SetDirtyAndModified();
+
                         // ExpireAndStop means to override default Delete handling (which is to go to InitialUpdater) by leaving the tombstoned record as current.
                         // Our SessionFunctionsWrapper.InPlaceUpdater implementation has already reinitialized-in-place or set Tombstone as appropriate and marked the record.
+
+                        // Try to transfer the record from the tag chain to the free record pool iff previous address points to invalid address.
```
hamdaankhalid, line 160: CRUX of the PR

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -518,8 +557,8 @@ private OperationStatus CreateNewRecordRMW<TInput, TOutput, TContext, TSessionFu
                 }
                 else
                 {
-                    // Else it was a CopyUpdater so call PCU
-                    if (!sessionFunctions.PostCopyUpdater(ref key, ref input, ref value, ref hlog.GetValue(newPhysicalAddress), ref output, ref rmwInfo, ref newRecordInfo))
```
hamdaankhalid, line 522: PCU was being called even when CU had previously returned false for RCU for tombstoning. I believe this was not the intended behavior so I have changed the code to not call PCU when RCU returns false and ExpireAndStop (tombstoning)

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -388,7 +413,21 @@ private OperationStatus CreateNewRecordRMW<TInput, TOutput, TContext, TSessionFu
                         forExpiration = true;
                     }
                     else if (rmwInfo.Action == RMWAction.ExpireAndStop)
-                        return OperationStatus.NOTFOUND;
+                    {
+                        if (allocOptions.ElideSourceRecord)
+                        {
+                            // Is there an issue in doing this? I only want to do this so HandleRecordElision does not trip on the Debug Assert
```
hamdaankhalid, line 419: Open question.

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
```diff
@@ -473,6 +516,10 @@ private OperationStatus CreateNewRecordRMW<TInput, TOutput, TContext, TSessionFu
                 }
                 if (rmwInfo.Action == RMWAction.ExpireAndStop)
                 {
+                    Debug.Assert(skipTombstoneAddition, "Should not have gone down RCU if NCU had already requested tombstoning." +
+                        "This block should only handle expiration/tombstoning via RCU.");
+                    skipTombstoneAddition = false;
+                    newRecordInfo.SetDirtyAndModified();
```
hamdaankhalid, line 522: another check for Ted to confirm, right usage

libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/InternalRMW.cs
hamdaankhalid: confirmed thisis okay

TedHartMS (approved):

PR #1190, opened 2025-05-07
Add protected mode.

prvyk:
Redis starts by default in 'protected mode' - binds to localhost only unless auth is specified. This PR adds the same to Garnet, except we don't check for auth - the user can explicitly bind elsewhere, and this override protected mode binding. 

Docker builds automatically disable protected mode and so bind by default to 0.0.0.0.

Note: CommandLineBooleanOption is added to get compatibility with redis's yes/no parsing.


libs/host/Configuration/Options.cs
```diff
@@ -533,6 +533,10 @@ internal sealed class Options
         [Option("enable-debug-command", Required = false, HelpText = "Enable DEBUG command for 'no', 'local' or 'all' connections")]
         public ConnectionProtectionOption EnableDebugCommand { get; set; }
 
+        [OptionValidation]
+        [Option("protected-mode", Required = false, HelpText = "Enable protected mode.")]
+        public CommandLineBooleanOption ProtectedMode { get; set; }
```
badrishc, line 538: Do we need Garnet's version of options to be `CommandLineBooleanOption` instead of a simple `bool`. RedisOptions makes sense for compatibility with redis.conf.

libs/host/Configuration/Options.cs
```diff
@@ -682,7 +686,8 @@ public GarnetServerOptions GetServerOptions(ILogger logger = null)
             var checkpointDir = CheckpointDir;
             if (!useAzureStorage) checkpointDir = new DirectoryInfo(string.IsNullOrEmpty(checkpointDir) ? (string.IsNullOrEmpty(logDir) ? "." : logDir) : checkpointDir).FullName;
 
-            if (!Format.TryParseAddressList(Address, Port, out var endpoints, out _) || endpoints.Length == 0)
+            if (!Format.TryParseAddressList(Address, Port, out var endpoints, out _, ProtectedMode == CommandLineBooleanOption.True)
```
badrishc, line 689: ProtectedMode == CommandLineBooleanOption.True || ProtectedMode == CommandLineBooleanOption.Yes?

libs/host/Configuration/Options.cs
badrishc: I think `CommandLineBooleanOption` may be fine here, leaving the comment for discussion and input from others.

prvyk: The reason was to allow CommandLine.Core to parse 'yes' as true. I couldn't find any nicer way of enhancing the parser. 

If we don't care for that compatibility, I can switch to a bool and switch the docker definition to '--protected-mode false'.

libs/host/Configuration/Options.cs
prvyk: It's defined to the same enum value, so comparing against True is sufficient.

badrishc (approved):

PR #1192, opened 2025-05-07
Reduce the BDN Chart Size from max items per chart from 50 to 25.  

darrenge:
Reduce the BDN Chart Size from max items per chart from 50 to 25.  This will keep net80 and net90 and it reduces the data file by half. This is in hope that we get around the BDN charting scalability problems.


darrenge:
One note - the old data (data points 26->50) are actually deleted from the data file.  If we still run into scalability problems with this fix, then we probably will need to scale back to net9.0 only.

darrenge:
Might do something bigger with this so closing out this PR

PR #1205, opened 2025-05-17
CollectionItemBroker Cleanup + Fixes

TalZaccai:
Addressing some comments / issues with the existing CollectionItemBroker implementation:
- When are keys removed from keysToObservers?
  - [x] Added logic for removing keys from keysToObservers when the observer queue is empty
  - [ ] Will add an additional event that triggers key removal in the main loop (in case the observers died and we never checked that key's queue again)
- HandleBrokerEvent is sync; if there are a large number of blocking sessions, can this be a scalaibility bottleneck?
  - This is by design, as the blocking commands are operating in a first-come-first-serve manner. We can change the logic here but that might be confusing as this behavior might not be expected by the user.
- Use named (e.g. NotStarted, Started, Disposed) constants for mainLoopTaskStatus. Badrish's checkin added a "2" value for Dispose() so this will improve readability.
  - [x] Done.
- Dispose() loops on Interlocked.CompareExchange but is not testing anything about the value before doing the CompareExchange, so a single Interlocked.Exchange should work
  - [x] Done. 

Also addressed in this PR - 
- [x] Handling WRONGTYPE errors in blocking commands as expected by the protocol (this is in place of PR #1130 which first mentioned the issue)
- [x] Adding tests to check proper WRONGTYPE handling in blocking commands
- [x] Handling "special" CollectionItemResult instances in previously missed commands
- [ ] Need to have all blocking commands covered in "forced unblock" tests (ClientUnblockBasicTest, MultipleClientsUnblockAndAddTest in RespTests)

### DIFF
--- a/libs/server/Objects/ItemBroker/CollectionItemBrokerEvent.cs
+++ b/libs/server/Objects/ItemBroker/CollectionItemBrokerEvent.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System.Collections.Concurrent;
-
 namespace Garnet.server
 {
     /// <summary>
@@ -22,15 +20,9 @@ namespace Garnet.server
         /// </summary>
         internal readonly byte[] Key;
 
-        /// <summary>
-        /// Observers
-        /// </summary>
-        internal readonly ConcurrentQueue<CollectionItemObserver> Observers;
-
-        public CollectionUpdatedEvent(byte[] key, ConcurrentQueue<CollectionItemObserver> observers)
+        public CollectionUpdatedEvent(byte[] key)
         {
             Key = key;
-            Observers = observers;
         }
     }
 
--- a/libs/server/Objects/ItemBroker/CollectionItemBroker.cs
+++ b/libs/server/Objects/ItemBroker/CollectionItemBroker.cs
@@ -46,8 +46,13 @@ namespace Garnet.server
         // Task for the main loop, we keep field for diagnostic purposes
         Task mainLoopTask = null;
 
-        // Flag to indicate if the main loop task has started
-        int mainLoopTaskStarted = 0;
+        // Integer to indicate main loop status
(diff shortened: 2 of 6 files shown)


prvyk:
Another thing I noticed in #1130 was that lists/sets should be deleted when empty -

https://github.com/microsoft/garnet/pull/1130/commits/407a80c7f47e65a6ed0f7ecd68a3a58455134c1b

lpush a 1
blpop a 0
exists a
(is 0 on redis, is 1 here. lpop is 0 on both redis and Garnet)

TalZaccai:
> Another thing I noticed in #1130 was that lists/sets should be deleted when empty -
> 
> [407a80c](https://github.com/microsoft/garnet/commit/407a80c7f47e65a6ed0f7ecd68a3a58455134c1b)
> 
> lpush a 1 blpop a 0 exists a (is 0 on redis, is 1 here. lpop is 0 on both redis and Garnet)

Ah good to know! Thanks for pointing it out :)

libs/server/Objects/ItemBroker/CollectionItemBrokerEvent.cs
```diff
@@ -1,58 +1,72 @@
...
+        [FieldOffset(16)]
+        internal readonly CollectionItemBrokerEventType EventType = CollectionItemBrokerEventType.NotSet;
 
         /// <summary>
-        /// The keys that the observer requests to subscribe on
+        /// Creates a CollectionUpdated event
         /// </summary>
-        internal byte[][] Keys { get; }
+        /// <param name="key">Key of updated collection</param>
+        public CollectionItemBrokerEvent(byte[] key)
```
badrishc, line 52: convert these to static helpers that return a struct, such as `static CollectionItemBrokerEvent NewCollectionUpdatedEvent(...)`

badrishc (approved):

libs/server/Objects/ItemBroker/CollectionItemBrokerEvent.cs
TalZaccai: Done

TedHartMS (approved):

PR #1208, opened 2025-05-20
Pass RESP3 to modules and custom commands

prvyk:
This PR changes command and module signatures so that we pass RespMemoryWriter to them instead of the ref tuple. This allows adjusting output to RESP3.

I considered the alternative of just passing an additional respProtocolVersion in the signature, and I think this is slightly nicer. We can erase some custom output code, authors get more output functions, and reallocation is handled for them. If they want the previous allocation behaviour (measure and then Rent exactly from the pool), they can get it by calling RespMemoryWriter.Realloc() themselves first.


libs/server/Custom/CustomRawStringFunctions.cs
```diff
@@ -2,15 +2,15 @@
...
+using Garnet.common;
 using Tsavorite.core;
 
 namespace Garnet.server
 {
     /// <summary>
     /// Base class for custom functions on raw strings
     /// </summary>
-    public abstract class CustomRawStringFunctions : CustomFunctions
+    public abstract class CustomRawStringFunctions
```
prvyk, line 13: Note CustomObjectFunctions does not inherit from CustomFunctions, and that CustomRawStringFunctions construction never fills in CustomFunctions's respServerSession field.

badrishc: Do we still need CustomFunctions if neither CustomObject not CustomRawString are using it? Can CustomFunctions be reduced or renamed based on where it is actually needed?

libs/server/Storage/Functions/ObjectStore/RMWMethods.cs
```diff
@@ -31,11 +30,17 @@ public bool NeedInitialUpdate(ref byte[] key, ref ObjectInput input, ref GarnetO
                     else
                     {
                         var customObjectCommand = GetCustomObjectCommand(ref input, type);
-                        (IMemoryOwner<byte> Memory, int Length) outp = (output.SpanByteAndMemory.Memory, 0);
-                        var ret = customObjectCommand.NeedInitialUpdate(key, ref input, ref outp);
-                        output.SpanByteAndMemory.Memory = outp.Memory;
-                        output.SpanByteAndMemory.Length = outp.Length;
-                        return ret;
+
+                        var writer = new RespMemoryWriter(functionsState.respProtocolVersion, ref output.SpanByteAndMemory);
```
badrishc, line 34: How about `using` pattern instead of `var ... try ... finally`?

libs/server/Storage/Functions/ObjectStore/RMWMethods.cs
```diff
@@ -57,11 +62,16 @@ public bool InitialUpdater(ref byte[] key, ref ObjectInput input, ref IGarnetObj
                 var customObjectCommand = GetCustomObjectCommand(ref input, type);
                 value = functionsState.GetCustomObjectFactory((byte)type).Create((byte)type);
 
-                (IMemoryOwner<byte> Memory, int Length) outp = (output.SpanByteAndMemory.Memory, 0);
-                var result = customObjectCommand.InitialUpdater(key, ref input, value, ref outp, ref rmwInfo);
-                output.SpanByteAndMemory.Memory = outp.Memory;
-                output.SpanByteAndMemory.Length = outp.Length;
-                return result;
+                var writer = new RespMemoryWriter(functionsState.respProtocolVersion, ref output.SpanByteAndMemory);
```
badrishc, line 65: How about `using` pattern instead of `var ... try ... finally`?

libs/server/Storage/Functions/ObjectStore/RMWMethods.cs
```diff
@@ -161,13 +171,18 @@ bool InPlaceUpdaterWorker(ref byte[] key, ref ObjectInput input, ref IGarnetObje
...
                         }
 
-                        (IMemoryOwner<byte> Memory, int Length) outp = (output.SpanByteAndMemory.Memory, 0);
                         var customObjectCommand = GetCustomObjectCommand(ref input, input.header.type);
-                        var result = customObjectCommand.Updater(key, ref input, value, ref outp, ref rmwInfo);
-                        output.SpanByteAndMemory.Memory = outp.Memory;
-                        output.SpanByteAndMemory.Length = outp.Length;
-                        return result;
-                        //return customObjectCommand.InPlaceUpdateWorker(key, ref input, value, ref output.spanByteAndMemory, ref rmwInfo);
+                        var writer = new RespMemoryWriter(functionsState.respProtocolVersion, ref output.SpanByteAndMemory);
```
badrishc, line 175: How about `using` pattern instead of `var ... try ... finally`?

libs/server/Custom/CustomRawStringFunctions.cs
prvyk: CustomProcedure and CustomTransactionProcedure still use CustomFunctions.

(Custom transactions get a CustomProcedureInput where the respversion is already transferred, so I thought there was not much need to touch there)

libs/server/Storage/Functions/ObjectStore/RMWMethods.cs
prvyk: I can't pass it by reference then?

badrishc: Oh well ... :)

libs/server/Custom/CustomRawStringFunctions.cs
badrishc: Maybe CustomProcedureBase?

prvyk: That sounds fine. Done.

badrishc (approved):

TedHartMS (approved):

Issue #1199, opened 2025-05-12
Garnet Worker net9.0 does not run


ryanheath:
### Describe the bug

The net9.0's v1.0.64 of the Garnet.worker.exe does not work or run.

It fails with exception:
`Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'System.Diagnostics.EventLog, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
File name: 'System.Diagnostics.EventLog, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
   at Microsoft.Extensions.Logging.EventLog.EventLogLogger..ctor(String name, EventLogSettings settings, IExternalScopeProvider externalScopeProvider)`

But the file `System.Diagnostics.EventLog.dll` is present in the Service folder of the release.

The net8.0 version does run.

One thing I noticed in the win-x64-based-readytorun.pubxml file, it seems the libs from net8.0 are also copied for the net9.0 version, maybe that is the bug?

See 
- [win-x64-based-readytorun.pubxml line 33](https://github.com/microsoft/garnet/blob/1b9a17444dd0f57d8673b0e50e72eb989837867c/main/GarnetServer/Properties/PublishProfiles/win-x64-based-readytorun.pubxml#L33C2-L33C133)
- [win-x64-based-readytorun.pubxml line 42](https://github.com/microsoft/garnet/blob/1b9a17444dd0f57d8673b0e50e72eb989837867c/main/GarnetServer/Properties/PublishProfiles/win-x64-based-readytorun.pubxml#L42C2-L42C108)

Also the net80\Service folder has a few dlls more than net90\Service folder.
For instance
- Microsoft.Bcl.Memory.dll
- System.Diagnostics.DiagnosticSource.dll
- System.IO.Pipelines.dll
- System.Text.Encodings.Web.dll
- System.Text.Json.dll

Side note, there are a lot of extra dlls, compared to the standalone executable of Garnet.
But that is probably any other story ...

### Steps to reproduce the bug

1. Download v1.0.64 asset file [win-x64-based-readytorun.zip](https://github.com/microsoft/garnet/releases/download/v1.0.64/win-x64-based-readytorun.zip)
2. In folder net9.0\Service there is file Garnet.worker.exe
3. Verify it does not run
4. In folder net8.0\Service there is file Garnet.worker.exe
[...]

PR #1209, opened 2025-05-20
Fix Garnet.Worker.exe failing on 9.0 in the published win-x64-based-readytorun.zip

darrenge:
The win-x64-based-readytorun.pubxml had hard coded net8.0 files that still worked for net9.0 until Garnet Release 61. Looks like nuget update caused 9.0 to fail. Just changing that hardcode net80 to $(TargetFramework) fixed the issue.

Verified fix by running release pipeline (without push to production) and verified win-x64-based-readytorun.zip


badrishc (approved):

ryanheath:
> win-x64-based-readytorun.zip

Where can I find this artifact?

darrenge:
It will be in next release ... adding comment to Issue so easy to find for folks: https://github.com/microsoft/garnet/issues/1199

PR #1214, opened 2025-05-22
ZRANGE BYLEX: Fix infinities processing, and invert exclusions when processing.

prvyk:
ZRANGEBYLEX + is supposed to be 'positive infinity' yet it was processed same as '-'.  Fix and add tests.

Also, exclusions ('[' and '(') weren't reversed, fix, test already checked this, but looked at the wrong result ("ZRANGE board [g (aaa BYLEX REV" command should return g as well).


libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -28,6 +28,13 @@ private struct ZRangeOptions
             public bool WithScores { get; set; }
         };
 
+        private enum SpecialRanges
```
TalZaccai, line 31: This could be a byte

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -1099,27 +1121,46 @@ private static bool TryParseParameter(ReadOnlySpan<byte> val, out double valueDo
...
+                    break;
                 default:
                     return false;
             }
+
+            if (limitChars.Length == 1)
+            {
+                if ((limitChars[0] == '-') || (limitChars[0] == '+'))
+                {
+                    // Doing this on '+' is intentional for compatibility.
```
TalZaccai, line 1157: Can you perhaps provide a more detailed comment here as to why this is required?

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -3527,76 +3527,157 @@ public void CanDoZRangeByScoreWithLimitLC(int bytesSent)
         }
 
         [Test]
-        public void CanDoZRangeByLex()
+        public async Task CanDoZRangeByLex()
```
TalZaccai, line 3530: Same comment as in PR #1143, let's stick with LC for now.

TalZaccai (changes requested):
Looks good overall, thanks!

test/Garnet.test/RespSortedSetTests.cs
prvyk: Switched back to LC.

PR #1218, opened 2025-05-23
Implementing Multi-DB support in StoreWrapper's ObjectCollectTask

TalZaccai:
This was an oversight in #1005, `StoreWrapper`'s `ObjectCollectTask` was only running collection on the default database. 
Thanks @hamdaankhalid for finding this 🥇


hamdaankhalid (approved):
LGTM

Issue #1220, opened 2025-05-23
New —protected-mode flag should be off by default in containers


danegsta:
### Describe the bug

The new required —protected-mode flag means that anyone running the 1.0 image with custom arguments (for auth, etc) will be broken until the new argument is added.

This is an issue in Aspire as we scaffold the 1.0 image and customize arguments, meaning any Aspire users with Garnet resources will be broken once they pull an updated image. 

Is there any way the default for protected mode can be changed to `no` in the container images?

### Steps to reproduce the bug

Run a ghcr.io/microsoft/garnet:1.0 container with custom arguments (but without the new —protected-mode no argument). The garnet instance won’t be accessible outside the container as it only binds to localhost. 

### Expected behavior

When run in a container, garnet should bind to all addresses even if —protected-mode no isn’t specified (the default in a container should be protected mode off). 

### Release version

v1.0.67

danegsta:
I know that —protected-mode is disabled in the default arguments for the images, but that doesn’t work if the arguments are being overridden.

badrishc:
As you mention, out containers indeed disable protected mode using arguments for images. Is it not possible for you to add to the arguments instead of overriding them?

badrishc:
Would a solution such as [this](https://github.com/redis/docker-library-redis/blob/master/6.2/alpine/Dockerfile#L79) where we modify the default.conf to disable protected mode, work? That way the entryline will not need to have the flag for the docker builds.

danegsta:
That seems like it would be ideal for container users if possible.

prvyk:
Can't you pass '--bind' '0.0.0.0' ? This will be accepted by previous versions, and overrides protected mode in practice in newer versions too?

badrishc:
> Can't you pass '--bind' '0.0.0.0' arguments? This will be accepted by previous versions, and overrides protected mode in newer versions too.

That won't help users of aspire that are on their current stable release, instead of main.

PR #1221, opened 2025-05-23
Disable protected mode by patching defaults.conf

badrishc:
Disable protected mode for Docker usage by patching `defaults.conf` instead of setting `--protected-mode no` via Dockerfile's `ENTRYLINE`. This change allows downstream consumers to override the arguments and still not lose the disabling of protected mode.

Updates version to 1.0.68 for new release

Fixes https://github.com/microsoft/garnet/issues/1220


badrishc:
`Dockerfile.nanoserver` does not work with this change, so this is on hold until we get some idea on how to make that work.

TalZaccai (approved):

TedHartMS (approved):

PR #1223, opened 2025-05-25
RespServerSession RESP3 update

prvyk:
This PR has two main goals:
1) Further update RESP3 support.
2) Save on boilerplate in RespServerSession output.

The former is done by fixing various cases for RESP3 null/double and adding verbatim string support. The latter is done by adding output functions to RespServerSession (in a separate file), but the PR holds off on use till performance implications are fully checked. Also some misc. RESP2 updates (redis-cli expects an extra newlines in CLIENT INFO/LIST, these commands should emit Bulk, not simple strings).

### DIFF
--- a/libs/server/Storage/Session/ObjectStore/Common.cs
+++ b/libs/server/Storage/Session/ObjectStore/Common.cs
@@ -229,9 +229,9 @@ namespace Garnet.server
         unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out string error, bool isScanOutput = false)
         {
             if (functionsState.respProtocolVersion >= 3)
-                return __ProcessResp3ArrayOutput(output, out error, isScanOutput);
+                return ProcessResp3ArrayOutput(output, out error, isScanOutput);
 
-            return __ProcessResp2ArrayOutput(output, out error, isScanOutput);
+            return ProcessResp2ArrayOutput(output, out error, isScanOutput);
         }
 
         /// <summary>
@@ -241,7 +241,7 @@ namespace Garnet.server
         /// <param name="error">A description of the error, if there is any</param>
         /// <param name="isScanOutput">True when the output comes from HSCAN, ZSCAN OR SSCAN command</param>
         /// <returns></returns>
-        private unsafe ArgSlice[] __ProcessResp2ArrayOutput(GarnetObjectStoreOutput outputFooter, out string error, bool isScanOutput)
+        private unsafe ArgSlice[] ProcessResp2ArrayOutput(GarnetObjectStoreOutput outputFooter, out string error, bool isScanOutput)
         {
             ArgSlice[] elements = default;
             error = default;
@@ -320,9 +320,13 @@ namespace Garnet.server
             return elements;
         }
 
-        private unsafe ArgSlice[] __ProcessResp3ArrayOutput(GarnetObjectStoreOutput output, out string error, bool isScanOutput)
+        private unsafe ArgSlice[] ProcessResp3ArrayOutput(GarnetObjectStoreOutput output, out string error, bool isScanOutput)
         {
-            bool IsSupportedArrayType(char c)
+            // We support arrays ('*'), RSEP3 sets ('~') and RESP3 maps ('%').
+            // All are returned as arrays.
+            // Returning other types will lead to unnecessary duplication of code,
+            // as we need to support RESP2 arrays in all cases anyway.
+            static bool IsSupportedArrayType(char c)
             {
                 return c is '*' or '~' or '%';
             }
@@ -357,21 +361,15 @@ namespace Garnet.server
                             return default;
                     }
 
-                    // We support arrays ('*'), RSEP3 sets ('~') and RESP3 maps ('%').
-                    // All are returned as arrays.
-                    // Returning other types will lead to unnecessary duplication of code,
-                    // as we need to support RESP2 arrays in all cases anyway.
                     else if (IsSupportedArrayType(c))
                     {
                         if (isScanOutput)
                         {
                             element = null;
                             len = 0;
-                            // Read the first two elements
-                            if (!RespReadUtils.TryReadUnsignedLengthHeader(out var outerArraySize, ref refPtr, end, c)
-                            // Read cursor value
-                             || !RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end)
-                                )
+                            // Try to read the array length and cursor value
+                            if (!RespReadUtils.TryReadUnsignedLengthHeader(out var outerArraySize, ref refPtr, end, c) ||
+                                !RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end))
                                 return default;
                         }
 


libs/server/Resp/RespServerSessionOutput.cs
```diff
@@ -0,0 +1,447 @@
...
+                    // Adjust number of bytes to copy, to MIN(space left on output buffer, bytes left to copy)
+                    int toCopy = bytesLeft;
+                    if (toCopy > destSpace)
+                        toCopy = destSpace;
+
+                    // Copy bytes to output buffer
+                    Buffer.MemoryCopy(src, dcurr, destSpace, toCopy);
+
+                    // Move cursor on output buffer and input memory, update bytes left
+                    dcurr += toCopy;
```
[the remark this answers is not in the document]
prvyk: This isn't something I've added. Is there something I'm missing to turn off in CodeQL?

badrishc:
This is a lot of refactoring and depending on .NET's force inlining capability. We need to make sure there are no performance regressions due to e.g. saturating the inline limits that .NET places internally.

Also, see the BDN results here and identify any regressions: https://github.com/microsoft/garnet/commit/ee395dc122413c6bbb15fc5b723c1f8d16ca63d6#comments

There seems to be general degradations all over the place.

libs/common/RespMemoryWriter.cs
```diff
@@ -398,6 +406,45 @@ public void WriteUtf8BulkString(ReadOnlySpan<char> chars)
...
+        /// If RESP2, write as Bulk String. If RESP3, write as Verbatim String with "txt" type.
+        /// </summary>
+        /// <param name="item"></param>
+        /// <param name="ext">String 3-letter type. If not supplied default is "txt"</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void WriteVerbatimString(scoped ReadOnlySpan<byte> item, scoped ReadOnlySpan<byte> ext = default)
+        {
+            if (resp3)
+            {
+                while (!RespWriteUtils.TryWriteVerbatimString(item, ext.IsEmpty ? "txt"u8 : ext, ref curr, end))
```
TalZaccai: Add this literal (`"txt"u8`) as a static readonly field.

libs/common/RespMemoryWriter.cs
```diff
@@ -398,6 +406,45 @@ public void WriteUtf8BulkString(ReadOnlySpan<char> chars)
                 ReallocateOutput();
         }
 
+        /// <summary>
+        /// Write Verbatim string to memory.
+        /// If RESP2, write as Bulk String. If RESP3, write as Verbatim String with "txt" type.
+        /// </summary>
+        /// <param name="item"></param>
```
TalZaccai: nit: `item` missing XML comment

libs/common/RespWriteUtils.cs
```diff
@@ -716,6 +728,33 @@ public static bool TryWriteEmptySet(ref byte* curr, byte* end)
...
         }
 
+        /// <summary>
+        /// Write verbatim string
+        /// </summary>
+        public static bool TryWriteVerbatimString(ReadOnlySpan<byte> str, ReadOnlySpan<byte> ext, ref byte* curr, byte* end)
+        {
+            Debug.Assert(ext.Length == 3);
+
+            var actualLength = 3 + 1 + str.Length;
```
TalZaccai, line 738: Better to add a comment here, something like - `"ext (3 bytes) + ':' (1 byte) + str"`

libs/common/RespWriteUtils.cs
```diff
@@ -716,6 +728,33 @@ public static bool TryWriteEmptySet(ref byte* curr, byte* end)
...
+            var itemDigits = NumUtils.CountDigits(actualLength);
+            var totalLen = 1 + itemDigits + 2 + actualLength + 2;
```
TalZaccai, line 744: Comment here too - `'=' (1 byte) + itemDigits + '\r\n' (2 bytes) + actualLength + '\r\n' (2 bytes)`

libs/server/Metrics/Slowlog/RespSlowlogCommands.cs
```diff
@@ -50,11 +50,9 @@ private bool NetworkSlowLogGet()
             int count = 10;
             if (parseState.Count == 1)
             {
-                if (!parseState.TryGetInt(0, out count))
+                if (!parseState.TryGetInt(0, out count) || (count < -1))
                 {
-                    while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
-                        SendAndReset();
-                    return true;
+                    return AbortWithErrorMessage(CmdStrings.RESP_ERR_COUNT_IS_OUT_OF_RANGE_N1);
```
TalZaccai, line 55: Shouldn't the original error should remain for `!parseState.TryGetInt(0, out count)`?

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -625,7 +625,7 @@ private void SortedSetRandomMember(ref ObjectInput input, ref GarnetObjectStoreO
 
                 if (withScores)
                 {
-                    writer.WriteDoubleBulkString(score);
+                    writer.WriteDoubleNumeric(score);
```
TalZaccai, line 640: I know that this is unrelated to this PR - but maybe you can take a look - 
In `RespWriteUtils.TryWriteDoubleNumeric` we're calculating `var itemDigits = NumUtils.CountDigits(bytesWritten);` but it's never used.

TalZaccai, line 640: I find this naming a bit confusing, since this may call either `TryWriteDoubleNumeric` or `TryWriteDoubleBulkString`. Should `WriteDoubleBulkString` even still be used?

libs/server/Resp/ACLCommands.cs
```diff
@@ -283,22 +274,18 @@ private bool NetworkAclWhoAmI()
...
-                Debug.Assert(aclAuthenticator.GetUserHandle()?.User != null);
+            var aclAuthenticator = (GarnetACLAuthenticator)_authenticator;
 
-                while (!RespWriteUtils.TryWriteSimpleString(aclAuthenticator.GetUserHandle().User.Name, ref dcurr, dend))
-                    SendAndReset();
-            }
+            // Return the name of the currently authenticated user.
+            Debug.Assert(aclAuthenticator.GetUserHandle()?.User != null);
+
+            WriteAsciiBulkString(aclAuthenticator.GetUserHandle().User.Name);
```
TalZaccai, line 282: Good catch, this should've always been a bulk string...

libs/server/Resp/ACLCommands.cs
```diff
@@ -120,22 +120,19 @@ private bool NetworkAclCat()
             // No additional args allowed
             if (parseState.Count != 0)
             {
-                while (!RespWriteUtils.TryWriteError($"ERR Unknown subcommand or wrong number of arguments for ACL CAT.", ref dcurr, dend))
-                    SendAndReset();
+                return AbortWithErrorMessage($"ERR Unknown subcommand or wrong number of arguments for ACL CAT.");
```
TalZaccai, line 117: While you're touching this file... Looks like the error messages here are wrong, they should all be `ERR wrong number of arguments for 'acl|XXX' command` (i.e. you can use `AbortWithWrongNumberOfArguments`)

libs/server/Resp/ArrayCommands.cs
```diff
@@ -372,11 +359,9 @@ private bool NetworkSCAN<TGarnetApi>(ref TGarnetApi storageApi)
                 return AbortWithWrongNumberOfArguments("SCAN");
 
             // Validate scan cursor
-            if (!parseState.TryGetLong(0, out var cursorFromInput))
+            if (!parseState.TryGetLong(0, out var cursorFromInput) || (cursorFromInput < 0))
```
TalZaccai, line 362: Another good catch!

libs/server/Resp/BasicCommands.cs
```diff
@@ -1053,22 +1037,18 @@ private bool NetworkCOMMAND_COUNT()
             // No additional args allowed
             if (parseState.Count != 0)
             {
-                var errorMsg = string.Format(CmdStrings.GenericErrWrongNumArgs, "COMMAND COUNT");
-                while (!RespWriteUtils.TryWriteError(errorMsg, ref dcurr, dend))
-                    SendAndReset();
+                return AbortWithWrongNumberOfArguments("COMMAND COUNT");
```
TalZaccai: nit: should be `"COMMAND|COUNT"`

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -437,21 +403,17 @@ private bool StringBitField<TGarnetApi>(ref TGarnetApi storageApi)
                         op = RespCommand.INCRBY;
                     else
                     {
-                        while (!RespWriteUtils.TryWriteError(
-                                   $"ERR Bitfield command {Encoding.ASCII.GetString(command)} not supported", ref dcurr,
-                                   dend))
-                            SendAndReset();
-                        return true;
+                        return AbortWithErrorMessage(
```
TalZaccai: nit: Should probably be a one-liner.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -245,9 +243,7 @@ private bool NetworkRENAME<TGarnetApi>(ref TGarnetApi storageApi)
             {
                 if (!parseState.GetArgSliceByRef(2).ReadOnlySpan.EqualsUpperCaseSpanIgnoringCase(CmdStrings.WITHETAG))
                 {
-                    while (!RespWriteUtils.TryWriteError($"ERR Unsupported option {parseState.GetString(2)}", ref dcurr, dend))
-                        SendAndReset();
-                    return true;
+                    return AbortWithErrorMessage($"ERR Unsupported option {parseState.GetString(2)}");
```
TalZaccai: Looks like this repeats itself a lot, might consider adding a string template in `CmdStrings`

libs/server/Resp/Objects/ListCommands.cs
```diff
@@ -84,11 +84,9 @@ private unsafe bool ListPop<TGarnetApi>(RespCommand command, ref TGarnetApi stor
             if (parseState.Count == 2)
             {
                 // Read count
-                if (!parseState.TryGetInt(1, out popCount))
+                if (!parseState.TryGetInt(1, out popCount) || (popCount < 0))
                 {
-                    while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER, ref dcurr, dend))
-                        SendAndReset();
-                    return true;
+                    return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_OUT_OF_RANGE);
```
TalZaccai, line 89: Should original error remain for `!parseState.TryGetInt(1, out popCount)`?

TalZaccai (changes requested):
Great work here! Thanks for spending time on this, looks good to me overall.

A main concern I have is that this is not really tested. We might want to start by adding a RESP3 test case wherever we call `ConnectionMultiplexer.Connect(TestUtils.GetConfig())` and set the protocol parameter in GetConfig. Thoughts?

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
prvyk: There are cases (especially in GEO code) where Redis outputs double as bulk string even in RESP3, which is silly, but it is what it is...

e.g.
https://redis.io/docs/latest/commands/geodist/

libs/server/Metrics/Slowlog/RespSlowlogCommands.cs
prvyk: In my testing, redis outputs 'ERR count should be greater than or equal to -1' even when it's not an integer at all.

prvyk:
> Great work here! Thanks for spending time on this, looks good to me overall.
> 
> A main concern I have is that this is not really tested. We might want to start by adding a RESP3 test case wherever we call `ConnectionMultiplexer.Connect(TestUtils.GetConfig())` and set the protocol parameter in GetConfig. Thoughts?

The question is what we test. Do we test that RESP3 output is working at all? Take a command with slightly differing output and ensure it's different? Or maybe test StackExchange is working in RESP3 mode?

There's a test for the first. Maybe we can join in testing the latter two by adding a parameter to certain testcases to set protocol version, and let StackExchange do the output test.

TalZaccai:
> > Great work here! Thanks for spending time on this, looks good to me overall.
> > A main concern I have is that this is not really tested. We might want to start by adding a RESP3 test case wherever we call `ConnectionMultiplexer.Connect(TestUtils.GetConfig())` and set the protocol parameter in GetConfig. Thoughts?
> 
> The question is what we test. Do we test that RESP3 output is working at all? Take a command with slightly differing output and ensure it's different? Or maybe test StackExchange is working in RESP3 mode?
> 
> There's a test for the first. Maybe we can join in testing the latter two by adding a parameter to certain testcases to set protocol version, and let StackExchange do the output test.

Yup I was more concerned with the latter for now, check that the client that we use does not break.

libs/server/Metrics/Slowlog/RespSlowlogCommands.cs
TalZaccai: Ugh, you have to love how consistent they are :)

libs/server/Resp/Objects/ListCommands.cs
prvyk: The ever-consistent Redis outputs this error even if argument is not an integer at all.

libs/server/Resp/KeyAdminCommands.cs
prvyk: Done.

libs/common/RespWriteUtils.cs
prvyk: Done.

libs/common/RespMemoryWriter.cs
prvyk: Done.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
prvyk: Removed it (must have gotten there from the equivalent bulk string method). Also, the formatter method used Utf8Formatter, when it could have used double.TryFormat since about NET8.

prvyk:
> > > Great work here! Thanks for spending time on this, looks good to me overall.
> > > A main concern I have is that this is not really tested. We might want to start by adding a RESP3 test case wherever we call `ConnectionMultiplexer.Connect(TestUtils.GetConfig())` and set the protocol parameter in GetConfig. Thoughts?
> > 
> > 
> > The question is what we test. Do we test that RESP3 output is working at all? Take a command with slightly differing output and ensure it's different? Or maybe test StackExchange is working in RESP3 mode?
> > There's a test for the first. Maybe we can join in testing the latter two by adding a parameter to certain testcases to set protocol version, and let StackExchange do the output test.
> 
> Yup I was more concerned with the latter for now, check that the client that we use does not break.

That turned out to be very useful. There were APIs which ran commands and read their output into variables. These commands could both output directly (so RESP3 change) and could be read, and the APIs was not able to read RESP3 output. The previous RESP3 PR had not accounted for this hack.

This was fixed with a mix of forcing some commands to RESP2 when we know the output is just read into variables (ZMPOP, ZRANGESTORE), and extending some reading code to handle RESP3 types and array-in-array (ProcessRespArrayOutput in Common.cs).

In https://github.com/microsoft/garnet/pull/1223/commits/515c35865213e49180d4603090bc78f87034a7b8 the entire testsuit was temporarily modified to run StackExchange via RESP3 and it passed (the two failures are sporadic and unrelated). This was later reverted to the default RESP2.

ZDIFF has a minor issue where ZDIFF z nx (z exists, nx does not) doesn't output z, but fixing it interfered with some tests so another time.

libs/server/Resp/ACLCommands.cs
prvyk: Done.

ACL CAT has different error message in reference so I kept it as is.

TalZaccai (approved):
LGTM! Thanks for addressing all my comments!

prvyk:
> LGTM! Thanks for addressing all my comments!

Two minor changes since then (following [06f6cd0](https://github.com/microsoft/garnet/pull/1223/commits/06f6cd02ecfb0668298a498fef47f26f431fb99e): #1241 showed that some APIs can/should read NULL and the underlying TryReadSignedLengthHeader wasn't updated for RESP3 - added some code for it, it doesn't conflict with that PR, nor with what the function already did (it already accounted for RESP2 NULL, and callers already checked length >= 0 when needed).

It also turned out that HELLO returned proto field as BulkString and not integer, and some client libraries don't like it (they check it to see HELLO 3 worked, and then they get "3", do "3" != 3 and panic). Fixed that. Also, id field was oddly hardcoded when it should be connection id, fixed that too.

TedHartMS (approved):

libs/common/RespWriteUtils.cs
```diff
@@ -704,6 +702,18 @@ public static bool TryWriteEmptyArray(ref byte* curr, byte* end)
...
 
+        /// <summary>
+        /// Write empty map
+        /// </summary>
+        public static bool TryWriteEmptyMap(ref byte* curr, byte* end)
+        {
+            if (4 > (int)(end - curr))
+                return false;
+
+            WriteBytes<uint>(ref curr, "%0\r\n"u8);
```
TalZaccai: This can be added to `RespStrings` maybe? Also the constants in `TryWriteZero` `TryWriteOne`?

libs/server/Resp/BasicCommands.cs
```diff
@@ -1594,29 +1569,20 @@ void ProcessHelloCommand(byte? respProtocolVersion, ReadOnlySpan<byte> username,
                     ("server", "redis"),
                     ("version", storeWrapper.redisProtocolVersion),
                     ("garnet_version", storeWrapper.version),
-                    ("proto", this.respProtocolVersion),
-                    ("id", 63),
+                    ("proto", (long)this.respProtocolVersion),
+                    ("id", Id),
```
TalZaccai, line 1573: @badrishc  was there a reason why this was hard-coded like this?

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -241,53 +241,84 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
                         }
 
                         // Get the number of elements
-                        if (!RespReadUtils.TryReadUnsignedArrayLength(out var arraySize, ref refPtr, outputPtr + outputSpan.Length))
+                        if (!RespReadUtils.TryReadUnsignedLengthHeader(out var arraySize, ref refPtr, end, c))
                             return default;
 
                         // Create the argslice[]
-                        elements = new ArgSlice[isScanOutput ? arraySize + 1 : arraySize];
+                        System.Collections.Generic.List<ArgSlice> elemlist = new(isScanOutput ? arraySize + 1 : arraySize);
```
TalZaccai: Why did we switch to a list here? This is less efficient.

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -241,53 +241,84 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                                if (RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end))
+                                {
+                                    elemlist.Add(new ArgSlice(element, len));
+                                }
                             }
                         }
+
+                        elements = [.. elemlist];
+                    }
+                    else if (c == ',')
```
TalZaccai: What is this new case for?

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -241,53 +241,84 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
 
-                    if (*refPtr == '-')
+                    if (c == '-')
                     {
-                        if (!RespReadUtils.TryReadErrorAsString(out error, ref refPtr, outputPtr + outputSpan.Length))
+                        if (!RespReadUtils.TryReadErrorAsString(out error, ref refPtr, end))
                             return default;
                     }
-                    else if (*refPtr == '*')
+                    else if ((c == '*') || (c == '~') || (c == '%'))
```
TalZaccai: We should probably have 3 different methods for these different types, we want to make sure that the type parsed is the expected one.

TalZaccai (changes requested):
Added a few more comments following the latest changes...

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: There are RESP3 cases where the modified output is array-in-array: a array for items and inside each item an array of 2 for key-value (Yes, they could have used a map and it doesn't make too much sense). The RESP2 equivalent is just one large array and that's what the calling code expects. So we spread every internal list into the large array and get a localized change. In that case, we don't know the required length in advance, so a list is used instead. We could have iterated twice to measure first and then allocate an array but that would also be inefficient.

example:

RESP3:
(Array length 2)
1) (Array length 2)
  1) "a"
  2) 1
2) (Array length 2)
  1) "b"
  2) 2
  
RESP2:
1) "a"
2) 1
3) "b"
4) 2
  
This code will always emit a single array with 4 elements ["a", 1, "b", 2] for both cases.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Some commands emit Doubles in RESP3. Doubles start with ',' in RESP3. e.g. ",1.2345\r\n". We read the result and pass it along much like a string.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: These all devolve to Array type in RESP2, and the caller can't make use of the extra type information since it will still have to keep supporting RESP2 and receive array in turn. If we say 'well, here we know it's really a Set', then we can do this for RESP3 too, no?

I'd love to be rid of this RESP reading hack entirely someday...

libs/common/RespWriteUtils.cs
prvyk: Done.

libs/server/Storage/Session/ObjectStore/Common.cs
TedHartMS: Please add the last three comments here as comments in the code

prvyk: Added.

libs/server/Resp/BasicCommands.cs
badrishc: No good reason, i just took what other resp servers were reporting here. it was unclear at the time what ID is supposed to mean.

libs/server/Storage/Session/ObjectStore/Common.cs
TalZaccai: Well you can tell if it's a nested array by looking at the first element. We don't want to degrade the performance of RESP2 array parsing as well as RESP3 non-nested array parsing.

libs/server/Storage/Session/ObjectStore/Common.cs
TalZaccai: I don't think I'm following... What does that have to do with parsing an array?

libs/server/Storage/Session/ObjectStore/Common.cs
TalZaccai: Probably some comment about which char prefix matches which type could also help the reader. Also, if we know the RESP version of the input, we should be able to call a simplified parsing method that only supports arrays for RESP2... Something to think about.

prvyk: Split parsing to RESP2 and RESP3 methods. The RESP2 method is the previous code with very minor changes.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Commands like ZINCRBY call the array parsing method despite not returning an array (hence the bulk string fallback). ZINCRBY returns double in RESP3, so function is updated to handle it.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done.

libs/server/Storage/Session/ObjectStore/Common.cs
TalZaccai: Shouldn't we fix the calls rather than support unexpected data? Plus if we support doubles this way, we should also support other data types like ints, booleans etc...

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                        if (!RespReadUtils.TryReadSignedLengthHeader(out var arraySize, ref refPtr, end, c))
+                            return default;
+
+                        if (arraySize < 0)
+                            return default;
+
+                        if (c == '%') // RESP3 Map
+                            arraySize *= 2;
+
+                        if (!isScanOutput && arraySize == 0)
```
TalZaccai, line 383: nit: can do this before the `if (c == '%')`

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+
+                            element = null;
+                            len = 0;
+                            // Read cursor value
+                            if (!RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end))
+                                return default;
+                        }
+
+                        // Get the number of elements
+                        if (!RespReadUtils.TryReadSignedLengthHeader(out var arraySize, ref refPtr, end, c))
```
TalZaccai, line 377: nit: can merge these 2 conditions with an `||`

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                        if (c == '%') // RESP3 Map
+                            arraySize *= 2;
+
+                        if (!isScanOutput && arraySize == 0)
+                            return [];
+
+                        // RESP3 command can output array-in-array cases.
+                        // We'll assume that if there's an array-in-array, it's always the first element
+                        // and that the shape is constant so as to calculate the required length in advance.
+                        var outerLen = 1;
```
TalZaccai: isn't this the innerLen?

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                        if (arraySize < 0)
+                            return default;
+
+                        if (c == '%') // RESP3 Map
+                            arraySize *= 2;
+
+                        if (!isScanOutput && arraySize == 0)
+                            return [];
+
+                        // RESP3 command can output array-in-array cases.
```
TalZaccai: Proposing the following comment instead:
```cs
// It is possible that the array elements consist of nested arrays.
// This code only supports nested arrays of a consistent dimension (i.e. not jagged), so we use the first element's dimension to infer the rest.
```

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+
+                        for (; i < elements.Length; i++)
+                        {
+                            element = null;
+                            len = 0;
+
+                            c = (char)*refPtr;
+
+                            // We still need to read the field to advance the pointer.
+                            if ((c == '*') || (c == '~') || (c == '%'))
```
TalZaccai: There needs to be a flag that specified that we're reading a nested array (based on the first element). And then we'll do something like:
```cs
if (isNestedArray)
{
   Debug.Assert((c == '*') || (c == '~') || (c == '%'));
   
   if (!RespReadUtils.TryReadUnsignedLengthHeader(out var len, ref refPtr, end, c))
      return default;

   Debug.Assert(len == innerLen)
}
```

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,151 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+
+                        if (!isScanOutput && arraySize == 0)
+                            return [];
+
+                        // RESP3 command can output array-in-array cases.
+                        // We'll assume that if there's an array-in-array, it's always the first element
+                        // and that the shape is constant so as to calculate the required length in advance.
+                        var outerLen = 1;
+                        c = (char)*refPtr;
+                        if ((c == '*') || (c == '~') || (c == '%'))
```
TalZaccai: Consider adding this to a method (e.g. `IsSupportedArrayType` or something) as it repeats itself in multiple locations.

TalZaccai (changes requested):
Added a few additional comments to the recent changes.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done (note the first item is read in advance so we exclude it. This never happens in scan output so we can just check != 0).

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: This means we do a useless *2 if %-1 comes along (no reason to, but still).

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done: Moved ZINCRBY processing to proper single output function, and also forced that api to RESP2 (it doesn't really output), so we don't need double processing code.

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done.

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -224,60 +224,78 @@ public unsafe GarnetStatus ObjectScan<TObjectContext>(GarnetObjectType objectTyp
...
+        }
+
+        /// <summary>
+        /// Converts an array of elements in RESP format to ArgSlice[] type
+        /// </summary>
+        /// <param name="outputFooter">The RESP format output object</param>
+        /// <param name="error">A description of the error, if there is any</param>
+        /// <param name="isScanOutput">True when the output comes from HSCAN, ZSCAN OR SSCAN command</param>
+        /// <returns></returns>
+        private unsafe ArgSlice[] __ProcessResp2ArrayOutput(GarnetObjectStoreOutput outputFooter, out string error, bool isScanOutput)
```
TalZaccai: naming: This is not a convention that we use in the code base, please rename to `ProcessResp2ArrayOutput` and `ProcessResp3ArrayOutput`

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,142 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                        return default;
+                    }
+
+                    if (c == '-')
+                    {
+                        if (!RespReadUtils.TryReadErrorAsString(out error, ref refPtr, end))
+                            return default;
+                    }
+
+                    // We support arrays ('*'), RSEP3 sets ('~') and RESP3 maps ('%').
```
TalZaccai: nit: comment can be moved to the `IsSupportedArrayType` declaration

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -287,7 +305,142 @@ unsafe ArgSlice[] ProcessRespArrayOutput(GarnetObjectStoreOutput output, out str
...
+                    {
+                        if (isScanOutput)
+                        {
+                            element = null;
+                            len = 0;
+                            // Read the first two elements
+                            if (!RespReadUtils.TryReadUnsignedLengthHeader(out var outerArraySize, ref refPtr, end, c)
+                            // Read cursor value
+                             || !RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end)
+                                )
```
TalZaccai: nit: spacing looks weird here.
```cs
// Try to read the array length and cursor value
if (!RespReadUtils.TryReadUnsignedLengthHeader(out var outerArraySize, ref refPtr, end, c) ||
    !RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr, end))
   return default;
```

TalZaccai (changes requested):
Couple of last nits (sorry...)

libs/server/Storage/Session/ObjectStore/Common.cs
prvyk: Done.

PR #1231, opened 2025-05-28
Validate Checkpoint Acquire

vazois:
Revisiting replication and fixing edge case as well as performing validation.

### Other improvements

* Add `ReplicaDisklessSyncFullSyncAofThreshold` configuration option
   Introduce a new configuration option for diskless replication, `ReplicaDisklessSyncFullSyncAofThreshold`, which sets a threshold for AOF replay size to determine when a full sync is required. The default value is set to "64m". The `NeedToFullSync` method is updated to incorporate this new threshold, and corresponding changes are made in `Options.cs`, `GarnetServerOptions.cs`, and various test files to ensure proper functionality and validation. The corresponding command line flag is `--repl-diskless-sync-full-sync-aof-threshold 64m`.

### DIFF
--- a/libs/cluster/Server/Replication/CheckpointStore.cs
+++ b/libs/cluster/Server/Replication/CheckpointStore.cs
@@ -85,7 +85,7 @@ namespace Garnet.cluster
                 {
                     if (!toDeletelogToken.Equals(logToken))
                     {
-                        logger.LogTrace("Deleting log token {toDeletelogToken}", toDeletelogToken);
+                        logger?.LogTrace("Deleting log token {toDeletelogToken}", toDeletelogToken);
                         ckptManager.DeleteLogCheckpoint(toDeletelogToken);
                     }
                 }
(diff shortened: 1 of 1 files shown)


libs/cluster/Server/Replication/CheckpointStore.cs
```diff
@@ -74,35 +75,29 @@ public void PurgeAllCheckpointsExceptEntry(CheckpointEntry entry = null)
...
-            {
-                if (!toDeleteIndexToken.Equals(indexToken))
+                // Delete index checkpoints
+                foreach (var toDeleteIndexToken in ckptManager.GetIndexCheckpointTokens())
                 {
-                    logger.LogTrace("Deleting index token {toDeleteIndexToken}", toDeleteIndexToken);
-                    ckptManager.DeleteIndexCheckpoint(toDeleteIndexToken);
+                    if (!toDeleteIndexToken.Equals(indexToken))
+                    {
+                        logger.LogTrace("Deleting index token {toDeleteIndexToken}", toDeleteIndexToken);
```
TalZaccai: `logger?.` x2

+ nit: `toDeleteLogToken`

PR #1236, opened 2025-06-01
ZUNION result must follow sorted set order invariant.

prvyk:
Sorted sets are sorted by their score in the ascending direction. ZUNION output does not current obey this property:

> zadd a 1 one 2 two
> zadd b 2 two 3 three
> zunion 2 a b

should be "one three two" but is "one two three" instead.

This PR fixes the output order for ZUNION to maintain the Sorted Set invariant. The simplest fix would be to just order the result on the parsing layer output. Yet I think the API should return a sensible result without requiring the caller to do an extra sort.

The change is a minimal change to keep the API shape while making it keep the ordering property if possible:
A) On NET8.0 we sort the Dictionary on output.
B) In NET9.0, we use OrderedDictionary.
C) The API is changed to return IDictionary instead of Dictionary to allow both results.


badrishc:
The ZUNION output type should perhaps be a `SortedSet<(double Score, byte[] Element)> sortedSet` so that users can run range queries on the returned object in addition to iterating over it?

prvyk:
> The ZUNION output type should perhaps be a `SortedSet<(double Score, byte[] Element)> sortedSet` so that users can run range queries on the returned object in addition to iterating over it?

I like that. That loses the ability to do result[key], but it's a reasonable exchange (The caller can still do result.Where(x => x.Element == "key")). Also allows dropping #if NET9_0_OR_GREATER.

TalZaccai:
I don't understand something here... Is this required by the protocol? Because I don't see that requirement in the ZUNION docs. 
Also, if we're implementing this logic for ZUNION, what about ZDIFF, ZINTER?

prvyk:
ZUNION docs:

"This command is similar to ZUNIONSTORE, but instead of storing the resulting sorted set, it is returned to the client."

https://redis.io/docs/latest/commands/zunion/

So the result is sorted set. Sorted sets are defined as "Sorted sets are sorted by their score in an ascending way. (etc.)"  (e.g. https://redis.io/docs/latest/commands/zadd/).

I guess we can read this as 'this is a sorted set, but we're returning it in a non sorted way _this_ time' but I don't think that's the intent. Also, the current situation is that "ZUNIONSTORE dest .." followed by "ZRANGE dest 0 -1" would return a different result to doing ZUNION on the sources directly.

>Also, if we're implementing this logic for ZUNION, what about ZDIFF, ZINTER?

Yes, these should have the same invariant too.

TalZaccai:
> ZUNION docs:
> 
> "This command is similar to ZUNIONSTORE, but instead of storing the resulting sorted set, it is returned to the client."
> 
> https://redis.io/docs/latest/commands/zunion/
> 
> So the result is sorted set. Sorted sets are defined as "Sorted sets are sorted by their score in an ascending way. (etc.)" (e.g. https://redis.io/docs/latest/commands/zadd/).
> 
> I guess we can read this as 'this is a sorted set, but we're returning it in a non sorted way _this_ time' but I don't think that's the intent. Also, the current situation is that "ZUNIONSTORE dest .." followed by "ZRANGE dest 0 -1" would return a different result to doing ZUNION on the sources directly.
> 
> > Also, if we're implementing this logic for ZUNION, what about ZDIFF, ZINTER?
> 
> Yes, these should have the same invariant too.

Alright, I'm good with it then. Let's just add ZDIFF & ZINTER to the PR. Verify that there aren't any other command that I might've missed ;)

badrishc:
> > The ZUNION output type should perhaps be a `SortedSet<(double Score, byte[] Element)> sortedSet` so that users can run range queries on the returned object in addition to iterating over it?
> 
> I like that. That loses the ability to do result[key], but it's a reasonable exchange (The caller can still do result.Where(x => x.Element == "key")). Also allows dropping #if NET9_0_OR_GREATER.

Ideally we would return an object on which we can do both range and index based access, exactly like our internal SortedSetObject, except that our internal object exposes a bunch of unnecessary things. What do you all think.

prvyk:
> > > The ZUNION output type should perhaps be a `SortedSet<(double Score, byte[] Element)> sortedSet` so that users can run range queries on the returned object in addition to iterating over it?
> > 
> > 
> > I like that. That loses the ability to do result[key], but it's a reasonable exchange (The caller can still do result.Where(x => x.Element == "key")). Also allows dropping #if NET9_0_OR_GREATER.
> 
> Ideally we would return an object on which we can do both range and index based access, exactly like our internal SortedSetObject, except that our internal object exposes a bunch of unnecessary things. What do you all think.

If we want to allow that I can think of two options:

A) Create a custom class taking code from SortedSetObject** (we wouldn't want to expose that). Users will need to get to the SortedSet to do the enumeration.
B) Use .NET 9s IAlternateEqualityComparer to allow the index-based access. e.g. A SortedDictionary<(double Score, byte[] Element), double Score> with an alternative equality for just Element. Users will need GetAlternateLookup() to do that lookup.

** Ideally, SortedSetObject itself would have a different custom datastructure as the current one is wasteful, but that's for another time.

PaulusParssinen:
> [...]
>
> If we want to allow that I can think of two options:
> 
> A) Create a custom class taking code from SortedSetObject** (we wouldn't want to expose that). Users will need to get to the SortedSet to do the enumeration. B) Use .NET 9s IAlternateEqualityComparer to allow the index-based access. e.g. A SortedDictionary<(double Score, byte[] Element), double Score> with an alternative equality for just Element. Users will need GetAlternateLookup() to do that lookup.
> 
> ** Ideally, SortedSetObject itself would have a different custom datastructure as the current one is wasteful, but that's for another time.

I agree with these in general and you may remember me having stab at this a while back https://github.com/microsoft/garnet/pull/1100.  Feel free to borrow from it.

However I think it's better for this to wait until we drop .NET 8 TFM, as it will get really perf regression -prone trying to #ifdef it away with overload resolution getting tricky (a bit more now with the expiration functionality on the sorted set being implemented as another dictionary lookup _IIRC_). Custom data-structure for the SortedSet could be the way forward regardless as you said.

test/Garnet.test/RespSortedSetTests.cs
```diff
@@ -1669,9 +1669,9 @@ public void CanDoZInterStoreWithSE()
         }
 
         [Test]
```
TalZaccai, line 1683: Since there are no test cases covering the ordering of ZINTER / ZDIFF's outputs,
I'd say add a couple of elements to the sets in `CanDoZInterWithSE` & `CanDoZDiff` so that we can verify the ordering...

TalZaccai (changes requested):
LGTM, one comment about testing

test/Garnet.test/RespSortedSetTests.cs
prvyk: I found a case where ZINTER broke the order invariant in current Garnet but not with the fix, so added new testcase.

For ZDIFF, I added another element to CanDoZDiff testcase. Though the way ZDIFF and Dictionary are implemented makes it keep the order in practice (I couldn't find a way to break it in current), so I don't know how useful the test really is.

ZDIFF starts with the first sorted set and then only removes elements, and in practice Dictionary keeps the original insertion order and removing elements alone apparently still keeps that order. One would need to Insert/Update elements afterwards to break it, and ZDIFF doesn't do that.

TalZaccai (approved):

PR #1241, opened 2025-06-06
Fix LPOP & HGET methods in custom procedure

Mimi8298:
Fix two issues:
- LPOP throws an ArgumentNullException when the list doesn't exist
- HGET throws an RespParsingException when the hash field doesn't exist


libs/server/Storage/Session/ObjectStore/ListOps.cs
```diff
@@ -94,7 +94,7 @@ public GarnetStatus ListPop<TObjectContext>(ArgSlice key, ListOperation lop, ref
            where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
         {
             var status = ListPop(key, 1, lop, ref objectStoreContext, out var elements);
-            element = elements.FirstOrDefault();
+            element = elements?.FirstOrDefault() ?? default;
```
TalZaccai: probably better to do: 
```cs
element = status == GarnetStatus.NOTFOUND ? default : elements.FirstOrDefault();
```

test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -222,6 +222,47 @@ public override unsafe void Main<TGarnetApi>(TGarnetApi garnetApi, ref CustomPro
         }
     }
 
+    public class ListPopCommandProc : CustomTransactionProcedure
```
TalZaccai: I think that instead of creating new procedures for these two specific commands, you can add the api calls to `TestProcedureHash` and `TestProcedureLists`

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -516,7 +516,7 @@ unsafe ArgSlice ProcessRespSingleTokenOutput(GarnetObjectStoreOutput output)
                 {
                     var refPtr = outputPtr;
 
-                    if (!RespReadUtils.TryReadPtrWithLengthHeader(ref element, ref len, ref refPtr,
+                    if (!RespReadUtils.TryReadPtrWithSignedLengthHeader(ref element, ref len, ref refPtr,
```
TalZaccai, line 519: This will return true if we're parsing a null value, and len will be -1. We shouldn't return an ArgSlice with a negative length, we should return a default in that case.

TalZaccai (changes requested):
Thanks for your contribution! Added a few comments.

test/Garnet.test/RespCustomCommandTests.cs
Mimi8298: I moved the tests in these procedures. Is it okay?

libs/server/Storage/Session/ObjectStore/Common.cs
```diff
@@ -516,9 +516,12 @@ unsafe ArgSlice ProcessRespSingleTokenOutput(GarnetObjectStoreOutput output)
...
                             outputPtr + outputSpan.Length))
                         return default;
+                    if (len < 0)
```
TalZaccai: nit: 
```cs
if (!RespReadUtils.TryReadPtrWithSignedLengthHeader(ref element, ref len, ref refPtr, 
   outputPtr + outputSpan.Length) || len < 0)
   return default;
```
Just a little more compact.

test/Garnet.test/RespCustomCommandTests.cs
TalZaccai: Way better! Thanks :)

libs/server/Storage/Session/ObjectStore/Common.cs
Mimi8298: Done 😀

PR #1242, opened 2025-06-06
Reduce allocation in random set index picking

PaulusParssinen:
Casting returning type to `Span<T>` doesn't remove the backing heap allocation. We have to hand the responsibility of allocating the buffer to the caller. Extracted from https://github.com/microsoft/garnet/pull/1100

We can sort get rid of these constants and calculate the threshold for the constant sized stackalloc on the fly with [`SpanOwner<T>`](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/high-performance/spanowner) like utility if this pattern gets more common in Garnet. I'm also using a bit conservative 1K threshold, could be higher if we wish so.


libs/server/Objects/Set/SetObjectImpl.cs
```diff
@@ -163,15 +163,18 @@ private void SetRandomMember(ref ObjectInput input, ref GarnetObjectStoreOutput
...
-            Span<int> indexes = default;
-
             if (count > 0)
             {
                 // Return an array of distinct elements
                 var countParameter = count > Set.Count ? Set.Count : count;
 
                 // The order of fields in the reply is not truly random
-                indexes = RandomUtils.PickKRandomIndexes(Set.Count, countParameter, seed);
+                const int StackallocThreshold = 256;
```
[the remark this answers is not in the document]
PaulusParssinen: Usually we want to have call-site specific granularity for the stack allocation thresholds depending on how "deep" into callstack we could be. Here we could probably extract it to be shared _somewhere_ but where would it really make sense?

We do have duplication in at method scope level though now that I look at it.

libs/common/RandomUtils.cs
```diff
@@ -18,20 +18,24 @@ public static class RandomUtils
         /// Pick k indexes from a collection of n items
```
TedHartMS: There's no longer a "k"

libs/common/RandomUtils.cs
```diff
@@ -43,46 +47,39 @@ public static Span<int> PickKRandomIndexes(int n, int k, int seed, bool distinct
...
                 }
             }
-
-            return result;
         }
 
-        private static Span<int> PickKRandomDistinctIndexesWithShuffle(int n, int k, int seed)
+        private static void PickKRandomDistinctIndexesWithShuffle(int n, Span<int> indexes, int seed)
         {
+            const int StackallocThreshold = 256;
```
TedHartMS: This is copied around a few places; how about centralizing it as a RandomUtils const

PaulusParssinen: `RandomUtils` makes sense as a place for it, thanks!

libs/common/RandomUtils.cs
```diff
@@ -11,27 +11,36 @@ namespace Garnet.common
...
+        /// </summary>
+        public const int IndexStackallocThreshold = 1024 / sizeof(int);
+
         // Threshold of k / n for choosing the random picking algorithm
-        private static readonly double KOverNThreshold = 0.1;
+        private const double KOverNThreshold = 0.1;
 
         /// <summary>
-        /// Pick k indexes from a collection of n items
+        /// Pick indexes from a collection of n items
```
TalZaccai: nit: might be good to alter the comment into something like ```/// Pick k indexes from a collection of n items into a given span of length k```

TalZaccai: Just altered it myself, hope it's ok :)

PaulusParssinen: That's fine, thanks!

PR #1244, opened 2025-06-09
Custom command name matching should be case-insensitive just like normal commands.

prvyk:
Following registration of a custom command, the command matching is case-sensitive, when it better be case-insensitive like normal commands.


libs/server/Custom/ExpandableMap.cs
```diff
@@ -283,6 +283,33 @@ internal static class ExpandableMapExtensions
         internal static bool MatchCommand<T>(this ExpandableMap<T> eMap, ReadOnlySpan<byte> cmd, out T value)
             where T : ICustomCommand
         {
+            // We need our own copy of EqualsUpperCaseSpanIgnoringCase, since the normal version
+            // would assert on non-ascii chars like the dot in 'json.get'.
+            static bool EqualsUpperCaseSpanIgnoringCase(ReadOnlySpan<byte> cmd, byte[] Name)
```
TalZaccai: This should probably be another helper method in `AsciiUtils` with an XML comment explaining that this version of the method supports all ASCII characters (could actually just be a flag in the original method).

TalZaccai: naming: param name should start with a lower-case letter (`Name`)

libs/server/Custom/ExpandableMap.cs
```diff
@@ -283,6 +283,33 @@ internal static class ExpandableMapExtensions
...
+
+                if (cmd.SequenceEqual(new ReadOnlySpan<byte>(Name)))
+                    return true;
+
+                for (var j = 0; j < cmd.Length; j++)
+                {
+                    var b1 = cmd[j];
+                    var b2 = Name[j];
+
+                    if ((b1 != b2) && ((b2 < 65) || (b2 > 90) || (b1 - 32 != b2)))
```
TalZaccai: For readability - either use the char literals or add a comment with the ASCII characters specified here.

libs/server/Custom/ExpandableMap.cs
```diff
@@ -295,7 +322,7 @@ internal static bool MatchCommand<T>(this ExpandableMap<T> eMap, ReadOnlySpan<by
             for (var i = 0; i < mapSize; i++)
             {
                 var currCmd = map[i];
-                if (currCmd != null && cmd.SequenceEqual(new ReadOnlySpan<byte>(currCmd.Name)))
+                if (EqualsUpperCaseSpanIgnoringCase(cmd, currCmd?.Name))
```
TalZaccai: No need to call the method if `currCmd == null`, you can leave that as part of the condition.

test/Garnet.test/RespCustomCommandTests.cs
```diff
@@ -365,13 +365,13 @@ public void CustomCommandTest2()
             // This conditional set should pass (prefix is greater)
             // New value is larger than existing
             var newValue4 = "foolargervalue4";
-            db.Execute("SETWPIFPGT", key, newValue4, BitConverter.GetBytes((long)4));
+            db.Execute("setwpifpgt", key, newValue4, BitConverter.GetBytes((long)4));
```
TalZaccai: These tests pass in main without the fix (when changing to the lower-case command), can you please check and see if there are any tests that we are missing that would've caught this? Thanks!

TalZaccai (changes requested):
Thanks for catching! Added a few comments.

libs/server/Custom/ExpandableMap.cs
prvyk: Used char literals.

test/Garnet.test/RespCustomCommandTests.cs
prvyk: Oh. I see. I didn't diagnose this right. What happens is this:

* I noticed 'json.get' fails.
* Changing the comparison does fix this, but this is unnecessary, because (I didn't check earlier) before custom commands are parsed, RespServerSession calls MakeUpperCase() first. So all comparisons should be against uppercase.
* So making the comparison later as case-sensitive isn't as bad as I though. 
  So why did 'json.get' fail?
  
* Because MakeUpperCase() aborts on the dot character, with the result 'json.get' is converted to 'JSON.get' and not 'JSON.GET'.

So the smaller fix would be to change MakeUpperCase() to allow dots.

Suggested new fix is this:
A) MakeUpperCase() searches for the command (it searches for first byte which is 'A' or above).
B) It tests every character. Earlier it aborted on every character below 'A' or above 'z'. Now we'll abort on whitespace or below chars or chars above 'z'.

Since commands are normally whitespace separated, there should be no functional change beyond accepting custom commands like 'json.get' or even 'now.do123!' in lowercase too.

libs/server/Custom/ExpandableMap.cs
prvyk: Resolving since new fix just changes MakeUpperCase().

test/Garnet.test/RespCustomCommandTests.cs
TalZaccai: Thanks for diagnosing this! Much appreciated!

TalZaccai (approved):

PR #1247, opened 2025-06-11
Implementing ScratchAllocationManager

TalZaccai:
`ScratchAllocationManager` is similar to `ScratchBufferManager`, but instead of creating a new buffer and copying the previous data when the buffer is full, it instead keeps a stack of the previous buffers and does not copy the previous data.
By doing so, `ScratchAllocationManager` ensures that any previously created `ArgSlice`s that have not been explicitly re-winded do not point to GCed memory.
The reason that both managers are currently necessary, is that `ScratchBufferManager` supports getting all the data in a continuous chunk of memory, where `ScratchAllocationManager` does not.


libs/server/ArgSlice/ScratchAllocationManager.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        void ExpandScratchBufferIfNeeded(int requiredLength)
+        {
+            if (currScratchBuffer.IsDefault || requiredLength > currScratchBuffer.Length - currScratchBuffer.scratchBufferOffset)
+                ExpandScratchBuffer(requiredLength);
+        }
+
+        void ExpandScratchBuffer(int requiredLength)
+        {
+            ScratchBuffer newScratchBuffer;
```
badrishc: since the buffer is meant to be used single threaded, there is no need to waste stack space on a local newScratchBuffer.

libs/server/ArgSlice/ScratchAllocationManager.cs
```diff
@@ -0,0 +1,292 @@
...
+                newScratchBuffer = backupScratchBuffer;
+                backupScratchBuffer = default;
+            }
+            // If we cannot use the backup buffer, we'll allocate a new buffer
+            else
+            {
+                // Length of new buffer is:
+                // If there is no current buffer - the closest power of 2 to the data length that is at least 64
+                // If there is a current buffer - the max between the closest power of 2 to the data length and twice the size of the current buffer.
+                var newLength = currScratchBuffer.IsDefault
```
badrishc: refactor into a helper: `currScratchBuffer = ScratchBuffer.Create(int requiredLength, int previousLength)`

test/Garnet.test/ScratchAllocationManagerTests.cs
```diff
@@ -0,0 +1,87 @@
...
...mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.\r\n\r\nLorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.\r\n\r\nLorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos.";
```
badrishc: string s = new string('a', size);

libs/server/ArgSlice/ScratchAllocationManager.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        ScratchBuffer currScratchBuffer;
+
+        readonly RefStack<ScratchBuffer> previousScratchBuffers = new();
+
+        // Max size of previously allocated unused buffer that we can reuse as necessary
+        readonly int backupScratchBufferMaxSize;
+
+        // A previously allocated unused buffer that we can reuse as necessary
+        ScratchBuffer backupScratchBuffer;
```
badrishc: `backupScratchBuffer` should not be necessary to maintain. On Reset, we can directly replace `currScratchBuffer` with largest buffer in the stack that is <= a configured `MaxInitialCapacity`.

libs/common/RefStack.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Garnet.common
+{
+    /// <summary>
+    /// Stack implementation supporting data access by reference
+    /// Note: this implementation is not thread-safe
+    /// </summary>
+    public class RefStack<T> where T : struct
+    {
+        // Default initial capacity of stack buffer
+        const int DefaultInitialCapacity = 64;
```
badrishc: too large, maybe should start with 1 or 2 and pay for doubling as we grow?

libs/common/RefStack.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Pop top item in stack
+        /// </summary>
+        /// <returns>Top item in stack</returns>
+        public T Pop()
+        {
+            if (count == 0) throw new InvalidOperationException("Stack contains no elements.");
+
```
TedHartMS: "throw" (and elsewhere, "return") should go on its own line

or if you really want a one-liner, 
            return count > 0 ? buffer[--count] : throw new InvalidOperationException("Stack contains no elements.");
:)  (doesn't work with "ref", though)

badrishc (approved):

TedHartMS (approved):

PR #1248, opened 2025-06-11
Avoid memory allocation for hash object operations wherever possible

Mimi8298:
Currently, most operations on Hash Objects allocate a `key` value to lookup the stored value. The aim here is to:
- avoid all memory allocations when the value already exists in the object (>= net.9)
- reuse the `key` instance for expireTime (>= net.9)
- avoid allocating a new `value` array when replacing a value of the same size

### DIFF
--- a/libs/server/Objects/Hash/HashObjectImpl.cs
+++ b/libs/server/Objects/Hash/HashObjectImpl.cs
@@ -135,11 +135,15 @@ namespace Garnet.server
                 if (countParameter > 0 && countParameter > count)
                     countParameter = count;
 
-                var absCount = Math.Abs(countParameter);
-                var indexes = RandomUtils.PickKRandomIndexes(count, absCount, seed, countParameter > 0);
+                var indexCount = Math.Abs(countParameter);
+
+                var indexes = indexCount <= RandomUtils.IndexStackallocThreshold ?
+                    stackalloc int[RandomUtils.IndexStackallocThreshold].Slice(0, indexCount) : new int[indexCount];
+
+                RandomUtils.PickKRandomIndexes(count, indexes, seed, countParameter > 0);
 
                 // Write the size of the array reply
-                writer.WriteArrayLength(withValues && (respProtocolVersion == 2) ? absCount * 2 : absCount);
+                writer.WriteArrayLength(withValues && (respProtocolVersion == 2) ? indexCount * 2 : indexCount);
 
                 foreach (var index in indexes)
                 {
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04, windows-2022, windows-2019 ]
+        os: [ ubuntu-22.04, ubuntu-24.04, windows-2022, windows-2025 ]
         framework: [ 'net8.0', 'net9.0' ]
         configuration: [ 'Debug', 'Release' ]
         test: [ 'Garnet.test', 'Garnet.test.cluster' ]
@@ -56,7 +56,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04, windows-2022, windows-2019 ]
+        os: [ ubuntu-22.04, ubuntu-24.04, windows-2022, windows-2025 ]
         framework: [ 'net8.0', 'net9.0' ]
         configuration: [ 'Debug', 'Release' ]
     steps:
--- a/benchmark/BDN.benchmark/Embedded/EmbeddedNetworkHandler.cs
+++ b/benchmark/BDN.benchmark/Embedded/EmbeddedNetworkHandler.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+#nullable disable
+
 using System.Diagnostics;
 using Garnet.common;
 using Garnet.networking;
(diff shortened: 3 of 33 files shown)


libs/common/RespMemoryWriter.cs
```diff
@@ -261,7 +261,7 @@ public void WriteInt64(long value)
         /// Write integer from bytes to memory.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void WriteIntegerFromBytes(byte[] resultBytes)
+        public void WriteIntegerFromBytes(ReadOnlySpan<byte> resultBytes)
```
prvyk: It would be nice if the argument was scoped. Makes it useful in more scenarios.

Mimi8298: Ah right! it's fixed now

PaulusParssinen:
Didn't think of using type alias for the key argument, makes this much cleaner than [my previous attempt](https://github.com/microsoft/garnet/pull/1105) 👍

Mimi8298:
@PaulusParssinen After the PR merge, maybe you can also commit your multi lookup optimization?

libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -19,7 +19,12 @@ private void HashGet(ref ObjectInput input, ref GarnetObjectStoreOutput output,
         {
             using var writer = new RespMemoryWriter(respProtocolVersion, ref output.SpanByteAndMemory);
 
-            var key = input.parseState.GetArgSliceByRef(0).SpanByte.ToByteArray();
+            var key = input.parseState.GetArgSliceByRef(0).SpanByte
```
TalZaccai: What about something like this to improve code readability? 

```cs
        private static ByteSpan GetParseStateArgumentByteSpan(ref ObjectInput input, int idx)
        {
            return input.parseState.GetArgSliceByRef(idx).SpanByte
#if NET9_0_OR_GREATER
                .AsReadOnlySpan();
#else
                .ToByteArray();
#endif
        }
```

libs/server/Objects/Hash/HashObject.cs
```diff
@@ -387,20 +416,28 @@ private void DeleteExpiredItemsWorker()
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private bool TryGetValue(byte[] key, out byte[] value)
+        private bool TryGetValue(ByteSpan key, out byte[] value)
         {
             value = default;
             if (IsExpired(key))
             {
                 return false;
             }
+#if NET9_0_OR_GREATER
```
TalZaccai, line 426: For better readability, common hash functionalities like add, remove, lookup etc can be extracted to helper methods.

TalZaccai (changes requested):
Functionally looks good to me, perhaps needs some slight tweaks to make code more readable. LMK what you think.

libs/server/Objects/Hash/HashObjectImpl.cs
Mimi8298: Done!

libs/server/Objects/Hash/HashObject.cs
Mimi8298: I'm having trouble seeing what you mean. In my view the Add, Remove, ContainsKey, etc. methods are already helpers for the hash object impl. I don't really see where additional helpers could be added (especially as they may only be used once)

Mimi8298: Ok I think I have a better solution, add extensions for dictionaries with byte[] like key.
```csharp
#if NET9_0_OR_GREATER
public static class ByteDictionaryExtensions
{
    private static Dictionary<byte[], TValue>.AlternateLookup<ReadOnlySpan<byte>> GetLookup<TValue>(this Dictionary<byte[], TValue> dictionary) => dictionary.GetAlternateLookup<ReadOnlySpan<byte>>();
    public static bool TryAdd<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key, [NotNullWhen(true)] TValue value) => GetLookup(dictionary).TryAdd(key, value);
    public static bool TryGetValue<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key, [NotNullWhen(true)] out TValue value) => GetLookup(dictionary).TryGetValue(key, out value);
    public static bool Remove<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key, [NotNullWhen(true)] out byte[] keyInstance, [NotNullWhen(true)] out TValue value) => GetLookup(dictionary).Remove(key, out keyInstance, out value);
    public static bool Remove<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key, [NotNullWhen(true)] out TValue value) => GetLookup(dictionary).Remove(key, out _, out value);
    public static bool ContainsKey<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key) => GetLookup(dictionary).ContainsKey(key);
    public static void Upsert<TValue>(this Dictionary<byte[], TValue> dictionary, ReadOnlySpan<byte> key, TValue value)
    {
        var lookup = GetLookup(dictionary);
        lookup[key] = value;
    }
}
#endif
```

It avoids the need for additional fields and can remove most defines. With the future [extension members](https://devblogs.microsoft.com/dotnet/csharp-exploring-extension-members/), we could add index extensions and have no more define.
Creating a new lookup shouldn't be very costly (internally, `GetAlternateLookup` checks whether the `comparer` is compatible and returns the structure).

Let me know if you want me to add this :)

TalZaccai: Not exactly what I meant, this is a cool idea though. However we'd probably need to do some benchmarking to make sure there's no perf hit.
I think the code can go in as-is and we can think of other improvements later.

Mimi8298: Okay!

PR #1256, opened 2025-06-17
Tighten up module loading and custom command registration

prvyk:
Garnet documentation says that custom command registration requires defining [ExtensionBinPaths](https://microsoft.github.io/garnet/docs/dev/custom-commands). 

This isn't what the code actually does. If no path is defined, a user can give an absolute path and the module/command will be loaded just fine if user has admin permissions or ACL is not set up.

It's a dangerous ability since user can load code to run under the server's context, and therefore I believe should be more restricted.

This PR:

* Adds a check to ensure module loading is disabled if paths are not defined. The loadmodulecs command line option is not subject to this restriction (it doesn't add much to ask a user to specify the path twice, once with loadmodulecs and once with extension-bin-path option).

* The loadmodulecs option did not check module signature, which seems counterintuitive and unsafe. The option text doesn't mention this, there's a switch to enable loading unsigned modules so the expectation is opposite (and no way to reenable the check), and Garnet should have a safe default (it's possible the module file is updated separately under the user's nose so the check is useful).

* Add supports to redis's enable-module-command directive for compatibility and as extra security. Add to tests too. <br />
It works same way as enable-debug-command: Access to commands MODULE and REGISTERCS is limited by connection type. Default is 'no' (disable the commands entirely), and other options are 'local' (only for local connections) or 'yes' (no limit except by ExtensionBinPath).

* Since I was already touching the same file, add support for DEBUG LOG for testing use.

### DIFF
--- a/.github/workflows/ci-bdnbenchmark.yml
+++ b/.github/workflows/ci-bdnbenchmark.yml
@@ -64,7 +64,14 @@ jobs:
         run: ./test/BDNPerfTests/run_bdnperftest.ps1 ${{ matrix.test }} ${{ matrix.framework }}
         shell: pwsh
         continue-on-error: false
-   
+  
+      - name: Random pause between tasks so multiple inserts don't run at the same time
+        run: |
+          $delay = Get-Random -Minimum 1 -Maximum 120
+          Write-Host "Sleeping for $delay seconds..."
+          Start-Sleep -Seconds $delay
+        shell: pwsh
+
       - name: Upload test results to artifacts
         uses: actions/upload-artifact@v4
         with:
--- a/Version.props
+++ b/Version.props
@@ -1,6 +1,6 @@
 <Project>
 	<!-- VersionPrefix property for builds and packages -->
 	<PropertyGroup>
-		<VersionPrefix>1.0.73</VersionPrefix>
+		<VersionPrefix>1.0.74</VersionPrefix>
 	</PropertyGroup>
 </Project>
--- a/libs/server/Databases/DatabaseManagerBase.cs
+++ b/libs/server/Databases/DatabaseManagerBase.cs
@@ -740,7 +740,7 @@ namespace Garnet.server
             }
 
             var scanFrom = StoreWrapper.objectStore.Log.ReadOnlyAddress;
-            var scanUntil = StoreWrapper.store.Log.TailAddress;
+            var scanUntil = StoreWrapper.objectStore.Log.TailAddress;
             (var deletedCount, var totalCount) = db.ObjectStoreExpiredKeyDeletionDbStorageSession.ObjectStoreExpiredKeyDeletionScan(scanFrom, scanUntil);
             Logger?.LogDebug("Object Store - Deleted {deletedCount} keys out {totalCount} records in range {scanFrom} to {scanUntil} for DB {id}", deletedCount, totalCount, scanFrom, scanUntil, db.Id);
 
--- a/libs/storage/Tsavorite/cs/src/core/ClientSession/ClientSession.cs
+++ b/libs/storage/Tsavorite/cs/src/core/ClientSession/ClientSession.cs
@@ -118,8 +118,7 @@ namespace Tsavorite.core
             // By the time Dispose is called, we should have no outstanding locks, so can use the BasicContext's sessionFunctions.
             _ = CompletePending(bContext.sessionFunctions, true);
 
-            if (store.RevivificationManager.IsEnabled)
-                MergeRevivificationStatsTo(ref store.RevivificationManager.stats, reset: true);
+            store.DisposeClientSession(ID);
         }
 
         /// <summary>
--- a/libs/storage/Tsavorite/cs/src/core/ClientSession/ManageClientSessions.cs
+++ b/libs/storage/Tsavorite/cs/src/core/ClientSession/ManageClientSessions.cs
@@ -43,6 +43,27 @@ namespace Tsavorite.core
         }
 
         /// <summary>
+        /// Dispose session with Tsavorite
+        /// </summary>
+        /// <param name="sessionID"></param>
+        /// <returns></returns>
+        internal void DisposeClientSession(int sessionID)
+        {
+            if (_activeSessions != null)
+            {
+                lock (_activeSessions)
+                {
+                    if (_activeSessions.TryGetValue(sessionID, out SessionInfo sessionInfo))
+                    {
+                        var session = sessionInfo.session;
+                        session.MergeRevivificationStatsTo(ref RevivificationManager.stats, reset: true);
+                        _ = _activeSessions.Remove(sessionID);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
         /// Dumps the revivification stats to a string.
         /// </summary>
         public string DumpRevivificationStats()


TalZaccai:
If no allowed bin paths are specified, no user should be able to register anything, so it is in fact disabled by default.
Which user should be allowed to run a specific command should be defined in the ACL.
I don't understand why we need this extra configuration parameter...

prvyk:
> If no allowed bin paths are specified, no user should be able to register anything, so it is in fact disabled by default. 

Ah, that's the intent? Because this doesn't work. Just use an absolute path for the module command (when this allowed bin path isn't defined at all) and Garnet will load anything.

> I don't understand why we need this extra configuration parameter... Which user should be allowed to run a specific command should be defined in the ACL.

Because the other mechanism didn't work and I wasn't sure if this was intentional. Also, ACL is disabled by default. The current situation is that module loading is enabled by default so long as the user uses an absolute path. (Also the extra parameter is used in redis and it's a bit different type of blocking)

Can I pivot this part to fixing the bin path check? My suggestion:

1) Fix the path check to disable anything when path is not defined, unless --loadmodulecs is used.
2) Keep the new 'check signatures for loadmodulecs option'.
3) Keep the new parameter for redis compat, but change it to allow by default and hide it.

TalZaccai:
> > If no allowed bin paths are specified, no user should be able to register anything, so it is in fact disabled by default.
> 
> Ah, that's the intent? Because this doesn't work. Just use an absolute path for the module command (when this allowed bin path isn't defined at all) and Garnet will load anything.
> 
> > I don't understand why we need this extra configuration parameter... Which user should be allowed to run a specific command should be defined in the ACL.
> 
> Because the other mechanism didn't work and I wasn't sure if this was intentional. Also, ACL is disabled by default. The current situation is that module loading is enabled by default so long as the user uses an absolute path. (Also the extra parameter is used in redis and it's a bit different type of blocking)
> 
> Can I pivot this part to fixing the bin path check? My suggestion:
> 
> 1. Fix the path check to disable anything when path is not defined, unless --loadmodulecs is used.
> 2. Keep the new 'check signatures for loadmodulecs option'.
> 3. Keep the new parameter for redis compat, but change it to allow by default and hide it.

Yeah if that's the case that's a very bad oversight on our part...
From a quick look it seems like in ModuleUtils.LoadAssemblies you need to return an error even when `allowedExtensionPaths` is null. (so ```if (allowedExtensionPaths == null || binaryFiles.Any(f =>...))```)

prvyk:
> > > If no allowed bin paths are specified, no user should be able to register anything, so it is in fact disabled by default.
> > Can I pivot this part to fixing the bin path check? My suggestion:
> > 
> > 1. Fix the path check to disable anything when path is not defined, unless --loadmodulecs is used.
> > 2. Keep the new 'check signatures for loadmodulecs option'.
> > 3. Keep the new parameter for redis compat, but change it to allow by default and hide it.
> 
> Yeah if that's the case that's a very bad oversight on our part... From a quick look it seems like in ModuleUtils.LoadAssemblies you need to return an error even when `allowedExtensionPaths` is null. (so `if (allowedExtensionPaths == null || binaryFiles.Any(f =>...))`)

Did this. Left an opening to specify an absolute path via --loadmodulecs. If you can start a server you have the same permissions anyway, so it's not a hole?

Without looking at the git history, I suspect that's the underlying cause: --loadmodulecs was left open as much as possible (to make testing modules easier?), but opening that path accidentally opened all the module loading paths.

libs/host/Configuration/Options.cs
```diff
@@ -542,16 +542,20 @@ internal sealed class Options : ICloneable
...
+        [Option("enable-module-command", Required = false, Hidden = true, HelpText = "Enable MODULE command for 'no', 'local' or 'all' connections")]
+        public ConnectionProtectionOption EnableModuleCommand { get; set; }
+
         [OptionValidation]
         [Option("protected-mode", Required = false, HelpText = "Enable protected mode.")]
         public CommandLineBooleanOption ProtectedMode { get; set; }
 
         [DirectoryPathsValidation(true, false)]
-        [Option("extension-bin-paths", Separator = ',', Required = false, HelpText = "List of directories on server from which custom command binaries can be loaded by admin users")]
+        [Option("extension-bin-paths", Separator = ',', Required = false, HelpText = "Set directories on server from which custom command binaries can be loaded by admin users from list")]
```
TalZaccai: nit: I don't see a reason to change this (and LoadModuleCS's) HelpText, I think it was better explained that we're expecting a list in these places. We should add that `EnableModuleCommand` should be turned on (in both cases).

libs/host/Configuration/Options.cs
```diff
@@ -542,16 +542,20 @@ internal sealed class Options : ICloneable
         [Option("enable-debug-command", Required = false, HelpText = "Enable DEBUG command for 'no', 'local' or 'all' connections")]
         public ConnectionProtectionOption EnableDebugCommand { get; set; }
 
+        [OptionValidation]
+        [Option("enable-module-command", Required = false, Hidden = true, HelpText = "Enable MODULE command for 'no', 'local' or 'all' connections")]
```
TalZaccai: Would be good to add that when enabled, source paths should be specified in `ExtensionBinPaths`.

libs/host/defaults.conf
```diff
@@ -362,6 +362,9 @@
 	/* Enable DEBUG command for clients - no/local/yes */
 	"EnableDebugCommand": "no",
 
+	/* Enable DEBUG command for clients - no/local/yes */
+	"EnableModuleCommand": "yes",
```
TalZaccai: If we're adding this option, it should probably be no by default to avoid confusion.

libs/host/GarnetServer.cs
```diff
@@ -321,7 +321,7 @@ private void LoadModules(CustomCommandManager customCommandManager)
 
                 var modulePath = moduleCSData[0];
                 var moduleArgs = moduleCSData.Length > 1 ? moduleCSData.Skip(1).ToArray() : [];
-                if (ModuleUtils.LoadAssemblies([modulePath], null, true, out var loadedAssemblies, out var errorMsg))
+                if (ModuleUtils.LoadAssemblies([modulePath], null, opts.ExtensionAllowUnsignedAssemblies, out var loadedAssemblies, out var errorMsg, ignorePathCheckWhenUndefined: true))
```
TalZaccai, line 324: Probably missing something here, but why are we ignoring the path check here?

libs/server/Resp/AdminCommands.cs
```diff
@@ -355,6 +356,22 @@ private bool TryRegisterCustomCommands(
         /// </summary>
         private bool NetworkRegisterCs(CustomCommandManager customCommandManager)
         {
+            if (parseState.Count < 6)
+            {
+                return AbortWithWrongNumberOfArguments(nameof(RespCommand.REGISTERCS));
+            }
+
+            if (
+                    (storeWrapper.serverOptions.EnableModuleCommand == ConnectionProtectionOption.No)
```
TalZaccai: nit: indents are wonky here

libs/server/Resp/AdminCommands.cs
```diff
@@ -519,8 +533,18 @@ private bool NetworkModuleLoad(CustomCommandManager customCommandManager)
         {
             if (parseState.Count < 1) // At least module path is required
             {
-                AbortWithWrongNumberOfArguments($"{RespCommand.MODULE}|{Encoding.ASCII.GetString(CmdStrings.LOADCS)}");
-                return true;
+                return AbortWithWrongNumberOfArguments($"{RespCommand.MODULE}|{Encoding.ASCII.GetString(CmdStrings.LOADCS)}");
+            }
+
+            if (
```
TalZaccai: Again, wonky indents :)

libs/server/Resp/AdminCommands.cs
```diff
@@ -712,63 +734,64 @@ private bool NetworkDebug()
...
-                throw new GarnetException(Microsoft.Extensions.Logging.LogLevel.Debug, panic: true);
+                throw new GarnetException(LogLevel.Debug, panic: true);
 
             if (command.EqualsUpperCaseSpanIgnoringCase(CmdStrings.ERROR))
             {
                 if (parseState.Count != 2)
                 {
-                    while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_WRONG_NUMBER_OF_ARGUMENTS, ref dcurr, dend))
-                        SendAndReset();
+                    return AbortWithWrongNumberOfArguments("error");
```
TalZaccai: Error should be `ERR unknown subcommand or wrong number of arguments for 'error'. Try DEBUG HELP.` (we should add a template for this probably)

libs/host/Configuration/Options.cs
```diff
@@ -542,16 +542,20 @@ internal sealed class Options : ICloneable
...
 
         [DirectoryPathsValidation(true, false)]
-        [Option("extension-bin-paths", Separator = ',', Required = false, HelpText = "List of directories on server from which custom command binaries can be loaded by admin users")]
+        [Option("extension-bin-paths", Separator = ',', Required = false, HelpText = "Set directories on server from which custom command binaries can be loaded by admin users from list")]
         public IEnumerable<string> ExtensionBinPaths { get; set; }
 
         [ModuleFilePathValidation(true, true, false)]
-        [Option("loadmodulecs", Separator = ',', Required = false, HelpText = "List of modules to be loaded")]
+        [Option("loadmodulecs", Separator = ',', Required = false, HelpText = "Load modules from list")]
         public IEnumerable<string> LoadModuleCS { get; set; }
```
TalZaccai: If something is defined here and `EnableModuleCommand` is disabled, we should abort the server with an error at start up.

test/Garnet.test/GarnetServerConfigTests.cs
```diff
@@ -249,7 +249,8 @@ public void ImportExportRedisConfigLocal()
             ClassicAssert.AreEqual(invalidOptions.Count, 0);
             ClassicAssert.AreEqual("127.0.0.1 -::1", options.Address);
             ClassicAssert.AreEqual(CommandLineBooleanOption.No, options.ProtectedMode);
-            ClassicAssert.AreEqual(ConnectionProtectionOption.Local, options.EnableDebugCommand);
+            ClassicAssert.AreEqual(ConnectionProtectionOption.Yes, options.EnableDebugCommand);
```
TalZaccai: Please add the following tests:
1. EnableModuleCommand is no & LoadModuleCS is not null - should fail at startup
2. EnableModuleCommand is no, ExtensionBinPaths in not null & trying to load a module from ExtensionBinPaths - Command should fail
3. EnableModuleCommand is yes, ExtensionBinPaths in null & trying to load a module - Command should fail

TalZaccai (changes requested):
Added some additional comments.

libs/host/defaults.conf
prvyk: Done.

libs/host/Configuration/Options.cs
prvyk: Done.

libs/server/Resp/AdminCommands.cs
prvyk: Done (though I preferred that method).

libs/host/GarnetServer.cs
prvyk: IMHO, the question is how much Garnet should make the user write and what is useful to check by default. 

If the user has access to start a server with --loadmodulecs option, they likely have already access to do --extension-bin-path (and --enable-module-command too) and even --extension-allow-unsigned.

Assuming a user wants to load an unsigned module from path, currently they can do just:
```
garnet-server --loadmodulecs /path/module
```

I thought allowing unsigned modules by default is not expected and it's safer to check by default (maybe the module file got switched somehow under the user running this), and so it would now be:
```
garnet-server --loadmodulecs /path/module --extension-allow-unsigned
```

(If the module is signed they can just do the first of course).

If we require the user to also do binpath and enable-module-command than it's:

```
garnet-server --loadmodulecs /path/module --extension-allow-unsigned --extension-bin-path /path --enable-module-command yes
```

I think this is excessive and contributes little but busywork. Also, in enable-module-command's case, wrong, there are setups that would want to disable the command and load a module on startup (on that in the other comment).

libs/host/Configuration/Options.cs
prvyk: It's possible a setup would want to load modules by default but disallow any change once a server has started. e.g. load the JSON module, but don't let any user load any other module from anywhere.

That would recommend a setup with --loadmodulecs but with the module command disabled.

TalZaccai: Makes sense. Thanks!

libs/host/GarnetServer.cs
TalZaccai: Ah ok I see what you're saying, so this is only for pre-loading modules. Makes sense.

test/Garnet.test/GarnetServerConfigTests.cs
prvyk: Added tests for 2 and 3. We agreed above 1 should not fail at startup.

PR #1262, opened 2025-06-20
Standardize timeout parsing with correct errors for too small or too large

prvyk:
There are a number of places where Garnet parses timeouts for blocking commands. These did not consistently error on negative values, and none of them errored on too large values (where .NET will raise an ArgumentOutOfRangeException and then Garnet closes the connection). Using a central SessionParseStateExtension for timeouts fixes all these issues neatly.

Note the .NET maximum timeout (int.MaxValue milliseconds) is far lower than the Valkey maximum timeout (63-bit integer seconds). I don't think the complexity justifies supporting this much? However, the .NET runtime maximum can be easily raised to uint32.MaxValue - 1 milliseconds, I'll be opening an issue on their tracker for this.


prvyk:
See https://github.com/dotnet/runtime/issues/116877 for the dotnet issue.

TalZaccai (approved):
Very nice clean up! Love it!

PR #1266, opened 2025-06-23
Fix SCRIPT|LOAD and EVAL performance regression

kevin-montrose:
In fixing [these bugs](https://github.com/microsoft/garnet/commit/f619caf9d11d19db91836fe09bbe53f79a1d73a5), a regression was introduced if the same script is repeatedly `SCRIPT|LOAD`'d or (more likely) `EVAL`'d.

The root cause is that the 2nd attempt at loading will, correctly, fail to populate the global cache. However this results in removing the newly compiled script from the session cache, so each subsequent call redoes all the compilation and cache construction work.

While strictly speaking repeated `SCRIPT|LOAD`s and `EVAL`s are already sub-par for performance, it's possible for a user to fall into this pattern legitimately due to the ephemeral nature of the script cache.  Thus it's worth fixing, IMO.

The fix is simple, we just consult the global cache for a reusable handle before attempt the load.

This also add `#Requires -Version 7` to `run_bdnperftests.ps1` so you get a nicer error message when (failing) to run it on the default Windows install of Powershell.

### DIFF
--- a/libs/server/Lua/LuaCommands.cs
+++ b/libs/server/Lua/LuaCommands.cs
@@ -121,7 +121,7 @@ namespace Garnet.server
             // that this is stack allocated is load bearing - if it moves, things will break
             Span<byte> digest = stackalloc byte[SessionScriptCache.SHA1Len];
             sessionScriptCache.GetScriptDigest(script.ReadOnlySpan, digest);
-            
+
             var scriptKey = new ScriptHashKey(digest);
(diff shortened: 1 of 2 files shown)


darrenge:
From my standpoint, the BDNs (and the corresponding .ps1 change) are all good. I ran BDN locally and in GH Actions and it passed. I don't want to label this as passed yet until others can review the actual fixes.

darrenge (approved):
Wait until others review the changes before merging as I just covered the BDN tests and the .ps1 file change.

badrishc (approved):

PR #1268, opened 2025-06-25
Support infinities in SortedSet

prvyk:
Redis/Valkey can accept 'inf' as a value:

>ZADD z inf a

Same for +inf/-inf.

This has some implications for addition:
>ZADD z INCR -inf a
>> ERR resulting score is not a number (NaN)

And yet:
>ZADD z2 -inf a
>ZINTER 2 z z2 WITHSCORES
> > "a"
> > 0

Infinity is also allowed in some Geo contexts:
>GEOSEARCH key FROMMEMBER location BYBOX inf 10 ft COUNT 3 ANY WITHCOORD WITHDIST DESC

This PR adds support for infinities in sorted sets, and add proper error messages where needed. Note that INCRBYFLOAT and HINCRBYFLOAT do not support infinities in reference implementations, and this PR does not add such support there.

### DIFF
--- a/libs/cluster/Server/ClusterUtils.cs
+++ b/libs/cluster/Server/ClusterUtils.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
-using System.ComponentModel;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.Extensions.Logging;
 using Tsavorite.core;
@@ -95,10 +95,14 @@ namespace Garnet.cluster
         {
             if (errorCode != 0)
             {
-                var errorMessage = new Win32Exception((int)errorCode).Message;
-                logger.LogError("[ClusterUtils] OverlappedStream GetQueuedCompletionStatus error: {errorCode} msg: {errorMessage}", errorCode, errorMessage);
+                var errorMessage = Utility.GetCallbackErrorMessage(errorCode, numBytes, context);
+                logger?.LogError("[ClusterUtils] OverlappedStream GetQueuedCompletionStatus error: {errorCode} msg: {errorMessage}", errorCode, errorMessage);
             }
+
             ((SemaphoreSlim)context).Release();
         }
+
+        [DllImport("libc")]
+        private static extern IntPtr strerror(int errnum);
     }
 }
--- a/libs/cluster/Server/Replication/CheckpointStore.cs
+++ b/libs/cluster/Server/Replication/CheckpointStore.cs
@@ -132,6 +132,8 @@ namespace Garnet.cluster
                 entry.metadata.objectStoreIndexToken = lastEntry.metadata.objectStoreIndexToken;
             }
 
+            _ = ValidateCheckpointEntry(entry);
+
             if (tail == null)
                 head = tail = entry;
             else
@@ -146,6 +148,25 @@ namespace Garnet.cluster
 
             if (safelyRemoveOutdated)
                 DeleteOutdatedCheckpoints();
+
+            bool ValidateCheckpointEntry(CheckpointEntry entry)
+            {
+                try
+                {
+                    if (!clusterProvider.replicationManager.TryAcquireSettledMetadataForMainStore(entry, out _, out _))
+                        throw new GarnetException("Failed to validate main store metadata at insertion");
+
+                    if (!clusterProvider.serverOptions.DisableObjects && !clusterProvider.replicationManager.TryAcquireSettledMetadataForObjectStore(entry, out _, out _))
+                        throw new GarnetException("Failed to validate object store metadata at insertion");
+
+                    return true;
+                }
+                catch (Exception ex)
+                {
+                    logger?.LogCheckpointEntry(LogLevel.Error, ex.Message, entry);
+                    return false;
+                }
+            }
         }
 
         /// <summary>
@@ -186,6 +207,8 @@ namespace Garnet.cluster
                         break;
                     clusterProvider.GetReplicationLogCheckpointManager(StoreType.Object).DeleteIndexCheckpoint(curr.metadata.objectStoreIndexToken);
                 }
+
+                logger?.LogCheckpointEntry(LogLevel.Warning, "Deleting outdated checkpoint", curr);
 
                 // At least one token can always be deleted thus invalidating the in-memory entry
                 var next = curr.next;
--- a/libs/cluster/Server/Replication/GarnetClusterCheckpointManager.cs
+++ b/libs/cluster/Server/Replication/GarnetClusterCheckpointManager.cs
@@ -85,7 +85,7 @@ namespace Garnet.cluster
             }
             catch (Exception ex)
             {
-                logger?.LogWarning(ex, "Best effort read of checkpoint metadata failed");
+                logger?.LogError(ex, "Best effort read of checkpoint metadata failed");
                 success = false;
             }
 
@@ -109,7 +109,7 @@ namespace Garnet.cluster
                 }
                 catch (Exception ex)
                 {
-                    logger?.LogWarning(ex, "Old format checkpoint metadata failed");
+                    logger?.LogError(ex, "Old format checkpoint metadata failed");
                     throw ex.InnerException;
                 }
 
--- a/libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
+++ b/libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 using System;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.Net;
 using System.Threading;
@@ -47,6 +46,8 @@ namespace Garnet.cluster
         private readonly ILogger logger = logger;
 
         public string errorMsg = default;
+
+        const int validateMetadataMaxRetryCount = 5;
 
         public void Dispose()
         {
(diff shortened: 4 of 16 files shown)


libs/common/RespReadUtils.cs
```diff
@@ -1306,5 +1306,45 @@ public static bool TryReadSerializedData(out byte[] key, out byte[] value, out l
 
             return true;
         }
+
+        /// <summary>
+        /// Read infinity value to number
```
TalZaccai: Would be informative to add what it parses and returns - i.e. reads `[+/-]INF` and returns `double.PositiveInfinity`/`double.NegativeInfinity` respectively.

libs/common/RespReadUtils.cs
```diff
@@ -1306,5 +1306,45 @@ public static bool TryReadSerializedData(out byte[] key, out byte[] value, out l
...
+        /// </summary>
+        /// <param name="value">input data</param>
+        /// <param name="number">If parsing was successful,contains positive or negative infinity</param>
+        /// <returns>True is infinity was read, false otherwise</returns>
+        public static bool TryReadInfinity(ReadOnlySpan<byte> value, out double number)
```
TalZaccai, line 1317: Since a call to this method is always preceded by a call to `NumUtils.TryParse`, why not add a `NumUtils.TryParseIncludeInfinity` which does both in the same call?

libs/common/RespReadUtils.cs
```diff
@@ -1306,5 +1306,45 @@ public static bool TryReadSerializedData(out byte[] key, out byte[] value, out l
...
+                {
+                    number = double.PositiveInfinity;
+                    return true;
+                }
+            }
+            else if (value.Length == 4)
+            {
+                if (value[0] == '+')
+                {
+                    if (value.Slice(1, 3).EqualsUpperCaseSpanIgnoringCase(RespStrings.INFINITY_UPPERCASE))
```
TalZaccai: Maybe add `RespStrings.POS_INFINITY` and `RespStrings.NEG_INFINITY` for simplicity?

libs/common/RespStrings.cs
```diff
@@ -10,6 +10,7 @@ public class RespStrings
         public static ReadOnlySpan<byte> EMPTYARRAY => "*0\r\n"u8;
         public static ReadOnlySpan<byte> EMPTYSET => "~0\r\n"u8;
         public static ReadOnlySpan<byte> EMPTYMAP => "%0\r\n"u8;
+        public static ReadOnlySpan<byte> INFINITY_UPPERCASE => "INF"u8;
```
TalZaccai: Please rename to `INFINITY`, the variable name suggests that it is uppercase (similar to `CmdStrings`)

libs/server/Resp/Parser/ParseUtils.cs
```diff
@@ -99,9 +99,9 @@ public static bool TryReadLong(ref ArgSlice slice, out long number)
         /// Parsed double
         /// </returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static double ReadDouble(ref ArgSlice slice)
+        public static double ReadDouble(ref ArgSlice slice, bool canBeInfinite)
```
TalZaccai, line 104: Please add XML comment (in both methods)

libs/server/Resp/Parser/SessionParseState.cs
```diff
@@ -415,21 +415,21 @@ public bool TryGetLong(int i, out long value)
         /// </summary>
         /// <returns></returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public double GetDouble(int i)
+        public double GetDouble(int i, bool canBeInfinite = true)
```
TalZaccai, line 418: Why is this true by default here? I'd say it should be false by default to not cause any unintended behavior.
Also please add XML comments here as well.

TalZaccai (changes requested):
Nicely done, added some nitpicky comments :)

libs/common/RespReadUtils.cs
prvyk: Done.

libs/common/RespStrings.cs
prvyk: Done. I wanted to disambiguate from the values that include the newline and wasn't sure of how to do that. Maybe a future diff would rename to EMPTYMAP_NEWLINE or something like that?

libs/server/Resp/Parser/ParseUtils.cs
prvyk: You mean the parameters? Done.

libs/server/Resp/Parser/SessionParseState.cs
prvyk: If we look at existing users of GetDouble and TryGetDouble, almost all of them are fine with accepting infinite values, either they accept them directly, or their existing range checks catch this (in timeout and latlong we'll get the value too large errors etc., redis does it too). In HINCRBYFLOAT we accept this to get the right error.

Only exception is calls from INCRBYFLOAT use, and even that should have allowed infinity to get the right error text ("ERR increment would produce NaN or Infinity"). Problem is that getting the right error out of RMW isn't trivial. I'll get back to it in a future diff.

libs/common/RespReadUtils.cs
prvyk: Done. It's NumUtils.TryParseWithInfinity().

libs/server/Resp/Parser/SessionParseState.cs
prvyk: Ok, we don't really need to care about RMW, just IsValidDouble() in PrivateMethods to get the right error.

Also found a way to ditch the TryGetDouble calls there. So all remaining users of function are users that can handle infinite value input. I could ditch canBeInfinite if desired...

libs/common/RespReadUtils.cs
prvyk: Tried to, it broke Debug tests due to the assertion there. It's possible to add a version of the comparer without the assertion if desired.

TalZaccai: Not sure I'm following, it broke what assertion where?

prvyk: The Debug.Assert(b2 is >= 65 and <= 90) in EqualsUpperCaseSpanIgnoringCase() in AsciiUtils.cs. POSITIVE/NEGATIVE infinity is +-inf and it didn't like the +/- characters.

TalZaccai: Why did I remember we had an overload for that that accepted non-alphabetic characters and now I can't find it?
In any case we can add an optional flag for that if you want...

libs/common/AsciiUtils.cs
```diff
@@ -84,6 +85,40 @@ public static bool EqualsUpperCaseSpanIgnoringCase(this ReadOnlySpan<byte> left,
...
+            return true;
+        if (left.Length != right.Length)
+            return false;
+
+        for (var i = 0; i < left.Length; i++)
+        {
+            var b1 = left[i];
+            var b2 = right[i];
+
+            if ((allowAllPrintableCharacters) && ((b2 is < 65 and >= 32) || (b2 is < 127 and > 96)))
```
TalZaccai: Perhaps more compact / better perf?
Also we probably want to fail here (not just Debug.Assert) if allowAllPrintableCharacters == false and we encounter a character out of range.

```cs
if (!allowNonAlphabeticChars && b2 is < 65 or > 90)
    return false;

if (b1 == b2 || ((!allowNonAlphabeticChars || b2 is >= 65 and <= 90) && b1 - 32 == b2))
    continue;
```

TalZaccai: Ha, you were changing the code as I was writing the comment :)

prvyk: I'm fine with doing it that way too. It's a bit clearer. Done.

libs/server/OperationError.cs
```diff
@@ -15,6 +15,10 @@ public enum OperationError : byte
         /// <summary>
         /// Operation failed due to incompatible type
         /// </summary>
-        INVALID_TYPE
+        INVALID_TYPE,
+        /// <summary>
+        /// Operation failed due to NaN/infinity
+        /// </summary>
+        NANORINFINITY
```
TalZaccai: Please rename to NAN_OR_INFINITY for readability

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
```diff
@@ -605,12 +605,23 @@ static bool IsValidNumber(int length, byte* source, Span<byte> output, out long
         static bool IsValidDouble(int length, byte* source, Span<byte> output, out double val)
         {
             // Check for valid number
-            if (!NumUtils.TryReadDouble(length, source, out val) || !double.IsFinite(val))
+            if (!NumUtils.TryReadDouble(length, source, out val))
```
TalZaccai: Why not use `TryParseWithInfinity` here?

libs/server/Resp/BasicCommands.cs
```diff
@@ -760,19 +760,22 @@ private bool NetworkIncrementByFloat<TGarnetApi>(ref TGarnetApi storageApi)
...
+            {
+                return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_NAN_INFINITY_INCR);
             }
 
             Span<byte> outputBuffer = stackalloc byte[NumUtils.MaximumFormatDoubleLength + 1];
             var output = ArgSlice.FromPinnedSpan(outputBuffer);
 
-            var input = new RawStringInput(RespCommand.INCRBYFLOAT, ref parseState, startIdx: 1);
+            var input = new RawStringInput(RespCommand.INCRBYFLOAT, ref parseState, startIdx: 1,
+                                           arg1: BitConverter.DoubleToInt64Bits(dbl));
```
TalZaccai: I don't know if it's worthwhile to do these conversions, might as well just parse the double again in the backend...

TalZaccai (changes requested):
Couple more comments

libs/server/OperationError.cs
prvyk: Done.

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
prvyk: Done.

libs/server/Resp/BasicCommands.cs
prvyk: This means we don't have to take into account a failure state in the second parsing later, and is very likely slightly faster. The conversions are just casts.

I can hide most mechanics (stackalloc, OperationError by casting from output.Span[0], and now arg1 cast) behind a new API call, which is better for this API anyway. Done.

TalZaccai (approved):

Issue #1269, opened 2025-06-26
Transactions don't support expiration commands


xangelix:
### Describe the bug

Hopefully this isn't duplicate, but I couldn't find anything that would suggest this is an expected error.

Inside a transaction, it appears some commands related to expiry are not known. This is not an issue on redis, causing my tested migration redis->garnet to fail.

### Steps to reproduce the bug

```
> HSET test ak av bk bv
(integer) 2

> HEXPIREAT test 1750899548 FIELDS 1 ak
1) "1"

> HPTTL test FIELDS 1 ak
1) "24796"

> MULTI
"OK"

> HEXPIREAT test 1750899548 FIELDS 1 bk
"ERR unknown command"
```

### Release version

`Garnet 1.0.73 64 bit; standalone mode`

### IDE

Redis official CLI
fred (redis rust library)

### OS version

Docker Alpine (GitHub repo image)

xangelix:
It's not entirely clear to me where this exception is from, but after testing those commands and some others ~30min ago, I noticed at some point this showed up in my garnet log:

```
[...]

PR #1271, opened 2025-06-26
MULTI/EXEC doesn't recognize or lock for certain valid RESP commands

prvyk:
This PR adds more commands to transaction manager locking so that they are recognized in MULTI/EXEC context. Helps with #1269 too.

Before this PR:

>multi
>zunion 2 a b
> >ERR unknown command

This PR is not a complete addition of all relevant commands, some commands are left aside for now, e.g. blocking commands (should check later if/how this works with multi/exec), CLUSTER, DUMP/RESTORE...

### DIFF
--- a/Version.props
+++ b/Version.props
@@ -1,6 +1,6 @@
 <Project>
 	<!-- VersionPrefix property for builds and packages -->
 	<PropertyGroup>
-		<VersionPrefix>1.0.77</VersionPrefix>
+		<VersionPrefix>1.0.78</VersionPrefix>
 	</PropertyGroup>
 </Project>


libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -64,142 +64,201 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
...
-                RespCommand.GEOHASH => SortedSetObjectKeys(SortedSetOperation.GEOHASH, inputCount),
-                RespCommand.GEODIST => SortedSetObjectKeys(SortedSetOperation.GEODIST, inputCount),
-                RespCommand.GEOPOS => SortedSetObjectKeys(SortedSetOperation.GEOPOS, inputCount),
-                RespCommand.GEORADIUS => GeoCommands(RespCommand.GEORADIUS, inputCount),
-                RespCommand.GEORADIUS_RO => GeoCommands(RespCommand.GEORADIUS_RO, inputCount),
+                RespCommand.APPEND => SingleKey(1, false, LockType.Exclusive),
+                RespCommand.BITCOUNT => SingleKey(1, false, LockType.Shared),
+                RespCommand.BITFIELD => SingleKey(1, false, LockType.Exclusive),
+                RespCommand.BITFIELD_RO => SingleKey(1, false, LockType.Shared),
+                RespCommand.BITPOS => SingleKey(1, false, LockType.Exclusive),
```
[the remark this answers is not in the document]
prvyk: Done.

TalZaccai:
Thanks! I appreciate what you're trying to do here, but why have we lost the original per-object handler design? It was more readable and it makes it now very difficult to figure out what was changed / added. Since this is quite delicate as we're dealing with locks here, I'd rather the PR changes to be clearer at this point so something doesn't get changed / added where it shouldn't have. Appreciate your help!

TalZaccai (changes requested):
Please see my comment. Thanks!

prvyk:
> Thanks! I appreciate what you're trying to do here, but why have we lost the original per-object handler design? It was more readable and it makes it now very difficult to figure out what was changed / added. Since this is quite delicate as we're dealing with locks here, I'd rather the PR changes to be clearer at this point so something doesn't get changed / added where it shouldn't have. Appreciate your help!

The problem is the original handlers were very confusing because they did an unnecessary use of *Operation enums, so occasionally the enums could be mismatched. e.g. the main switch calling for SortedSetKeys(SortedSetOperation.SOMETHING) and the switch in SortedSetKeys() not checking for SortedSetOperation.SOMETHING. Another problem was with commands that didn't have an *Operation enum thanks to how they were implemented, but needed different handling here. (The Operation enum has only 5 bits for objects, so it's a limited resource).

What if I return the per-object design, but use RespCommand enum instead of *Operation? Would that be fine? So:

Switch (cmd)
{
RespComman.LLEN => ListCommands(cmd),
RespCommand.SADD => SetCommands(cmd),
...
}

etc.?

TalZaccai:
> > Thanks! I appreciate what you're trying to do here, but why have we lost the original per-object handler design? It was more readable and it makes it now very difficult to figure out what was changed / added. Since this is quite delicate as we're dealing with locks here, I'd rather the PR changes to be clearer at this point so something doesn't get changed / added where it shouldn't have. Appreciate your help!
> 
> The problem is the original handlers were very confusing because they did an unnecessary use of *Operation enums, so occasionally the enums could be mismatched. e.g. the main switch calling for SortedSetKeys(SortedSetOperation.SOMETHING) and the switch in SortedSetKeys() not checking for SortedSetOperation.SOMETHING. Another problem was with commands that didn't have an *Operation enum thanks to how they were implemented, but needed different handling here. (The Operation enum has only 5 bits as we know, so it's a limited resource).
> 
> What if I return the per-object design, but use RespCommand enum instead of *Operation? Would that be fine? So:
> 
> Switch (cmd) { RespComman.LLEN => ListCommands(cmd), RespCommand.SADD => SetCommands(cmd), ... }
> 
> etc.?

Yeah I think that should be ok, and let's hold off on sorting the commands lexicographically, it just makes diffing harder :)

vazois:
We should add tests for this. Ensure that it works properly and newly added commands trigger a failure if the txnManager does not cover them.

prvyk:
> We should add tests for this. Ensure that it works properly and newly added commands trigger a failure if the txnManager does not cover them.

It's too many commands to test each one, but I can add a few tests, especially for writing commands where the WATCH interaction is important...

The idea of adding a test for new commands is great; It should probably be something like the commandDocs test - run GetKeys for each command, and if the result is -1 and not in the exclusion list, fail.

vazois:
> > We should add tests for this. Ensure that it works properly and newly added commands trigger a failure if the txnManager does not cover them.
> 
> It's too many commands to test each one, but I can add a few tests, especially for writing commands where the WATCH interaction is important...
> 
> The idea of adding a test for new commands is great; It should probably be something like the commandDocs test - run GetKeys for each command, and if the result is -1 and not in the exclusion list, fail.

I understand. We just need a small test that executes a transaction with the corresponding command. The concept should be similar to ACL tests (https://github.com/microsoft/garnet/blob/main/test/Garnet.test/Resp/ACL/RespCommandTests.cs).

Example

```csharp
        [Test]
        [TestCase("GET", new[] { "mykey" })]
        [TestCase("SET", new[] { "mykey", "myvalue" })]
        public void TxnRespCommands(string command, string[] args)
        {

            var tran = db.CreateTransaction();
            var t = tran.ExecuteAsync("GET", ["mykey"]);
            var committed = tran.Execute();
            ClassicAssert.IsTrue(committed);
        }
```

Maybe an LLM can generate most of the code here (i.e. command and paramaters

prvyk:
> > > We should add tests for this. Ensure that it works properly and newly added commands trigger a failure if the txnManager does not cover them.
> > 
> > 
> > It's too many commands to test each one, but I can add a few tests, especially for writing commands where the WATCH interaction is important...
> > The idea of adding a test for new commands is great; It should probably be something like the commandDocs test - run GetKeys for each command, and if the result is -1 and not in the exclusion list, fail.
> 
> I understand. We just need a small test that executes a transaction with the corresponding command. The concept should be similar to ACL tests (https://github.com/microsoft/garnet/blob/main/test/Garnet.test/Resp/ACL/RespCommandTests.cs).

I took a different path, by creating commandlines out of COMMAND INFO/DOCS output, and then always discarding the result. There's no "transaction.Execute" with StackExchange, but doing MULTI/DISCARD with GarnetClientSession worked.

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -165,8 +165,8 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
                 RespCommand.SETEXXX => SingleKey(1, false, LockType.Exclusive),
                 RespCommand.DEL => ListKeys(inputCount, StoreType.All, LockType.Exclusive),
                 RespCommand.DELIFGREATER => SingleKey(1, false, LockType.Exclusive),
-                RespCommand.EXISTS => SingleKey(1, false, LockType.Shared),
-                RespCommand.RENAME => SingleKey(1, false, LockType.Exclusive),
+                RespCommand.EXISTS => ListKeys(inputCount, StoreType.All, LockType.Shared),
+                RespCommand.RENAME => ListKeys(1, StoreType.All, LockType.Exclusive),
```
TalZaccai: I think we should have a `SingleKey` method that accepts a `StoreType` and then you could apply the same logic as in `ListKeys`. i.e.:
```cs
if (storeType is StoreType.Main or StoreType.All)
    SaveKeyEntryToLock(key, false, type);
if (storeType is StoreType.Object or StoreType.All && !objectStoreBasicContext.IsNull)
    SaveKeyEntryToLock(key, true, type);
```

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -418,10 +549,34 @@ private int MSETKeys(int inputCount, bool isObject, LockType type)
         }
 
         /// <summary>
-        /// Returns a list of keys for *STORE commands (e.g. SUNIONSTORE, ZINTERSTORE etc.)
+        /// Returns a list of keys for set *STORE commands (e.g. SUNIONSTORE, SINTERSTORE etc.)
         /// Where the first key's value is written to and the rest of the keys' values are read from.
         /// </summary>
-        private int XSTOREKeys(int inputCount, bool isObject)
+        private int SSTOREKeys(int inputCount, bool isObject, int offset = 0)
```
TalZaccai: Let's rename it to something more informative, like `SingleWriteKeyListReadKeys`...

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -418,10 +549,34 @@ private int MSETKeys(int inputCount, bool isObject, LockType type)
...
+        {
+            if (inputCount > offset)
```
TalZaccai: nit: if `inputCount <= offset` we can just return early.

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -418,10 +549,34 @@ private int MSETKeys(int inputCount, bool isObject, LockType type)
...
+
+            return inputCount;
+        }
+
+        /// <summary>
+        /// Returns a list of keys for complex *STORE commands (e.g. ZUNIONSTORE, ZINTERSTORE etc.)
+        /// Where the first key's value is written to and the rest of the keys' values are read from.
+        /// We can get number of read keys by checking the second argument.
+        /// </summary>
+        private int ZSTOREKeys(int inputCount, bool isObject)
```
TalZaccai: Let's rename this to `SingleWriteKeyListReadKeysWithCount` or something of the sort...

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -175,35 +175,124 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
...
+                    return 1;
+
+                error = System.Text.Encoding.ASCII.GetBytes(string.Format(
+                    CmdStrings.GenericErrCommandDisallowedWithOption, RespCommand.DEBUG, "enable-debug-command"));
+                return -1;
+            }
+
             return command switch
             {
+                RespCommand.CONFIG_GET => 1,
```
TalZaccai: I'm not sure some of these commands should be allowed within a transaction. I flagged these, please verify:
RespCommand.CONFIG_GET
RespCommand.CONFIG_REWRITE
RespCommand.CONFIG_SET
RespCommand.CLIENT_KILL
RespCommand.FLUSHALL
RespCommand.FLUSHDB
RespCommand.HELLO
RespCommand.LATENCY_RESET
RespCommand.REPLICAOF
RespCommand.SECONDARYOF
RespCommand.SLOWLOG_RESET

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -258,99 +347,141 @@ private int GeoCommands(RespCommand command, int inputCount)
...
+                RespCommand.ZRANK => SingleKey(1, true, LockType.Shared),
+                RespCommand.ZRANGE => SingleKey(1, true, LockType.Shared),
+                RespCommand.ZREVRANK => SingleKey(1, true, LockType.Shared),
+                RespCommand.ZREMRANGEBYLEX => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.ZREMRANGEBYRANK => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.ZREMRANGEBYSCORE => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.ZLEXCOUNT => SingleKey(1, true, LockType.Shared),
+                RespCommand.ZPOPMIN => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.ZRANDMEMBER => SingleKey(1, true, LockType.Shared),
+                RespCommand.ZDIFF => ListKeys(true, LockType.Shared),
```
TalZaccai: number of keys should match the `numkeys` parameter, same comment for ZINTER, ZINTERCARD, ZUNION, ZMPOP, LMPOP...

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -258,99 +347,141 @@ private int GeoCommands(RespCommand command, int inputCount)
...
+                RespCommand.GEOADD => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.GEOHASH => SingleKey(1, true, LockType.Shared),
+                RespCommand.GEODIST => SingleKey(1, true, LockType.Shared),
+                RespCommand.GEOPOS => SingleKey(1, true, LockType.Shared),
```
TalZaccai: Where did GEOSEARCH go?

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -175,35 +175,124 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
...
                 RespCommand.PERSIST => SingleKey(1, false, LockType.Exclusive),
                 RespCommand.MGET => ListKeys(inputCount, StoreType.Main, LockType.Shared),
                 RespCommand.MSET => MSETKeys(inputCount, false, LockType.Exclusive),
                 RespCommand.MSETNX => MSETKeys(inputCount, false, LockType.Exclusive),
                 RespCommand.UNLINK => ListKeys(inputCount, StoreType.All, LockType.Exclusive),
                 RespCommand.GETDEL => SingleKey(1, false, LockType.Exclusive),
                 RespCommand.APPEND => SingleKey(1, false, LockType.Exclusive),
-                _ => AdminCommands(command)
+
+                RespCommand.BITFIELD_RO => SingleKey(1, false, LockType.Shared),
```
TalZaccai: General comment: it's confusing when and integer denotes offset and when it denotes count. We should probably add the param name for each of these (as well as the boolean params) for better readability

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -258,99 +347,141 @@ private int GeoCommands(RespCommand command, int inputCount)
...
+                RespCommand.SRANDMEMBER => SingleKey(1, true, LockType.Shared),
+                RespCommand.SPOP => SingleKey(1, true, LockType.Exclusive),
+                RespCommand.SISMEMBER => SingleKey(1, true, LockType.Shared),
+                RespCommand.SMISMEMBER => SingleKey(1, true, LockType.Shared),
+                RespCommand.SSCAN => SingleKey(1, true, LockType.Shared),
+                RespCommand.SUNION => ListKeys(inputCount, StoreType.Object, LockType.Shared),
+                RespCommand.SUNIONSTORE => SSTOREKeys(inputCount, true),
+                RespCommand.SDIFF => ListKeys(inputCount, StoreType.Object, LockType.Shared),
+                RespCommand.SDIFFSTORE => SSTOREKeys(inputCount, true),
+                RespCommand.SMOVE => ListKeys(inputCount, StoreType.Object, LockType.Exclusive),
```
TalZaccai: Should only read 2 keys

test/Garnet.test/TransactionTests.cs
```diff
@@ -221,6 +223,169 @@ public void LargeTxnWatch([Values(512, 2048, 4096)] int size)
             }
         }
 
+        [Test]
+        public async Task TxnCommandCoverage()
```
TalZaccai, line 227: This is great! Thanks for adding!

libs/server/Transaction/TxnKeyManager.cs
prvyk: Even before this PR, it was mapped in the previous switch to GeoCommands. Moved it now here to make it clearer.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Fixed.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Only one location denotes offset. I added the param name there to make it clear. Also changed prototypes to either make the StoreType clear or optional (for functions that practically are used with only one type of lock).

libs/server/Transaction/TxnKeyManager.cs
prvyk: Tested these with redis, all returned QUEUED and were later executed. Couldn't test all thoroughly, but hello/flush*/client kill/config/*reset worked.

libs/server/Transaction/TxnKeyManager.cs
prvyk: I'm not sure I understand? ListKey read the keynum from the RESP commandline. I changed the function name to make it clearer.

libs/server/Transaction/TxnKeyManager.cs
TalZaccai: Discussed it with the team, we're of the same opinion that we shouldn't support these commands unless given a good reason why we should. Specifically FLUSHALL and FLUSHDB require a database-wide key lock, which we don't support at the moment.

libs/server/Transaction/TxnKeyManager.cs
TalZaccai: For the commands I listed here, you're calling `ListKeys(int inputCount, StoreType storeType, LockType type)`, where you should be calling the `ListKeys` overload that you're talking about, unless I'm missing something...

prvyk: Most of that list called the right overloads, except for ZINTER and ZINTERCARD. Fixed.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Removed these commands from the function.

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -64,146 +64,224 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
...
+                RespCommand.GETIFNOTMATCH => SingleKey(StoreType.Main, LockType.Shared),
+                RespCommand.GETWITHETAG => SingleKey(StoreType.Main, LockType.Shared),
+                RespCommand.SET => SingleKey(StoreType.Main, LockType.Exclusive),
+                RespCommand.SETIFMATCH => SingleKey(StoreType.Main, LockType.Exclusive),
+                RespCommand.SETIFGREATER => SingleKey(StoreType.Main, LockType.Exclusive),
+                RespCommand.GETRANGE => SingleKey(StoreType.Main, LockType.Shared),
+                RespCommand.SETRANGE => SingleKey(StoreType.Main, LockType.Exclusive),
+                RespCommand.PFADD => SingleKey(StoreType.Main, LockType.Exclusive),
                 RespCommand.PFCOUNT => ListKeys(inputCount, StoreType.Main, LockType.Shared),
                 RespCommand.PFMERGE => ListKeys(inputCount, StoreType.Main, LockType.Exclusive),
```
TalZaccai: PFMERGE writes to the first key and reads from the rest

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -258,112 +336,164 @@ private int GeoCommands(RespCommand command, int inputCount)
...
+                RespCommand.SISMEMBER => SingleKey(StoreType.Object, LockType.Shared),
+                RespCommand.SMISMEMBER => SingleKey(StoreType.Object, LockType.Shared),
+                RespCommand.SSCAN => SingleKey(StoreType.Object, LockType.Shared),
+                RespCommand.SUNION => ListKeys(inputCount, StoreType.Object, LockType.Shared),
+                RespCommand.SUNIONSTORE => SingleWriteKeyListReadKeys(inputCount),
+                RespCommand.SDIFF => ListKeys(inputCount, StoreType.Object, LockType.Shared),
+                RespCommand.SDIFFSTORE => SingleWriteKeyListReadKeys(inputCount),
+                RespCommand.SMOVE => ListKeys(2, StoreType.Object, LockType.Exclusive),
+                RespCommand.SINTER => ListKeys(inputCount, StoreType.Object, LockType.Shared),
+                RespCommand.SINTERCARD => ListKeys(inputCount, StoreType.Object, LockType.Shared),
```
TalZaccai: SINTERCARD should also call the other ListKeys overload

libs/server/Transaction/TxnKeyManager.cs
prvyk: Fixed.

libs/server/Transaction/TxnKeyManager.cs
prvyk: Assumed it's like SINTER... Very consistent of them. Fixed.

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -64,146 +64,224 @@ internal int GetKeys(RespCommand command, int inputCount, out ReadOnlySpan<byte>
...
+                    return 1;
+
+                error = System.Text.Encoding.ASCII.GetBytes(string.Format(
+                    CmdStrings.GenericErrCommandDisallowedWithOption, RespCommand.DEBUG, "enable-debug-command"));
+                return -1;
+            }
+
             return command switch
             {
+                RespCommand.COMMAND => 1,
```
TalZaccai: Looking at this list again, SCAN, KEYS, DBSIZE, MEMORY_USAGE are also problematic here. I think let's just leave it as it was before, I don't see why any of these commands should be called from a txn context as most of them are not data related.

TalZaccai (changes requested):
Added a couple of comments

libs/server/Transaction/TxnKeyManager.cs
prvyk: Done.

Sometimes adding non-data commands to Txn context is useful, e.g. print execution stage and output time of execution if any of the parts takes too much time.

TalZaccai: Sure, but if we're not careful it can lead to some unexpected behavior / results, so maybe we should just see if anyone is even interested in these...

TalZaccai (approved):

PR #1279, opened 2025-07-04
Dynamically Set Memory Limits with CONFIG SET

TalZaccai:
This PR enables an administrator to dynamically set the following parameters using the `CONFIG SET` command:
- `MemorySize` (parameter: `memory`)
- `IndexSize` (parameter: `index`)
- `ObjectStoreLogMemorySize` (parameter: `obj-log-memory`)
- `ObjectStoreIndexSize` (parameter: `obj-index`)
- `ObjectStoreHeapMemorySize` (parameter: `obj-heap-memory`)


libs/server/Metrics/Info/GarnetInfoMetrics.cs
```diff
@@ -61,37 +61,40 @@ private void PopulateServerInfo(StoreWrapper storeWrapper)
...
+            var main_store_read_cache_size = 0L;
+            long total_main_store_size;
 
-            var object_store_index_size = -1L;
-            var object_store_log_memory_size = -1L;
-            var object_store_read_cache_log_memory_size = -1L;
-            var object_store_heap_memory_size = -1L;
-            var object_store_read_cache_heap_memory_size = -1L;
-            var total_object_store_size = -1L;
+            var disableObj = storeWrapper.serverOptions.DisableObjects;
```
TedHartMS, line 70: maybe change to
    var initialSize = storeWrapper.serverOptions.DisableObjects ? -1L : 0L;

test/Garnet.test/RespConfigTests.cs
```diff
@@ -0,0 +1,241 @@
...
+            TestUtils.DeleteDirectory(TestUtils.MethodTestDir);
+        }
+
+        [Test]
+        [TestCase(true, "16g", "32g", "64g", "g4")]
+        [TestCase(true, "9gB", "28GB", "33G", "2gBB")]
+        [TestCase(false, "16m", "32m", "64m", "3bm")]
+        [TestCase(false, "5MB", "30M", "128mb", "44d")]
+
+        public void ConfigSetMemorySizeTest(bool mainStore, string smallerSize, string largerSize, string largerThanBufferSize, string malformedSize)
```
badrishc: Can use `var info = TestUtils.GetStoreAddressInfo(server)` to get addresses (see `PersistCopyUpdateTest`) and verify that after expanding the memory size, HeadAddress does not move forward when sufficient new records are inserted at tail (because the log size is made larger).

test/Garnet.test/RespConfigTests.cs
```diff
@@ -0,0 +1,241 @@
...
+    {
+        GarnetServer server;
+        private string memorySize = "17g";
+        private string indexSize = "64m";
+        private string objectStoreLogMemorySize = "17m";
+        private string objectStoreHeapMemorySize = "32m";
+        private string objectStoreIndexSize = "8m";
+
+        [SetUp]
+        public void Setup()
```
badrishc, line 40: Verify that this works:
* Start server with log size X, insert keys
* Take checkpoint and kill (similar to `SeSaveRecoverTest`)
* Recover server with smaller & larger size Y
* Verify keys

test/Garnet.test/RespConfigTests.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        [Test]
+        [TestCase(true, "16g", "32g", "64g", "g4")]
+        [TestCase(true, "9gB", "28GB", "33G", "2gBB")]
+        [TestCase(false, "16m", "32m", "64m", "3bm")]
+        [TestCase(false, "5MB", "30M", "128mb", "44d")]
+
+        public void ConfigSetMemorySizeTest(bool mainStore, string smallerSize, string largerSize, string largerThanBufferSize, string malformedSize)
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
```
badrishc, line 76: Another test: start at a particular log size and write enough records that HeadAddress moves forward. Now reduce the log size, and we should verify that HeadAddress moves forward as we evict records (and empty pages) to make the log memory footprint (tail - head) match the configured size.

libs/server/ServerConfig.cs
```diff
@@ -171,39 +188,191 @@ private bool NetworkCONFIG_SET()
...
+            var confMemorySize = ServerOptions.ParseSize(
+                    mainStore ? storeWrapper.serverOptions.MemorySize
+                        : storeWrapper.serverOptions.ObjectStoreLogMemorySize, out _);
+
+            // If the new memory size is the same as the configured memory size, nothing to do
+            if (newMemorySize == confMemorySize)
+                return;
+
+            // Calculate the buffer size based on the configured memory size
+            var adjConfMemorySize = ServerOptions.PreviousPowerOf2(confMemorySize);
```
TedHartMS: I think you got this pattern from GarnetServerOptions.MemorySizeBits? In that case the "PreviousPowerOf2->test for equality -> *=2" is done to adjust emptyPageCount. Here (and in ConfigSetMemorySizeTest), it seems you can just do NextPowerOf2, and maybe add a relevant comment in GarnetServerOptions.MemorySizeBits

libs/server/ServerConfig.cs
```diff
@@ -171,39 +188,191 @@ private bool NetworkCONFIG_SET()
...
+            if (confMemorySize != adjConfMemorySize)
+                confMemorySize = adjConfMemorySize * 2;
+
```
badrishc, line 261: All relevant tests can be parameterized by Reviv turned on (full reviv, not just in chain) to ensure they are robust.

test/Garnet.test/TestUtils.cs
```diff
@@ -1080,5 +1097,30 @@ public static StoreAddressInfo GetStoreAddressInfo(IServer server, bool includeR
...
+
+        /// <summary>
+        /// Get a random alphanumeric string of specified length
+        /// </summary>
+        /// <param name="len"></param>
+        /// <returns></returns>
+        public static string GetRandomString(int len)
+        {
+            const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+            return new string([.. Enumerable.Repeat(chars, len).Select(s => s[Random.Next(s.Length)])]);
```
prvyk: The initialization of Random above doesn't specify or log the seed. Would a failure be easily repeatable?

Note that if we don't care about using a specific seed for a test, might as well use RandomNumberGenerator.GetString().

TalZaccai: True, I just copied this one from another part of the code, updated it to use RandomNumberGenerator. I doubt that repeating the random seq in this context would be required, it's mainly used to generate random keys.

libs/server/Metrics/Info/GarnetInfoMetrics.cs
```diff
@@ -62,37 +62,41 @@ private void PopulateServerInfo(StoreWrapper storeWrapper)
...
+
+            var initialSize = storeWrapper.serverOptions.DisableObjects ? -1L : 0L;
```
TedHartMS: nit: change this to use disableObj

libs/server/Servers/ServerOptions.cs
```diff
@@ -253,6 +256,7 @@ public static long ParseSize(string value)
                         if (char.ToLower(c) == suffix[i])
                         {
                             result *= (long)Math.Pow(1024, i + 1);
+                            bytesRead++;
                             return result;
                         }
                     }
```
TedHartMS, line 267: Suggest checking for 'b' here and if found, bytesRead++, else at least an Assert and leave bytesRead unincremented. Then below, TryParseSize (and potentially another caller) does not need the check for 'b'

libs/server/Storage/SizeTracker/CacheSizeTracker.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft Corporation.
+﻿
```
TedHartMS: Didn't "format" complain about a leading empty line? (It does for trailing ones..)

libs/common/AsciiUtils.cs
```diff
@@ -114,4 +114,30 @@ public static bool EqualsUpperCaseSpanIgnoringCase(this ReadOnlySpan<byte> left,
...
+        {
+            var b1 = left[i];
+            var b2 = right[i];
+
+            if (!allowNonAlphabeticChars && b2 is < 97 or > 122)
+                return false;
+
+            if (b1 == b2 || ((!allowNonAlphabeticChars || b2 is >= 97 and <= 122) && (b1 + 32 == b2)))
+                continue;
+            return false;
```
TedHartMS: Seems a lot of "if"s. Can it be simplified?
if (b2 is >= 97 and <= 122)
    return b1 == b2 || b1 + 32 == b2;
return allowNonAlphabeticChars && b1 == b2;

libs/server/Storage/SizeTracker/CacheSizeTracker.cs
TalZaccai: Nah you need to run code cleanup for that, and that'll change a lot of things in a lot of places...

PR #974, opened 2025-01-29
JSON Module rewrite with custom JSONPath implementation

Vijay-Nirmal:
As part of this PR, the JSON module implementation has been rewritten using a custom implementation of JsonPath to achieve better performance and compatibility with Redis. As part of this rewrite other parts of the Json module have also been rewritten

### Main Changes:

* Custom implementation of JSONPath evaluator, which is a port of Newtonsoft.Json with many optimizations and some bug fixes
* Implemented full support for JSON.SET command but there is a known issue which should prevent this PR from merging
* Implemented full support for JSON.GET command but there is a limitation in STJ that doesn't allow us to customize the output format using custom characters. Most likely we can never support this feature in Garnet if we use STJ https://github.com/dotnet/runtime/issues/111899
* Added separate benchmark for JsonPath evaluator and Json comments with a big JSON
* Removed `JsonPath.Net` package reference

### Miscellaneous:

* Made `CmdStrings` class public in `CmdStrings.cs` to allow external access.
* Changed `ExistOptions` enum to public in `RespEnums.cs` for broader accessibility.

### Known Issues:

* If value that needs to be updated is `null` using `JSON.SET` command then the update will fail, it can't find the parent object to update the value. This is a limitation of the custom JSONPath implementation because of the way the Null is represented in JsonNode. Will find a way to support this in future without affecting the performance much. This issue shouldn't block the PR from being merged as this scenario was failing in the current implementation as well.
* `Utf8JsonWriter` doesn't allow enough customizability to support `INDENT`, `NEWLINE` and `SPACE` options in JSON.GET command. We can't inherit the Utf8JsonWriter class as well as it's a sealed class to write our own logic to customize it. This feature will not be supported in STJ in future as well https://github.com/dotnet/runtime/issues/111899#issuecomment-2619676646.  We have only two options, either Garnet can never support the customizability provided by Redis for certain JSON commands, or Garnet should use Newtonsoft.Json (which supports customisation). Both of these options are not ideal.

### TODO

- [x] Initial setup
- [x] Adding more test cases
- [x] Code cleanup and code documentation
- [x] Creating a separate document page for JSON module
- [x] And other things I forgot ;-)

### Custom JSONPath Benchmark

| Method                           | JsonPath                                          |         Mean |       Error |         Ratio |  Allocated | Alloc Ratio |
| -------------------------------- | ------------------------------------------------- | -----------: | ----------: | ------------: | ---------: | ----------: |
[...]

PR #1281, opened 2025-07-06
JsonPath Improvements

jsneedles:
I noticed #974 recently, and as a big user of JsonPath in C# -- I had to take a look.

There were a few very plain bugs that were present... But as I dug in more I realized that there was opportunity to make the JsonPath more closely aligned with the standard/rfc (https://www.rfc-editor.org/rfc/rfc9535) 

## Main Changes: 

- Added support for `Scan` filters for Index / Slicer / Multiple Index / Multiple Slicer
- Aligned `scan` recursive behavior with the spec (don't return self) and treat `.*` and `[*]` the same) 
- Added support for negative / from-the-end indices like `$.books[-1]` 
- Added support for `in` (case-insensitive) operator as well as `!` for NOT
- Added support for query filter groupings like `$[?(@.a && (@.b || @.c || @.z))]` - explicit parentheses threw errors previously
- Added support for parsing array literals (useful for `in`) so you can say `$[?(@.a in [1,2,3])]` 
- Fixes for a few matching bugs, especially for null handling 
- Supporting parsing quoted strings in single or double quotes
- Escaping check for unicode characters / code points

## New Tests
- I added a few new tests to validate my first fixes, but realized it'd be easier to lean on the JsonPath regression suite (https://cburgmer.github.io/json-path-comparison/).

There's an extensive set of cases that covers a wide range of parsing & evaluation scenarios, that this now successfully passes (mostly) 😄 .

There's a few remaining scenarios that fail, but wanted to get feedback / opinions on... Namely `large number indexers`. Right now we parse as `int` and it fails, if we switched that to `long`, the parsing would succeed, but I think trying to actually evaluate the index of an array that long won't work (and is pretty unrealistic anyhow). 

image

I currently left the tests without a consensus that threw errors as `inconclusive`.

## Future

There's definitely room for optimization, but I ensured the benchmark results are in line with the results before my PR. 

- the `in` operator especially right now is naive, it could definitely be made more efficient!

Would love to gather some feedback and I'll tweak accordingly!


jsneedles:

prvyk:
> There's a few remaining scenarios that fail, but wanted to get feedback / opinions on... Namely `large number indexers`. Right now we parse as `int` and it fails, if we switched that to `long`, the parsing would succeed, but I think trying to actually evaluate the index of an array that long won't work (and is pretty unrealistic anyhow).

I guess (without reading the code), there are some array length checks to handle indexes outside the array, and that ranges are capped?

If so, I think the proper solution would be to read as long, but internally set a int sentinel value for too large values, and then let the length checks handle this. Note that the C# maximum is Array.MaxValue which is a tiny bit smaller than int.MaxValue so there's room for large sentinel values that do not need to worry about collisions with an existing index - or about changing the variable type. Possibly no need to add special sentinel checks. The advantage is being slightly more compatible without complicating the implementation code.

There was a similar scenario with *SCAN cursor values and I made a similar choice to change the code to accept longs even though builtin Garnet objects would never get there as they are implemented now. Though I just changed to long throughout (there was very little to change) and didn't need a sentinel. I suspect JsonPath is a bit more complicated though?

(There was some extra motivation so it's not a complete parallel: It's possible to design custom objects and maybe someone might choose to use long cursors. Cursor values are opaque and allowed to pack more information than a single counter so the array/collection limits aren't definitive).

TalZaccai:
@Vijay-Nirmal Could you also take a look since this refers to your original PR?

jsneedles:
> > There's a few remaining scenarios that fail, but wanted to get feedback / opinions on... Namely `large number indexers`. Right now we parse as `int` and it fails, if we switched that to `long`, the parsing would succeed, but I think trying to actually evaluate the index of an array that long won't work (and is pretty unrealistic anyhow).
> 
> I guess (without reading the code), there are some array length checks to handle indexes outside the array, and that ranges are capped?
> 
> If so, I think the proper solution would be to read as long, but internally set a int sentinel value for too large values, and then let the length checks handle this. Note that the C# maximum is Array.MaxValue which is a tiny bit smaller than int.MaxValue so there's room for large sentinel values that do not need to worry about collisions with an existing index - or about changing the variable type. Possibly no need to add special sentinel checks. The advantage is being slightly more compatible without complicating the implementation code.
> 
> There was a similar scenario with *SCAN cursor values and I made a similar choice to change the code to accept longs even though builtin Garnet objects would never get there as they are implemented now. Though I just changed to long throughout (there was very little to change) and didn't need a sentinel. I suspect JsonPath is a bit more complicated though?
> 
> (There was some extra motivation so it's not a complete parallel: It's possible to design custom objects and maybe someone might choose to use long cursors. Cursor values are opaque and allowed to pack more information than a single counter so the array/collection limits aren't definitive).

Thanks! 

I think I got the gist of what you're suggesting is to just "cheat" 😄  -- so I went ahead and did that.

Since it's unrealistic for any array of these lengths to be had (and I never knew about `Array.MaxLength` so thx! )-- I just changed it to parse as a `long` and if it's outside the `Array.MaxLength` range, I just capped it at that. 

The other option of changing to `long` throughout would obviously work, but then when we go to do the array indexer on the `JsonArray` - it wouldn't work because it expects an `int` 

Let me know if you were thinking of something different.

prvyk:
> 
> Thanks!
> 
> I think I got the gist of what you're suggesting is to just "cheat" 😄 -- so I went ahead and did that.
> 
...
> 
> Let me know if you were thinking of something different.

Yep, I was thinking of what you did. It gives us the right result in any realistic case - a C# array can't contain more items.

I suspect that if such large values are ever used in real code it's because someone is using large values in a slice as a stand-in for infinite instead of doing an open start/end (maybe a jsonpath template is buried so deep they can't modify it easily, and can only supply integers to a function/API that creates it. 'Eh, let's give +-(2^53-1) and call it a day. We'll never have so many items anyway.').

prvyk:
As a style thing, you may wish to have a tiny function that parses the _expression with start, length, and returns an int (internally it parses as long and caps). That would save the casts later on and repeating this code.

modules/GarnetJSON/JSONPath/JsonPath.cs
```diff
@@ -360,18 +366,17 @@ private PathFilter ParseArrayIndexer(char indexerCloseChar)
                         }
 
                         var indexer = _expression.AsSpan(start, length);
-                        int index = int.Parse(indexer, CultureInfo.InvariantCulture);
+                        int index = EnsureIndexWithinArrayBounds(long.Parse(indexer, CultureInfo.InvariantCulture));
```
TalZaccai: I would probably do `bool TryParseIndex(string value, out int index)` that does the long.Parse + validation, that way it is contained within that method, in case we ever want to change it. Also, if it returns false we should probably throw an exception rather than suppress it.

TalZaccai (changes requested):
LGTM overall, added one comment. 
Probably need to run dotnet format, there are a bunch of extra whitespaces all over :)

modules/GarnetJSON/JSONPath/JsonPath.cs
jsneedles: Thanks! I went ahead and made that change.

jsneedles:
> LGTM overall, added one comment. Probably need to run dotnet format, there are a bunch of extra whitespaces all over :)

Thanks! 

I just ran `dotnet format` and don't see any changes 🤔   Also did `dotnet format --verify-no-changes` a bunch of times and no errors.

(I've not used `dotnet format` before, so maybe im missing some step/setup)

TalZaccai:
> > LGTM overall, added one comment. Probably need to run dotnet format, there are a bunch of extra whitespaces all over :)
> 
> Thanks!
> 
> I just ran `dotnet format` and don't see any changes 🤔 Also did `dotnet format --verify-no-changes` a bunch of times and no errors.
> 
> (I've not used `dotnet format` before, so maybe im missing some step/setup)

No worries, I ended up running code cleanup and that seems to have done the trick, hope that's ok.

PR #1288, opened 2025-07-09
Add Txn Support for HEXPIRE, HPEXPIRE, HPEXPIREAT

vazois:
Fixes issue #1269

### DIFF
--- a/libs/server/Transaction/TxnKeyManager.cs
+++ b/libs/server/Transaction/TxnKeyManager.cs
@@ -145,6 +145,10 @@ namespace Garnet.server
                 RespCommand.HSETNX => HashObjectKeys((byte)HashOperation.HSETNX),
                 RespCommand.HSTRLEN => HashObjectKeys((byte)HashOperation.HSTRLEN),
                 RespCommand.HVALS => HashObjectKeys((byte)HashOperation.HVALS),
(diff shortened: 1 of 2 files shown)


prvyk:
There are far more missing commands in Txn than these relatively rarely used commands. In Hash, H*TLL, HSCAN; Much of SortedSet operations (including ZMPOP and ZUNION); A third of List operations (including LPOS and LMOVE) and so on. We can see in a side-by-side diff of #1271 how much was missing (and that's still missing the blocking commands for instance):

https://github.com/microsoft/garnet/pull/1271/files?diff=split

vazois:
> There are far more missing commands in Txn than these relatively rarely used commands. In Hash, H*TLL, HSCAN; Much of SortedSet operations (including ZMPOP and ZUNION); A third of List operations (including LPOS and LMOVE) and so on. We can see in a side-by-side diff of #1271 how much was missing (and that's still missing the blocking commands for instance):
> 
> https://github.com/microsoft/garnet/pull/1271/files?diff=split

yes, that might be the case. This PR fixes HEXPIRE issue only as titled.
There was no clear indication that someone was working on the issue since no one was assigned to it.
I will assign you for the rest of the commands

TalZaccai (approved):

PR #1290, opened 2025-07-09
Fix local connection protection check

prvyk:
Following #1256, the check for --enable-X-command=local was accidentally inverted, and it passed only remote connections and not local connections. Fix this and add a test to ensure this remains correct.


test/Garnet.test/GarnetServerConfigTests.cs
```diff
@@ -868,7 +925,7 @@ public async Task MultiTcpSocketTest()
             var addresses = Dns.GetHostAddresses(hostname);
             addresses = [.. addresses, IPAddress.IPv6Loopback, IPAddress.Loopback];
 
-            var endpoints = addresses.Select(address => new IPEndPoint(address, TestUtils.TestPort)).ToArray();
+            var endpoints = addresses.Distinct().Select(address => new IPEndPoint(address, TestUtils.TestPort)).ToArray();
```
prvyk, line 928: Very very rarely, there's an 'address already in use' failure here in github runs. Maybe some runner configuration returns a duplicate or loopback address? Adding a Distinct() may help here. It won't hurt.

vazois (approved):

PR #1296, opened 2025-07-15
Fixes for expiry-related commands

TalZaccai:
* Moved expiry time determination from storage to processing layer to enable proper AOF recovery of expiration times
* Encoding expiry timestamp in millisecond granularity + expiration option into store input arguments
* Commands affected: `[P]EXPIRE[AT]`, `Z[P]EXPIRE[AT]`, `H[P]EXPIRE[AT]`.
* Added tests to ensure AOF recovery of expiration times


libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -434,7 +435,17 @@ private bool NetworkEXPIRE<TGarnetApi>(RespCommand command, ref TGarnetApi stora
                 }
             }
 
-            var input = new RawStringInput(command, ref parseState, startIdx: 1, arg1: (byte)expireOption);
+            // Convert timeout to milliseconds
+            if (command == RespCommand.EXPIRE)
+                timeout *= 1000;
+
+            // Convert to expiration time
+            var expirationInMs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() + timeout;
```
vazois: I am wondering if we can simplify this
```suggestion
command == RespCommand.EXPIRE ? DateTimeOffset.UtcNow.AddSeconds(timeout).ToUnixTimeMilliseconds() :DateTimeOffset.UtcNow.AddMilliseconds(timeout).ToUnixTimeMilliseconds();
```

libs/server/ExpirationUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+            return (expiration, expirationOption);
+        }
+
+        /// <summary>
+        /// Encode a unix expiration timestamp in milliseconds and an ExpireOption enum into two integers
+        /// </summary>
+        /// <param name="expirationTimeInMs">Unix expiration timestamp in milliseconds</param>
+        /// <param name="expirationOption">Expiration option</param>
+        /// <returns>Encoded tail and head of expiration time and expiration option</returns>
+        internal static (int encodedTail, int encodedHead) EncodeExpirationToTwoInt32(long expirationTimeInMs, ExpireOption expirationOption)
```
vazois: why not use one format where everything is encoded in a single long value?

libs/server/Resp/Objects/HashCommands.cs
```diff
@@ -645,15 +626,24 @@ private unsafe bool HashExpire<TGarnetApi>(RespCommand command, ref TGarnetApi s
                 return AbortWithErrorMessage(Encoding.ASCII.GetBytes(string.Format(CmdStrings.GenericErrMustMatchNoOfArgs, "numFields")));
             }
 
-            var fieldsParseState = parseState.Slice(currIdx, numFields);
+            // Convert expiration to milliseconds
+            if (command == RespCommand.HEXPIRE || command == RespCommand.HEXPIREAT)
+                expiration *= 1000;
```
vazois: nit; maybe have them in separate switch statement to be clear what calculations are being done.

libs/server/Resp/Objects/SortedSetCommands.cs
```diff
@@ -1803,8 +1786,19 @@ private unsafe bool SortedSetExpire<TGarnetApi>(RespCommand command, ref TGarnet
                 return AbortWithErrorMessage(CmdStrings.GenericErrMustMatchNoOfArgs, "numMembers");
             }
 
+            // Convert to milliseconds
+            if (command == RespCommand.ZEXPIRE || command == RespCommand.ZEXPIREAT)
```
vazois: nit; maybe have them in separate switch statement to be clear what calculations are being done.

libs/server/Resp/KeyAdminCommands.cs
```diff
@@ -391,7 +391,8 @@ private bool NetworkEXPIRE<TGarnetApi>(RespCommand command, ref TGarnetApi stora
             }
 
             var key = parseState.GetArgSliceByRef(0);
-            if (!parseState.TryGetInt(1, out _))
+
+            if (!parseState.TryGetInt(1, out var timeout))
```
prvyk: Shouldn't this be TryGetLong?

(I wanted to comment on not checking negative values, but it turns out Redis allow them here and not in HEXPIRE...)

libs/server/ExpirationUtils.cs
TalZaccai: Because ObjectInput has 2 integer arguments

libs/server/Resp/KeyAdminCommands.cs
TalZaccai: Ha! Probably a good idea anyways...

vazois (approved):

badrishc (approved):

PR #1297, opened 2025-07-16
Add Timeout at Replica Sync Remote Comms

vazois:
This PR adds missing timeout for over the network communications between primary and secondary.

### DIFF
--- a/libs/server/Databases/DatabaseManagerBase.cs
+++ b/libs/server/Databases/DatabaseManagerBase.cs
@@ -220,7 +220,7 @@ namespace Garnet.server
         {
             try
             {
-                DoCompaction(db);
+                DoCompaction(db, isFromCheckpoint: true, logger);
                 var lastSaveStoreTailAddress = db.MainStore.Log.TailAddress;
                 var lastSaveObjectStoreTailAddress = (db.ObjectStore?.Log.TailAddress).GetValueOrDefault();
 
@@ -439,12 +439,13 @@ namespace Garnet.server
         /// </summary>
         /// <param name="db">Database to run compaction on</param>
         /// <param name="logger">Logger</param>
-        protected void DoCompaction(GarnetDatabase db, ILogger logger = null)
+        /// <param name="isFromCheckpoint">True if called from checkpointing, false if called from background task</param>
+        protected void DoCompaction(GarnetDatabase db, bool isFromCheckpoint = false, ILogger logger = null)
         {
             try
             {
-                // Periodic compaction -> no need to compact before checkpointing
-                if (StoreWrapper.serverOptions.CompactionFrequencySecs > 0) return;
+                // If periodic compaction is enabled and this is called from checkpointing, skip compaction
+                if (isFromCheckpoint && StoreWrapper.serverOptions.CompactionFrequencySecs > 0) return;
 
                 DoCompaction(db, StoreWrapper.serverOptions.CompactionMaxSegments,
                     StoreWrapper.serverOptions.ObjectStoreCompactionMaxSegments, 1,


Garnet.sln
```diff
@@ -1,4 +1,4 @@
-﻿
+﻿﻿
```
TalZaccai: Can probably undo the changes here?

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -278,7 +278,7 @@ public async Task<bool> SendCheckpoint()
                     clusterProvider.replicationManager.PrimaryReplId,
                     localEntry.ToByteArray(),
                     beginAddress,
-                    checkpointAofBeginAddress).ConfigureAwait(false);
+                    checkpointAofBeginAddress).WaitAsync(timeout, cts.Token);
```
TalZaccai: Would it make sense to add the `.ConfigureAwait(false)` after `WaitAsync`?

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -571,7 +572,7 @@ private unsafe (SectorAlignedMemory, int) ReadInto(IDevice device, ulong address
                 device.ReadAsync(address, (IntPtr)pbuffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
             else
                 device.ReadAsync(segmentId, address, (IntPtr)pbuffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
-            semaphore.Wait();
+            semaphore.Wait(clusterProvider.clusterManager.clusterTimeout, cts.Token);
```
TalZaccai: Should this be async as well?

Garnet.sln
vazois: Oops, my mistake. I reverted back to the original

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
vazois: Yes, it would. I added it back after the WaitAsync

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
vazois: I had to remove the unsafe specifier for this method, but I think it should be ok to make it WaitAsync.

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -559,19 +564,22 @@ private async Task SendObjectFiles(GarnetClientSession gcs, Guid token, Checkpoi
...
-                device.ReadAsync(segmentId, address, (IntPtr)pbuffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
-            semaphore.Wait();
+            unsafe
+            {
+                if (segmentId == -1)
+                    device.ReadAsync(address, (IntPtr)pbuffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
+                else
+                    device.ReadAsync(segmentId, address, (IntPtr)pbuffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
+            }
+            await semaphore.WaitAsync(clusterProvider.clusterManager.clusterTimeout, cts.Token);
```
TalZaccai: Can add ConfigureAwait(false) here too?

vazois: done!

priyanjgupta:
Sorry for this basic question I'm still getting familiar with the replication internals.
**Scenario**: What happens if a timeout occurs while the secondary is actively processing a file segment?
What happens:
**Primary side**: Throws TimeoutException → Exception is caught in the general handler → Returns false (marking the operation as failed)
**Secondary side**: Continues processing the segment → May successfully complete the write operation
Can this create any inconsistencies ? 
Is there a protocol-level mechanism that handles such mid-operation timeouts to prevent inconsistencies?
And if not, should the replication logic be concerned about such scenarios?
From my understanding, this might not lead to issues because a timeout causes `NetworkClusterInitiateReplicaSync` to fail, and the secondary would reset it's state. If replication is retried, it would reprocess the file from the beginning.
**_Could you please confirm whether the secondary safely rewrites the data in such cases, rather than rejecting it or ending up in an inconsistent state?Just want to validate that my understanding of the safety mechanism is correct._**
Thanks

vazois:
> Sorry for this basic question I'm still getting familiar with the replication internals. **Scenario**: What happens if a timeout occurs while the secondary is actively processing a file segment? What happens: **Primary side**: Throws TimeoutException → Exception is caught in the general handler → Returns false (marking the operation as failed) **Secondary side**: Continues processing the segment → May successfully complete the write operation Can this create any inconsistencies ? Is there a protocol-level mechanism that handles such mid-operation timeouts to prevent inconsistencies? And if not, should the replication logic be concerned about such scenarios? From my understanding, this might not lead to issues because a timeout causes `NetworkClusterInitiateReplicaSync` to fail, and the secondary would reset it's state. If replication is retried, it would reprocess the file from the beginning. **_Could you please confirm whether the secondary safely rewrites the data in such cases, rather than rejecting it or ending up in an inconsistent state?Just want to validate that my understanding of the safety mechanism is correct._** Thanks

This is a good observation and yes, the primary should rewrite the contents of the checkpoint because it starts rewriting from scratch. I added a test that demonstrates this.
When primary fails for any reason, the secondary should receive a response and cleanup any acquired locks and resources etc.
In that case, when re-initiating the sync path the files associated with the checkpoint will be overwritten.
If the secondary is unreachable, I added a timeout to the initiateReplicaSyncTask in order to avoid the secondary being stuck.

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -465,22 +469,23 @@ private async Task SendFileSegments(GarnetClientSession gcs, Guid token, Checkpo
...
+            logger?.LogInformation("<Begin sending checkpoint file segments {guid} {type} {startAddress} {endAddress} {batchSize}", token, type, startAddress, endAddress, batchSize);
             try
             {
                 while (startAddress < endAddress)
                 {
                     var num_bytes = startAddress + batchSize < endAddress ?
                         batchSize :
                         (int)(endAddress - startAddress);
-                    var (pbuffer, readBytes) = ReadInto(device, (ulong)startAddress, num_bytes);
+                    var (pbuffer, readBytes) = await ReadInto(device, (ulong)startAddress, num_bytes);
```
TalZaccai: `.ConfigureAwait(false)`?

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -523,9 +528,9 @@ private async Task SendObjectFiles(GarnetClientSession gcs, Guid token, Checkpoi
                     while (startAddress < size)
                     {
                         var num_bytes = startAddress + batchSize < size ? batchSize : (int)(size - startAddress);
-                        var (pbuffer, readBytes) = ReadInto(device, (ulong)startAddress, num_bytes, segment);
+                        var (pbuffer, readBytes) = await ReadInto(device, (ulong)startAddress, num_bytes, segment);
```
TalZaccai: `.ConfigureAwait(false)`?

libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSession.cs
```diff
@@ -523,9 +528,9 @@ private async Task SendObjectFiles(GarnetClientSession gcs, Guid token, Checkpoi
...
 
-                        resp = await gcs.ExecuteSendFileSegments(fileTokenBytes, (int)type, startAddress, pbuffer.GetSlice(readBytes), segment).ConfigureAwait(false);
+                        resp = await gcs.ExecuteSendFileSegments(fileTokenBytes, (int)type, startAddress, pbuffer.GetSlice(readBytes), segment).WaitAsync(replicaSyncTimeout, cts.Token).ConfigureAwait(false);
```
TalZaccai: nit: line too long

libs/cluster/Server/Replication/ReplicaOps/ReceiveCheckpointHandler.cs
```diff
@@ -65,6 +74,10 @@ public void ProcessFileSegments(int segmentId, Guid token, CheckpointFileType ty
 
             Debug.Assert(writeIntoCkptDevice != null);
             WriteInto(writeIntoCkptDevice, (ulong)startAddress, data, data.Length, segmentId);
+
+#if DEBUG
+            ExceptionInjectionHelper.WaitOnClear(ExceptionInjectionType.Replication_Timeout_On_Receive_Checkpoint).GetAwaiter().GetResult();
```
TalZaccai: `.ConfigureAwait(false)` before `GetAwaiter()`?

libs/cluster/Server/Replication/ReplicaOps/ReceiveCheckpointHandler.cs
```diff
@@ -77,27 +90,37 @@ public void ProcessFileSegments(int segmentId, Guid token, CheckpointFileType ty
...
-            else
-                device.WriteAsync((IntPtr)pbuffer.aligned_pointer, segmentId, address, (uint)numBytesToWrite, IOCallback, null);
-            writeCheckpointSemaphore.Wait();
+                if (segmentId == -1)
+                    device.WriteAsync((IntPtr)pbuffer.aligned_pointer, address, (uint)numBytesToWrite, IOCallback, null);
+                else
+                    device.WriteAsync((IntPtr)pbuffer.aligned_pointer, segmentId, address, (uint)numBytesToWrite, IOCallback, null);
 
-            pbuffer.Return();
+                _ = writeCheckpointSemaphore.Wait(replicaSyncTimeout, cts.Token);
```
TalZaccai, line 113: Why not do a WaitAsync here?

libs/cluster/Server/Replication/ReplicaOps/ReceiveCheckpointHandler.cs
```diff
@@ -77,27 +90,37 @@ public void ProcessFileSegments(int segmentId, Guid token, CheckpointFileType ty
...
+            }
+            catch (Exception ex)
+            {
+                logger?.LogError(ex, $"{nameof(ReceiveCheckpointHandler)}.WriteInto");
+                throw new Exception(ex.Message);
```
TalZaccai: Why not just rethrow ex?

libs/common/ExceptionInjectionHelper.cs
```diff
@@ -87,5 +87,20 @@ public static async Task WaitOnCondition(ExceptionInjectionType exceptionType)
                     await Task.Yield();
             }
         }
+
+        /// <summary>
+        /// Wait on clear condition
+        /// </summary>
+        /// <param name="exceptionType"></param>
+        /// <returns></returns>
+        public static async Task WaitOnClear(ExceptionInjectionType exceptionType)
```
TalZaccai: nit: Not sure if we have a proper naming convention for async methods, but should probably end with `Async`

libs/common/ExceptionInjectionHelper.cs
```diff
@@ -87,5 +87,20 @@ public static async Task WaitOnCondition(ExceptionInjectionType exceptionType)
...
+        {
+            var flag = ExceptionInjectionTypes[(int)exceptionType];
```
TalZaccai: nit: Maybe a bit cleaner?
```cs
var flag = ExceptionInjectionTypes[(int)exceptionType];
while (flag)
{
    await Task.Yield();
    flag = ExceptionInjectionTypes[(int)exceptionType];
}
```

TalZaccai (changes requested):
Added a few comments

libs/cluster/Server/Replication/ReplicaOps/ReceiveCheckpointHandler.cs
vazois: Because the execution is synchronous, due to writing from the network buffer. We will have to make ProcessMessage async at the top level.

libs/common/ExceptionInjectionHelper.cs
vazois: actually, I need to remove if because it is redundant

libs/cluster/Server/Replication/ReplicaOps/ReceiveCheckpointHandler.cs
vazois: You are right, the exception should be captured at ProcessMessage level. I will make the change.

PR #1300, opened 2025-07-16
Update publishing to use normal publish and SingleFile for Windows service

prvyk:
As of v1.0.78, a handrolled publish target is used to publish the Windows service for win x64 platform. So it can't use settings like SingleFile or ReadyRun (at least I couldn't manage to). The target is also incompatible with an artifact layout.
This PR changes createbinaries script and win=x64 publish target so:

* Normal publishing** is used for Windows service, making publish options available.
* Since a normal target is used, we can enable publishing Windows service as singlefile.
* The handrolled publish target is removed.

In addition the following changes are done too:

* Publish Windows service for win-arm64 too (the same reasoning for publishing with x64 applies I believe).
* Added parameters for script to allow setting up destination and build paths (default is identical to current script).
* Change Zipfile compression to use Deflate64. Windows 7+ supports it, so all supported Windows platforms can decompress. Compresses better with only a small change to running time (less than a second extra, result is very close to what -mx=9 with normal Deflate would achieve).

** It runs dotnet publish separately for the service. If that's a problem we could use a traversal project or msbuildproj files to only run one dotnet publish, but I doubt that's a problem - maybe it costs a little in runtime that's all.

### DIFF
--- a/.azure/pipelines/createbinaries.ps1
+++ b/.azure/pipelines/createbinaries.ps1
@@ -1,12 +1,12 @@
-<#$f
+<#
 .SYNOPSIS
-    This script is designed to publish GarnetServer into various platforms. 
+	This script is designed to publish GarnetServer into various platforms.
 
 .DESCRIPTION
 
-    Script to publish the GarnetServer executable from various Profiles. It will clean up all the files not needed and by default it will zip them all up in the output directory. 
+	Script to publish the GarnetServer executable from various Profiles. It will clean up all the files not needed and by default it will zip them all up in the output directory.
 	
-    Parameter: mode
+.PARAMETER mode
 		0 (default) = do both publish and zip
 		1 = publish only
 		2 = zip only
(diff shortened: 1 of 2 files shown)


prvyk:
I checked a singlefile build could run as a Windows service (sc create binpath= ...), it does work. 

Aside, I wonder what portable.7z is meant for. The portable publishing profile doesn't have a RID, so it just publishes an executable based on the build runner. @darrenge Do you know what it's for?

darrenge (changes requested):
I tried running this in the pipeline and I got this error:

Copy-Item : Cannot find path 
'D:\a\1\s\main\GarnetServer\bin\Release\net8.0\publish\win-x64\runtimes\win-x64\native\native_device.dll' because it 
does not exist.
At D:\a\1\s\.azure\pipelines\createbinaries.ps1:78 char:3
+         Copy-Item -Path $nativeRuntimePathFile -Destination $publishP ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (D:\a\1\s\main\G...tive_device.dll:String) [Copy-Item], ItemNotFoundExce 
   ption
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.CopyItemCommand

darrenge:
> I checked a singlefile build could run as a Windows service (sc create binpath= ...), it does work.
> 
> Aside, I wonder what portable.7z is meant for. The portable publishing profile doesn't have a RID, so it just publishes an executable based on the build runner. @darrenge Do you know what it's for?

As for portable.7z, it’s typically a compressed archive containing the output of a portable publish — meaning:
What portable.7z Usually Contains
- The compiled .dlls
- Supporting .json files like .deps.json and .runtimeconfig.json
- Any referenced assemblies
- Possibly config files or static assets
This format is handy for:
- CI/CD pipelines: Easy to transfer and unpack on target machines
- Cross-platform distribution: Since it’s not tied to a specific RID, it can run anywhere with a compatible runtime
- Minimal footprint: Compared to self-contained builds, it’s smaller and faster to publish

prvyk:
> I tried running this in the pipeline and I got this error:
> 
> Copy-Item : Cannot find path 'D:\a\1\s\main\GarnetServer\bin\Release\net8.0\publish\win-x64\runtimes\win-x64\native\native_device.dll' because it does not exist. At D:\a\1\s.azure\pipelines\createbinaries.ps1:78 char:3

Odd, it kept working locally (both Linux and Windows)... The fix I've pushed should work with the pipeline too. If I'm diagnosing this right. If it doesn't, is there a way to see the pipeline's build artifacts so we can see where these files went during publish?

>Cross-platform distribution: Since it’s not tied to a specific RID, it can run anywhere with a compatible runtime

In practice, the executable is compiled per the platform of the builder and will not run on other platforms, which let me wonder what the purpose is... Maybe I get it now. The user often can't run the executable, but "dotnet GarnetServer.dll" will run crossplatform. I forgot a .NET program could be started that way too.

darrenge (changes requested):
Now seeing this error:
Copy-Item : Cannot find path 
  Garnet.cluster -> D:\a\1\s\libs\cluster\bin\Release\net8.0\Garnet.cluster.dll
'D:\a\1\s\main\GarnetServer\bin\Release\net8.0\publish\win-x64\runtimes\win-x64\native\native_device.dll' because it 
does not exist.
At D:\a\1\s\.azure\pipelines\createbinaries.ps1:78 char:3
+         Copy-Item -Path $nativeRuntimePathFile -Destination $publishP ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (D:\a\1\s\main\G...tive_device.dll:String) [Copy-Item], ItemNotFoundExce 
   ption
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.CopyItemCommand

prvyk:
I've made another attempt. I suspect the issue was dotnet publish not playing nice when the former build was at the destination directory (i.e. not cleaned up in advance).

darrenge (changes requested):
Latest error:
runtime D:\a\1\s/main/GarnetServer/bin/Release/net8.0/publish/win-x64/runtimes/win-x64/native/native_device.dll for 
win-x64 does not exist
At D:\a\1\s\.azure\pipelines\createbinaries.ps1:86 char:4
+             throw "runtime $nativeRuntimePathFile for $publishFolder  ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (runtime D:\a\1\... does not exist:String) [], RuntimeException
    + FullyQualifiedErrorId : runtime D:\a\1\s/main/GarnetServer/bin/Release/net8.0/publish/win-x64/runtimes/win-x64/n 
   ative/native_device.dll for win-x64 does not exist

prvyk:
win-x64 is the last stage in the build. If anything was wrong it should have failed before. My theory is this:

A) The pipeline runs dotnet build -c Release on the GarnetServer csproj earlier (which isn't actually required for anything since we now do publish directly).
B) What's 'special' is that the runner is using win x64 architecture.
C) The build interferes with the publish step later (they don't use the same publish profile, yet publish is still 'lazy' and reuses files).

I've pushed an attempted fix (both remove all previous build artifacts and remove unnecessary step from external pipeline). Thanks for testing this! I can't recreate git pipeline environment to test it myself...

darrenge:
I am wondering if it really is worth it to get these changes in.

prvyk:
> I am wondering if it really is worth it to get these changes in.

You have a point. I believe these changes are worthwhile, but doing ping-pong like that isn't.

I'll need to either reproduce the environment, or do this very incrementally. I think I'll close this and maybe try to get some smaller things in separately in the future. Sorry, didn't expect this problem at all.

darrenge:
I agree ... good call with closing it out. I do a lot of pipeline work and it is like this. It is a hassle almost like the old "Punch Card" days.

PR #1303, opened 2025-07-18
INCRBYFLOAT: fix double's char counting and new API

prvyk:
This PR fixes two problems:

1.
> INCRBYFLOAT key 0.1
> > Error: Server closed the connection
> > (Backend) System.ArgumentException: Destination is too short. (Parameter 'destination') 

Problem is NumUtils.CountCharsInDouble() uses Log10(value) and 0.1 counts as -1 digit.

2. Memory allocation/pinning for the new API was wrong: It should be done by caller, otherwise the runtime may mess with the memory once the API call was done but before the result being printed to the user (this was masked by the runtime usually using similar address and not messing with it, and by NUnit using same global context between tests - so once it was initialized the runtime kept the address).

### DIFF
--- a/libs/server/API/GarnetApi.cs
+++ b/libs/server/API/GarnetApi.cs
@@ -250,11 +250,9 @@ namespace Garnet.server
             => Increment(key, out output, -decrementCount);
 
         /// <inheritdoc />
-        public GarnetStatus IncrementByFloat(ArgSlice key, out ArgSlice output, double val,
-                                             Span<byte> pinnedOutputBuffer)
+        public GarnetStatus IncrementByFloat(ArgSlice key, ref ArgSlice output, double val)
         {
             SessionParseState parseState = default;
-            output = ArgSlice.FromPinnedSpan(pinnedOutputBuffer);
 
             var input = new RawStringInput(RespCommand.INCRBYFLOAT, ref parseState, BitConverter.DoubleToInt64Bits(val));
             _ = Increment(key, ref input, ref output);
@@ -267,7 +265,6 @@ namespace Garnet.server
             switch (errorFlag)
             {
                 case OperationError.INVALID_TYPE:
-                    return GarnetStatus.WRONGTYPE;
                 case OperationError.NAN_OR_INFINITY:
                     return GarnetStatus.WRONGTYPE;
                 default:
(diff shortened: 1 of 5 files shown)


libs/server/API/GarnetApi.cs
```diff
@@ -250,27 +250,26 @@ public GarnetStatus Decrement(ArgSlice key, out long output, long decrementCount
             => Increment(key, out output, -decrementCount);
 
         /// <inheritdoc />
-        public GarnetStatus IncrementByFloat(ArgSlice key, out ArgSlice output, double val)
+        public GarnetStatus IncrementByFloat(ArgSlice key, out ArgSlice output, double val,
```
TalZaccai: Can't we do this like in BasicCommands.NetworkIncrement? That way we just pass the ArgSlice by ref instead of having 2 parameters.

libs/server/API/GarnetApi.cs
```diff
@@ -250,27 +250,26 @@ public GarnetStatus Decrement(ArgSlice key, out long output, long decrementCount
...
 
             switch (errorFlag)
             {
-                case OperationError.SUCCESS:
-                    return GarnetStatus.OK;
-                case OperationError.NAN_OR_INFINITY:
-                    return GarnetStatus.WRONGTYPE;
                 case OperationError.INVALID_TYPE:
                     return GarnetStatus.WRONGTYPE;
+                case OperationError.NAN_OR_INFINITY:
```
TalZaccai, line 268: nit: Can just do a fallthrough here since both return WRONGTYPE

libs/server/API/GarnetApi.cs
```diff
@@ -279,20 +278,22 @@ public GarnetStatus IncrementByFloat(ArgSlice key, out ArgSlice output, double v
...
+                case GarnetStatus.OK:
+                    _ = NumUtils.TryReadDouble(_output.ReadOnlySpan, out dbl);
                     break;
-                case OperationError.NAN_OR_INFINITY:
-                    dbl = double.NaN;
+                case GarnetStatus.WRONGTYPE:
+                default:
+                    var errorFlag = (OperationError)_output.Span[0];
+
+                    if (errorFlag == OperationError.NAN_OR_INFINITY)
```
TalZaccai: nit: ```dbl = errorFlag == OperationError.NAN_OR_INFINITY ? double.NaN : 0;```

TalZaccai (changes requested):
Couple of small comments

libs/server/API/GarnetApi.cs
prvyk: Excellent idea. Done.

libs/server/API/GarnetApi.cs
prvyk: Done.

PR #1305, opened 2025-07-18
Fixing replica checkpointing divergence with explicit CommitAOF calls

### LINKED ISSUE #1301: Out of band CommitAOF cause cluster replication failure
### Describe the bug

If you got a cluster setup, and you call `StoreApi.CommitAOF()` on a replica, replication breaks.

It breaks pretty badly too, will not recover until you remove replication (with `REPLICAOF NO ONE`) and re-replicate (with `CLUSTER REPLICATE <node id>`) the failing node.

### Some Background

We're hosting `GarnetServer` in a little wrapper process, the wrapper process periodically calls `GarnetServer.Store.CommitAOF` on a task, and there are multiple processes (on multiple machines) configured in a cluster with a replication factor of 1 (every primary has 1 replica).

Our intent with this periodic commit is to handle a "cold recovery" case with reduced (but not zero) data loss.  If both the primary and its replica go down, and only the replica node comes back up we'd like some confidence that the replica will have only dropped (some small replication lag + the commit frequency) worth of data when it is promoted to replace the (still dead) primary.  We want this because, for many use cases, there's a big (sometimes catastrophically big) difference between a completely empty/cold cache and one that merely has a lower than expected hit rate.

### The Bug

Looking back through tests and notes, I think this was working in October 2024 - but I also think that was probably by accident as I've looked more into how replication works.

The bug as of at least `.76` is calling `StoreApi.CommitAOF` on a replica immediately prevents future replication due an AOF desync.  This manifests as something like:

>  Divergent AOF Stream recordLength:20; previousAddress:492; currentAddress:492; nextAddress:512; tailAddress:540

getting logged ([thrown here](https://github.com/microsoft/garnet/blob/297d0a3532f75204d424bd3d47225e13bdc873af/libs/cluster/Server/Replication/ReplicaOps/ReplicationReplicaAofSync.cs#L78)) on the replica.  This eventually kills the [`ReplicaSyncTask` on the primary](https://github.com/microsoft/garnet/blob/297d0a3532f75204d424bd3d47225e13bdc873af/libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs#L98).  The primary doesn't appear to attempt to re-establish that connection without outside prompting (either a restart, cluster modification, or [looking at the source] a `CLUSTER AOFSYNC` command).

I've put [this in a branch](https://github.com/microsoft/garnet/tree/replicaCheckpointDivergence) too, but it's basically:
```csharp
// Use case here is, outside of the cluster, period COMMITAOFs are requested.
// Done so in recovery scenarios, if a primary does NOT come back there's still confidence
// a replica has recent data.

var replica_count = 1;// Per primary
var primary_count = 1;
var nodes_count = primary_count + primary_count * replica_count;
ClassicAssert.IsTrue(primary_count > 0);

context.CreateInstances(nodes_count, disableObjects: false, enableAOF: true, useTLS: true, tryRecover: false, FastAofTruncate: true, CommitFrequencyMs: -1);
context.CreateConnection(useTLS: true);
var (shards, _) = context.clusterTestUtils.SimpleSetupCluster(primary_count, replica_count, logger: context.logger);

shards = context.clusterTestUtils.ClusterShards(0, context.logger);
ClassicAssert.AreEqual(1, shards.Count);
ClassicAssert.AreEqual(1, shards[0].slotRanges.Count);
ClassicAssert.AreEqual(0, shards[0].slotRanges[0].Item1);
ClassicAssert.AreEqual(16383, shards[0].slotRanges[0].Item2);

var primaryEndPoint = (IPEndPoint)context.endpoints[0];
var replicaEndPoint = (IPEndPoint)context.endpoints[1];

// Check cluster is in good state pre-commit
{
    var firstVal = Guid.NewGuid().ToString();
    var setPrimaryRes = (string)context.clusterTestUtils.Execute(primaryEndPoint, "SET", ["test-key", firstVal]);
    ClassicAssert.AreEqual("OK", setPrimaryRes);
    var getPrimaryRes = (string)context.clusterTestUtils.Execute(primaryEndPoint, "GET", ["test-key"]);
    ClassicAssert.AreEqual(firstVal, getPrimaryRes);

    context.clusterTestUtils.WaitForReplicaAofSync(0, 1);

    var readonlyReplicaRes = (string)context.clusterTestUtils.Execute(replicaEndPoint, "READONLY", []);
    ClassicAssert.AreEqual("OK", readonlyReplicaRes);
    var getReplicaRes = (string)context.clusterTestUtils.Execute(replicaEndPoint, "GET", ["test-key"]);
    ClassicAssert.AreEqual(firstVal, getReplicaRes);
}

// Commit outside of cluster logic
context.nodes[0].Store.CommitAOF();
context.nodes[1].Store.CommitAOF();  // <--- comment this out, and test passes

// Check cluster remains in good state
{
    var secondVal = Guid.NewGuid().ToString();
    var setPrimaryRes = (string)context.clusterTestUtils.Execute(primaryEndPoint, "SET", ["test-key2", secondVal]);
    ClassicAssert.AreEqual("OK", setPrimaryRes);
    var getPrimaryRes = (string)context.clusterTestUtils.Execute(primaryEndPoint, "GET", ["test-key2"]);
    ClassicAssert.AreEqual(secondVal, getPrimaryRes);

    // This dies horribly (if context.nodes[1].Store.CommitAOF() remains uncommented), so use a Sleep instead
    //context.clusterTestUtils.WaitForReplicaAofSync(0, 1);
    Thread.Sleep(5_000);

    var readonlyReplicaRes = (string)context.clusterTestUtils.Execute(replicaEndPoint, "READONLY", []);
    ClassicAssert.AreEqual("OK", readonlyReplicaRes);
    var getReplicaRes = (string)context.clusterTestUtils.Execute(replicaEndPoint, "GET", ["test-key2"]);
    ClassicAssert.AreEqual(secondVal, getReplicaRes);
}
```

### Expected behavior

I would expect calls to `StoreApi.CommitAOF` to not break cluster replication on "success".

### Screenshots

_No response_

### Release version

v1.0.76

### IDE

_No response_

### OS version

Linux, specially `CBL-Mariner 2.0.20250701 (x86_64)`.
This reproduces on Windows 11 too.

### Additional context

To my mind, there are a few things to figure out:
 - [ ] Should `StoreApi.CommitAOF` work on a replica?  If not, how should it fail?
       * I say no - there should be a `TryCommitAOF` which returns `false` if called on a replica.
       * It is difficult for a hosting process to know if the current node is a replica, and impossible for it to prevent a node's role from changing before a call to `CommitAOF`.
 - [ ] How should a replica behave if its local AOF has diverged from its primary?
       * I would expect a `WARNING` log, and for the replica to discard its local AOF.  A replica having writes not on is primary is an illegal state, and leaving a replica "up" but not capable of recovering is a dangerous state to be in.
       * I could see an argument for a setting to control this behavior.
 - [ ] Will a primary recover from `AofSyncTaskInfo.ReplicaSyncTask` being killed due to replica action?
       * This one is hard to sus out.  Based on logging and a quick read of the code, I don't _think_ it will.  But it should.
       * Though without changes to divergent AOF behavior, the primary will still never succeed in replicating.

I can see argument the other way for all of these, but those are my takes.

kevin-montrose:
Fixes #1301

TODOs:
 - [x] Repro test for basic issue
 - [x] Prevent role changes during commits
 - [x] Test for async divergence killing `ReplicaSyncTask`
 - [x] ReplicaSync task recovery (behind a setting, `ClusterReplicationReestablishmentTimeout`)
 - [x] Tests for new config
 - [x] Test for "cold" recovery scenario

## Fix 1: Prevent CommitAOF on Replicas

Described in more detail in #1301, and replicated in test `ClusterReplicationManualCheckpointing`.  Basically, if you take a commit on replica it's AOF file will diverge and future replication will fail.

This prevents taking that commit with `StoreApi`, holding a (new) lock on the cluster config while doing the check to prevent a TOCTOU issue.

The [`ClusterReplicationManualCheckpointing` test](https://github.com/microsoft/garnet/blob/192e0d5415c32ba8d822edbfcc7931d429080565/test/Garnet.test.cluster/ReplicationTests/ClusterReplicationBaseTests.cs#L1278) validates this fix.

## Fix 2: Allow Replicas to resume synchronization with Primaries automatically

A discovery while investigating the first issue is that _should_ something interrupt the stream of `CLUSTER|APPENDLOG` commands from a Primary to a Replica the stream will not be resumed without intervention.  This is a bit tricky, because detecting that replication has broken isn't trivial.

This fix adds a new setting (`ClusterReplicationReestablishmentTimeout`) which, when non-zero, enables a Replica to detect when replication has failed and re-issue sync resuming commands (either `CLUSTER|initiate_replica_sync` or `CLUSTER|attach_sync`) against the Primary.

This is subtle, so I'm going to spell out my understanding of replication and what I'm changing.

When we a Replica is being synced with a Primary (either because it started in that state, or `CLUSTER` commands were added to change it into a Replica):
 1. The Replica connects to the Primary.
 2. The Replica sends (either, depending on settings) `CLUSTER|attach_sync` or `CLUSTER|initiate_replica_sync` to the Primary
 3. Upon receiving that message, the Primary opens a connection to the Replica and begins replicating the underlying log by sending `CLUSTER|APPEND_LOG` commands on a background task
     * This task is implemented with the [`AofSyncTaskInfo.ReplicaSyncTask`](https://github.com/microsoft/garnet/blob/67bcf6024bf106ae54cca3554fba711b746e0004/libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs#L98) method
     * If the `ReplicaSyncTask` faults, it disposes the `GarnetClient` it was using to send the `CLUSTER|APPEND_LOG` commands

I am making the following changes:
 1. `IClusterSession` now tracks if it's being used to `CLUSTER|APPEND_LOG` commands
    * Along with existing tracking of remote node ids, this allows us to identify _on the Replica_ if we're actually synching with a Primary - not just configured to, but actively doing so
    * The `Dispose` in the `ReplicaSyncTask` method (which runs on the Primary, remember) ensures that the `IClusterSession` on the Replica will be torn down if something goes wrong in replication on either end (that is, a fault in the Primary's `TsavoriteLogScanSingleIterator` will kill the task just as an error in the Replica's handling of the `CLUSTER|APPEND_LOG` command will)
 2. When a Replica [receives a `CLUSTER|GOSSIP`](https://github.com/microsoft/garnet/blob/192e0d5415c32ba8d822edbfcc7931d429080565/libs/cluster/Session/RespClusterBasicCommands.cs#L408) and has `ClusterReplicationReestablishmentTimeout > 0`, we check to see if it is actively syncing with its Primary per 1
     * We wait for a `CLUSTER|GOSSIP` from the Primary so we don't attempt to synchronize if the Primary is likely to be down
     * This is also throttled, so we don't attempt to re-establish replication too frequently
     * This is implemented in [`ReplicationManager.EnsureReplication`](https://github.com/microsoft/garnet/blob/192e0d5415c32ba8d822edbfcc7931d429080565/libs/cluster/Server/Replication/ReplicationManager.cs#L177)
  3. On the Primary side, duplicate synchronization streams are prevented by the [`ReplicaSyncSessionTaskStore`](https://github.com/microsoft/garnet/blob/main/libs/cluster/Server/Replication/PrimaryOps/ReplicaSyncSessionTaskStore.cs) rejecting duplicate sync sessions for the same Replica.

The net effect should be that, in the event a Replica is not synching with its primary _but is communicating with it_ the Replica will resume that synchronization quickly without manual intervention.

The [`ReplicaSyncTaskFaultsRecoverAsync` test](https://github.com/microsoft/garnet/blob/192e0d5415c32ba8d822edbfcc7931d429080565/test/Garnet.test.cluster/ReplicationTests/ClusterReplicationBaseTests.cs#L1344C27-L1344C60) validates this.  Note this test depends on exception injection (to fault synchronization on the Replica side of things), so is DEBUG only.

There are a few open questions here:
 - [x] Is `EnsureReplication` correctly resuming synchronization?  In particular, disk vs diskless replication and `force` and `tryAddReplica` parameters are a bit opaque to me.  Also, maybe we DRY that whole thing up - there are a few different call sites doing the same conditional checks.
 - [x] Can we prevent a TOCTOU race in `EnsureReplication`?  It's possible with current code for the role of the Replica to change, or for it to start replicating a different primary between deciding to resync and actually issuing the commands.  I think the correct thing is probably to hold the `recoverLock` but the existing lock acquisitions in synchronization don't play nicely there.

## Fix 3: Tests for relevant scenarios

To keep our (now firmed up) deployment scenario from breaking in the future, I'm adding tests to cover our topology and failure cases.

The big change here is allowing Replicas to [restore from disk (gated behind `ClusterReplicaResumeWithData`)](https://github.com/microsoft/garnet/blob/43b8b33a57f28b35d79f4d209661786b69d283fb/libs/cluster/Server/Replication/ReplicationManager.cs#L468), and NOT resetting the log when syncing with a Primary [until after the connection from Replica -> Primary is established](https://github.com/microsoft/garnet/blob/d17d4134ab014a3b7088f85df78f3d954c443e19/libs/cluster/Server/Replication/ReplicaOps/ReplicaReceiveCheckpoint.cs#L85).

### DIFF
--- a/libs/server/Objects/SortedSetGeo/GeoHash.cs
+++ b/libs/server/Objects/SortedSetGeo/GeoHash.cs
@@ -303,8 +303,8 @@ namespace Garnet.server
         /// </summary>
         public static bool GetDistanceWhenInRectangle(double widthMts, double heightMts, double latCenterPoint, double lonCenterPoint, double lat2, double lon2, ref double distance)
         {
-            var lonDistance = Distance(lat2, lon2, latCenterPoint, lon2);
-            var latDistance = Distance(lat2, lon2, lat2, lonCenterPoint);
+            var lonDistance = Distance(lat2, lon2, lat2, lonCenterPoint);
+            var latDistance = Distance(lat2, lon2, latCenterPoint, lon2);
             if (lonDistance > widthMts / 2 || latDistance > heightMts / 2)
             {
                 return false;
--- a/modules/GarnetJSON/GarnetJsonObject.cs
+++ b/modules/GarnetJSON/GarnetJsonObject.cs
@@ -40,8 +40,12 @@ namespace GarnetJSON
     /// </summary>
     public class GarnetJsonObject : CustomObjectBase
     {
-        private static readonly JsonSerializerOptions DefaultJsonSerializerOptions = new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
-        private static readonly JsonSerializerOptions IndentedJsonSerializerOptions = new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping, WriteIndented = true };
+        private static readonly JsonSerializerOptions DefaultJsonSerializerOptions =
+            new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
+
+        private static readonly JsonSerializerOptions IndentedJsonSerializerOptions =
+            new JsonSerializerOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping, WriteIndented = true };
+
         private static readonly byte[] OpenBoxBracket = Encoding.UTF8.GetBytes("[");
         private static readonly byte[] CloseBoxBracket = Encoding.UTF8.GetBytes("]");
         private static readonly byte[] OpenCurlyBracket = Encoding.UTF8.GetBytes("{");
@@ -105,7 +109,9 @@ namespace GarnetJSON
         /// <summary>
         /// Disposes the <see cref="GarnetJsonObject"/> instance.
         /// </summary>
-        public override void Dispose() { }
+        public override void Dispose()
+        {
+        }
 
         /// <inheritdoc/>
         public override unsafe void Scan(long start, out List<byte[]> items, out long cursor, int count = 10,
@@ -122,7 +128,8 @@ namespace GarnetJSON
         /// <param name="newLine">The string to use for new lines.</param>
         /// <param name="space">The string to use for spaces.</param>
         /// <returns>True if the operation is successful; otherwise, false.</returns>
-        public bool TryGet(ReadOnlySpan<ArgSlice> paths, List<byte[]> output, out ReadOnlySpan<byte> errorMessage, string? indent = null, string? newLine = null, string? space = null)
+        public bool TryGet(ReadOnlySpan<ArgSlice> paths, List<byte[]> output, out ReadOnlySpan<byte> errorMessage,
+            string? indent = null, string? newLine = null, string? space = null)
         {
             if (paths.Length == 1)
             {
@@ -137,6 +144,7 @@ namespace GarnetJSON
                 {
                     output.Add(Comma);
                 }
+
                 isFirst = false;
 
                 output.Add(DoubleQuotes);
@@ -148,6 +156,7 @@ namespace GarnetJSON
                     return false;
                 }
             }
+
             output.Add(CloseCurlyBracket);
 
             errorMessage = default;
@@ -165,7 +174,8 @@ namespace GarnetJSON
         /// <param name="newLine">The string to use for new lines. (ignored with generic format)</param>
         /// <param name="space">The string to use for spaces. (ignored with generic format)</param>
         /// <returns>True if the operation is successful; otherwise, false.</returns>
-        public bool TryGet(ReadOnlySpan<byte> path, List<byte[]> output, out ReadOnlySpan<byte> errorMessage, string? indent = null, string? newLine = null, string? space = null)
+        public bool TryGet(ReadOnlySpan<byte> path, List<byte[]> output, out ReadOnlySpan<byte> errorMessage,
+            string? indent = null, string? newLine = null, string? space = null)
         {
             try
             {
@@ -177,7 +187,10 @@ namespace GarnetJSON
 
                 if (path.Length == 0)
                 {
-                    output.Add(JsonSerializer.SerializeToUtf8Bytes(rootNode, indent is null && newLine is null && space is null ? DefaultJsonSerializerOptions : IndentedJsonSerializerOptions));
+                    output.Add(JsonSerializer.SerializeToUtf8Bytes(rootNode,
+                        indent is null && newLine is null && space is null
+                            ? DefaultJsonSerializerOptions
+                            : IndentedJsonSerializerOptions));
                     return true;
                 }
 
@@ -192,10 +205,15 @@ namespace GarnetJSON
                     {
                         output.Add(Comma);
                     }
+
                     isFirst = false;
 
-                    output.Add(JsonSerializer.SerializeToUtf8Bytes(item, indent is null && newLine is null && space is null ? DefaultJsonSerializerOptions : IndentedJsonSerializerOptions));
+                    output.Add(JsonSerializer.SerializeToUtf8Bytes(item,
+                        indent is null && newLine is null && space is null
+                            ? DefaultJsonSerializerOptions
+                            : IndentedJsonSerializerOptions));
                 }
+
                 output.Add(CloseBoxBracket);
                 return true;
             }
@@ -216,7 +234,8 @@ namespace GarnetJSON
         /// <param name="errorMessage">The error message if the operation fails.</param>
         /// <returns>The result of the set operation.</returns>
         /// <exception cref="JsonException">Thrown when there is an error in JSON processing.</exception>
-        public SetResult Set(ReadOnlySpan<byte> path, ReadOnlySpan<byte> value, ExistOptions existOptions, out ReadOnlySpan<byte> errorMessage)
+        public SetResult Set(ReadOnlySpan<byte> path, ReadOnlySpan<byte> value, ExistOptions existOptions,
+            out ReadOnlySpan<byte> errorMessage)
         {
             try
             {
@@ -253,7 +272,8 @@ namespace GarnetJSON
                     }
 
                     // Find parent node using parent path
-                    var parentNode = rootNode.SelectNodes(GetParentPath(pathStr, out var pathParentOffset)).FirstOrDefault();
+                    var parentNode = rootNode.SelectNodes(GetParentPath(pathStr, out var pathParentOffset))
+                        .FirstOrDefault();
                     if (parentNode is null)
                     {
                         return SetResult.ConditionNotMet;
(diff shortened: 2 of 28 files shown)


libs/cluster/Server/Replication/ReplicaOps/ReplicaReceiveCheckpoint.cs
```diff
@@ -35,14 +36,16 @@ public bool TryReplicateDiskbasedSync(
             bool background,
             bool force,
             bool tryAddReplica,
+            bool upgradeLock,
+            bool allowReplicaResetOnFailure,
```
vazois: as discussed, maybe we can add all these options in a struct?

libs/cluster/Server/Replication/ReplicationManager.cs
```diff
@@ -167,6 +171,93 @@ public ReplicationManager(ClusterProvider clusterProvider, ILogger logger = null
...
+                if (Interlocked.CompareExchange(ref lastEnsureReplicationAttempt, newLastEnsureReplicationAttempt, oldLastEnsureReplicationAttempt) != oldLastEnsureReplicationAttempt)
+                {
+                    // Another attempt was made, somehow, so bail
+                    return;
+                }
+
+                logger.LogInformation("Beginning resync to {primaryId} after replication session failed", primaryId);
+
+                ReadOnlySpan<byte> errorMessage;
+                var success =
```
vazois: I think we need to make async if it is possible because it executes synchronously on the gossip path and recovery could take some time.
It should be safe to use background: true or create your own background task. You just need to make sure the lock is released when necessary or its release is delegated to the TryReplicate call.

libs/cluster/Session/RespClusterBasicCommands.cs
```diff
@@ -404,6 +404,12 @@ private bool NetworkClusterGossip(out bool invalidParameters)
                     SendAndReset();
             }
 
+            // After each GOSSIP, ensure cluster connections for replication are in a good state
+            if (Server is GarnetServerBase garnetServer)
```
vazois, line 408: Can you make use a flag here to make it user opt-in?

kevin-montrose: There is a config (`ClusterReplicationReestablishmentTimeout`, disabled when 0) it's just checked in the `ReplicationManager` instead of here.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -506,6 +506,22 @@ public class GarnetServerOptions : ServerOptions
...
 
+        /// <summary>
+        /// Seconds between attempts to re-establish replication between a Primary and Replica if the replication connection
+        /// has faulted.
+        /// 
+        /// 0 disables.
+        /// 
+        /// Attempts will only be made if Primary and Replica are exchanging GOSSIP messages.
+        /// </summary>
+        public int ClusterReplicationReestablishmentTimeout;
```
vazois: I think we should initialize this with the defaults.conf value. @TalZaccai will leaving it empty create an issue with how the options are being parsed?

vazois (approved):
Looks good! Minor comment waiting for @TalZaccai to confirm in GarnetServerOptions.cs has to have all parameters initialized with the defaults.conf value instead of leaving it empty.
Should we bump the version?

libs/server/Servers/GarnetServerOptions.cs
kevin-montrose: The default initialization here will match the `defaults.conf`, so is this just a style concern?

PR #1310, opened 2025-07-23
Added metrics for transaction commands received and executed successfully.

Xizt:
Since we are using transactions extensively in few of our clusters, we need metrics related to transaction commands that were received, executed to completion to setup appropriate monitoring and alerting on top of it.


Xizt:
@vazois @TalZaccai please help with review and feedback.

libs/server/Custom/CustomRespCommands.cs
```diff
@@ -25,9 +25,14 @@ private bool TryTransactionProc(byte id, CustomTransactionProcedure proc, int st
 
             LatencyMetrics?.Start(LatencyMetricsType.TX_PROC_LAT);
 
+
+            sessionMetrics?.incr_total_transaction_commands_received();
+
             var procInput = new CustomProcedureInput(ref parseState, startIdx: startIdx, respVersion: respProtocolVersion);
             if (txnManager.RunTransactionProc(id, ref procInput, proc, ref output))
             {
+                sessionMetrics?.incr_total_transaction_executed_successfully();
```
vazois: I am thinking we should count the number of failures instead, to not put too much overhead on the success path.

libs/server/Metrics/GarnetSessionMetrics.cs
```diff
@@ -47,10 +47,20 @@ public void incr_total_write_commands_processed(byte cmd) { }
 
         public void add_total_read_commands_processed(ulong count) { }
 
+        public void incr_total_transaction_commands_received(ulong count = 1) { }
+
+        public void incr_total_transaction_executed_successfully(ulong count = 1) { }
+
+        public void incr_total_transaction_transactions(ulong count = 1) { }
```
vazois: do we use this somewhere?

Xizt: Didn't find any references. Added as we have for similarity to other metrics

libs/server/Custom/CustomRespCommands.cs
Xizt: Updated, thanks

libs/server/Metrics/GarnetSessionMetrics.cs
vazois: I am not sure what you mean byt similar to other references. Maybe the code is deprecated for those. If we are not using it for something specific, we should remove it.

Xizt: @vazois  Will remove NulLGarnetSessionMetrics struct and IMetrics interface as well since there's no other implementation for it. Do you know of any reason it existed? Is GarnetSessionMetrics intended to implement IMetrics? It doesn't currently.

vazois: I was referring to the definition incr_total_transaction_transactions() not the whole interface implementation. Now that I checked it, it seems that it is not used at the moment. So, it should be safe to remove. Unless anyone else has any comments @badrishc, @TalZaccai.

badrishc: At some point we wanted to template by this struct in order to make metrics overhead zero via JIT. that's still of interest, but if you want to remove the struct entirely that is fine too

Xizt: Removed.

PaulusParssinen:
While we're doing this, can we please turn those metrics to just be properties with proper .NET casing? 😅

vazois:
> While we're doing this, can we please turn those metrics to just be properties with proper .NET casing? 😅

That will be great if we can establish that there is no perf implication because some of those metrics are on the critical path. However, this is not the goal of this PR, so I will merge it as is. 
You are more that welcome to add a PR for this ☺️

vazois (approved):

vazois (approved):

Issue #1299, opened 2025-07-16
Observing high CPU in certain scenarios for large clusters.


Xizt:
### Describe the bug

Context: During deployments, our pods in k8s go in terminating state (sigterm). We ensure that if the pod that goes in terminating state, should get its role changed if primary. Our control plane detects that a particular pod is in terminating state and takes appropriate action. Scenarios for the control plane:
Terminating pod is
1. Primary for a shard: We issue failover for secondary. 
2. Secondary for a shard: If it's a secondary, we issue REPLICAOF NO ONE.

We also keep querying CLUSTER NODES (currently frequency is 0.5 req/s) to garnet pod to check what the status is and if the action has control plane actions has been completed. (This querying only happens if pod is in terminating state). 
If the pod is not serving a slot anymore, we allow the terminating garnet pod to do an Environment.Exit
 
Observation: For a terminating primary, when secondary is issued a FAILOVER, primary gets high CPU. This is a large cluster with 150+ pods and high gossip during this period.
 
Issue 1: In gossip main, we have lot of gossip happening, during which a lock is obtained in FlushConfig. We found out that 58 threads were block on FlushConfig lock. One thread which got the lock access, is waiting synchronously on a semaphore and the IOCallback did not get called. This is confirmed by inspecting the value of numPending being 1 in the memory dump. Had the semaphore been released, the value is also decremented.
 
Issue2: During gossip, in initConnections, we see that the node keeps trying to gossip with unreachable nodes. This results in exceptions. In exception handling, in clusterConnectionStore, we do TryRemove(nodeId) which obtains a write lock and when trying to dispose the network pool, we do a Thread.Yield. At the same time, lot of threads are processing CLUSTER NODES command on same node which keep waiting trying to obtain readlock in GetConnectionInfo. 

Have shared dumps internally.

### Steps to reproduce the bug

These issue arise during cases as described above and in certain scenarios.

msft-paddy14:
@vazois  This one is doing a linear search under a lock for gossip and  besides the flushing it might cause bottlenecks as it's used in paths for gossip, meet, metrics etc. It might become bottleneck for large clusters - especially during setup when multiple nodes try to do MEET. Should this be a ConcurrencyDictionary to avoid global locking? I see that we need random iteration capabilties incase of sampled gossip, maybe it makes sense to keep linear one for those scenarios where size is smaller?

[...]

PR #1315, opened 2025-07-25
High CPU Utilization from Gossip Fix

vazois:
This PR addresses #1299 

- [x] Expose parameter to skip writing to disk with flushConfig or perform flushConfig in a background task.
- [x] Rewrite GarnetClusterConnection to use a dictionary for fast search


libs/cluster/Server/Gossip/GarnetClusterConnectionStore.cs
```diff
@@ -53,28 +54,50 @@ public void Dispose()
...
+        /// Caller responsible for locking.
+        /// </summary>
+        /// <param name="nodeId">Node-id to search for.</param>
+        /// <param name="conn">Connection object returned on success otherwise null.</param>
+        /// <returns>True if connection is found otherwise false.</returns>
+        private bool UnsafeGetConnection(string nodeId, out GarnetServerNode conn)
+        {
+            conn = null;
+            if (_disposed) return false;
+            for (var i = 0; i < numConnection; i++)
```
Xizt: Understanding the reason behind the connections being an array instead of something like a ConcurrentDictionary. Is it because the number of connections is expected to be significantly low to impact the search?

libs/cluster/Server/Gossip/GarnetClusterConnectionStore.cs
```diff
@@ -212,26 +202,29 @@ public bool TryRemove(string nodeId)
...
-                if (_conn.NodeId.Equals(nodeId, StringComparison.OrdinalIgnoreCase))
-                {
-                    conn = _conn;
-                    return true;
-                }
+                _lock.WriteLock();
+
+                if (_disposed) return;
+
+                for (var i = 0; i < numConnection; i++)
```
Xizt: nit: UnsafeDisposeConnection that can be reused in Dispose and CloseAll

libs/cluster/Server/ClusterManager.cs
```diff
@@ -84,10 +96,37 @@ public unsafe ClusterManager(ClusterProvider clusterProvider, ILogger logger = n
...
-            GossipSamplePercent = opts.GossipSamplePercent;
+            GossipSamplePercent = serverOptions.GossipSamplePercent;
+
+            // Run Backround task
+            if (serverOptions.ConfigFlushFrequency > 0)
+                Task.Run(() => FlushTask());
+
+            async Task FlushTask()
+            {
+                var flushConfigFrequency = TimeSpan.FromMicroseconds(serverOptions.ConfigFlushFrequency);
```
Xizt: The option input is in milliseconds.

libs/cluster/Server/ClusterManager.cs
```diff
@@ -84,10 +96,37 @@ public unsafe ClusterManager(ClusterProvider clusterProvider, ILogger logger = n
...
+                {
+                    Interlocked.Increment(ref numActiveTasks);
+                    while (true)
+                    {
+                        if (ctsGossip.IsCancellationRequested)
+                            return;
+
+                        if (Interlocked.CompareExchange(ref flushCount, 0, 0) > 0)
+                            ClusterUtils.WriteInto(clusterConfigDevice, pool, 0, currentConfig.ToByteArray(), logger: logger);
+
```
Xizt, line 121: I see that FlushConfig() obtains a lock while doing the WriteInto operation. Does this action require the same?

libs/cluster/Server/Gossip/GarnetClusterConnectionStore.cs
vazois: Sure, make sense.

libs/cluster/Server/ClusterManager.cs
vazois: No, it does not need a lock because there will be only one background task when ConfigFlushFrequency > 0. Ideally, for your use case you will be using -1, so nothing is flushed.

libs/cluster/Server/ClusterManager.cs
vazois: typo, thanks!

libs/cluster/Server/Gossip/GarnetClusterConnectionStore.cs
vazois: I feel that access should be pretty fast if the number of items is in the order of 100s - 1000s. The other issue is empirically we found that ConcurrentDictionary not being that performant. I guess since gossip is not on the critical path it should be fine. Last thing using our own lock gives us the ability to favor Readers (accessing connection stats) over Writers (adding a new connection). We can probably get away with that if we use regular Dictionary with our lock.

Is it possible to provide a repro that demonstrates high CPU utilization so we can profile somehow the different options?

Xizt: I see that paddy has already mentioned similar concern on the issue. https://github.com/microsoft/garnet/issues/1299#issuecomment-3116478748

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -196,6 +196,11 @@ public class GarnetServerOptions : ServerOptions
         /// </summary>
         public int ClusterTimeout = 60;
 
+        /// <summary>
+        /// How frequently to flush config unto disk to persist updates. =-1: never (memory only), =0: immediately (every update performs flush), >0: frequency in ms
+        /// </summary>
+        public int ConfigFlushFrequency = 0;
```
badrishc: Can either make this of seconds granularity and rename to `ConfigFlushFrequencySecs` or leave as milliseconds if that fine-grained granularity is deemed necessary. In the latter case, it can be renamed to `ConfigFlushFrequencyMs` (to be consistent with other similar frequency configs). Same comment holds for Options.cs.

vazois: Makes sense. I will leave it milliseconds because seems more appropriorate.

badrishc (approved):

Issue #1313, opened 2025-07-24
Commit Frequency Secs option does not work


hamdaankhalid:
### Describe the bug

--compaction-freq option via command line doesn't work when I have it enabled. I was able to RCA and it looks like current code reuses the same copmaction method for checkpointing and even when compaction background task calls it, it early exits because it thinks checkpointing called compaction.

### Steps to reproduce the bug

Run GarnetServer with --compaction-freq 10 and whatever compaction type you want. Put a debug in the compaction method and see how it early exits.

### Expected behavior

Compaction should run at interval

s-syasir:
Hey, here's a solution I came up with. Lmk if it works.
https://github.com/microsoft/garnet/pull/1316

PR #1316, opened 2025-07-26
Fix compaction frequency [#1313]

s-syasir:
Updated the DoCompaction method in DatabaseManagerBase.cs to use a boolean to indicate whether the call comes from the checkpoint or is the periodic task. This solves the bug of the periodic task being skipped over because the previous check was only of the frequency


hamdaankhalid (approved):
Thanks for fixing this.
Congrats on your first commit to Garnet 🥇

PR #1317, opened 2025-07-27
Implement struct.pack struct.upack and struct.size

tarazou9:
Implement LUA `struct.pack` `struct.upack` and `struct.size`. The code is based on, not copied from, the Lua functions as Redis. [Link here](https://github.com/redis/redis/blob/unstable/deps/lua/src/lua_struct.c#L297)

How it's tested

- [ x ] Tested Garnet OSS locally
- [ x ] Unit tests
- [ x ] Run Redis and compare all cases included in the unit tests.


libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
             }
         }
 
+        [StructLayout(LayoutKind.Sequential)]
```
kevin-montrose: As this is heavily based on Lua's impl (as [copied into Redis](https://github.com/redis/redis/blob/unstable/deps/lua/src/lua_struct.c)), let's move the re-implemented bits into a new partial (`LuaRunner.Functions.Struct.cs` maybe?) and include [the license text](https://github.com/redis/redis/blob/db4fc2a83309bf8b65e25deedfac0ff71d67e4b8/deps/lua/src/lua_struct.c#L407) (along with a comment that the code is based on, not copied from, the Lua stuff) there.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+
+        internal enum Endian
+        {
+            BIG = 0,
+            LITTLE = 1,
+        }
+
+        internal static class Native
+        {
+            public static int Endian => BitConverter.IsLittleEndian ? 1 : 0; // Matches LITTLE = 1, BIG = 0
```
kevin-montrose: nit: Since this is a constant, `{get;} = ` instead of `=>`.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+        {
+            if (optIx >= format.Length || !char.IsDigit(format[optIx]))
+                return defaultValue;
+
+            int result = 0;
+            while (optIx < format.Length && char.IsDigit(format[optIx]))
+            {
+                int digit = format[optIx] - '0';
+                if (result > (int.MaxValue / 10) || result * 10 > (int.MaxValue - digit))
+                    // Integral size overflow
```
kevin-montrose: nit: Garnet is inconstistent around bracing style, but a 2-line no-brace block is rough. 
```csharp
{
  // Integral size overflow
  return self.LuaWrappedError(1, self.constStrs.ErrBadArg);
}
```

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+                totalSize += toAlign;
+
+                while (toAlign-- > 0)
+                {
+                    AddAlignmentPadding(this);
+                }
+
+                switch (opt)
+                {
+                    case 'b': case 'B':
```
kevin-montrose: nit: consistent `case` layout - let's put each `case foo:` on it's own line, like is done with `'c'` and `'s'`.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+            state.ExpectLuaStackEmpty();
+
+            var ret = scratchBufferBuilder.ViewFullArgSlice().ReadOnlySpan;
+            if (!state.TryPushBuffer(ret))
+            {
+                return LuaWrappedError(1, constStrs.OutOfMemory);
+            }
+
+            return 1;
+
```
kevin-montrose: nit: extra new line

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+                        break;
+                }
+
+                self.state.Remove(stackIndex);
+
+                constStrRegisteryIndex = -1;
+                return true;
+            }
+
+            static bool TryEncodeFloatingPoint(LuaRunner self, int stackIndex, Header h, out int constStrRegisteryIndex)
```
kevin-montrose: `FloatingPoint` is ambiguous here, covering `float`, `double`, and `decimal`.  I believe the intent here is just [`float`](https://learn.microsoft.com/en-us/dotnet/api/system.single?view=net-8.0), so rename to `TryEncodeSingle`.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+
+                var value = self.state.CheckNumber(stackIndex);
+
+                var into = self.scratchBufferBuilder.ViewRemainingArgSlice(4).Span;
+
+                if (h.Endian == Native.Endian) // header endian == native endian
+                {
+                    // If header endian matches native endian, pick the method that matches native order
+                    if (BitConverter.IsLittleEndian)
+                        BinaryPrimitives.WriteDoubleLittleEndian(into[0..], value);
```
kevin-montrose: Should there be `WriteSingleXXX` calls?  This is gonna write 8 bytes, not 4.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+
+            var numLuaArgs = state.StackTop;
+
+            if (numLuaArgs == 0 || !TryGetFormat(this, 1, out var format))
+            {
+                return LuaWrappedError(0, constStrs.BadArgPack);
+            }
+
+            // Use stackIndex 1 become when try get format, it also removes format from the lus stack
+            state.KnownStringToBuffer(1, out var data);
```
kevin-montrose: How do we know this is a string?

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+            if (numLuaArgs >= 3) // Only check if the 3rd argument is passed
+            {
+                if (state.Type(3) != LuaType.Number || state.CheckNumber(3) <= 0)
+                {
+                    return LuaWrappedError(0, constStrs.BadArgPack);
+                }
+                pos = (int)state.CheckNumber(3);
+                pos--;
+                // slice data to start from pos
+                data = data[pos..];
```
kevin-montrose: We need a range check here, throwing an exception is really bad from Lua.

libs/server/Lua/LuaRunner.Functions.cs
```diff
@@ -2773,6 +2774,813 @@ static bool TryDecodeLargeMap(LuaRunner self, ref ReadOnlySpan<byte> data, out i
...
+
+                self.state.PushNumber(result);
+                data = data[1..];
+                decodedCount++;
+
+                constStrErrId = -1;
+                return true;
+            }
+            
+            static bool TryDecodeFloatingPoint(LuaRunner self, Header h, int size, ref ReadOnlySpan<byte> data, ref int decodedCount, out int constStrErrId)
```
kevin-montrose: Same note on `FloatingPoint` naming.

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: Makes sense. Refactored it to a partial file.

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: Yes, I meant float, not the floating point. Thank you for correcting!

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: This is a bug, thank you for spotting this! Fixed.

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: I assume it's a string from here 
https://github.com/redis/redis/blob/ecd5e639ed4d35f71c3e67889879aee0e8fb3919/deps/lua/src/lua_struct.c#L297

Added an assertion to check the type, not sure if I need to throw a lua error here?

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: the data is truncated is removed. I use a copy slice instead so it's less confusing.

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,368 @@ public void Struct()
...
+        }
+
+        [Test]
+        public void StructUnpack()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase();
+
+            var unpackDC0Res = db.ScriptEvaluate("return { struct.unpack('dc0', '\\x1F\\x85\\xEB\\x51\\xB8\\x1E\\x09\\x40dynamic') }");
+            ClassicAssert.AreEqual("dyn", (string)unpackDC0Res[0]);
```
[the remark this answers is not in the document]
tarazou9: No, Copilot is wrong. The size is 3.14 cast into 3, so only get the first three chars of the string, that's the purpose of this test.
When "c0" is passed, for "unpack", it will retrieve size from previous arg.

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,368 @@ public void Struct()
...
+            ClassicAssert.AreEqual("ERR Lua encountered an error: bad argument to loadstring", excFormatString.Message);
+
+            var excFormatCharacter = ClassicAssert.Throws<RedisServerException>(() => db.ScriptEvaluate("return struct.size('c0')"));
+            ClassicAssert.AreEqual("ERR Lua encountered an error: bad argument to loadstring", excFormatCharacter.Message);
+
+            var excBadControlOptions = ClassicAssert.Throws<RedisServerException>(() => db.ScriptEvaluate("return struct.size('@I')"));
+            ClassicAssert.AreEqual("ERR Lua encountered an error: bad argument to loadstring", excBadControlOptions.Message);
+
+            var excBadFormat = ClassicAssert.Throws<RedisServerException>(() => db.ScriptEvaluate("return struct.size(123)"));
+            ClassicAssert.AreEqual("ERR Lua encountered an error: bad argument to loadstring", excBadControlOptions.Message);
```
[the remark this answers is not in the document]
tarazou9: Fixed

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,944 @@
...
+                decodedCount++;
+
+                constStrErrId = -1;
+                return true;
+            }
+
+            static bool TryDecodeSingle(LuaRunner self, Header h, int size, int pos, ReadOnlySpan<byte> data, ref int decodedCount, out int constStrErrId)
+            {
+                var slice = data.Slice(pos);
+                float value = h.Endian == Native.Endian ? BinaryPrimitives.ReadSingleLittleEndian(slice) : BinaryPrimitives.ReadSingleBigEndian(slice);
```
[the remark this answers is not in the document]
tarazou9: This is a legit bug, fixed

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,944 @@
...
+                if (l < size)
+                {
+                    // string too short, expected at least {size} bytes but got {l}
+                    constStrRegisteryIndex = self.constStrs.BadArgPack;
+                    return false;
+                }
+
+                var into = self.scratchBufferBuilder.ViewRemainingArgSlice(data.Length).Span;
+                data.CopyTo(into[0..]);
+                self.scratchBufferBuilder.MoveOffset(data.Length);
```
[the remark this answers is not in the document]
tarazou9: This is legit bug, fixed.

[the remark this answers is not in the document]
tarazou9: This is legit bug, fixed.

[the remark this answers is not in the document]
tarazou9: bug, fixed

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,944 @@
...
+                decodedCount++;
+
+                constStrErrId = -1;
+                return true;
+            }
+
+            static bool TryDecodeDouble(LuaRunner self, Header h, int size, int pos, ReadOnlySpan<byte> data, ref int decodedCount, out int constStrErrId)
+            {
+                var slice = data.Slice(pos);
+                double value = h.Endian == Native.Endian ? BinaryPrimitives.ReadDoubleLittleEndian(slice) : BinaryPrimitives.ReadDoubleBigEndian(slice);
```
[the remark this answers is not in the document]
tarazou9: bug, fixed

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,368 @@ public void Struct()
...
+            var unpackSignedIntRes = (int[])db.ScriptEvaluate("return { struct.unpack('i4', '\\x9D\\xFF\\xFF\\xFF') }");
+            ClassicAssert.True(unpackSignedIntRes.SequenceEqual([-99, 5]));
+
+            var unpackUnsignedIntRes = (int[])db.ScriptEvaluate("return { struct.unpack('I2', '\\x00\\x02') }");
+            ClassicAssert.True(unpackUnsignedIntRes.SequenceEqual([512, 3]));
+
+            var unpackRes = (int[])db.ScriptEvaluate("return { struct.unpack('>hB', '\\x12\\x34\\x56') }");
+            ClassicAssert.True(unpackRes.SequenceEqual([0x1234, 0x56, 4]));
+
+            var unpackZeroTerminatedStringRes = db.ScriptEvaluate("return { struct.unpack('Bc0', '\\x07dynamic') }");
```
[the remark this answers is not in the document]
tarazou9: no, this is intended.

libs/server/Lua/LuaStateWrapper.cs
```diff
@@ -283,7 +283,7 @@ internal LuaStatus PCall(int args, int rets)
             // We have to copy this off, as once we PCall curStackTop could be modified
             var oldStackTop = StackTop;
 
-            var res = NativeMethods.PCall(state, args, rets);
+            LuaStatus res = NativeMethods.PCall(state, args, rets);
```
kevin-montrose: nit: pointless change.

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,981 @@
...
+        /// Entry point for struct.size from a Lua script.
+        /// </summary>
+        internal int StructSize(nint luaStatePtr)
+        {
+            state.CallFromLuaEntered(luaStatePtr);
+
+            var numLuaArgs = state.StackTop;
+            if (numLuaArgs == 0 || !TryGetFormat(this, 1, out var format))
+            {
+                return LuaWrappedError(1, constStrs.BadArgLoadString);
```
kevin-montrose: Seems like the wrong error message here?

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,981 @@
...
+                {
+                    return LuaWrappedError(1, errOptSizeIndex);
+                }
+
+                int toAlign = GetToAlign(totalSize, h.Align, opt, size);
+                totalSize += toAlign;
+
+                if (opt == 's')
+                {
+                    return LuaWrappedError(1, constStrs.BadArgLoadString);
```
kevin-montrose: Same thing.

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,981 @@
...
+                }
+
+                if (opt == 'c' && size == 0)
+                {
+                    return LuaWrappedError(1, constStrs.BadArgLoadString);
```
kevin-montrose: Same thing.

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,981 @@
...
+            state.CallFromLuaEntered(luaStatePtr);
+
+            var numLuaArgs = state.StackTop;
+
+            if (numLuaArgs < 2 || !TryGetFormat(this, 1, out var format))
+            {
+                return LuaWrappedError(0, constStrs.BadArgUnpack);
+            }
+
+            Debug.Assert(state.Type(1) == LuaType.String, "Expected string");
```
kevin-montrose: We must _check_ the type and return an explicit error - there' s no guarantee the caller didn't mess up and pass a non-string, and throwing here will crash the process.

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,407 @@ public void Struct()
...
+        }
+
+        [Test]
+        public void StructUnpack()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase();
+
+            var unpackDC0Res = db.ScriptEvaluate("return { struct.unpack('dc0', '\\x1F\\x85\\xEB\\x51\\xB8\\x1E\\x09\\x40dynamic') }");
+            ClassicAssert.AreEqual("dyn", (string)unpackDC0Res[0]);
```
kevin-montrose, line 2897: This appears to diverge from Redis behavior, which return `["", 9]`.

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,407 @@ public void Struct()
...
+            ClassicAssert.AreEqual(12, (long)unpackDC0Res[1]);
+
+            var unpackHRes = (int[])db.ScriptEvaluate("return { struct.unpack('HH', '\\x01\\x00\\x02\\x00') }");
```
kevin-montrose: Likewise, this appears to diverge - Redis returns `[12408, 30769, 5]`.

test/Garnet.test/LuaScriptTests.cs
```diff
@@ -2724,6 +2724,407 @@ public void Struct()
...
+            ClassicAssert.True(unpackHRes.SequenceEqual([1, 2, 5]));
+
+            var unpackCharactersRes = db.ScriptEvaluate("return { struct.unpack('c5', 'hello') }");
+            ClassicAssert.AreEqual("hello", (string)unpackCharactersRes[0]);
+
+            var unpackBigEndianRes = (int[])db.ScriptEvaluate("return { struct.unpack('>hB', '\\x34\\x12\\x56') }");
```
kevin-montrose: This one too - please run these against Redis.  You can just change the `ConnectionMultiplexer` line to point to a local instance.

For completeness, here's the Redis version and host I'm comparing against:
```
redis_version:8.0.3
redis_git_sha1:00000000
redis_git_dirty:1
redis_build_id:680cac7fe8eb957c
redis_mode:standalone
os:Linux 5.15.167.4-microsoft-standard-WSL2 x86_64
arch_bits:64
```

libs/server/Lua/LuaRunner.Functions.Struct.cs
```diff
@@ -0,0 +1,981 @@
...
+            BIG = 0,
+            LITTLE = 1,
+        }
+
+        internal static class Native
+        {
+            public static int Endian { get; } = BitConverter.IsLittleEndian ? 1 : 0; // Matches LITTLE = 1, BIG = 0
+        }
+
+        static bool TryGetFormat(LuaRunner self, int stackIndex, out string format)
```
kevin-montrose: It would be preferable if we don't allocate a `string` here.

We know the format string is either all ASCII, or invalid - so we don't need the encoder either.

Because we're removing from the stack we do need a copy.  I suggest something like:
```csharp
static bool TryGetFormat(LuaRunner self, int stackIndex, ref Span<byte> format, out byte[] rentedArr)
{
    if (self.state.Type(stackIndex) != LuaType.String)
    {
        format = default;
        rentedArr = null;
        return false;
    }

    self.state.KnownStringToBuffer(stackIndex, out var data);
    if (data.Length > format.Length)
    {
        format = rentedArr = ArrayPool<byte>.Shared.Rent(data.Length);
        format = format[..data.Length];
        data.CopyTo(format);
    }
    else
    {
        rentedArr = null;
        data.CopyTo(format);
        format = format[..data.Length];
    }

    self.state.Remove(stackIndex);
    return true;
}
```

And then `stackalloc` some small buffer format (maybe 64-bytes) to pass in, and return the rented array if non-null (in a `finally` presumably).

libs/server/Lua/LuaStateWrapper.cs
tarazou9: During debugging, I saw a case when "res" resturns a string "OK" and below in the if statement `if (res != LuaStatus.OK || rets < 0)`, it returns true because `res != LusaStatus.OK` return true. Not sure why it didn't cast correctly.

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: So for the methods that shared between `pack` and `unpack`, I'm not sure what's the best error messages to use, that's why I used this..... Any suggestions?

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: Replied in the above.

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: Replied in the above.

test/Garnet.test/LuaScriptTests.cs
tarazou9: So the value for the double is 3.14, which converts to an int it becomes 3. I believe that's what Redis is doing.

test/Garnet.test/LuaScriptTests.cs
tarazou9: I did manually run **all** the cases included in the unit tests against a local redis and compared all of them. Let me double check, maybe I overlooked.

test/Garnet.test/LuaScriptTests.cs
tarazou9: Because the string `\\x01\\x00\\x02\\x00` in C# becomes "\x01\x00\x02\x00" when interpreted by Lua in Redis via EVAL. That gives you the correct 4 bytes as expected. I believe the unit test above is doing the right thing.

image

test/Garnet.test/LuaScriptTests.cs
tarazou9: Same thing as the above. 

image

test/Garnet.test/LuaScriptTests.cs
tarazou9: oh I think it's the same thing `\\` VS. `\`

image

libs/server/Lua/LuaRunner.Functions.Struct.cs
kevin-montrose: Just define a new one, `BadArgSize` or something - they're all hooked up in `ConstantStringRegistryIndexes`.

test/Garnet.test/LuaScriptTests.cs
kevin-montrose: Hmm, strange.  Lemme dig a little more - I just swapped the `ConnectionMultiplexer` setup in the tests themselves, looks like you're running in `redis-cli`?

tarazou9: Yes, I'm running redis-cli in the docker container

kevin-montrose: It's the escape sequences.

Redis's Lua (5.1) (apparently, I did not know this) expects decimal escape sequences - not hex.

So instead of:
`return { struct.unpack('dc0', '\\x1F\\x85\\xEB\\x51\\xB8\\x1E\\x09\\x40dynamic') }`
it'd be
`return { struct.unpack('dc0', '\31\133\235\81\184\30\9\64dynamic') }`

Tested against Redis (8.0.3) again:
image

Lua 5.4 Supporting hex escapes is an interesting find... that's observably different behavior between Redis and Garnet.  Filing that away as a potential problem in the future - but irrelevant for this PR.

For now, just convert to decimal escapes so we're verifying matching behavior.

libs/server/Lua/LuaStateWrapper.cs
kevin-montrose: I'd guess stack corruption during that debugging session then - there's no overload of `NativeMethods.PCall` that returns a `string` and no conversion operator for `LuaStatus` to `string`.

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: Added the check.

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: Defined a `BadArgFormat` and updated.

libs/server/Lua/LuaRunner.Functions.Struct.cs
tarazou9: Updated! Thank you for the helpful suggestion!

libs/server/Lua/LuaRunner.Functions.cs
tarazou9: Updated.

libs/server/Lua/LuaStateWrapper.cs
tarazou9: Yeah okay, that makes sense. Reverted this.

test/Garnet.test/LuaScriptTests.cs
tarazou9: Okay, this explains it. I updated all unit tests, thank you for the finding!

PR #1318, opened 2025-07-28
Enforce stricter client name limits also for HELLO command

prvyk:
Redis/valkey enforce limits on client names. These limits were not enforced on HELLO SETNAME command, and even when enforced, current limits allowed tab characters and so on. This PR enforces stricter limits for both commands setting client name.

### DIFF
--- a/libs/server/SessionParseStateExtensions.cs
+++ b/libs/server/SessionParseStateExtensions.cs
@@ -104,7 +104,13 @@ namespace Garnet.server
         {
             clientName = parseState.GetString(idx);
 
-            if (string.IsNullOrEmpty(clientName))
+            if (clientName == null)
+            {
+                return false;
+            }
+
+            // Reference allows clearing client name
+            if (clientName == string.Empty)
             {
                 return true;
             }
--- a/libs/server/SessionParseStateExtensions.cs
+++ b/libs/server/SessionParseStateExtensions.cs
@@ -94,6 +94,33 @@ namespace Garnet.server
         }
 
         /// <summary>
+        /// Parse client name from parse state at specified index.
+        /// </summary>
+        /// <param name="parseState">The parse state</param>
+        /// <param name="idx">The argument index</param>
+        /// <param name="clientName">Client name</param>
+        /// <returns>True if value parsed successfully</returns>
+        internal static bool TryGetClientName(this SessionParseState parseState, int idx, out string clientName)
+        {
+            clientName = parseState.GetString(idx);
+
+            if (string.IsNullOrEmpty(clientName))
+            {
+                return true;
+            }
+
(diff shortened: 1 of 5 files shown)


libs/server/SessionParseStateExtensions.cs
```diff
@@ -93,6 +93,33 @@ public static bool TryGetLatencyMetricsType(this SessionParseState parseState, i
...
+        /// <param name="idx">The argument index</param>
+        /// <param name="clientName">Client name</param>
+        /// <returns>True if value parsed successfully</returns>
+        internal static bool TryGetClientName(this SessionParseState parseState, int idx, out string clientName)
+        {
+            clientName = parseState.GetString(idx);
+
+            if (string.IsNullOrEmpty(clientName))
+            {
+                return true;
```
vazois, line 115: Shouldn't this be false? We don't allow NullOrEmpty strings right?

prvyk: Redis/valkey allow this.

>CLIENT SETNAME a
>CLIENT GETNAME
> >a

>CLIENT SETNAME ""
>CLIENT GETNAME
> >(nil)

I don't think we can actually get null here, in practice it's always a string, empty or not (but the check doesn't hurt).

libs/server/SessionParseStateExtensions.cs
```diff
@@ -93,6 +93,33 @@ public static bool TryGetLatencyMetricsType(this SessionParseState parseState, i
...
+            }
+
+            foreach (var c in clientName)
+            {
+                if (c < 33 || c > 126)
```
vazois, line 122: This doesn't seem equivalent to the condition you removed at the caller. Is this an improvement?
Why don't you use the conditions that you removed at the caller? i.e. name.Contains(' ');
Otherwise, add some comments explaining what is not allowed.

prvyk: Reference error message is:

> ERR Client names cannot contain spaces, newlines or special characters.

Which is more than just space. So I limited to printable characters excluding space.

Added comment to source to explain this.

libs/server/SessionParseStateExtensions.cs
vazois: It is best to use the same conditions that were used at the caller (CLIENTSETNAME), unless of course you are fixing a bug.
This way we avoid a regression.

prvyk: Well, this works with redis and thus so should be allowed.

Added comment to explain this + test to ensure clearing (the other setname test already tested an empty name, I wanted to check clearing an existing name too).

vazois (approved):

PR #1319, opened 2025-07-30
Resetting replication recovery when cluster reset is issued

Xizt:


Xizt:
@vazois We discussed an approach where we can reset the SingleWriteMultiReader. I thought about giving that a try instead of calling EndRecovery() but that too has challenges that the thread which was stuck after calling BeginRecovery() but hadn't called EndRecovery() will forever have a thread spinning? 
Calling EndRecovery() as well doesn't seem full proof given that recoverLock is still obtained by the other thread either gets released before can cause RESET thread to spin wait forever or vice versa. 
I understand your previous changes with timeouts will help to recover from replication stuck issue. Do you have any reommendations for this PR ?

vazois:
> @vazois We discussed an approach where we can reset the SingleWriteMultiReader. I thought about giving that a try instead of calling EndRecovery() but that too has challenges that the thread which was stuck after calling BeginRecovery() but hadn't called EndRecovery() will forever have a thread spinning? Calling EndRecovery() as well doesn't seem full proof given that recoverLock is still obtained by the other thread either gets released before can cause RESET thread to spin wait forever or vice versa. I understand your previous changes with timeouts will help to recover from replication stuck issue. Do you have any reommendations for this PR ?

Yes this is correct. You have to think about the background task that is running and is holding that lock.
I think we need to add a separate cancellation token that can be cancelled and reset when necessary because right now the attach sync request is tied to the cancellation toke of the replication manager.
This needs to be done for both disk-based and diskless replication.
Then this cancellation token will trigger the exception handling and release the lock so you don't have to release it manually.
You just need to review the different possible states and appropriatly release the cts there.
Here is an example:

https://github.com/microsoft/garnet/compare/main...vazois:vazois/cts-replication?expand=1

I also noticed that on Connect I forgot to pass the storeWrapper.serverOptions.ReplicaAttachTimeout. That might be also true for diskless sync path. Please add this in your PR also

https://github.com/microsoft/garnet/blob/e5826216c959e39a3f0ac61eae12f6e652e9e3ea/libs/cluster/Server/Replication/ReplicaOps/ReplicaReceiveCheckpoint.cs#L95

vazois:
> > @vazois We discussed an approach where we can reset the SingleWriteMultiReader. I thought about giving that a try instead of calling EndRecovery() but that too has challenges that the thread which was stuck after calling BeginRecovery() but hadn't called EndRecovery() will forever have a thread spinning? Calling EndRecovery() as well doesn't seem full proof given that recoverLock is still obtained by the other thread either gets released before can cause RESET thread to spin wait forever or vice versa. I understand your previous changes with timeouts will help to recover from replication stuck issue. Do you have any reommendations for this PR ?
> 
> Yes this is correct. You have to think about the background task that is running and is holding that lock. I think we need to add a separate cancellation token that can be cancelled and reset when necessary because right now the attach sync request is tied to the cancellation toke of the replication manager. This needs to be done for both disk-based and diskless replication. Then this cancellation token will trigger the exception handling and release the lock so you don't have to release it manually. You just need to review the different possible states and appropriatly release the cts there. Here is an example:
> 
> https://github.com/microsoft/garnet/compare/main...vazois:vazois/cts-replication?expand=1
> 
> I also noticed that on Connect I forgot to pass the storeWrapper.serverOptions.ReplicaAttachTimeout. That might be also true for diskless sync path. Please add this in your PR also
> 
> https://github.com/microsoft/garnet/blob/e5826216c959e39a3f0ac61eae12f6e652e9e3ea/libs/cluster/Server/Replication/ReplicaOps/ReplicaReceiveCheckpoint.cs#L95

Also check ExceptionInjectionHelper to see how you can create some test cases for this.

libs/cluster/Server/Replication/ReplicationManager.cs
```diff
@@ -457,6 +458,16 @@ public void EndRecovery(RecoveryStatus nextRecoveryStatus, bool downgradeLock)
             }
         }
 
+        public void ResetRecovery()
+        {
+            switch (currentRecoveryStatus)
+            {
+                case RecoveryStatus.ClusterReplicate:
```
vazois, line 462: I think you need to add the other options also other than NoRecovery and ReadRole.

badrishc:
Ping on getting these review comments addressed, thanks.

libs/cluster/Server/Replication/ReplicationManager.cs
Xizt: Added.

vazois (approved):
Great work!

PR #1324, opened 2025-08-05
More tiny SortedSet fixes

prvyk:
1. ZINTERCARD z LIMIT 0 is defined as 'unlimited' in docs, yet code enforced 0 as limit. Same for SINTERCARD.

https://redis.io/docs/latest/commands/zintercard/

>When provided with the optional LIMIT argument (which defaults to 0 and means unlimited)

2. ZREVRANGEBYSCORE didn't reverse exclusions (i.e. open intervals) which can lead to wrong results.

3. ZRANGE BYLEX/BYSCORE LIMIT (something) 0 should return an empty array. 
ZRANGE LIMIT (negative) (count) also acts this way in practice. An extension to make it work from the end makes sense, this PR makes it follows reference behaviour though.

4. Some style changes for SortedSet code - use tuple provided name where possible, use more efficient Dictionary.Remove overload.

### DIFF
--- a/libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
+++ b/libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
@@ -7,6 +7,8 @@ using System.Diagnostics;
 using System.Linq;
 using Garnet.common;
 
+using LimitDef = (int offset, int countLimit);
+
 namespace Garnet.server
 {
     /// <summary>
@@ -22,9 +24,9 @@ namespace Garnet.server
             public bool ByScore { get; set; }
             public bool ByLex { get; set; }
             public bool Reverse { get; set; }
-            public (int, int) Limit { get; set; }
-            public bool ValidLimit { get; set; }
             public bool WithScores { get; set; }
+            public bool ValidLimit { get; set; }
+            public LimitDef Limit { get; set; }
         };
 
         private enum SpecialRanges : byte
@@ -215,11 +217,10 @@ namespace Garnet.server
                 var value = input.parseState.GetArgSliceByRef(i).ReadOnlySpan;
                 var valueArray = value.ToArray();
 
-                if (!sortedSetDict.TryGetValue(valueArray, out var key))
+                if (!sortedSetDict.Remove(valueArray, out var key))
                     continue;
 
                 output.Header.result1++;
-                sortedSetDict.Remove(valueArray);
                 sortedSet.Remove((key, valueArray));
                 _ = TryRemoveExpiration(valueArray);
 
@@ -588,13 +589,13 @@ namespace Garnet.server
             // Using to list to avoid modified enumerator exception
             foreach (var item in sortedSet.Skip(start).Take(elementCount).ToList())
             {
-                if (sortedSetDict.Remove(item.Item2, out var key))
+                if (sortedSetDict.Remove(item.Element, out var key))
                 {
-                    sortedSet.Remove((key, item.Item2));
+                    sortedSet.Remove((key, item.Element));
 
-                    UpdateSize(item.Item2, false);
+                    UpdateSize(item.Element, false);
                 }
-                TryRemoveExpiration(item.Item2);
+                TryRemoveExpiration(item.Element);
             }
 
             // Write the number of elements
@@ -937,7 +938,7 @@ namespace Garnet.server
             bool validLimit,
             bool rem,
             out int errorCode,
-            (int, int) limit = default)
+            LimitDef limit = default)
         {
             var elementsInLex = new List<(double, byte[])>();
 
@@ -961,7 +962,7 @@ namespace Garnet.server
 
             if (minValueInfinity == SpecialRanges.InfiniteMax ||
                 maxValueInfinity == SpecialRanges.InfiniteMin ||
-                (validLimit && (limit.Item1 < 0 || limit.Item2 == 0)))
+                (validLimit && (limit.offset < 0 || limit.countLimit == 0)))
             {
                 errorCode = 0;
                 return elementsInLex;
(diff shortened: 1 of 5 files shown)


libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -7,6 +7,8 @@
 using System.Linq;
 using Garnet.common;
 
+using LimitDef = (int offset, int countLimit);
```
TalZaccai: For simplicity, let's change the naming to `Limit = (int offset, int count)`, that way it corresponds with the command syntax.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -959,7 +960,9 @@ private void SortedSetCollect(ref ObjectInput input, ref GarnetObjectStoreOutput
                 (maxValueExclusive, minValueExclusive) = (minValueExclusive, maxValueExclusive);
             }
 
-            if (minValueInfinity == SpecialRanges.InfiniteMax || maxValueInfinity == SpecialRanges.InfiniteMin)
+            if (minValueInfinity == SpecialRanges.InfiniteMax ||
+                maxValueInfinity == SpecialRanges.InfiniteMin ||
+                (validLimit && (limit.offset < 0 || limit.countLimit == 0)))
```
TalZaccai: Shouldn't it be `(limit.countLimit < 0 && limit.offset == 0)`?

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
```diff
@@ -1040,41 +1042,42 @@ private void SortedSetCollect(ref ObjectInput input, ref GarnetObjectStoreOutput
...
         {
             if (doReverse)
             {
                 (minValue, maxValue) = (maxValue, minValue);
+                (minExclusive, maxExclusive) = (maxExclusive, minExclusive);
             }
 
             List<(double, byte[])> scoredElements = new();
-            if (sortedSet.Max.Item1 < minValue)
+            if ((validLimit && (limit.offset < 0 || limit.countLimit == 0)) ||
```
TalZaccai: Shouldn't it be `(limit.countLimit == 0 && limit.offset < 0)`?

libs/server/Storage/Session/ObjectStore/SetOps.cs
```diff
@@ -985,7 +985,7 @@ public GarnetStatus SetIntersectLength(ReadOnlySpan<ArgSlice> keys, int? limit,
                 var status = SetIntersect(keys, ref setObjectStoreLockableContext, out var result);
                 if (status == GarnetStatus.OK && result != null)
                 {
-                    count = limit.HasValue ? Math.Min(result.Count, limit.Value) : result.Count;
+                    count = limit > 0 ? Math.Min(result.Count, limit.Value) : result.Count;
```
TalZaccai, line 988: `limit.HasValue && limit > 0`?

libs/server/Storage/Session/ObjectStore/SortedSetOps.cs
```diff
@@ -1335,7 +1335,7 @@ public GarnetStatus SortedSetIntersectLength(ReadOnlySpan<ArgSlice> keys, int? l
             var status = SortedSetIntersect(keys, null, SortedSetAggregateType.Sum, out var pairs);
             if (status == GarnetStatus.OK && pairs != null)
             {
-                count = limit.HasValue ? Math.Min(pairs.Count, limit.Value) : pairs.Count;
+                count = limit > 0 ? Math.Min(pairs.Count, limit.Value) : pairs.Count;
```
TalZaccai, line 1338: Same

TalZaccai (changes requested):
Couple of comments

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
prvyk: Redis/valkey behaviour is that a negative limit count is ignored and returns all values, while a zero limit count returns empty regardless of offset. A zero offset is just the starting point. A negative offset returns empty regardless of limit count. So the current code is accurate.

> ZADD board 10 a 20 b 30 c 40 d 50 e 60 f
> zrangebylex board - + LIMIT 0 -1 (returns everything)
> zrangebylex board - + LIMIT 0 0 (returns empty array)
> zrangebylex board - + LIMIT -1 0 (returns empty array)
> zrangebylex board - + LIMIT -1 1 (returns empty array)

libs/server/Storage/Session/ObjectStore/SetOps.cs
prvyk: If limit is null, (limit > 0) is false. The current code is accurate. I could edit for clarity?

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
prvyk: https://github.com/microsoft/garnet/pull/1324#discussion_r2255555370

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
TalZaccai: Ahhh ok you're returning an empty list here, sorry, I thought you were returning the full collection. Sorry, my bad.

libs/server/Storage/Session/ObjectStore/SetOps.cs
TalZaccai: Meh, not a big deal.

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
prvyk: Done.

TalZaccai (approved):

libs/server/Objects/SortedSet/SortedSetObjectImpl.cs
Montiwa11: - Thank you.

PR #1326, opened 2025-08-07
Bug Fix: Corrected checkpoint task used in WaitOrDie for object store snapshot

priyanjgupta:
This PR fixes a bug in the diskless replication checkpoint flow where the wrong task `mainStoreCheckpointTask` was being passed to the `WaitOrDie` method during object store snapshotting.

Bug:
Previously, during object store checkpointing, `WaitOrDie` was incorrectly invoked with `mainStoreCheckpointTask`, which corresponds to the **main store checkpoint**. This could result in:
- Incorrect progress checks: Monitoring main store progress while object store is being snapshotted
- Potential hangs or premature success handling: WaitOrDie would complete based on main store status, not object store
- Misleading exception triggers: "Object store checkpoint stream failed" would trigger based on main store result
- Main store checkpoint task being awaited twice: Once for main store, once incorrectly for object store


vazois (approved):
Good catch!

PR #1328, opened 2025-08-07
HELLO command: Abort protocol change if there's an authentication failure or any other validation failure

prvyk:
RESP3 spec suggests that a failed auth may prevent protocol version change.

https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md

>Note that even if the protocol is supported, the HELLO command may return an error and perform no action (so the connection will remain in RESP2 mode) in case the authentication credentials are wrong:

Observed reference behaviour is that every validation or authentication failure prevents protocol change (or client name change).
To match this, reorder code in NetworkHELLO() and ProcessHelloMessage() so that verification and then authentication precede every action, so no protocol change is done if user authentication fails.

### DIFF
--- a/test/Garnet.test/RespTests.cs
+++ b/test/Garnet.test/RespTests.cs
@@ -3849,21 +3849,14 @@ namespace Garnet.test
         }
 
         [Test]
-        public void HelloTest2()
+        public void HelloAuthErrorTest()
         {
             using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig(protocol: RedisProtocol.Resp2));
             var db = redis.GetDatabase(0);
 
             // Failed HELLO should not change current protocol
-            try
-            {
-                db.Execute("HELLO", "3", "AUTH", "NX", "NX");
-                ClassicAssert.Fail("Should never reach this line, user does not exist");
-            }
-            catch
-            {
-                // Passthrough
-            }
+            Assert.Throws<RedisServerException>(() => db.Execute("HELLO", "3", "AUTH", "NX", "NX"),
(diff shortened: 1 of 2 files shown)


test/Garnet.test/RespTests.cs
```diff
@@ -3848,6 +3848,34 @@ public void HelloTest1()
...
         }
 
+        [Test]
+        public void HelloTest2()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig(protocol: RedisProtocol.Resp2));
+            var db = redis.GetDatabase(0);
+
+            // Failed HELLO should not change current protocol
+            try
```
TalZaccai: Cleaner way to do this is with ```Assert.Throws<RedisServerException>(() => db.Execute("HELLO", "3", "AUTH", "NX", "NX"), <expected-error>)```

test/Garnet.test/RespTests.cs
```diff
@@ -3848,6 +3848,34 @@ public void HelloTest1()
             ClassicAssert.AreEqual("master", (string)resultDict["role"]);
         }
 
+        [Test]
+        public void HelloTest2()
```
TalZaccai: Can pick a more informative name for the test, like `HelloAuthErrorTest`

test/Garnet.test/RespTests.cs
prvyk: Done.

test/Garnet.test/RespTests.cs
prvyk: Done.

TalZaccai (approved):

PR #1331, opened 2025-08-08
Implement ACL GENPASS

prvyk:
The PR implements ACL GENPASS + tests. It's very easy to set up when .NET already implements the random routine for us and there's no DB access.

### DIFF
--- a/libs/server/Resp/ACLCommands.cs
+++ b/libs/server/Resp/ACLCommands.cs
@@ -382,17 +382,18 @@ namespace Garnet.server
             var length = 64;
             if (parseState.Count == 1)
             {
-                if (!parseState.TryGetLong(0, out var chars))
+                if (!parseState.TryGetLong(0, out var bits))
                 {
                     return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
                 }
 
-                if ((chars <= 0) || (chars > 4096))
+                if ((bits <= 0) || (bits > 4096))
                 {
                     return AbortWithErrorMessage("ERR ACL GENPASS argument must be the number of bits for the output password, a positive number up to 4096"u8);
                 }
 
-                length = (chars % 4 == 0) ? (int)chars / 4 : (int)chars / 4 + 1;
+                // Bits gets rounded to the next multiple of 4 and converted to bytes to get the password length
+                length = (int)(bits / 4) + (bits % 4 == 0 ? 0 : 1);
             }
 
             WriteAsciiBulkString(System.Security.Cryptography.RandomNumberGenerator.GetHexString(length, true));
--- a/libs/server/Resp/ACLCommands.cs
+++ b/libs/server/Resp/ACLCommands.cs
@@ -368,6 +368,38 @@ namespace Garnet.server
         }
 
         /// <summary>
+        /// Processes ACL GENPASS subcommand.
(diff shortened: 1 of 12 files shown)


libs/server/Resp/ACLCommands.cs
```diff
@@ -367,6 +367,38 @@ private bool NetworkAclSave()
...
+            if (parseState.Count > 1)
+            {
+                return AbortWithWrongNumberOfArguments("acl|genpass");
+            }
+
+            // Default length
+            var length = 64;
+            if (parseState.Count == 1)
+            {
+                if (!parseState.TryGetLong(0, out var chars))
```
TalZaccai: nit: rename to bits, as this is not the actual # of chars

libs/server/Resp/ACLCommands.cs
```diff
@@ -367,6 +367,38 @@ private bool NetworkAclSave()
...
+                {
+                    return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
+                }
+
+                if ((chars <= 0) || (chars > 4096))
+                {
+                    return AbortWithErrorMessage("ERR ACL GENPASS argument must be the number of bits for the output password, a positive number up to 4096"u8);
+                }
+
+                length = (chars % 4 == 0) ? (int)chars / 4 : (int)chars / 4 + 1;
```
TalZaccai: Add a comment: ```// Bits gets rounded to the next multiple of 4 and converted to bytes to get the password length```

TalZaccai: Might be a bit clearer: ```length = (int)(chars / 4) + (chars % 4 == 0 ? 0 : 1);```

test/Garnet.test/Resp/ACL/BasicTests.cs
```diff
@@ -111,6 +111,19 @@ public async Task BasicUsersTest()
             ClassicAssert.Contains("default", users);
         }
 
+        [Test]
+        public async Task BasicGenPassTest()
+        {
```
TalZaccai, line 116: Need to test edge cases and error conditions as well (e.g. 0, 4096, 4097, abcd, ...)

prvyk: Done. Couldn't use Assert.Throws, the delegate NUnit expects isn't compatible with GarnetClientSession.

libs/server/Resp/ACLCommands.cs
prvyk: I was conflicted on what's clearer and preferred doing the check outside. But inside will do as well. Done.

test/Garnet.test/Resp/ACL/BasicTests.cs
TalZaccai: Not sure I understand... Why not do for example: 
```cs
Assert.ThrowsAsync<Exception>(async () => await c.ExecuteAsync("ACL", "GENPASS", "abcd"),
    Encoding.ASCII.GetString(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER));
```
?

prvyk: Ah, I forgot ThrowAsync existed... Done.

TalZaccai (approved):

PR #1332, opened 2025-08-08
Sort commands in transaction manager

prvyk:
No functional change here beyond sorting the commands in TxnKeyManager to make it clearer what is and isn't supported.

(It's easy to verify these are all moved lines using git diff --color-moved. The extra deletions in the stats are just empty lines being removed)


vazois (approved):
This is good. However, I think the TxnManager should leverage RespCommandsInfo, similar to how the slot verification implementation works.
It is probably a refactoring job for another PR.

PR #1333, opened 2025-08-09
Scan and get distribution of hybrid log scan

hamdaankhalid:
Add an info command to scan your hybrid log


vazois:
how expensive is this operation? INFO command might be called periodically from operator side to get the state of the instance. So it is best to avoid putting there commands that require heavy processing.

hamdaankhalid:
> how expensive is this operation? INFO command might be called periodically from operator side to get the state of the instance. So it is best to avoid putting there commands that require heavy processing.

It is not a part of default INFO set

PR #1335, opened 2025-08-12
Implement several HELP subcommands

prvyk:
This PR adds HELP subcommands to all commands with subcommands that are supported at least in part but didn't have HELP implemented despite it being implemented in the reference implementations.

### DIFF
--- a/website/docs/commands/analytics.md
+++ b/website/docs/commands/analytics.md
@@ -234,6 +234,16 @@ Array reply: a list of active channels, optionally matching the specified patter
 
 ---
 
+### PUBSUB HELP
+
+```bash
+    PUBSUB HELP
+```
+
+The PUBSUB HELP command returns a helpful text describing the different subcommands.
+
+---
+
 ### PUBSUB NUMPAT
 #### Syntax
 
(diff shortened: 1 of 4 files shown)


libs/server/Lua/LuaCommands.cs
```diff
@@ -271,6 +271,34 @@ private bool NetworkScriptFlush()
...
+        /// <returns></returns>
+        private bool NetworkScriptHelp()
+        {
+            if (parseState.Count != 0)
+            {
+                return AbortWithWrongNumberOfArguments("script|help");
+            }
+
+            WriteHelp(
+                "SCRIPT <subcommand> [<arg> [value] [opt] ...]. Subcommands are:",
```
TalZaccai, line 287: We should add a RespCommandsHelp.json resource file that gets loaded on an as-needed basis, similarly to Docs/Info. This file can also get populated automagically by the CommandInfoUpdater tool.

TalZaccai (changes requested):

TalZaccai:
Hey, any updates on this one? Thanks!

prvyk:
I've been busy... I'll get back to it soon.

PR #1278, opened 2025-07-04
Implement the inline command format

prvyk:
This PR implements parsing command line supplied using Redis's inline command format.

https://redis.io/docs/latest/develop/reference/protocol-spec/#inline-commands

The inline command format allows people to connect with telnet and just write commands, without requiring other command line tools. It can be useful to avoid *-cli quoting and to see raw responses. Also a very few poorly written clients may send the inline format, so it can be useful there too.

The basic implementation strategy was to take the current code, spread it out a bit, and in the same point after seeing the buffer does not contain a RESP string, try to parse as inline command: We separate the buffer to an array of ArgSlice pointing to the original buffer. Each ArgSlice pointing to a 'word' of the original command line. The first two 'words' are printed to a temporary RESP formatted string, and then we run the regular parsing on it. If a command was matched, we initialize the parseState from the ArgSlice array. The intention is to keep the main path fast and (almost) untouched, while the inline path does the necessary work to match itself to the rest of the code. It was also preferred to not implement another command matching for this.

---

### Examples:

> telnet localhost 6379

>ECHO 'hello world!'
> >$12
> >hello world!

>SET key "a\x0Ab"
> >+OK

>GET key
> >$3
> >a
> >x

The PR implements reading the format fully (as far as I can tell, given it's not very specified), with quoting and escaping, matching even rather odd reference behaviour.

>PI"NG"
> >+PONG

>PING \\"a b"
> >$4
> >\\a b

>"ping"
> >+PONG

Another thing to note is that Garnet doesn't close the connection on protocol errors (unlike redis/valkey), so it's much nicer to type there compared to references which will close the connection whenever quotes are mismatched.

PR #1336, opened 2025-08-13
Avoid unnecessary/buggy readHead access during command processing

prvyk:
This PR contains two changes to avoid access to bytesRead, readHead or direct parsing during command parsing after the RespCommand is decided and parseState is set:

1. ListReadKeysWithCount() in TxnKeyManager doesn't need to reparse the input. It has all required information already in respSession.parseState variable. We can then erase GetCommandAsArgSlice() as it's no longer necessary.

2. MakeUpperCase() in RespServerSession used bytesRead - readHead for length. Which is wrong when called from GetOperationDirection(). There the actual length is only the argSlice length which is usually far smaller. This accidentally works for typical inputs (probably fails for crafted inputs), fix it by supplying the actual length.

(Both of these are from #1278, but are logically independent fixes we can push separately).


libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -487,17 +488,20 @@ private int ListKeys(int inputCount, StoreType storeType, LockType type)
...
-            if (!success) return -2;
+            if (respSession.parseState.Count == 0)
+                return -2;
 
-            if (!NumUtils.TryParse(numKeysArg.ReadOnlySpan, out int numKeys)) return -2;
+            if (!respSession.parseState.TryGetInt(0, out var numKeys))
+                return -2;
 
-            for (var i = 0; i < numKeys; i++)
+            if (numKeys + mandatoryArgs + 1 > respSession.parseState.Count)
```
TalZaccai: This seems unrelated to the role of `TxnKeyManager`, why are we verifying additional argument count here?

prvyk: Right, there's no need to. Redis/valkey don't do this either.

TalZaccai (approved):

PR #1340, opened 2025-08-13
Add `global.json`

JamieMagee:
This adds a `global.json` file to enforce a minimum version of the .NET SDK.

This project contains a mix of .NET 8 and 9 projects, but newer versions of the SDK are backwards compatible, and can build projects targeting older versions of .NET.

I also simplified the GitHub Actions workflows, as `actions/setup-dotnet` looks at `global.json` by default. So there is no need to explicitly specify `dotnet-version` or `global-json-file`.

Finally, I added a Dependabot configuration which will keep `global.json` up-to-date as new security patches are released for .NET.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "dotnet-sdk"
+    directory: "/"
+    schedule:
+      interval: "monthly"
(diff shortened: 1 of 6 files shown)


darrenge (approved):
Did some spot checks and looks like it is working as hoped. Tal also looked it over and agrees but since she pushed a change to this PR, she can't approve the PR.

badrishc:
Great work, thanks!

PR #1344, opened 2025-08-14
Update `BenchmarkDotNet` from `0.14.0` to `0.15.2`

JamieMagee:
The benchmarks were failing after #1340. Updating BenchmarkDotNet from `0.14.0` to `0.15.2` seems to fix the issue. I can't see a relevant issue in [the changelogs][1], and unfortunately, I don't have time to `git bisect` it to find out.

EDIT: I am 99% sure it is this change: https://github.com/dotnet/BenchmarkDotNet/pull/2645

[1]: https://benchmarkdotnet.org/changelog/index.html


darrenge (approved):
Looks good. Ran BDN tests all passed

PR #1346, opened 2025-08-14
Sketch out VectorAPI support

kevin-montrose:
Redis has recently added a Vector Set API, exploring implementing it.

Some assumptions:
 - Actual Vector Set implementation is outside of .NET, I've mocked it up against as an interface
    * Long term that interface goes away and we just pick an implementation, it's there to keep me honest about the abstraction
 - All Vector Set data is in the MainStore, so replication and checkpoint "just work"
 - While creating a Vector Set is a write op, adding a vector to an existing set is a read op
   * This is modeled as taking Shared locks around `VADD`, `VSIM`, etc., and only upgrading to an exclusive lock when we need create an index

Open questions:
 - [ ] Need to conceal Vector Set data from other MainStore ops
 - [ ] Need to prevent attributes and neighbor lists from clobbering raw vectors, probably also part of ^
 - [ ] Implement remaining vector commands
 - [ ] Pull in an actual implementation to work against
 - [ ] I hate the `[ThreadStatic]` but keeping the context on the stack is hard too; other options?

### DIFF
--- a/libs/server/API/GarnetApi.cs
+++ b/libs/server/API/GarnetApi.cs
@@ -21,19 +21,22 @@ namespace Garnet.server
     /// <summary>
     /// Garnet API implementation
     /// </summary>
-    public partial struct GarnetApi<TContext, TObjectContext> : IGarnetApi, IGarnetWatchApi
+    public partial struct GarnetApi<TContext, TObjectContext, TVectorContext> : IGarnetApi, IGarnetWatchApi
         where TContext : ITsavoriteContext<SpanByte, SpanByte, RawStringInput, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator>
         where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
+        where TVectorContext : ITsavoriteContext<SpanByte, SpanByte, VectorInput, SpanByte, long, VectorSessionFunctions, MainStoreFunctions, MainStoreAllocator>
     {
         readonly StorageSession storageSession;
         TContext context;
         TObjectContext objectContext;
+        TVectorContext vectorContext;
 
-        internal GarnetApi(StorageSession storageSession, TContext context, TObjectContext objectContext)
+        internal GarnetApi(StorageSession storageSession, TContext context, TObjectContext objectContext, TVectorContext vectorContext)
         {
             this.storageSession = storageSession;
             this.context = context;
             this.objectContext = objectContext;
+            this.vectorContext = vectorContext;
         }
 
         #region WATCH
@@ -479,6 +482,26 @@ namespace Garnet.server
         /// <inheritdoc />
         public bool ResetScratchBuffer(int offset)
             => storageSession.scratchBufferBuilder.ResetScratchBuffer(offset);
+        #endregion
+
+        #region VectorSet commands
+
+        /// <inheritdoc />
+        public GarnetStatus VectorSetAdd(SpanByte key, int reduceDims, ReadOnlySpan<float> values, ArgSlice element, VectorQuantType quantizer, int buildExplorationFactor, ArgSlice attributes, int numLinks, out VectorManagerResult result)
+        => storageSession.VectorSetAdd(key, reduceDims, values, element, quantizer, buildExplorationFactor, attributes, numLinks, out result);
+
+        /// <inheritdoc />
+        public GarnetStatus VectorSetValueSimilarity(SpanByte key, ReadOnlySpan<float> values, int count, float delta, int searchExplorationFactor, ReadOnlySpan<byte> filter, int maxFilteringEffort, ref SpanByteAndMemory outputIds, ref SpanByteAndMemory outputDistances, out VectorManagerResult result)
+        => storageSession.VectorSetValueSimilarity(key, values, count, delta, searchExplorationFactor, filter, maxFilteringEffort, ref outputIds, ref outputDistances, out result);
+
+        /// <inheritdoc />
+        public GarnetStatus VectorSetElementSimilarity(SpanByte key, ReadOnlySpan<byte> element, int count, float delta, int searchExplorationFactor, ReadOnlySpan<byte> filter, int maxFilteringEffort, ref SpanByteAndMemory outputIds, ref SpanByteAndMemory outputDistances, out VectorManagerResult result)
+        => storageSession.VectorSetElementSimilarity(key, element, count, delta, searchExplorationFactor, filter, maxFilteringEffort, ref outputIds, ref outputDistances, out result);
+
+        /// <inheritdoc/>
+        public GarnetStatus VectorEmbedding(SpanByte key, ReadOnlySpan<byte> element, ref SpanByteAndMemory outputDistances)
+        => storageSession.VectorEmbedding(key, element, ref outputDistances);
+
         #endregion
     }
 }
--- a/libs/server/Resp/Vector/VectorManager.cs
+++ b/libs/server/Resp/Vector/VectorManager.cs
@@ -1,0 +1,608 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Buffers;
+using System.Buffers.Binary;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
+using Tsavorite.core;
+
+namespace Garnet.server
+{
+    using MainStoreAllocator = SpanByteAllocator<StoreFunctions<SpanByte, SpanByte, SpanByteComparer, SpanByteRecordDisposer>>;
+    using MainStoreFunctions = StoreFunctions<SpanByte, SpanByte, SpanByteComparer, SpanByteRecordDisposer>;
+
+    internal sealed unsafe class DummyService : IVectorService
+    {
+        private const byte FullVector = 0;
+        private const byte NeighborList = 1;
+        private const byte QuantizedVector = 2;
+        private const byte Attributes = 3;
+
+        private sealed class ByteArrayEqualityComparer : IEqualityComparer<byte[]>
+        {
+            public static readonly ByteArrayEqualityComparer Instance = new();
+
+            private ByteArrayEqualityComparer() { }
+
+            public bool Equals(byte[] x, byte[] y)
+            => x.AsSpan().SequenceEqual(y);
+
+            public int GetHashCode([DisallowNull] byte[] obj)
+            {
+                var hash = new HashCode();
+                hash.AddBytes(obj);
+
+                return hash.ToHashCode();
+            }
+        }
+
+        private readonly ConcurrentDictionary<nint, (VectorReadDelegate Read, VectorWriteDelegate Write, VectorDeleteDelegate Delete, ConcurrentDictionary<byte[], byte> Members)> data = new();
+
+        /// <inheritdoc/>
+        public bool UseUnmanagedCallbacks { get; } = false;
+
+        /// <inheritdoc/>
+        public nint CreateIndexUnmanaged(ulong context, uint dimensions, uint reduceDims, VectorQuantType quantType, uint buildExplorationFactor, uint numLinks, delegate* unmanaged[Cdecl]<ulong, byte*, nuint, byte*, nuint, int> readCallback, delegate* unmanaged[Cdecl]<ulong, byte*, nuint, byte*, nuint, bool> writeCallback, delegate* unmanaged[Cdecl]<ulong, byte*, nuint, bool> deleteCallback)
+        => throw new NotImplementedException();
+
+        /// <inheritdoc/>
+        public nint CreateIndexManaged(ulong context, uint dimensions, uint reduceDims, VectorQuantType quantType, uint buildExplorationFactor, uint numLinks, VectorReadDelegate readCallback, VectorWriteDelegate writeCallback, VectorDeleteDelegate deleteCallback)
+        {
+            var ptr = (nint)(context + 17); // some arbitrary non-multiple of 4 to mess with things
+
+            if (!data.TryAdd(ptr, new(readCallback, writeCallback, deleteCallback, new(ByteArrayEqualityComparer.Instance))))
+            {
+                throw new InvalidOperationException("Shouldn't be possible");
+            }
+
+            return ptr;
+        }
+
+        /// <inheritdoc/>
+        public void DropIndex(ulong context, nint index)
+        {
+            if (!data.TryRemove(index, out _))
+            {
+                throw new InvalidOperationException("Attempted to drop index that was already dropped");
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool Insert(ulong context, nint index, ReadOnlySpan<byte> id, ReadOnlySpan<float> vector, ReadOnlySpan<byte> attributes)
+        {
+            var (_, write, _, members) = data[index];
+
+            // save vector data
+            _ = members.AddOrUpdate(id.ToArray(), static (_) => 0, static (key, old) => (byte)(old + 1));
+            _ = write(context + FullVector, id, MemoryMarshal.Cast<float, byte>(vector));
+
+            if (!attributes.IsEmpty)
+            {
+                _ = write(context + Attributes, id, attributes);
+            }
+
+            return true;
+        }
+
+        /// <inheritdoc/>
+        public int SearchVector(ulong context, nint index, ReadOnlySpan<float> vector, float delta, int searchExplorationFactor, ReadOnlySpan<byte> filter, int maxFilteringEffort, Span<byte> outputIds, Span<float> outputDistances, out nint continuation)
+        {
+            var (read, _, _, members) = data[index];
+
+            // Hack, just use a fixed sized buffer for now
+            Span<byte> memberData = stackalloc byte[128];
+
+            var matches = 0;
+            var remainingOutputIds = outputIds;
+            var remainingDistances = outputDistances;
+
+            // We don't actually do the distance calc, this is just for testing
+            foreach (var member in members.Keys)
+            {
+                var len = read(context + FullVector, member, memberData);
+                if (len == 0)
+                {
+                    continue;
+                }
+
+                var asFloats = MemoryMarshal.Cast<byte, float>(memberData[..len]);
+                if (member.Length > remainingOutputIds.Length + sizeof(int))
+                {
+                    // This is where a continuation would be set
+                    throw new NotImplementedException();
+                }
+
+                BinaryPrimitives.WriteInt32LittleEndian(remainingOutputIds, member.Length);
+                remainingOutputIds = remainingOutputIds[sizeof(int)..];
+                member.AsSpan().CopyTo(remainingOutputIds);
+                remainingOutputIds = remainingOutputIds[member.Length..];
+
+                remainingDistances[0] = (float)Random.Shared.NextDouble();
+                remainingDistances = remainingDistances[1..];
+                matches++;
+
+                if (remainingDistances.IsEmpty)
+                {
+                    break;
+                }
+            }
+
+            continuation = 0;
+            return matches;
+        }
+
+        /// <inheritdoc/>
+        public int SearchElement(ulong context, nint index, ReadOnlySpan<byte> id, float delta, int searchExplorationFactor, ReadOnlySpan<byte> filter, int maxFilteringEffort, Span<byte> outputIds, Span<float> outputDistances, out nint continuation)
+        {
+            var (read, _, _, members) = data[index];
+
+            // Hack, just use a fixed sized buffer for now
+            Span<byte> memberData = stackalloc byte[128];
+            var len = read(context + FullVector, id, memberData);
+            if (len == 0)
+            {
+                continuation = 0;
+                return 0;
+            }
+
+            var vector = MemoryMarshal.Cast<byte, float>(memberData[..len]);
+            return SearchVector(context, index, vector, delta, searchExplorationFactor, filter, maxFilteringEffort, outputIds, outputDistances, out continuation);
+        }
+
+        /// <inheritdoc/>
+        public int ContinueSearch(ulong context, nint index, nint continuation, Span<byte> outputIds, Span<float> outputDistances, out nint newContinuation)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc/>
+        public bool TryGetEmbedding(ulong context, nint index, ReadOnlySpan<byte> id, Span<float> dimensions)
+        {
+            var (read, _, _, _) = data[index];
+
+            return read(context + FullVector, id, MemoryMarshal.Cast<float, byte>(dimensions)) != 0;
+        }
+    }
+
+    public enum VectorManagerResult
+    {
+        Invalid = 0,
+
+        OK,
+        BadParams,
+        Duplicate,
+        MissingElement,
+    }
+
+    /// <summary>
+    /// Methods for managing an implementation of various vector operations.
+    /// </summary>
+    internal static class VectorManager
+    {
+        internal const int IndexSizeBytes = Index.Size;
+
+        [StructLayout(LayoutKind.Explicit, Size = Size)]
+        private struct Index
+        {
+            internal const int Size = 33;
+
+            [FieldOffset(0)]
+            public ulong Context;
+            [FieldOffset(8)]
+            public ulong IndexPtr;
+            [FieldOffset(16)]
+            public uint Dimensions;
+            [FieldOffset(20)]
+            public uint ReduceDims;
+            [FieldOffset(24)]
+            public uint NumLinks;
+            [FieldOffset(28)]
+            public uint BuildExplorationFactor;
+            [FieldOffset(32)]
+            public VectorQuantType QuantType;
+        }
+
+        private static readonly unsafe delegate* unmanaged[Cdecl]<ulong, byte*, nuint, byte*, nuint, int> ReadCallbackPtr = &ReadCallbackUnmanaged;
+        private static readonly unsafe delegate* unmanaged[Cdecl]<ulong, byte*, nuint, byte*, nuint, bool> WriteCallbackPtr = &WriteCallbackUnmanaged;
+        private static readonly unsafe delegate* unmanaged[Cdecl]<ulong, byte*, nuint, bool> DeleteCallbackPtr = &DeleteCallbackUnmanaged;
+
+        private static readonly VectorReadDelegate ReadCallbackDel = ReadCallbackManaged;
+        private static readonly VectorWriteDelegate WriteCallbackDel = WriteCallbackManaged;
+        private static readonly VectorDeleteDelegate DeleteCallbackDel = DeleteCallbackManaged;
+
+        private static readonly IVectorService Service = new DummyService();
+
+        private static ulong NextContextValue;
+
+        [ThreadStatic]
+        private static StorageSession ActiveThreadSession;
+
+        /// <summary>
+        /// Get a new unique context for a vector set.
+        /// 
+        /// This value is guaranteed to not be shared by any other vector set in the store.
+        /// </summary>
+        /// <returns></returns>
+        private static ulong NextContext()
+        => Interlocked.Add(ref NextContextValue, 4);
+
+        [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
+        private static unsafe int ReadCallbackUnmanaged(ulong context, byte* keyData, nuint keyLength, byte* writeData, nuint writeLength)
+        => ReadCallbackManaged(context, MemoryMarshal.CreateReadOnlySpan(ref *keyData, (int)keyLength), MemoryMarshal.CreateSpan(ref *writeData, (int)writeLength));
+
+        [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
+        private static unsafe bool WriteCallbackUnmanaged(ulong context, byte* keyData, nuint keyLength, byte* writeData, nuint writeLength)
+        => WriteCallbackManaged(context, MemoryMarshal.CreateReadOnlySpan(ref *keyData, (int)keyLength), MemoryMarshal.CreateReadOnlySpan(ref *writeData, (int)writeLength));
+
+        [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
+        private static unsafe bool DeleteCallbackUnmanaged(ulong context, byte* keyData, nuint keyLength)
+        => DeleteCallbackManaged(context, MemoryMarshal.CreateReadOnlySpan(ref *keyData, (int)keyLength));
+
+        private static int ReadCallbackManaged(ulong context, ReadOnlySpan<byte> key, Span<byte> value)
+        {
+            ref var ctx = ref ActiveThreadSession.vectorContext;
+            var keySpan = SpanByte.FromPinnedSpan(key);
+            VectorInput input = new();
+            var outputSpan = SpanByte.FromPinnedSpan(value);
+
+            var status = ctx.Read(ref keySpan, ref input, ref outputSpan);
+            if (status.IsPending)
+            {
+                CompletePending(ref status, ref outputSpan, ref ctx);
+            }
+
+            if (status.Found)
+            {
+                return outputSpan.Length;
+            }
+
+            return 0;
+        }
+
+        private static bool WriteCallbackManaged(ulong context, ReadOnlySpan<byte> key, ReadOnlySpan<byte> value)
+        {
+            ref var ctx = ref ActiveThreadSession.vectorContext;
+            var keySpan = SpanByte.FromPinnedSpan(key);
+            VectorInput input = new();
+            var valueSpan = SpanByte.FromPinnedSpan(value);
+
+            Span<byte> output = stackalloc byte[1];
+            var outputSpan = SpanByte.FromPinnedSpan(output);
+
+            var status = ctx.Upsert(ref keySpan, ref input, ref valueSpan, ref outputSpan);
+            if (status.IsPending)
+            {
+                CompletePending(ref status, ref outputSpan, ref ctx);
+            }
+
+            return status.IsCompletedSuccessfully;
+        }
+
+        private static bool DeleteCallbackManaged(ulong context, ReadOnlySpan<byte> key)
+        {
+            ref var ctx = ref ActiveThreadSession.vectorContext;
+            var keySpan = SpanByte.FromPinnedSpan(key);
+
+            var status = ctx.Delete(ref keySpan);
+            Debug.Assert(!status.IsPending, "Deletes should never go async");
+
+            return status.IsCompletedSuccessfully;
+        }
+
+        private static void CompletePending<TContext>(ref Status status, ref SpanByte output, ref TContext objectContext)
+            where TContext : ITsavoriteContext<SpanByte, SpanByte, VectorInput, SpanByte, long, VectorSessionFunctions, MainStoreFunctions, MainStoreAllocator>
+        {
+            objectContext.CompletePendingWithOutputs(out var completedOutputs, wait: true);
+            var more = completedOutputs.Next();
+            Debug.Assert(more);
+            status = completedOutputs.Current.Status;
+            output = completedOutputs.Current.Output;
+            Debug.Assert(!completedOutputs.Next());
+            completedOutputs.Dispose();
+        }
+
+        /// <summary>
+        /// Construct a new index, and stash enough data to recover it with <see cref="ReadIndex"/>.
+        /// </summary>
+        internal static void CreateIndex(
+            uint dimensions,
+            uint reduceDims,
+            VectorQuantType quantType,
+            uint buildExplorationFactory,
+            uint numLinks,
+            ref SpanByte indexValue)
+        {
+            var context = NextContext();
+
+            // Enforce defaults, which match Redis; see https://redis.io/docs/latest/commands/vadd/
+            quantType = quantType == VectorQuantType.Invalid ? VectorQuantType.Q8 : quantType;
+            buildExplorationFactory = buildExplorationFactory == 0 ? 200 : buildExplorationFactory;
+            numLinks = numLinks == 0 ? 16 : numLinks;
+
+            nint indexPtr;
+            if (Service.UseUnmanagedCallbacks)
+            {
+                unsafe
+                {
+                    indexPtr = Service.CreateIndexUnmanaged(context, dimensions, reduceDims, quantType, buildExplorationFactory, numLinks, ReadCallbackPtr, WriteCallbackPtr, DeleteCallbackPtr);
+                }
+            }
+            else
+            {
+                indexPtr = Service.CreateIndexManaged(context, dimensions, reduceDims, quantType, buildExplorationFactory, numLinks, ReadCallbackDel, WriteCallbackDel, DeleteCallbackDel);
+            }
+
+            ref var asIndex = ref Unsafe.As<byte, Index>(ref MemoryMarshal.GetReference(indexValue.AsSpan()));
+            asIndex.Context = context;
+            asIndex.Dimensions = dimensions;
+            asIndex.ReduceDims = reduceDims;
+            asIndex.QuantType = quantType;
+            asIndex.BuildExplorationFactor = buildExplorationFactory;
+            asIndex.NumLinks = numLinks;
+            asIndex.IndexPtr = (ulong)indexPtr;
+        }
+
+        internal static void ReadIndex(
+            ReadOnlySpan<byte> indexValue,
+            out ulong context,
+            out uint dimensions,
+            out uint reduceDims,
+            out VectorQuantType quantType,
+            out uint buildExplorationFactor,
+            out uint numLinks,
+            out nint indexPtr
+        )
+        {
+            Debug.Assert(indexValue.Length == Index.Size, "Index value is incorrect, implies vector set index is probably corrupted");
+
+            ref var asIndex = ref Unsafe.As<byte, Index>(ref MemoryMarshal.GetReference(indexValue));
+
+            context = asIndex.Context;
+            dimensions = asIndex.Dimensions;
+            reduceDims = asIndex.ReduceDims;
+            quantType = asIndex.QuantType;
+            buildExplorationFactor = asIndex.BuildExplorationFactor;
+            numLinks = asIndex.NumLinks;
+            indexPtr = (nint)asIndex.IndexPtr;
+
+            Debug.Assert((context % 4) == 0, "Context not as expected, vector set index is probably corrupted");
+        }
+
+        /// <summary>
+        /// Add a vector to a vector set encoded by <paramref name="indexValue"/>.
+        /// 
+        /// Assumes that the index is locked in the Tsavorite store.
+        /// </summary>
+        /// <returns>Result of the operaiton.</returns>
+        internal static VectorManagerResult TryAdd(
+            StorageSession currentStorageSession,
+            ReadOnlySpan<byte> indexValue,
+            ReadOnlySpan<byte> element,
+            ReadOnlySpan<float> values,
+            ReadOnlySpan<byte> attributes,
+            uint providedReduceDims,
+            VectorQuantType providedQuantType,
+            uint providedBuildExplorationFactor,
+            uint providedNumLinks
+        )
+        {
+            ActiveThreadSession = currentStorageSession;
+            try
+            {
+                ReadIndex(indexValue, out var context, out var dimensions, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out var indexPtr);
+
+                if (dimensions != values.Length)
+                {
+                    return VectorManagerResult.BadParams;
+                }
+
+                if (providedReduceDims != 0 && providedReduceDims != reduceDims)
+                {
+                    return VectorManagerResult.BadParams;
+                }
+
+                if (providedQuantType != VectorQuantType.Invalid && providedQuantType != quantType)
+                {
+                    return VectorManagerResult.BadParams;
+                }
+
+                if (providedBuildExplorationFactor != 0 && providedBuildExplorationFactor != buildExplorationFactor)
+                {
+                    return VectorManagerResult.BadParams;
+                }
+
+                if (providedNumLinks != 0 && providedNumLinks != numLinks)
+                {
+                    return VectorManagerResult.BadParams;
+                }
+
+                var insert =
+                    Service.Insert(
+                        context,
+                        indexPtr,
+                        element,
+                        values,
+                        attributes
+                    );
+
+                if (insert)
+                {
+                    return VectorManagerResult.OK;
+                }
+
+                return VectorManagerResult.Duplicate;
+            }
+            finally
+            {
+                ActiveThreadSession = null;
+            }
+        }
+
+        /// <summary>
+        /// Perform a similarity search given a vector to compare against.
+        /// </summary>
+        internal static VectorManagerResult ValueSimilarity(
+            StorageSession currentStorageSession,
+            ReadOnlySpan<byte> indexValue,
+            ReadOnlySpan<float> values,
+            int count,
+            float delta,
+            int searchExplorationFactor,
+            ReadOnlySpan<byte> filter,
+            int maxFilteringEffort,
+            ref SpanByteAndMemory outputIds,
+            ref SpanByteAndMemory outputDistances
+        )
+        {
+            ActiveThreadSession = currentStorageSession;
+            try
+            {
+                ReadIndex(indexValue, out var context, out var dimensions, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out var indexPtr);
+
+                // Make sure enough space in distances for requested count
+                if (count > outputDistances.Length)
+                {
+                    if (!outputDistances.IsSpanByte)
+                    {
+                        outputDistances.Memory.Dispose();
+                    }
+
+                    outputDistances = new SpanByteAndMemory(MemoryPool<byte>.Shared.Rent(count));
+                }
+
+                var found =
+                    Service.SearchVector(
+                        context,
+                        indexPtr,
+                        values,
+                        delta,
+                        searchExplorationFactor,
+                        filter,
+                        maxFilteringEffort,
+                        outputIds.AsSpan(),
+                        MemoryMarshal.Cast<byte, float>(outputDistances.AsSpan()),
+                        out var continuation
+                    );
+
+                if (continuation != 0)
+                {
+                    // TODO: paged results!
+                    throw new NotImplementedException();
+                }
+
+                outputDistances.Length = sizeof(float) * found;
+
+                return VectorManagerResult.OK;
+            }
+            finally
+            {
+                ActiveThreadSession = null;
+            }
+        }
+
+        /// <summary>
+        /// Perform a similarity search given a vector to compare against.
+        /// </summary>
+        internal static VectorManagerResult ElementSimilarity(
+            StorageSession currentStorageSession,
+            ReadOnlySpan<byte> indexValue,
+            ReadOnlySpan<byte> element,
+            int count,
+            float delta,
+            int searchExplorationFactor,
+            ReadOnlySpan<byte> filter,
+            int maxFilteringEffort,
+            ref SpanByteAndMemory outputIds,
+            ref SpanByteAndMemory outputDistances
+        )
+        {
+            ActiveThreadSession = currentStorageSession;
+            try
+            {
+                ReadIndex(indexValue, out var context, out var dimensions, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out var indexPtr);
+
+                // Make sure enough space in distances for requested count
+                if (count * sizeof(float) > outputDistances.Length)
+                {
+                    if (!outputDistances.IsSpanByte)
+                    {
+                        outputDistances.Memory.Dispose();
+                    }
+
+                    outputDistances = new SpanByteAndMemory(MemoryPool<byte>.Shared.Rent(count * sizeof(float)));
+                }
+
+                var found =
+                    Service.SearchElement(
+                        context,
+                        indexPtr,
+                        element,
+                        delta,
+                        searchExplorationFactor,
+                        filter,
+                        maxFilteringEffort,
+                        outputIds.AsSpan(),
+                        MemoryMarshal.Cast<byte, float>(outputDistances.AsSpan()),
+                        out var continuation
+                    );
+
+                if (continuation != 0)
+                {
+                    // TODO: paged results!
+                    throw new NotImplementedException();
+                }
+
+                outputDistances.Length = sizeof(float) * found;
+
+                return VectorManagerResult.OK;
+            }
+            finally
+            {
+                ActiveThreadSession = null;
+            }
+        }
+
+        internal static bool TryGetEmbedding(StorageSession currentStorageSession, ReadOnlySpan<byte> indexValue, ReadOnlySpan<byte> element, ref SpanByteAndMemory outputDistances)
+        {
+            ActiveThreadSession = currentStorageSession;
+            try
+            {
+                ReadIndex(indexValue, out var context, out var dimensions, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out var indexPtr);
+
+                // Make sure enough space in distances for requested count
+                if (dimensions * sizeof(float) > outputDistances.Length)
+                {
+                    if (!outputDistances.IsSpanByte)
+                    {
+                        outputDistances.Memory.Dispose();
+                    }
+
+                    outputDistances = new SpanByteAndMemory(MemoryPool<byte>.Shared.Rent((int)dimensions * sizeof(float)), (int)dimensions * sizeof(float));
+                }
+                else
+                {
+                    outputDistances.Length = (int)dimensions * sizeof(float);
+                }
+
+                return
+                    Service.TryGetEmbedding(
+                        context,
+                        indexPtr,
+                        element,
+                        MemoryMarshal.Cast<byte, float>(outputDistances.AsSpan())
+                    );
+            }
+            finally
+            {
+                ActiveThreadSession = null;
+            }
+        }
+    }
+}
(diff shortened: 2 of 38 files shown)


badrishc:
This all makes sense. In storage-v2, we have the notion of a [LogRecord](https://github.com/microsoft/garnet/blob/tedhar/object-allocator/website/docs/dev/tsavorite/logrecord.md) record format, that already includes optional fields such as ETag and Expiration. Adding a namespace would be analogous to this, with a couple of differences:

* We use one RecordInfo bit to indicate whether the record has a namespace field
* If true, then the NameSpace byte (or larger if needed) exists as part of the record, roughly: <RecordInfo, key, value, etag?,  expiration?, namespace?, ...>
* Tsavorite APIs are adjusted to accept the optional namespace (in addition to key/expiration/etag).
* All hash codes and key equalities in Tsavorite would need to incorporate the namespace in their computations.
* Certain namespaces (e.g., namespace numbers starting with "11") would be reserved for vector-set usage. Within that sub-namespace, the vector-sets can partition bits as they want.

TBD: how to handle larger namespace names in the same framework (e.g., "/user/foo/"), and is that useful/necessary to make as a first-class citizen versus users directly incorporating in the actual key.

prvyk:
>Adding a namespace would be analogous to this, with a couple of differences:
> 
> ...
> 
> TBD: how to handle larger namespace names in the same framework (e.g., "/user/foo/"), and is that useful/necessary to make as a first-class citizen versus users directly incorporating in the actual key.

Namespaces can be used for a lot of things, like replacing the current numbered database implementation with something that is supported cluster-wide as in valkey (In that case they end up in the same AOF file I guess?). Also, quite a few RESP-accepting DBs have a namespace implementation of sorts, e.g. [kvrocks](https://kvrocks.apache.org/docs/namespace/). I had an idea of combining ACLs and database numbers (IMHO, this would be usually better than redis's prefix ACLs, because it doesn't require client to cooperate), and namespaces would be just as natural here.

vazois:
> Once Main & Object stores are merged, we should rework this like so:
> 
> * Use a bit in `RecordInfo` to indicate "has namespace"
> * Map Vector Set contexts to namespaces, using a similar byte sequence idea to group Vector Sets into their own class of namespaces
>   
>   * Something akin to "all namespaces that start `0b10xx_xxxx` are Vector Sets"
> * All elements of a Vector Set go into the same namespace
>   
>   * In practice this will be multiple namespaces per Vector Set since each one needs at least quantized vectors and neighbor lists

This sounds more like a data type instead of a namespace. I propose to treat it as such and have the option to include Bitmaps and Hyperloglog on this. In fact, we might need to think about the possibility of adding other specially encoded complex data types as bulk strings.
Namespace can exist separately

kevin-montrose:
> > Once Main & Object stores are merged, we should rework this like so:
> > 
> > * Use a bit in `RecordInfo` to indicate "has namespace"
> > * Map Vector Set contexts to namespaces, using a similar byte sequence idea to group Vector Sets into their own class of namespaces
> >   
> >   * Something akin to "all namespaces that start `0b10xx_xxxx` are Vector Sets"
> > * All elements of a Vector Set go into the same namespace
> >   
> >   * In practice this will be multiple namespaces per Vector Set since each one needs at least quantized vectors and neighbor lists
> 
> This sounds more like a data type instead of a namespace. I propose to treat it as such and have the option to include Bitmaps and Hyperloglog on this. In fact, we might need to think about the possibility of adding other specially encoded complex data types as bulk strings. Namespace can exist separately

You're right here - there are two orthogonal concerns:
 1. Vector Set top-level key, which _is_ visible to other RESP commands (like `TYPE`, `DEL`, etc.) but needs to be distinguishable from other non-Vector Set data
 2. Vector element data _in_ a Vector Set, which needs to be in the main store but inaccessible to other commands
 
 This branch currently uses a namespace byte in `SpanByte` for 2, and a bit in `RecordInfo` for 1.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -0,0 +1,922 @@
...
+                completedOutputs.Dispose();
+            }
+        }
+
+        /// <summary>
+        /// Wait until all ops passed to <see cref="HandleVectorSetAddReplication"/> have completed.
+        /// </summary>
+        internal void WaitForVectorOperationsToComplete()
+        {
+            while (Interlocked.CompareExchange(ref replicationReplayPendingVAdds, 0, 0) != 0)
```
badrishc: Can avoid the long spin using something like a ManualResetEventSlim that is signaled by the reader when it atomically decrements replicationReplayPendingVAdds to 0.

libs/server/AOF/AofProcessor.cs
```diff
@@ -428,13 +435,24 @@ static void StoreRMW(BasicContext<SpanByte, SpanByte, RawStringInput, SpanByteAn
...
+
+            // VADD requires special handling, shove it over to the VectorManager
+            if (storeInput.header.cmd == RespCommand.VADD)
+            {
+                vectorManager.HandleVectorSetAddReplication(obtainServerSession, ref key, ref storeInput);
+                return;
+            }
+            else
+            {
+                vectorManager.WaitForVectorOperationsToComplete();
```
badrishc, line 406: Need to also WaitForVectorOperationsToComplete at the end of the ongoing AOF batch, so that the FromPinnedSpan keys etc being used by worker threads stay safe at the end.

kevin-montrose: Currently I'm just making copies before moving the op onto a task.

Avoiding that would be an optimization, so I'll make a note.

libs/storage/Tsavorite/cs/src/core/VarLen/SpanByte.cs
```diff
@@ -25,8 +25,10 @@ public unsafe struct SpanByte
         private const int UnserializedBitMask = 1 << 31;
```
tiagonapoli, line 25: nit: maybe update <remarks> in the class description - it's currently mentioning max payload length 1GB, but now with namespace bit it will be ~500MB - Unless that's expected to change with Store V2

kevin-montrose: Store V2 changes all of this, so I'm not too concerned about that particular nit right now.

libs/server/Storage/Session/MainStore/VectorStoreOps.cs
```diff
@@ -0,0 +1,279 @@
...
+            Span<byte> indexSpan = stackalloc byte[VectorManager.IndexSizeBytes];
+
+            using (vectorManager.ReadVectorIndex(this, ref key, ref input, indexSpan, out var status))
+            {
+                if (status != GarnetStatus.OK)
+                {
+                    return status;
+                }
+
+                // After a successful read we add the vector while holding a shared lock
```
tiagonapoli: nit
```suggestion
                // After a successful read we remove the vector while holding a shared lock
```

libs/server/API/GarnetApi.cs
```diff
@@ -21,19 +21,22 @@ namespace Garnet.server
...
-    public partial struct GarnetApi<TContext, TObjectContext> : IGarnetApi, IGarnetWatchApi
+    public partial struct GarnetApi<TContext, TObjectContext, TVectorContext> : IGarnetApi, IGarnetWatchApi
         where TContext : ITsavoriteContext<SpanByte, SpanByte, RawStringInput, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator>
         where TObjectContext : ITsavoriteContext<byte[], IGarnetObject, ObjectInput, GarnetObjectStoreOutput, long, ObjectSessionFunctions, ObjectStoreFunctions, ObjectStoreAllocator>
+        where TVectorContext : ITsavoriteContext<SpanByte, SpanByte, VectorInput, SpanByte, long, VectorSessionFunctions, MainStoreFunctions, MainStoreAllocator>
     {
         readonly StorageSession storageSession;
         TContext context;
         TObjectContext objectContext;
+        TVectorContext vectorContext;
```
tiagonapoli: Maybe I'm missing something - Looks like `vectorContext` declared here is not used anywhere?

libs/server/Storage/Session/MainStore/VectorStoreOps.cs
```diff
@@ -0,0 +1,279 @@
...
+                if (status != GarnetStatus.OK)
+                {
+                    dimensions = 0;
+                    return status;
+                }
+
+                // No need to recreate, all of this data is available to Garnet alone
+
+                // After a successful read we add the vector while holding a shared lock
+                // That lock prevents deletion, but everything else can proceed in parallel
```
tiagonapoli: I'm a bit confused, does this comment apply for VDIM? In VectorManager.ReadIndex we're only interpreting the index metadata raw bytes, right? No locks as well, correct?

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -0,0 +1,762 @@
...
+            //        indexPtr,
+            //        element,
+            //        MemoryMarshal.Cast<byte, float>(outputDistances.AsSpan())
+            //    );
+        }
+
+        /// <summary>
+        /// Helper to complete read/writes during vector set op replay that go async.
+        /// </summary>
+        private static void CompletePending(ref Status status, ref SpanByteAndMemory output, ref BasicContext<SpanByte, SpanByte, RawStringInput, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator> context)
```
tiagonapoli: Looks like this overload is not being used anywhere (from visual studio's 'Find All References')

libs/server/Storage/Session/MainStore/VectorStoreOps.cs
kevin-montrose: Copy/pasta error - it's not about adding, it's about reading metadata.

We still need to acquire the lock because that's how we (generally) validate the that we're working on a Vector Set.  We could optimize that away, but VDIM is marginal enough (and read locks fast enough) that it hasn't been done.

I will correct comment and make a note about some missing (ie. `WRONGTYPE`) behavior.

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
```diff
@@ -242,12 +247,12 @@ void CopyRespToWithInput(ref RawStringInput input, ref SpanByte value, ref SpanB
                     throw new GarnetException($"Not enough space in {input.header.cmd} buffer");
 
                 case RespCommand.TTL:
-                    var ttlValue = ConvertUtils.SecondsFromDiffUtcNowTicks(value.MetadataSize > 0 ? value.ExtraMetadata : -1);
+                    var ttlValue = ConvertUtils.SecondsFromDiffUtcNowTicks(value.MetadataSize == 8 ? value.ExtraMetadata : -1);
```
tiagonapoli, line 252: For my learning, why these changes to check `MetadataSize == 8`?

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,272 @@
...
+        internal VectorSessionFunctions(FunctionsState functionsState)
+        {
+            this.functionsState = functionsState;
+        }
+
+        #region Deletes
+        /// <inheritdoc />
+        public bool SingleDeleter(ref SpanByte key, ref SpanByte value, ref DeleteInfo deleteInfo, ref RecordInfo recordInfo)
+        {
+            recordInfo.ClearHasETag();
```
tiagonapoli, line 33: For my learning, in which situations VectorSessionFunctions.SingleDeleter would be called?

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
kevin-montrose: Before Vector Sets `MetadataSize > 0` indicated a TTL, after Vector Sets `MetadataSize == 1` indicates a namespace, `MetadataSize == 8` indicates a TLL.

This will go away with Store v2 which supports TTLs and Namespaces in a less hacky way.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: When a delete occurs and no other modification is happening to the targeted record.

The only distinction between the two `XDeleter` methods is that we only want to increment the watch map (which is used for Redis Transactions) if the concurrent deleter _actually deleted_ the record.  This can be false if two simultaneous deletes come in for the same record.  With a single deleter we always know the delete happened, so no need for the check.

I copied this behavior from `MainSessionFunctions` iirc.

libs/server/Auth/GarnetAadAuthenticator.cs
```diff
@@ -11,6 +11,8 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.IdentityModel.Validators;
+//using Microsoft.IdentityModel.Tokens;
+//using Microsoft.IdentityModel.Validators;
```
tiagonapoli: Should these be deleted?

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -48,7 +48,7 @@ public unsafe void VerifyKeyOwnership(ArgSlice key, LockType type)
             if (!clusterEnabled) return;
 
             var readOnly = type == LockType.Shared;
-            if (!respSession.clusterSession.NetworkIterativeSlotVerify(key, readOnly, respSession.SessionAsking))
+            if (!respSession.clusterSession.NetworkIterativeSlotVerify(key, readOnly, respSession.SessionAsking, isVectorSetWriteCommand: false)) // TODO: Is it ok to ignore Vector Set-y-ness of the key?
```
vazois: nit; if this is completed remove TODO comment

libs/server/Transaction/TxnRespCommands.cs
```diff
@@ -60,7 +60,7 @@ private bool NetworkEXEC()
                 endReadHead = txnManager.txnStartHead;
 
                 txnManager.GetKeysForValidation(recvBufferPtr, out var keys, out int keyCount, out bool readOnly);
-                if (NetworkKeyArraySlotVerify(keys, readOnly, keyCount))
+                if (NetworkKeyArraySlotVerify(keys, readOnly, isVectorSetWriteCommand: false, keyCount)) // TODO: We should actually verify if commands contained are Vector Set writes
```
vazois: nit; if this is completed remove TODO comment

libs/cluster/Server/Migration/MigrateOperation.cs
```diff
@@ -18,16 +20,28 @@ internal sealed partial class MigrateOperation
             public MainStoreScan mss;
             public ObjectStoreScan oss;
 
+            public readonly ConcurrentDictionary<byte[], byte[]> vectorSetsIndexKeysToMigrate;
+#if NET9_0_OR_GREATER
+            private readonly ConcurrentDictionary<byte[], byte[]>.AlternateLookup<ReadOnlySpan<byte>> vectorSetsIndexKeysToMigrateLookup;
```
vazois: I cannot find this being used somewhere or referenced other than at the constructor?!

libs/cluster/Server/Migration/MigrateOperation.cs
```diff
@@ -193,7 +280,13 @@ public void DeleteKeys()
                 {
                     foreach (var key in sketch.argSliceVector)
                     {
-                        var spanByte = key.SpanByte;
+                        if (key.MetadataSize == 1)
```
vazois, line 277: Not deleting the keys makes them visible for updates after DeleteKeys since the sketch in migrating slots is reset.
Will this create a problem with the namespace keys? Can they be updated or overwritten?

libs/cluster/Server/Migration/MigrateSessionSlots.cs
```diff
@@ -68,6 +119,60 @@ async Task<bool> CreateAndRunMigrateTasks(StoreType storeType, long beginAddress
...
+                                        {
+                                            logger?.LogCritical("Failed to migrate Vector Set key {key} during migration", keySpan);
+                                            return false;
+                                        }
+
+                                        gcs.SetClusterMigrateHeader(_sourceNodeId, _replaceOption, isMainStore: true, isVectorSets: true);
+                                    }
+
+                                    // Delete the index on this node now that it's moved over to the destination node
+                                    migrateOperation[0].DeleteVectorSet(ref keySpan);
```
vazois: before this point there is not gcs.SendAndReset call unless the buffer is full. Should we SendAndReset before deleting the data, in the event a crash happens, and the data in the network buffer is lost?

libs/cluster/Server/Migration/MigrateSessionSlots.cs
```diff
@@ -68,6 +119,60 @@ async Task<bool> CreateAndRunMigrateTasks(StoreType storeType, long beginAddress
                     _cts.Cancel();
                     return false;
                 }
+
+                // Handle migration of discovered Vector Set keys now that they're namespaces have been moved
+                if (storeType == StoreType.Main)
```
vazois: you might need to put this inside the try above so any exceptions are caught and handled appropriately

libs/cluster/Server/Migration/Sketch.cs
```diff
@@ -44,6 +44,19 @@ public bool TryHashAndStore(Span<byte> key)
             return true;
         }
 
+        public bool TryHashAndStore(ulong ns, Span<byte> key)
```
vazois, line 47: nit; add a description on this

libs/server/ArgSlice/ArgSliceVector.cs
```diff
@@ -29,7 +31,32 @@ public bool TryAddItem(Span<byte> item)
...
+
+            items.Add(argSlice.SpanByte);
+            return true;
+        }
+
+        /// <summary>
+        /// Try to add ArgSlice
+        /// </summary>
+        /// <param name="item"></param>
+        /// <returns>True if it succeeds to add ArgSlice, false if maxCount has been reached.</returns>
```
vazois, line 44: nit; add missing parameter description for ns

libs/cluster/Server/Migration/MigrateSession.cs
```diff
@@ -338,6 +343,8 @@ public bool TryRecoverFromFailure()
             // This will execute the equivalent of SETSLOTRANGE STABLE for the slots of the failed migration task
             ResetLocalSlot();
 
+            // TODO: Need to relinquish any migrating Vector Set contexts from target node
```
vazois, line 346: is this still a TODO?

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -43,6 +44,7 @@ bool CanServeSlot(RespCommand cmd)
             storeWrapper.clusterProvider.ExtractKeySpecs(commandInfo, cmd, ref parseState, ref csvi);
             csvi.readOnly = cmd.IsReadOnly();
             csvi.sessionAsking = SessionAsking;
+            csvi.isVectorSetWriteCommand = cmd is RespCommand.VADD or RespCommand.VREM or RespCommand.VSETATTR;
```
vazois: Looking at how this is used, it seems that we wait on the whole slot to be stable instead of a given key. Can we make this part of the command properties in the json file? Something like SlotStableCheck vs KeyStableCheck.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -1337,6 +1402,14 @@ public bool CopyUpdater(ref SpanByte key, ref RawStringInput input, ref SpanByte
                     CopyValueLengthToOutput(ref newValue, ref output, functionsState.etagState.etagSkippedStart);
                     break;
 
+                case RespCommand.VADD:
+                    Debug.Assert(input.arg1 is VectorManager.VADDAppendLogArg or VectorManager.MigrateElementKeyLogArg or VectorManager.MigrateIndexKeyLogArg, "Unexpected CopyUpdater call on VADD key");
+                    break;
```
tiagonapoli, line 1421: For my learning:

My understanding is that when we do RMW, one of these callbacks will be called, depending on the HybridLog state:
- InitialUpdater: Key doesn't exist yet, it will create a new log entry and appended to the HybridLog tail
- InPlaceUpdater: Key is in the mutable range of the hybrid log, so value can be read/updated in place
- CopyUpdater: Key exists but it's not in the mutable range, so a Read-Copy-Updated needs to be done on the value and a mutable entry will be added to the HybridLog tail.

Why we handle DeleteAfterDropArg/RecreateIndexArg in InPlaceUpdater but not in CopyUpdater? Is it guaranteed that when we issue RWM for these operations the key will be in the mutable range of the HybridLog?

libs/server/Resp/Vector/DiskANNService.cs
```diff
@@ -0,0 +1,319 @@
...
+            }
+            else
+            {
+                throw new NotImplementedException($"{vectorType}");
+            }
+
+            var attributes_data = Unsafe.AsPointer(ref MemoryMarshal.GetReference(attributes));
+            var attributes_len = attributes.Length;
+
+            return NativeDiskANNMethods.insert(context, index, (nint)id_data, (nuint)id_len, vectorType, (nint)vector_data, (nuint)vector_len, (nint)attributes_data, (nuint)attributes_len) == 1;
```
tiagonapoli, line 86: For my learning:

These variable length types such as id_data need to be pinned, right? Insert itself doesn't pin then, so the caller must ensure that these are pinned, correct? Should we document that?

libs/cluster/Server/Migration/MigrateOperation.cs
kevin-montrose: Ah yeah, dead code from earlier tracking of vector sets.  Superseded by the `ignoreLookup` stuff later - removing...

libs/cluster/Server/Migration/MigrateOperation.cs
kevin-montrose: Vector Set writes on the slot are blocked once we're migrating it, so no writes should be possible to the namespaced keys.

libs/cluster/Session/SlotVerification/ClusterSlotVerify.cs
```diff
@@ -69,12 +74,20 @@ ClusterSlotVerificationResult SingleKeyReadSlotVerify(ref ClusterConfig config,
...
-            ClusterSlotVerificationResult SingleKeyReadWriteSlotVerify(ref ClusterConfig config, ref ArgSlice keySlice)
+            ClusterSlotVerificationResult SingleKeyReadWriteSlotVerify(bool isVectorSetWriteCommand, ref ClusterConfig config, ref ArgSlice keySlice)
             {
                 var _slot = slot == -1 ? ArgSliceUtils.HashSlot(ref keySlice) : (ushort)slot;
+
+            tryAgain:
                 var IsLocal = config.IsLocal(_slot, readWriteSession: readWriteSession);
                 var state = config.GetState(_slot);
 
+                if (isVectorSetWriteCommand && state is SlotState.IMPORTING or SlotState.MIGRATING)
```
vazois: Is it possible to push this further down to avoid impacting the common path? Maybe at line 108 "return state switch...".
This depends of course on what, we need to prioritize for fast checks. I am assuming GET should be a first class citizen.

libs/cluster/Session/SlotVerification/RespClusterSlotVerify.cs
```diff
@@ -95,13 +95,13 @@ private void WriteClusterSlotVerificationMessage(ClusterConfig config, ClusterSl
         /// <param name="dend"></param>
         /// <param name="count"></param>
         /// <returns></returns>
```
vazois, line 99: nit; don't forget to update description when adding a new parameter

libs/cluster/Session/SlotVerification/RespClusterSlotVerify.cs
```diff
@@ -95,13 +95,13 @@ private void WriteClusterSlotVerificationMessage(ClusterConfig config, ClusterSl
...
-        public bool NetworkKeyArraySlotVerify(Span<ArgSlice> keys, bool readOnly, byte sessionAsking, ref byte* dcurr, ref byte* dend, int count = -1)
+        public bool NetworkKeyArraySlotVerify(Span<ArgSlice> keys, bool readOnly, byte sessionAsking, bool isVectorSetWriteCommand, ref byte* dcurr, ref byte* dend, int count = -1)
```
vazois: nit; might be better to name the new parameter waitForStableSlot (seems more descriptive)

libs/common/ReadOptimizedLock.cs
```diff
@@ -0,0 +1,409 @@
...
+
+            // Successfully acquired all shards exclusively
+
+            // Throwing away half the hash shouldn't affect correctness since we do the same thing when processing the full hash
+            lockToken = (int)hash;
+
+            return true;
+        }
+
+
```
vazois: nit; remove newline here

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+                    Debug.Assert(indexConfig.IsSpanByte, "Should never need to move index onto the heap");
+                }
+                catch
+                {
+                    vectorSetLocks.ReleaseSharedLock(sharedLockToken);
+
+                    throw;
+                }
+
+                var needsRecreate = readRes == GarnetStatus.OK && NeedsRecreate(indexConfig.AsReadOnlySpan());
```
vazois: NeedsRecreate may throw an exception. Make sure that the lock is released in that case.

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+                {
+                    if (!vectorSetLocks.TryPromoteSharedLock(keyHash, sharedLockToken, out var exclusiveLockToken))
+                    {
+                        // Release the SHARED lock if we can't promote and try again
+                        vectorSetLocks.ReleaseSharedLock(sharedLockToken);
+
+                        continue;
+                    }
+
+                    ReadIndex(indexSpan, out var indexContext, out var dims, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out _, out _);
```
vazois: Ensure this also does not create an issue with locking when exception is triggered.

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+                return new(in vectorSetLocks, sharedLockToken);
+            }
+        }
+
+        /// <summary>
+        /// Utility method that will read vector set index out, create one if it doesn't exist, or RECREATE one if needed.
+        /// 
+        /// Returns a disposable that prevents the index from being deleted while undisposed.
+        /// </summary>
+        internal ReadVectorLock ReadOrCreateVectorIndex(
```
vazois, line 242: The potential issue here is that the lock is not released in certain scenarios that could trigger an exception as in the previous method.

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+            var indexConfig = SpanByteAndMemory.FromPinnedSpan(indexSpan);
+
+            // Get the index
+            var acquiredLock = AcquireExclusiveLocks(storageSession, ref key);
+            try
+            {
+                status = storageSession.Read_MainStore(ref key, ref input, ref indexConfig, ref storageSession.basicContext);
+            }
+            catch
+            {
```
vazois, line 466: should we be setting the status explicitly here to something other than GarnetStatus.OK to avoid double dispose later on?

vazois: Also since this exception is caught, maybe add a log message here.

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+                    }
+
+                    ReadIndex(indexSpan, out var indexContext, out var dims, out var reduceDims, out var quantType, out var buildExplorationFactor, out var numLinks, out _, out _);
+
+                    input.arg1 = RecreateIndexArg;
+
+                    nint newlyAllocatedIndex;
+                    unsafe
+                    {
+                        newlyAllocatedIndex = Service.RecreateIndex(indexContext, dims, reduceDims, quantType, buildExplorationFactor, numLinks, ReadCallbackPtr, WriteCallbackPtr, DeleteCallbackPtr, ReadModifyWriteCallbackPtr);
```
vazois: Also here, is there a possibility of an uncaught exception?

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+            catch
+            {
+                acquiredLock.Dispose();
+
+                throw;
+            }
+
+            if (status != GarnetStatus.OK)
+            {
+                // This can happen is something else successfully deleted before we acquired the lock
```
vazois: nit; typo

libs/server/Resp/Vector/VectorManager.Locking.cs
```diff
@@ -0,0 +1,447 @@
...
+
+                acquiredLock.Dispose();
+                return default;
+            }
+
+            return acquiredLock;
```
vazois, line 481: I saw the caller is a using statement, will this result in a double dispose in the event that the acquiredLock is disposes inside the catch block?

libs/server/Storage/Functions/MainStore/RMWMethods.cs
kevin-montrose: I believe this is just an oversight actually - in practice all our tests are staying in the mutable portion unless we're explicitly testing recreation after a restore.

So this logic probably needs to be duplicated - checking...

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: That exception should actually be an assert - was converted when debugging some weirdness and never rolled back.  Will make `NeedsRecreate` infallible.

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: Same thing, should be infallible.

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: Exception is rethrown, so status and second dispose won't run and logging will happen higher up.  Unless I'm misunderstanding?

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: No - this is a thin wrapper around DiskANN, exceptions won't be raised.  They wouldn't work on Linux regardless, but they're also a contract violation so just crashing is correct.

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: We re-throw on L433, so the lock won't be both disposed and returned.

libs/server/Resp/Vector/VectorManager.Locking.cs
kevin-montrose: With the throw -> assert conversion in ReadIndex, I don't think there are any paths to leave a lock still held.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
```diff
@@ -0,0 +1,316 @@
...
+            if (status.IsPending)
+            {
+                CompletePending(ref status, ref outputSpan, ref ctx);
+            }
+
+            return status.IsCompletedSuccessfully ? (byte)1 : default;
+        }
+
+        [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
+        private static unsafe byte DeleteCallbackUnmanaged(ulong context, nint keyData, nuint keyLength)
```
tiagonapoli, line 245: nit: this is covered in the documentation, but should we have a short method description with arguments / return value description here as well?

libs/server/Resp/Vector/VectorManager.Callbacks.cs
```diff
@@ -0,0 +1,316 @@
...
+
+            var status = ctx.RMW(ref keyWithNamespace, ref input);
+            if (status.IsPending)
+            {
+                SpanByte ignored = default;
+
+                CompletePending(ref status, ref ignored, ref ctx);
+            }
+
+            return status.IsCompletedSuccessfully ? (byte)1 : default;
```
tiagonapoli: For my learning - what happens if operation doesn't complete successfully? Does diskann-garnet have some error handling / logging / retry?

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// <inheritdoc />
+        public bool ConcurrentWriter(ref SpanByte key, ref VectorInput input, ref SpanByte src, ref SpanByte dst, ref SpanByte output, ref UpsertInfo upsertInfo, ref RecordInfo recordInfo)
+        => SpanByteFunctions<VectorInput, SpanByte, long>.DoSafeCopy(ref src, ref dst, ref upsertInfo, ref recordInfo, 0);
+
+        #endregion
+
+        #region RMW
+        /// <inheritdoc />
+        public int GetRMWInitialValueLength(ref VectorInput input)
+        => sizeof(byte) + sizeof(int) + input.WriteDesiredSize;
```
tiagonapoli: I didn't follow the additional `sizeof(byte) + sizeof(int)` here - why is it added?

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+        /// <inheritdoc />
+        public void PostSingleDeleter(ref SpanByte key, ref DeleteInfo deleteInfo) { }
+        #endregion
+
+        #region Reads
+        /// <inheritdoc />
+        public bool SingleReader(ref SpanByte key, ref VectorInput input, ref SpanByte value, ref SpanByte dst, ref ReadInfo readInfo)
+        {
+            Debug.Assert(key.MetadataSize == 1, "Should never read a non-namespaced value with VectorSessionFunctions");
```
tiagonapoli, line 57: For my learning: Just confirming - This also applies to all other functions right? RMW, Write, etc, are always namespaced for VectorSessionFunctions?

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,272 @@
...
+            if (input.ReadDesiredSize > 0)
+            {
+                Debug.Assert(dst.Length >= value.Length, "Should always have space for vector point reads");
+
+                dst.Length = value.Length;
+                value.AsReadOnlySpan(functionsState.etagState.etagSkippedStart).CopyTo(dst.AsSpan());
+            }
+            else
+            {
+                input.ReadDesiredSize = value.Length;
```
tiagonapoli, line 79: For my understanding: in which scenario ReadDesiredSize is zero?

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,272 @@
...
+                if (dst.Length >= value.Length)
+                {
+                    value.AsReadOnlySpan(functionsState.etagState.etagSkippedStart).CopyTo(dst.AsSpan());
+                    dst.Length = value.Length;
+                }
```
tiagonapoli, line 84: Do we need to return any type of error/cancellation if destination lenght is not enough?

tiagonapoli: ah wait, this flow here (when input.Callback == 0) is only used for ContextMetadata reading?

libs/server/Resp/Vector/VectorManager.Locking.cs
vazois: oh yeah you are right on this. I am closing since it is non-issue. Maybe add a small description here.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            Span<byte> distinctKey = stackalloc byte[key.Length + 1];
+            var keyWithNamespace = SpanByte.FromPinnedSpan(distinctKey);
+            keyWithNamespace.MarkNamespace();
+            keyWithNamespace.SetNamespaceInPayload((byte)context);
+            key.CopyTo(keyWithNamespace.AsSpan());
+
+            ref var ctx = ref ActiveThreadSession.vectorContext;
+
+        tryAgain:
```
tiagonapoli: Just curious - why the choice for goto?

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -0,0 +1,743 @@
...
+                updateToDroppableVectorSet.header.cmd = RespCommand.VADD;
+
+                var update = storageSession.basicContext.RMW(ref key, ref updateToDroppableVectorSet);
+                if (!update.IsCompletedSuccessfully)
+                {
+                    throw new GarnetException("Failed to make Vector Set delete-able, this should never happen but will leave vector sets corrupted");
+                }
+
+                // Actually delete the value
+                var del = storageSession.basicContext.Delete(ref key);
```
tiagonapoli, line 451: For my learning: what happens if a crash or something happens in the middle of  `TryDeleteVectorSet` - That VectorSet will be corrupted? Or from the AOF, Garnet will be able to recover to a valid state?

For example, what happens if we delete a vector set, but fail to set ContextMetadata with the cleanUp flag because a restart happens first - is it possible that all DiskANN data will remain there and not cleaned up?

Maybe I'm missing something, but I was thinking if we should first RMW ContextMetadata with the cleanup flag to ensure that even upon restart we'll know that a particular context needs cleanup

kevin-montrose: That's a good catch - we will leave the Vector Set in a corrupted state if a crash happens between the RMW completing and the CleanupDroppedIndex completing.

The fix is a little trickier - just moving the ContextMetadata marking isn't enough.  While that would let us resume a cleanup of element data after crashing (since incomplete cleanups are enqueued at process start), we wouldn't be able find the index or replication keys.

The RMW is what starts the delete process, so resumption of delete on first touch probably the path forward.  Needs a good test, so gonna get that in place first before fixing.

libs/server/Storage/Session/MainStore/VectorStoreOps.cs
```diff
@@ -0,0 +1,276 @@
...
+                result = vectorManager.TryAdd(indexSpan, element.ReadOnlySpan, valueType, values.ReadOnlySpan, attributes.ReadOnlySpan, (uint)reduceDims, quantizer, (uint)buildExplorationFactor, (uint)numLinks, out errorMsg);
+
+                if (result == VectorManagerResult.OK)
+                {
+                    // On successful addition, we need to manually replicate the write
+                    vectorManager.ReplicateVectorSetAdd(ref key, ref input, ref basicContext);
+                }
```
tiagonapoli: Just confirming my understanding: On an extreme case where a crash happens between TryAdd and ReplicateVectroSetAdd it will still be ok because TryAdd won't persist anything, being a READ, right? So nothing was really commited and there won't be inconsistencies?

kevin-montrose: In Redis cluster (and thus Garnet), in general, writes are not durable - so while there _is_ a gap here where a write might happen on a primary and never sync to a replica, that gap already exists.

The equivalent gap for non-Vector Set ops is if we crash between a write op and the PostXXX method which writes to the AOF and then the persistence of the AOF.

In both cases _a_ fix is to resync a replica after the AOF is empty - syncing the Tsavorite log will restore any missed writes.  That would recover the half-executed write.

However in practice a primary crashing will cause the replica to get promoted, so the write will just be lost most of the time.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
kevin-montrose: I was probably futzing about with an interlocked version at some point and kept the goto.  It is a smell here, lemme replace that with a while(true).

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: Never now.  Might have been possible before reads from DiskANN were batched.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: Correct, the assert should be duplicated onto appropriate paths - will fix.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: It's `namespace` + `ttl` + `actual data`.

The TTL bits are unused currently but were kept because a) that better with how MainSessionFunctions so some helpers assume it's here b) we thought maybe TTL was going to be necessary.

In store v2 this should go away or least be "the same" as everything else since TTL and namespaces are first-class there.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
kevin-montrose: In theory DiskANN will fail the outer operation, and hopefully do some cleanup.  We'd hand back the failure response to whatever RESP command is running.

That's untested at the moment, not sure of the state of the diskann-garnet code w.r.t. to failures.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
kevin-montrose: They've been changing often enough that I'd rather not duplicate for now, lest they be more prone to getting out of sync.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: Actually, thinking about this longer... I dunno that the `byte` makes sense at all - that may just be left over from earlier exploration and I conflated it with the namespace byte added later.  I'll poke at it.

libs/server/Resp/Vector/VectorManager.Replication.cs
```diff
@@ -0,0 +1,541 @@
...
+            }
+
+            if (!res.IsCompletedSuccessfully)
+            {
+                logger?.LogCritical("Failed to inject replication write for VADD into log, result was {res}", res);
+                throw new GarnetException("Couldn't synthesize Vector Set add operation for replication, data loss will occur");
+            }
+
+            // Helper to complete read/writes during vector set synthetic op goes async
+            static void CompletePending(ref Status status, ref SpanByteAndMemory output, ref TContext context)
```
tiagonapoli: nit: looks like this function repeats 3 times on this file, should we have a common method?

libs/server/Resp/Vector/VectorManager.Replication.cs
```diff
@@ -0,0 +1,541 @@
...
+        internal void DropVectorSetReplicationKey<TContext>(SpanByte key, ref TContext context)
+            where TContext : ITsavoriteContext<SpanByte, SpanByte, RawStringInput, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator>
+        {
+            Span<byte> keyWithNamespaceBytes = stackalloc byte[key.Length + 1];
+            var keyWithNamespace = SpanByte.FromPinnedSpan(keyWithNamespaceBytes);
+            keyWithNamespace.MarkNamespace();
+            keyWithNamespace.SetNamespaceInPayload(0);
+            key.AsReadOnlySpan().CopyTo(keyWithNamespace.AsSpan());
+
+            Span<byte> dummyBytes = stackalloc byte[4];
```
tiagonapoli: nit: looks like it's unused

libs/server/Resp/Vector/VectorManager.Replication.cs
kevin-montrose: Consequence of these all being in different files and then consolidated into VectorManager over time.  Will DRY up.

libs/server/AOF/AofProcessor.cs
```diff
@@ -70,10 +74,11 @@ public AofProcessor(
         {
             this.storeWrapper = storeWrapper;
 
-            var replayAofStoreWrapper = new StoreWrapper(storeWrapper, recordToAof);
+            replayAofStoreWrapper = new StoreWrapper(storeWrapper, recordToAof);
```
vazois: maybe leave this as a local to prevent from people using it for any other purpose.

libs/server/Resp/Vector/VectorManager.Replication.cs
```diff
@@ -0,0 +1,541 @@
...
+
+            static void StartReplicationReplayTasks(VectorManager self, Func<RespServerSession> obtainServerSession)
+            {
+                self.logger?.LogInformation("Starting {numTasks} replication tasks for VADDs", self.replicationReplayTasks.Length);
+
+                for (var i = 0; i < self.replicationReplayTasks.Length; i++)
+                {
+                    // Allocate session outside of task so we fail "nicely" if something goes wrong with acquiring them
+                    var allocatedSession = obtainServerSession();
+                    if (allocatedSession.activeDbId != self.dbId && !allocatedSession.TrySwitchActiveDatabaseSession(self.dbId))
```
vazois, line 338: currently cluster supports a single database, we should validate that the VectorManager.dbId = 0

libs/server/Resp/Vector/VectorManager.Replication.cs
```diff
@@ -0,0 +1,541 @@
...
+                    {
+                        allocatedSession.Dispose();
+                        throw new GarnetException($"Could not switch replication replay session to {self.dbId}, replication will fail");
+                    }
+
+                    self.replicationReplayTasks[i] = Task.Factory.StartNew(
```
vazois, line 344: I might have missed it, but we need a way to cancel these tasks in the event the replication stream breaks (i.e. failover).

libs/server/Resp/Vector/VectorManager.Replication.cs
```diff
@@ -0,0 +1,541 @@
...
+                if (Interlocked.CompareExchange(ref replicationReplayStarted, 1, 0) == 0)
+                {
+                    StartReplicationReplayTasks(this, obtainServerSession);
+                }
+            }
+
+            // We need a running count of pending VADDs so WaitForVectorOperationsToComplete can work
+            _ = Interlocked.Increment(ref replicationReplayPendingVAdds);
+            replicationBlockEvent.Reset();
+            var queued = replicationReplayChannel.Writer.TryWrite(new(keyBytes, dims, reduceDims, valueType, valuesBytes, elementBytes, quantizer, buildExplorationFactor, attributesBytes, numLinks));
```
vazois: How does this interact with the checkpoint replay? If we are replaying a VADD and the next replay op is a checkpoint, do, we wait for the background VADDs to complete? If not, I think we should wait for all background replays to complete before taking a checkpoint at replay.

kevin-montrose: We wait for VADDs to complete in ProcessAofRecordInternal if we're not applying a StoreRMW entry, and we wait for VADDs to complete if the StoreRMW entry is for something besides a VADD.

libs/server/AOF/AofProcessor.cs
kevin-montrose: I'm not sure I follow - we need this around to generate new `RespServerSession`s in `ObtainServerSession()`.

libs/server/Resp/Vector/VectorManager.Replication.cs
kevin-montrose: What's the signal for that?  Looking at `TryFailover` and nothing is jumping out - seems like AofProcessor (and its associated RespServerSession) is kept live as long as the `ClusterProvider` is.

Since the AOF stream would be interrupted during failover, I'm not sure aborting the task is actually necessary - they'll just stay pending until/unless a new AOF stream is received.

libs/server/AOF/AofProcessor.cs
vazois: maybe use a delegate or a class wrapper that contains the replayStoreWrapper. As long as we can prevent people using this instance of the storeWrapper directly within the AofProcessor.

```suggestion
            var replayAofStoreWrapper = new StoreWrapper(storeWrapper, recordToAof);
            this.obtainServerSessionDelegate = () => ObtainServerSession(replayAofStoreWrapper);
```

libs/server/Resp/Vector/VectorManager.Replication.cs
vazois: There is a cancellationToken in ReplicaReplayTask.cs. This was not passed down to ProcessAofRecordInternal because we didn't have any background tasks in the AofProcessor.cs (as far as I know).
So, you may use that.

libs/server/Resp/Vector/VectorManager.Replication.cs
vazois: That is great! Thanks!

libs/common/CountingEventSlim.cs
```diff
@@ -0,0 +1,92 @@
...
+                if (addRes <= 0)
+                {
+                    // Some thread is about to Set, undo and wait
+
+                    _ = Interlocked.Decrement(ref count);
+                    _ = Thread.Yield();
+                }
+                else
+                {
+                    resetEvent.Reset();
```
tiagonapoli, line 48: I'm wondering if we should protect resetEvent.Reset() with some kind of lock as well, if we're creating a more generic CountingEventSlim.

Maybe it's a scenario too extreme, but what if (let me know if it makes sense or maybe I'm missing something):

Initial conditions:
- count = 1
- resetEvent is signaled, so it's blocking operations from running
- a background thread is actively decrementing count

Serialized execution:
1. [Thread1]: Calls Increment, executes line 38 and increments count to 2, addRes = 2, then it suspends
2. [BackgroundThread] Decrements 2 times, getting count to zero, then does the CompareExchange to set the value and calls `resetEvent.Set()`
3. [Thread1] Goes out of suspension, it still sees addRes = 2, so it will enter the `else` block and will call `resetEvent.Reset()`. Now even though count == 0, `resetEvent` will block execution

This problem doesn't happen with the AOF VADD processing logic because we complete `CountingEventSlim.Increment` before we add work for the BackgroundThreads, so they can't decrement to 0 in the middle of `CountingEventSlim.Increment`. But if CountingEventSlim is used in the future with something else, maybe a 'deadlock' could happen, right?

libs/server/Transaction/TxnKeyManager.cs
```diff
@@ -50,7 +50,7 @@ public unsafe void VerifyKeyOwnership(ArgSlice key, LockType type)
             if (!clusterEnabled) return;
 
             var readOnly = type == LockType.Shared;
-            if (!respSession.clusterSession.NetworkIterativeSlotVerify(key, readOnly, respSession.SessionAsking))
+            if (!respSession.clusterSession.NetworkIterativeSlotVerify(key, readOnly, respSession.SessionAsking, waitForStableSlot: false)) // TODO: Is it ok to ignore Vector Set-y-ness of the key?
```
tiagonapoli: nit: Is this TODO pending?

libs/server/Storage/Functions/MainStore/ReadMethods.cs
```diff
@@ -17,14 +17,44 @@ public bool SingleReader(
...
+                }
+            }
+
+            // GET is used in a number of non-RESP contexts, which messes up existing logic
+            //
+            // Easiest to mark the actually-RESP commands with a < 0 arg1 and roll back to old logic
+            // after the Vector Set checks
+            //
+            // TODO: This is quite hacky, but requires a bunch of non-Vector Set changes - do those and remove
+            if (input.arg1 < 0 && cmd == RespCommand.GET)
```
tiagonapoli, line 53: For my learning: I didn't understand this workaround - what's the issue that it's solving?

libs/common/CountingEventSlim.cs
kevin-montrose: If I follow your scenario the Increment from 1 -> 2 hasn't yet completed, but Decrement is called twice.

This would be misusing the type, since Decrement must be paired with previous calls to Increment.

To illustrate, your scenario implies:

Earlier on any thread:
 - Increment()  // count == 1 when complete
 
Thread1:
  1. Increment() // count is 2 (if before `Thread2:1`), 1 (if after `Thread2:1`), or 0 (if after `Thread2:2`)
  
Thread2:
  1. Decrement() // count is 0 (if before `Thread1:1`), or 1 (if after `Thread1:1`)
  2. Decrement() // count is -1 (if before `Thread1:1`) or 0 (if after `Thread1:1`)
   
That -1 on the last line is illegal.

Sure, a suspension in `Thread1:1` would cause problems - but so would a suspension _before_ `Thread1:1`.  It's just a misuse unless there's some other synchronization forcing `Thread2:2` to follow `Thread1:1` - which would prevent the race.

tiagonapoli: Ahhh, yes, makes sense - I see now that the scenario I described would only make sense if `CountingEventSlim` were used alone, without some other data structure like ReaderWriter, and if `CountingEventSlim` provided some kind of `TryDecrement` or `CurrentCount`. Without that, the scenario I described would be misuse. Thanks for the clarification

libs/server/Storage/Functions/MainStore/ReadMethods.cs
kevin-montrose: The issue is that a number of parts of Garnet use Read variants to get data to parse or modify.  Oftentimes without specifying a `RespCommand` as part of `RawStringInput`.

That causes a lot of problems that were revealed when I fixed some `WRONGTYPE` issues - so what I did was make sure "real" GETs now always have the `RespCommand` specified so we can do `WRONGTYPE` checks... but that breaks other code that assumes (implicitly, often) `RespCommand.None`.  So I hack around and set arg1 < 0 when doing these old-NONE now-GET reads, and then reset to NONE here.

Huge hack - but the alternative is to make a bunch of changes to non-Vector Set commands.  This PR is already a monster, soooo...  hack it  is.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
```diff
@@ -0,0 +1,404 @@
...
+            }
+
+            return sizeof(byte) + sizeof(int) + effectiveWriteDesiredSize;
+        }
+        /// <inheritdoc />
+        public int GetRMWModifiedValueLength(ref SpanByte value, ref VectorInput input)
+        {
+            if (input.WriteDesiredSize < 0)
+            {
+                // Add to value, this is a dynamically sized type
```
tiagonapoli, line 226: Just for my understanding - I don't fully understand the Tsavorite value sizing yet

Could we possibly have an unbounded size for the value stored on this case? For each time we add or delete to the list of deleting vector sets, GetRMWModifiedValueLength is saying that the new list size will be `value.Length + abs(input.WriteDesiredSize)` - strictly increasing - so if Tsavorite actually allocates the new size and that's unused, but Tsavorite doesn't know it, the next time we call it, we will increase again and again?

tiagonapoli: ahhh, maybe we forgot to use RecordInfo?
> > /// <param name="recordInfo">A reference to the RecordInfo for the record; used for variable-length record length modification</param>

libs/server/Resp/Vector/VectorManager.Cleanup.cs
```diff
@@ -0,0 +1,386 @@
...
+            }
+
+            if (isAdding)
+            {
+                // Not already added, so slap it in
+                BinaryPrimitives.WriteUInt64LittleEndian(remaining, context);
+                BinaryPrimitives.WriteInt32LittleEndian(remaining[sizeof(ulong)..], len);
+
+                key.CopyTo(remaining[(sizeof(ulong) + sizeof(int))..]);
+            }
```
tiagonapoli, line 232: I saw in some other places where we have variable length data these functions calls:
```cs
                    rmwInfo.ClearExtraValueLength(ref recordInfo, ref value, value.TotalSize);
                    value.UnmarkExtraMetadata();
                    value.ShrinkSerializedLength(metadataSize + inputValue.Length + EtagConstants.EtagSize);
                    rmwInfo.SetUsedValueLength(ref recordInfo, ref value, value.TotalSize);
```
Do we need to manage the used length here as well? I added another comment asking about possibly unbounded lenght increase, maybe these would solve this issue? (if it is an issue)

tiagonapoli: Does DiskANN have variable sized data values as well? I see it has RMW operation as a callback:

Image

I'm wondering if we need to do this shrinking / set used value as well?

website/docs/dev/vector-sets.md
```diff
@@ -0,0 +1,421 @@
...
+To fix that, synthetic writes against related keys are made after an insert or remove.  These writes are against the same Vector Set key, but in namespace 0.  See `VectorManager.ReplicateVectorSetAdd` and `VectorManager.ReplicateVectorSetRemove` for details.
+
+> [!IMPORTANT]
+> There is a failure case here where we crash between the insert operation completing and the replication operation completing.
+>
+> This appears to simply extend a window that already existed between when a Tsavorite operation completed and an entry was written to the AOF.
+> This needs to confirmed - if it is not the case, handling this failure needs to be figured out.
+
+> [!IMPORTANT]
+> This code assumes a Vector Set under the empty string is illegal.  That needs to be tested against Redis, and if it's not true we need to use
```
tiagonapoli: I tried out these commands in local redis and it worked, really weird to have empty string as valid vector set name :(
Maybe a TODO for v2?
```
> VADD "" REDUCE 5 VALUES 10 1.1 2.2 3.3 4.4 5.5 6.6 7.7 8.8 9.9 10.10 element1 M 32
(integer) 1
> VINFO ""
1) "quant-type"
2) "int8"
3) "hnsw-m"
4) "32"
5) "vector-dim"
6) "5"
7) "projection-input-dim"
8) "10"
9) "size"
10) "1"
```

libs/server/Resp/Vector/VectorManager.Cleanup.cs
kevin-montrose: DiskANN is fixed size everywhere except attributes, but those are always upserts.

That said, unbounded growth is a concern for the pending deletes.  Lemme poke at that a bit.

libs/server/Storage/Functions/MainStore/VectorSessionFunctions.cs
kevin-montrose: Yeah, there was a case here where the delete tracking key could grow unbounded.  A reboot would fix that, but long running processes with lots of Vector Set deletes would break future deletes eventually.

This has been fixed.

PR #1370, opened 2025-08-28
Some fixes to database checkpointing

### LINKED ISSUE #1342: BGSAVE Already in Progress 100% CPU
### Describe the bug

In Redis, if there is a BGSAVE already in progress, it will raise an exception that includes the phrase "already in progress".

In Garnet, if there is a BGSAVE already in progress, it appears to try to save anyway. This works until there are more threads than CPUs available. Once there are more BGSAVEs in progress than CPUs available, Garnet locks up at 100% CPU and never recovers.

I think it would be fine to just raise an exception if there is a BGSAVE already in progress. But something needs to check to verify that it won't go into 100% CPU and lock up. Never mind that my application shouldn't be trying to do that many BGSAVEs at the same time. I can resolve that. But this is a potential denial of service attack that needs to be addressed.

In the mean time, if anyone else runs into this and wants a workaround, I'm checking the INFO["store"]["SystemState"]["state"] and only proceeding with a BGSAVE call if the state is "REST".

I've tried creating a simple example to demonstrate the issue, but it doesn't fail on small amounts of data. The application I'm running is loading in a series of up to 1,000 HKEYs and supporting lists and sets with about 100K of data per source file. Once I process 8 to 10 files concurrently, each one issuing a BGSAVE when it loads or is removed, then I see Garnet lock up at 100% CPU and no recovery.

### Expected behavior

Expected an "already in progress" exception to be raised on the second concurrent BGSAVE call. Alternatively, expected Garnet to not lock up, no matter how many BGSAVE calls are Phase.IN_PROGRESS.

### Screenshots

_No response_

### Release version

v1.0.79

### IDE

VS Code latest

### OS version

Problem occurs when using GarnetServer on both MacOS and Windows 10 IoT Enterprise.

### Additional context

_No response_

TalZaccai:
Couple of fixes to database checkpointing:
- `TakeCheckpoint` / `TakeDatabasesCheckpointAsync` methods were waiting for in-progress checkpoint to complete - fixed to break early if there's already a checkpoint in progress
- `MultiDatabaseManager.TakeCheckpoint` was ignoring background parameter - fixed
- Access to `dbIdsToCheckpoint` was not thread safe - added `multiDbCheckpointingLock` for synchronization

- Added tests to verify early termination when in-progress checkpointing is taking place, in both single and multi-db contexts.

Should fix #1342


libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -152,20 +155,32 @@ public override bool TakeCheckpoint(bool background, ILogger logger = null, Canc
...
+            if (activeDbIdsMapSize > 1)
             {
-                databasesContentLock.ReadUnlock();
+                if (!multiDbCheckpointingLock.TryWriteLock())
+                    return false;
             }
 
-            return true;
+            var activeDbIdsMapSnapshot = activeDbIds.Map;
+            Array.Copy(activeDbIdsMapSnapshot, dbIdsToCheckpoint, activeDbIdsMapSize);
```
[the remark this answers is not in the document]
TalZaccai: This should be protected by using `TryGetDatabasesContentReadLock` beforehand

libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -152,20 +155,40 @@ public override bool TakeCheckpoint(bool background, ILogger logger = null, Canc
             var lockAcquired = TryGetDatabasesContentReadLock(token);
             if (!lockAcquired) return false;
 
-            try
-            {
-                var activeDbIdsMapSize = activeDbIds.ActualSize;
-                var activeDbIdsMapSnapshot = activeDbIds.Map;
-                Array.Copy(activeDbIdsMapSnapshot, dbIdsToCheckpoint, activeDbIdsMapSize);
+            var checkpointLockTaken = false;
```
TedHartMS: Why the inconsistency of introducing this variable here instead of re-checking activeDbidsMapSize as is done in TaskCheckpointBasedOnAofSizeLimitAsync?

libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -175,7 +198,8 @@ public override bool TakeCheckpoint(bool background, int dbId, ILogger logger =
             var databasesMapSnapshot = databases.Map;
             Debug.Assert(dbId < databasesMapSize && databasesMapSnapshot[dbId] != null);
 
-            if (!TryPauseCheckpointsContinuousAsync(dbId, token).GetAwaiter().GetResult())
+            // Check if checkpoint already in progress
```
TedHartMS, line 201: Nit--the comment on this call on line 239 is wrong (but I can't enter a comment on that line)

libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -967,31 +1001,25 @@ private async Task TakeDatabasesCheckpointAsync(int dbIdsCount, ILogger logger =
                 var currIdx = 0;
                 while (currIdx < dbIdsCount)
                 {
-                    var dbId = dbIdsToCheckpoint[currIdx];
+                    var idx = currIdx++;
```
TedHartMS: It looks like this could just be a for loop on currIdx with no need for idx--it's not captured

libs/server/Databases/MultiDatabaseManager.cs
TalZaccai: Good catch, I needed to fix it there too. Since if the lock wasn't taken successfully we don't want to release it.

TedHartMS (approved):

PR #1372, opened 2025-08-31
[Bugfix] AOF Replay Double Replaying Finalize 

hamdaankhalid:
AOF currently double replays the finalize phase. This PR fixes the issue, by handling AOF commits from Finalize as separate portions of replay instead of grouping finalize calls with Transaction at replay.

Finalize needs to log to AOF because while the server is alive, the finalize blocks AOF commits show the ordering that occurred. During replay we respect that exact ordering by calling the sequence of finalize.

### DIFF
--- a/website/docs/dev/onboarding.md
+++ b/website/docs/dev/onboarding.md
@@ -104,6 +104,8 @@ RESP representation:
     **Tsavorite** and **Garnet** rely heavily on these two types for allocating data in memory and then transfer it on the network layer. Understanding and familiarity with both of them will be very helpful for a better understanding of the code in general.
 
     * [Documentation about Span](https://learn.microsoft.com/en-us/dotnet/api/system.span-1?view=net-7.0)
+ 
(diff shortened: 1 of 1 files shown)


libs/server/Transaction/TransactionManager.cs
```diff
@@ -227,8 +227,14 @@ internal bool RunTransactionProc(byte id, ref CustomProcedureInput procInput, Cu
             {
                 try
                 {
-                    // Run finalize procedure at the end
-                    proc.Finalize(garnetTxFinalizeApi, ref procInput, ref output);
+                    // Run finalize procedure at the end.
+                    // If the transaction was invoked during AOF replay skip the finalize step altogether
+                    // Finalize logs to AOF accordingly, so let the replay pick up the commits from AOF as
+                    // part of normal AOF replay.
```
badrishc, line 233: Put a comment in the transaction proc base class/interface that Finalize is considered post transaction processing and therefore is not executed at recovery time. Instead, the individual Tsavorite commands are logged and replayed through the AOF.

PR #1403, opened 2025-10-06
Enable codeql for cpp

Mathos1432:
There are a few tsavorite files in the repo that aren't getting scanned by CodeQL. This enabled CodeQl on the repo for c-cpp, with build mode none.


darrenge (approved):
This makes sense to do.

PR #1412, opened 2025-10-16
fix txn replay for fuzzy region

vazois:
This PR fixes couple of issues:

1. Our existing transaction replay mechanism keeps track of the operations of a given transaction in a dictionary indexed by sessionID. When a replica encounters a checkpoint marker, it will start buffering the transaction commit markers so it can replay them after encountering the CheckpointEnd marker.
In that situation, our current solution might intermingle operations between two different transactions from the same session.
2. When replaying the individual operations of a transaction at the replica, there is no locking. In that case, the results of the operations become immediately visible which not desirable given that a reader might see partial results.

In this PR, I am adding the following classes:

- TransactionGroup
       Keeps track of the list of transaction operations
- AofReplayContext
       Keeps track of the replay context and associated objects/buffers used to replay individual operations and transactions.       
- AofReplayCoordinator
       Implements the mechanism to keep track of active transactions, gather transaction operations and coordinate transaction replay.

### DIFF
--- a/Version.props
+++ b/Version.props
@@ -1,6 +1,6 @@
 <Project>
 	<!-- VersionPrefix property for builds and packages -->
 	<PropertyGroup>
-		<VersionPrefix>1.0.89</VersionPrefix>
+		<VersionPrefix>1.0.90</VersionPrefix>
 	</PropertyGroup>
 </Project>


libs/server/AOF/AofEntryType.cs
```diff
@@ -77,18 +77,7 @@ public enum AofEntryType : byte
...
-        /// <summary>
-        /// Deprecated with unified checkpointing: Checkpoint for object store start
-        /// </summary>
-        ObjectStoreCheckpointStartCommit = 0x31,
-        /// <summary>
-        /// Deprecated with unified checkpointing: Checkpoint for object store end
-        /// </summary>
-        ObjectStoreCheckpointEndCommit = 0x33,
-        #endregion
+        FlushDb = 0x61
```
kevin-montrose: nit: keep a trailing comma for better diffing.

libs/server/AOF/AofProcessor.cs
```diff
@@ -113,268 +100,221 @@ public void Dispose()
...
         {
-            logger?.LogInformation("Begin AOF recovery for DB ID: {id}", db.Id);
-            return RecoverReplay(db, untilAddress);
-        }
-
-        private long RecoverReplay(GarnetDatabase db, long untilAddress)
-        {
-            // Begin replay for specified database
-            logger?.LogInformation("Begin AOF replay for DB ID: {id}", db.Id);
+            Stopwatch swatch = new();
```
kevin-montrose: Prefer `Stopwatch.GetTimestamp()` and `Stopwatch.GetElapsedTime()` to allocating a `Stopwatch` instance.

libs/server/AOF/AofProcessor.cs
```diff
@@ -113,268 +100,221 @@ public void Dispose()
...
-                int count = 0;
+                logger?.LogInformation("Begin AOF recovery for DB ID: {id}", db.Id);
+                return RecoverReplay(db, untilAddress);
+            }
+            finally
+            {
+                var seconds = swatch.ElapsedMilliseconds / 1000.0;
+                var aofSize = db.AppendOnlyFile.TailAddress - db.AppendOnlyFile.BeginAddress;
+                var recordsPerSec = total_number_of_replayed_records / seconds;
+                var GiBperSecs = (aofSize / seconds) / (double)1_000_000_000;
```
kevin-montrose: nit: casing - local variables shouldn't start with a capital.

libs/server/AOF/AofProcessor.cs
```diff
@@ -398,160 +338,180 @@ private void SwitchActiveDatabaseContext(GarnetDatabase db, bool initialSetup =
             }
         }
 
-        static void StoreUpsert(BasicContext<SpanByte, SpanByte, RawStringInput, SpanByteAndMemory, long, MainSessionFunctions, MainStoreFunctions, MainStoreAllocator> basicContext,
-            RawStringInput storeInput, byte* ptr)
+        static int HeaderSize() => sizeof(AofHeader);
```
kevin-montrose: Why a method here?  Could just put `sizeof(AofHeader)` everywhere - or to be terser a `static readonly int` field (which should end up treated as a constant by the JIT).

libs/server/AOF/ReplayCoordinator/AofReplayContext.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using Garnet.common;
+using Garnet.networking;
+
+namespace Garnet.server
+{
+    /// <summary>
+    /// AofReplayContext
+    /// </summary>
+    public class AofReplayContext
```
kevin-montrose: nit: prefer to `sealed` all classes where inheritance is not expected.

libs/server/AOF/AofProcessor.cs
vazois: Yeah, it is not needed here. I cherry picked it from my other branch where I introduced AofExtendedHeader and will have to pick between those two during runtime. I can remove for this PR.

libs/server/AOF/ReplayCoordinator/AofReplayCoordinator.cs
```diff
@@ -0,0 +1,283 @@
...
+                    // Start by saving transaction keys for locking
+                    SaveTransactionGroupKeysToLock(txnManager, txnGroup);
+
+                    // Start transaction
+                    _ = txnManager.Run(internal_txn: true);
+
+                    // Process in parallel transaction group
+                    ProcessTransactionGroupOperations(aofProcessor, txnManager.LockableContext, txnManager.ObjectStoreLockableContext, txnGroup, asReplica);
+
+                    // Commit (NOTE: need to ensure that we do not write to log here)
```
badrishc: Can we add a debug assert or similar here to make sure we do not break this invariant of not writing to log here.

badrishc (approved):

libs/server/AOF/ReplayCoordinator/AofReplayCoordinator.cs
vazois: good point, added!

badrishc (approved):

badrishc (approved):

PR #1413, opened 2025-10-16
Fix: snapshot checkpoint flush corner case

badrishc:
Consider the case where there is nothing to flush to the snapshot file because the flushed region already contains everything up to finalLogicalAddress.


TedHartMS (approved):
Looks good; it's consistent with the behavior for other cases

PR #1414, opened 2025-10-17
Prevent Replica from taking a checkpoint at TaskCheckpointBasedOnAofSizeLimitAsync

vazois:


libs/server/Databases/SingleDatabaseManager.cs
```diff
@@ -215,11 +219,18 @@ public override async Task TaskCheckpointBasedOnAofSizeLimitAsync(long aofSizeLi
...
-
             try
             {
+                // Checkpoint will be triggered from AOF replay
+                if (storeWrapper.clusterProvider.IsReplica())
+                {
+                    logger?.LogInformation("Replica skipping {method}", nameof(TaskCheckpointBasedOnAofSizeLimitAsync));
+                    return;
+                }
+
```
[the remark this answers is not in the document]
vazois: PauseCheckpoint implicitly blocks BeginRecovery which prevents replica changing role.
Hence it is necessary to execute in that order to ensure that the role of the node persists until the end of the checkpoint.

libs/host/Configuration/Options.cs
```diff
@@ -858,6 +862,7 @@ public GarnetServerOptions GetServerOptions(ILogger logger = null)
                 CommitFrequencyMs = CommitFrequencyMs,
                 WaitForCommit = WaitForCommit.GetValueOrDefault(),
                 AofSizeLimit = AofSizeLimit,
+                AofSizeLimitEnforceFrequency = AofSizeLimitEnforceFrequency,
```
badrishc: These should be `AofSizeLimitEnforceFrequencyMs` here and other related places, similar to `AofReplicationRefreshFrequencyMs`.

libs/host/defaults.conf
```diff
@@ -171,6 +171,9 @@
 	/* Maximum size of AOF (rounds down to power of 2) after which unsafe truncation will be applied. Left empty AOF will grow without bound unless a checkpoint is taken */
 	"AofSizeLimit" : "",
 
+	/* Frequency (in ms) of execution of the AutoCheckpointBasedOnAofSizeLimit background task. */
```
badrishc: Does this need to be milliseconds or ok to be seconds?

vazois: Not necessary. I was thinking that the AOF might grow very rapidly, hence a high frequency of truncation is needed. Though I guess at the granularity of 1 second it should be fine. I will switch this to seconds

libs/host/Configuration/Options.cs
vazois: I will switch this to seconds

badrishc (approved):

PR #1416, opened 2025-10-23
Leverage NUnit Assert.Throws

johnthcall:
Instead of the pattern of try catch with and asserting on the caught exception leverage Assert.Throws from NUnit. With this approach if an Assert.Fail at the end of the try block is missing a regression won't silently pass. Since this is a style change feel free to close if the team likes the existing pattern.

This PR only changes Sync flows as NUnit does have an Assert.ThrowsAsync but it's return is not a Task and so it performs Sync over Async under the hood. Happy to make that change as well if it is desired.

### DIFF
--- a/test/Garnet.test/GarnetBitmapTests.cs
+++ b/test/Garnet.test/GarnetBitmapTests.cs
@@ -107,7 +107,7 @@ namespace Garnet.test
             ClassicAssert.IsFalse(db.StringGetBit(key, 8999));
             ClassicAssert.IsTrue(db.StringGetBit(key, 9999));
 
-            var ex = Assert.Throws<RedisServerException>(() => db.Execute("SETBIT", key, "-1", "1"),);
+            var ex = Assert.Throws<RedisServerException>(() => db.Execute("SETBIT", key, "-1", "1"));
             ClassicAssert.AreEqual("ERR bit offset is not an integer or out of range",
                                    ex.Message);
         }


test/Garnet.test.cluster/ClusterConfigTests.cs
```diff
@@ -79,21 +79,16 @@ public void ClusterForgetAfterNodeRestartTest()
...
-                Assert.Fail("Cluster forget call shouldn't have succeeded for an invalid node id.");
-            }
-            catch (Exception ex)
-            {
-                Assert.That(ex.Message == "ERR I don't know about node 1ip23j89123no.");
-            }
+            var ex = Assert.Throws<RedisServerException>(() => server.Execute("cluster", args),
+                "Cluster forget call shouldn't have succeeded for an invalid node id.");
+
+            Assert.That(ex.Message == "ERR I don't know about node 1ip23j89123no.");
```
TalZaccai: For a better debugging experience, please use `Assert.That(ex.Message, Is.EqualTo(<message>));`. Otherwise it'll give a generic error, something like "Expected True but was False" rather than "Expected message was "abcd" but got "efgh"".

test/Garnet.test.cluster/RedirectTests/ClusterSlotVerificationTests.cs
```diff
@@ -591,12 +534,9 @@ void SERedisTRYAGAINTest(BaseCommand command)
                     ConfigureSlotForMigration();
                     try
                     {
-                        _ = context.clusterTestUtils.GetServer(requestNodeIndex).Execute(command.Command, command.GetSingleSlotRequest(), CommandFlags.NoRedirect);
-                    }
-                    catch (Exception ex)
-                    {
+                        var ex = Assert.Throws<RedisServerException>(() => context.clusterTestUtils.GetServer(requestNodeIndex).Execute(command.Command, command.GetSingleSlotRequest(), CommandFlags.NoRedirect),
```
TalZaccai, line 537: Looks like there are still a couple of try/catch blocks in this method?

TedHartMS (approved):

PR #1429, opened 2025-11-03
Corporate security flagged issue - Enable NuGet Transitive Pinning

darrenge:
This was found in ADO repo so need to fix in GH repo so it mirrors it there.  

From Corporate:
Enable NuGet Transitive Pinning (🔒 deterministic deps, 🛡️ faster CG fixes)

What & Why
Enable NuGet’s Transitive Pinning to effectively promote a transitive dependency to a top-level dependency implicitly on your behalf when necessary.
Result: Ensures central management of transitive dependencies and a fixed graph that makes Component Governance (security) findings easier to triage and remediate.


darrenge:
It looks like this check will fail until other security issues are merged

Issue #89, opened 2024-03-22
Investigate use of RandomAccess API to implement storage device in Tsavorite


PaulusParssinen:
We discussed this a bit earlier in [the Garnet discord](https://aka.ms/garnet-discord) but I'm opening issue for tracking it here.

In .NET 6, [the File IO in the BCL was essentially entirely rewritten](https://devblogs.microsoft.com/dotnet/file-io-improvements-in-dotnet-6/), including introduction of new `RandomAccess` low-level API. The `RandomAccess` API is __extremely__ optimized low-level offset based thread-safe IO wrapper that depending on the platform, either uses `pread(v)`/`pwrite(v)` for Unix or Overlapped IO for Windows.

We should investigate if `RandomAccess` based storage device implementation can be used to replace `LocalStorageDevice` that only supports Windows currently.

AlgorithmsAreCool:
I've personally had great success with `RandomAccess`, but I'm curious about the perceived advantages of this addition? Performance? Consolidation?

PR #1430, opened 2025-11-06
Storage device improvements

badrishc:
* Add new storage device based on .NET's [RandomAccess](https://learn.microsoft.com/en-us/dotnet/api/system.io.randomaccess) class
* Add device benchmark
* Use `GetQueuedCompletionStatusEx` in `LocalStorageDevice`

Fixes https://github.com/microsoft/garnet/issues/89

### DIFF
--- a/libs/server/API/GarnetApi.cs
+++ b/libs/server/API/GarnetApi.cs
@@ -351,6 +351,13 @@ namespace Garnet.server
         /// <inheritdoc />
         public GarnetStatus Read_ObjectStore(ref byte[] key, ref ObjectInput input, ref GarnetObjectStoreOutput output)
             => storageSession.Read_ObjectStore(ref key, ref input, ref output, ref objectContext);
+
+        public void ReadWithPrefetch<TBatch>(ref TBatch batch, long userContext = default)
+            where TBatch : IReadArgBatch<SpanByte, RawStringInput, SpanByteAndMemory>
+#if NET9_0_OR_GREATER
+            , allows ref struct
+#endif
+            => storageSession.ReadWithPrefetch(ref batch, ref context, userContext);
         #endregion
 
         #region Bitmap Methods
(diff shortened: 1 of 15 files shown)


PaulusParssinen:
Very glad to see this being investigated!

Related issue https://github.com/microsoft/garnet/issues/89

libs/host/Configuration/Options.cs
```diff
@@ -507,6 +507,10 @@ internal sealed class Options : ICloneable
         [Option("use-native-device-linux", Required = false, HelpText = "Use native device on Linux for local storage")]
         public bool? UseNativeDeviceLinux { get; set; }
 
+        [OptionValidation]
+        [Option("use-random-access-device", Required = false, HelpText = "Use device based on .NET RandomAccess for local storage")]
```
vazois: Is it possible to introduce an enum for all different device types?

badrishc: ok done, please check

benchmark/Device.benchmark/DeviceUtils.cs
```diff
@@ -0,0 +1,70 @@
...
+            var pbuffer = pool.Get((int)numBytesToWrite);
+            fixed (byte* bufferRaw = buffer)
+            {
+                Buffer.MemoryCopy(bufferRaw, pbuffer.aligned_pointer, size, size);
+            }
+            using var semaphore = new SemaphoreSlim(0);
+            device.WriteAsync((IntPtr)pbuffer.aligned_pointer, address, (uint)numBytesToWrite, IOCallback, semaphore);
+            semaphore.Wait();
+
+            pbuffer.Return();
```
vazois: should we put this inside a finally?

badrishc: done

vazois (approved):

AlgorithmsAreCool:
I'm curious how performance is via the new `RandomAccess` based device?

PR #1460, opened 2025-12-06
Streams Draft PR

hamdaankhalid:
This PR aims to stabilize the experimental support for streams, and add new features to Streams that are actively needed by me (XREVRANGE, and RANGE).

I will be using memtier to benchmark stream performance.


hamdaankhalid:
Clsoing this PR to work on the forked branch: https://github.com/microsoft/garnet/pull/1461

PR #1462, opened 2025-12-07
Remove unnecessary unsafe from GetGeoHashCode

PaulusParssinen:
Turns out this was unnecessary. https://godbolt.org/z/86jszKEY1


badrishc:
Good find!

PR #1469, opened 2025-12-15
Implement VINFO

tiagonapoli:


test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1481,6 +1481,64 @@ public void RecreateIndexesOnRestore()
...
+            ClassicAssert.AreEqual(32, (int)res2[9]);
+            ClassicAssert.AreEqual("size", (string)res2[10]);
+            ClassicAssert.AreEqual(0, (long)res2[11]);
+
+            // VINFO on non-existent key returns null
+            var res5 = db.Execute("VINFO", ["nonexistent"]);
+            ClassicAssert.IsTrue(res5.IsNull);
+
+            // VINFO on wrong type returns error
+            // TODO: add WRONGTYPE test once implemented
```
kevin-montrose: Can we get this done?

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1481,6 +1481,64 @@ public void RecreateIndexesOnRestore()
 
         // TODO: FLUSHDB needs to cleanup too...
 
+        [Test]
+        public void VINFO()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase();
+
+            // Create a vector set with known parameters
```
kevin-montrose: I suggest trying all combinations of quantizers, reduce, effort (default and explicit values), and num links (default and explicit values) and making sure they're correct in VINFO.

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1481,6 +1481,64 @@ public void RecreateIndexesOnRestore()
...
+            var res1 = db.Execute("VADD", ["foo", "REDUCE", "5", "VALUES", "10", "1.0", "2.0", "3.0", "4.0", "5.0", "6.0", "7.0", "8.0", "9.0", "10.0", new byte[] { 0, 0, 0, 0 }, "Q8", "EF", "16", "M", "32"]);
+            ClassicAssert.AreEqual(1, (int)res1);
+
+            // Get VINFO - should return an array of 12 elements (6 key-value pairs)
+            var res2 = (RedisValue[])db.Execute("VINFO", ["foo"]);
+            ClassicAssert.AreEqual(12, res2.Length);
```
kevin-montrose: This makes an assumption about ordering, probably map these into a dictionary and then validate so ordering isn't a hard assumption.

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1481,6 +1481,64 @@ public void RecreateIndexesOnRestore()
...
+            ClassicAssert.AreEqual("q8", (string)res2[1]);
+            ClassicAssert.AreEqual("input-vector-dimensions", (string)res2[2]);
+            ClassicAssert.AreEqual(10, (int)res2[3]);
+            ClassicAssert.AreEqual("reduced-dimensions", (string)res2[4]);
+            ClassicAssert.AreEqual(5, (int)res2[5]);
+            ClassicAssert.AreEqual("build-exploration-factor", (string)res2[6]);
+            ClassicAssert.AreEqual(16, (int)res2[7]);
+            ClassicAssert.AreEqual("num-links", (string)res2[8]);
+            ClassicAssert.AreEqual(32, (int)res2[9]);
+            ClassicAssert.AreEqual("size", (string)res2[10]);
```
kevin-montrose: Add a TODO that size isn't implemented yet.

libs/server/Resp/Vector/RespServerSessionVectors.cs
```diff
@@ -1004,11 +1004,47 @@ private bool NetworkVINFO<TGarnetApi>(ref TGarnetApi storageApi)
...
+                return true;
+            }
+            
+            if (res == GarnetStatus.WRONGTYPE)
+            {
+                WriteError("ERR Not a Vector Set"u8);
+                return true;
+            }
 
+            string quantTypeStr = quantType switch
```
kevin-montrose: This allocates a bunch of strings and then has to map things to ASCII.  Not a high priority command, but we should follow best practice here and do this alloc free.

Pattern is to only use the `ReadOnlySpan<byte>` taking overloads (and the `u8` suffix) for constant strings.  For the numbers, use `TryWriteIntXXAsBulkString` and `SendAndReset()` with a while loop.

kevin-montrose: nit: use type inference (ie. `var`) in the Garnet project.  This differs from some other coding standards, but its preferred here.

libs/server/Storage/Session/MainStore/VectorStoreOps.cs
```diff
@@ -272,5 +272,44 @@ internal unsafe GarnetStatus VectorSetDimensions(SpanByte key, out int dimension
...
+        [SkipLocalsInit]
+        internal unsafe GarnetStatus VectorSetInfo(SpanByte key,
+            out VectorQuantType quantType,
+            out uint vectorDimensions,
+            out uint reducedDimensions,
+            out uint buildExplorationFactor,
+            out uint numberOfLinks,
+            out long size)
+        {
+            parseState.InitializeWithArgument(ArgSlice.FromPinnedSpan(key.AsReadOnlySpan()));
```
kevin-montrose: nit: this could be `new(ref key)`, I believe the `ArgSlice` internal constructor is visible.

libs/server/Resp/Vector/RespServerSessionVectors.cs
```diff
@@ -1004,11 +1004,47 @@ private bool NetworkVINFO<TGarnetApi>(ref TGarnetApi storageApi)
...
+                return AbortWithWrongNumberOfArguments("VINFO");
+            }
 
-            while (!RespWriteUtils.TryWriteDirect(CmdStrings.RESP_OK, ref dcurr, dend))
-                SendAndReset();
+            var key = parseState.GetArgSliceByRef(0);
+            var res = storageApi.VectorSetInfo(key, out VectorQuantType quantType, out uint vectorDimensions, out uint reducedDimensions, out uint buildExplorationFactor, out uint numLinks, out long size);
+            if (res == GarnetStatus.NOTFOUND)
+            {
+                WriteNull();
```
kevin-montrose: This is subtly wrong - see [`VINFO` Returns](https://redis.io/docs/latest/commands/vinfo/#return-information).

If `RespProtocolVersion == 3` this should return the null (`_\r\n`), but in RESP 2 it writes the "null array" (`*-1\r\n`).  This is wrapped up in `WriteNullArray()`, so just call that instead.

 Many (but not all) libraries will paper over that mismatch, so understandable it got missed.

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1481,6 +1481,64 @@ public void RecreateIndexesOnRestore()
...
+            ClassicAssert.AreEqual(0, (long)res2[11]);
+
+            // VINFO on non-existent key returns null
+            var res5 = db.Execute("VINFO", ["nonexistent"]);
```
kevin-montrose: Per prior comment, give this a test with both RESP2 and RESP3 (you'll need to be explicit with connections to test that).  SE.Redis does expose types on RedisResult, though I'm not 100% sure if you can distinguish the null array.  Regardless, exercise the code path.

libs/server/Resp/Vector/RespServerSessionVectors.cs
tiagonapoli: Nice catch, fixed

libs/server/Resp/Vector/RespServerSessionVectors.cs
tiagonapoli: Fixed, it's now using alloc free - Since I supposed this command wouldn't be used a lot I just did simplest solution

Question - There are some helper methods with [AggressiveInlining] that wrap the `SendAndReset()` with while loop logic, e.g.:
```cs
        private void WriteSimpleString(scoped ReadOnlySpan<byte> simpleString)
        {
            while (!RespWriteUtils.TryWriteSimpleString(simpleString, ref dcurr, dend))
                SendAndReset();
        }
```
I'm using them, is it ok?

test/Garnet.test/RespVectorSetTests.cs
tiagonapoli: Fixed - Added a test that covers RESP2 and RESP3

libs/server/Resp/Vector/RespServerSessionVectors.cs
tiagonapoli: Fixed

test/Garnet.test/RespVectorSetTests.cs
tiagonapoli: Fixed - It's now testing multiple combinations of params

test/Garnet.test/RespVectorSetTests.cs
tiagonapoli: Fixed

test/Garnet.test/RespVectorSetTests.cs
tiagonapoli: Fixed

libs/server/Resp/Vector/RespServerSessionVectors.cs
```diff
@@ -1015,11 +1015,47 @@ private bool NetworkVINFO<TGarnetApi>(ref TGarnetApi storageApi)
...
+                return true;
+            }
 
+            var quantTypeSpan = quantType switch
+            {
+                VectorQuantType.NoQuant => "f32"u8,
+                VectorQuantType.Bin => "bin"u8,
+                VectorQuantType.Q8 => "q8"u8,
+                VectorQuantType.XPreQ8 => "xpreq8"u8,
+                _ => "invalid"u8,
```
kevin-montrose: nit: let's throw here - I'd rather fill up logging than return something strange.

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -94,7 +95,7 @@ public void WrongTypeForVectorSetOpsOnNonVectorSetKeys()
                             // TODO: Implement when VGETATTR works
                             continue;
                         case RespCommand.VINFO:
-                            // TODO: Implement when VGETATTR works
+                            exc = ClassicAssert.Throws<RedisServerException>(() => db.Execute("VINFO", ["foo"]));
                             continue;
```
kevin-montrose: This needs to be a `break;` now or we don't run the test.

test/Garnet.test/RespVectorSetTests.cs
```diff
@@ -1566,6 +1567,149 @@ public void RecreateIndexesOnRestore()
...
+                                ClassicAssert.AreEqual(values["build-exploration-factor"], expectedEf);
+                                ClassicAssert.AreEqual(values["num-links"], expectedNumLinks);
+                                // TODO: Update once DiskANN exposes card()
+                                ClassicAssert.AreEqual(values["size"], "0");
+
+                                // Modify first value for second add (change from "1.0" to "2.0")
+                                vectorValues[2] = "2.0";
+
+                                // Add another element and try again
+                                res = db.Execute("VADD", GenerateVADDOptions(fooKey, quantizer, reduceValueToUse, ef, numLinks, vectorValues.ToArray(), [0, 0, 0, 0]));
```
kevin-montrose: I suspect you wanted this element to be `[1, 0, 0, 0]`, otherwise `VADD` doing an update instead of add.  Can't tell it now, but in the future the size result would be wrong.

tiagonapoli: Nice catch, fixed

test/Garnet.test/RespVectorSetTests.cs
tiagonapoli: Nice catch

libs/server/Resp/Vector/RespServerSessionVectors.cs
tiagonapoli: Fixed

libs/server/Resp/Vector/RespServerSessionVectors.cs
kevin-montrose: That's fine - there's a push and pull kinda going on where some folks reach for the (older) while-loop style and some use helpers.  At some point maybe we do a cleanup and switch to helpers everywhere, but that'd need some profiling work.

kevin-montrose (approved):

PR #1476, opened 2025-12-18
Implement VSETATTR

tiagonapoli:


hailangx:
@copilot does the VSETATTR  and vadd converge on the attributes write ?

PR #1478, opened 2025-12-19
New ETag Commands Syntax + ETag Object Support

TalZaccai:
This PR introduces a notion of _"meta commands"_ - commands that "envelop" existing data commands.
The meta commands introduced in this PR are ETag related - _EXECWITHETAG_ (execute a command and add the dest key's etag to the output), as well as _EXECIFMATCH_, _EXECIFNOTMATCH_, _EXECIFGREATER_ (conditionally-execute a command based on the dest key's ETag in relation to the ETag meta-argument).

These command are able to replace the existing etag-related commands and command arguments, and in-turn support many more (including object commands).
For example:
`DELIFGREATER key etag` -> `EXECIFGREATER etag DEL key`
`GETIFNOTMATCH key etag` -> `EXECIFNOTMATCH etag GET key`
`GETWITHETAG key` -> `EXECWITHETAG GET key`
`SETIFMATCH key value etag` -> `EXECIFMATCH etag SET key value`

To achieve this logic, we parse the meta command and its meta-arguments into the `SessionParseState`. Currently this logic supports a constant number of meta-arguments which is defined in the `Arity` property in the meta-command's `RespCommandInfo` (arg count = arity - 1).
When the parse state gets handed to the appropriate `ISessionFunctions` method, the method decides whether the operation should get executed (in the case of conditional-execution) and to what value should the record's ETag advance (if at all).
The "special output" (in case the ETag needs to be written along with the command output) is handled by the `ISessionFunctions` method in the main and unified case, and in the `IGarnetObject.Operate` method in the object case.

_To-do:_
- [x] Add new supported meta-commands, remove existing etag commands and command arguments & update usages
- [x] Implement meta-command and meta-argument parsing in the RESP command parser.
- [x] Update `SessionParseState` to parse and hold meta-arguments
- [x] Update `InputHeader` to include meta-command and update commands
- [x] Implement ETag-related meta-command handling in `ISessionFunctions` for all 3 contexts
- [x] Implement meta-command-based ETag output for all object commands
- [x] Handle ETags in operations that do not call into `ISessionFunctions` (internal transactions)
- [x] Further clean-up of ETag logic in `ISessionFunctions`
- [x] Introduce leaner AOF serialization for commands with single long meta-argument
- [x] Support programmatic commands supplying long etag without serializing to parse state

_Testing & Benchmarking:_
- [x] Expand existing main-store ETag tests to include all meta-commands
- [x] Verify all write commands result in ETag advancement
- [x] Verify ETag output added to all commands when meta-command present
- [x] Verify running meta-commands within transactions
- [ ] Verify benchmarking numbers on commands without ETags
- [x] Add benchmarking logic for meta-commands with basic commands / basic commands with ETags


libs/server/InputHeader.cs
```diff
@@ -18,24 +18,105 @@ namespace Garnet.server
...
+        /// <summary>
+        /// No flags set
+        /// </summary>
+        None = 0,
+
         /// <summary>
         /// Flag indicating an operation intending to add an etag for a RAWSTRING command.
         /// </summary>
-        WithEtag = 16,
+        WithEtag = 1,
```
badrishc: is this still needed in RespInputFlags?

libs/server/Objects/Types/IGarnetObject.cs
```diff
@@ -21,9 +21,12 @@ public interface IGarnetObject : IHeapObject
         /// </summary>
         /// <param name="input"></param>
         /// <param name="output"></param>
+        /// <param name="respProtocolVersion"></param>
+        /// <param name="execOp"></param>
+        /// <param name="updatedEtag"></param>
         /// <param name="sizeChange"></param>
         /// <returns></returns>
-        bool Operate(ref ObjectInput input, ref ObjectOutput output, byte respProtocolVersion, out long sizeChange);
+        bool Operate(ref ObjectInput input, ref ObjectOutput output, byte respProtocolVersion, bool execOp, long updatedEtag, out long sizeChange);
```
badrishc: add comments on what execOp and updatedEtag mean. ideally Operate and calls like it should not accept arguments about etags specifically. they just receive a meta command and its arguments.

libs/server/InputHeader.cs
```diff
@@ -62,105 +139,94 @@ public struct RespInputHeader
...
+        internal RespMetaCommand MetaCmd;
+
         /// <summary>
         /// Create a new instance of RespInputHeader
         /// </summary>
         /// <param name="cmd">Command</param>
+        /// <param name="metaCmd">Meta command</param>
         /// <param name="flags">Flags</param>
-        public RespInputHeader(RespCommand cmd, RespInputFlags flags = 0)
+        public RespInputHeader(RespCommand cmd, RespMetaCommand metaCmd, RespInputFlags flags = 0)
```
badrishc: meta command could be 0 by default, to simplify the common cases where nobody cares about meta commands.

libs/server/Resp/BasicCommands.cs
```diff
@@ -621,13 +599,10 @@ private bool NetworkSET_Conditional<TGarnetApi>(RespCommand cmd, int expiry, Pin
                       ? TimeSpan.FromMilliseconds(expiry).Ticks
                       : TimeSpan.FromSeconds(expiry).Ticks);
 
-            var input = new StringInput(cmd, ref parseState, startIdx: 1, arg1: inputArg);
+            var input = new StringInput(cmd, metaCommand, ref parseState, startIdx: 1, arg1: inputArg);
```
badrishc: it is unclear whether parseState here contains the wrapping command as well. ideally we have a parse state for the actual command that is no different from before, and an optional extra parse state for the outer wrapper shell.

libs/server/InputHeader.cs
TalZaccai: No that's not used any longer

libs/server/Objects/Types/IGarnetObject.cs
TalZaccai: I agree, that irks me too, the issue here is that Operate sets the output of the command which may need to include the etag (and there are different scenarios depending on the Operate logic, for instance it may need to only output an error without the etag even when one is requested).

libs/server/InputHeader.cs
TalZaccai: I wanted to err on the side of caution here, I didn't want it to be easy to just forget to input the meta command. If you want to ignore it, you should do so explicitly.

libs/server/Resp/BasicCommands.cs
TalZaccai: parseState contains both, and they are seperately accessed. The meta-parameters preceed the "standard" parameters on the buffer. You would access the standard parameters as you would before (indexes unchanged) and you access the meta-parameters by explicitly specifying `isMetaArg: true`.

libs/server/InputHeader.cs
```diff
@@ -19,23 +19,29 @@ namespace Garnet.server
...
+        None = 0,
 
         /// <summary>
         /// Flag indicating a SET operation that returns the previous value (for strings).
         /// </summary>
-        SetGet = 32,
+        SetGet = 1,
 
         /// <summary>
         /// Deterministic
```
TedHartMS, line 32: expand--what does this mean

libs/server/InputHeader.cs
```diff
@@ -62,6 +64,21 @@ public struct RespInputHeader
         [FieldOffset(2)]
         internal RespInputFlags flags;
 
+        [FieldOffset(3)]
+        internal byte SubId;
```
TedHartMS, line 80: what does this do

libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -78,31 +77,68 @@ private void HashGetAll(ref ObjectOutput output, byte respProtocolVersion)
             }
         }
 
-        private void HashDelete(ref ObjectInput input, ref ObjectOutput output)
+        private void HashDelete(ref ObjectInput input, ref ObjectOutput output, byte respProtocolVersion)
         {
+            var removed = 0;
```
TedHartMS: What's the benefit to replacing the on-the-fly increment of Result1 by introducing a local that is later assigned to output.Result1? Is there something that relies on Result1 not being changed until a successful completion? (This pattern has been introduced in a lot of places)

libs/server/Objects/Hash/HashObjectImpl.cs
```diff
@@ -78,31 +77,68 @@ private void HashGetAll(ref ObjectOutput output, byte respProtocolVersion)
...
             var key = GetByteSpanFromInput(ref input, 0);
-            output.result1 = TryGetValue(key, out var hashValue) ? hashValue.Length : 0;
+            var length = TryGetValue(key, out var hashValue) ? hashValue.Length : 0;
+
+            if (!input.header.CheckSkipRespOutputFlag())
+            {
+                using var writer = new RespMemoryWriter(respProtocolVersion, ref output.SpanByteAndMemory);
+                writer.WriteInt32(length);
+            }
+
```
TedHartMS: Does this need to set ObjectOutputFlags.ValueUnchanged? (Similar for ListLength and likely others)

libs/server/Resp/RespServerSession.cs
```diff
@@ -55,6 +55,8 @@ internal sealed unsafe partial class RespServerSession : ServerSessionBase
 
         internal SessionParseState parseState;
         internal SessionParseState customCommandParseState;
+        internal MetaCommandInfo metaCommandInfo;
+        private long etag;
```
TedHartMS, line 63: add comment

libs/server/InputHeader.cs
TalZaccai: That's not related to this PR, I just changed the numeric values

libs/server/Objects/Hash/HashObjectImpl.cs
TalZaccai: At this point it was mainly for readability to understand what is being stored, but I can revet these changes.

libs/server/Storage/Functions/MainStore/ReadMethods.cs
```diff
@@ -25,26 +25,35 @@ public bool Reader<TSourceLogRecord>(in TSourceLogRecord srcLogRecord, ref Strin
...
-            var cmd = input.header.cmd;
-            var value = srcLogRecord.ValueSpan; // reduce redundant length calculations
-            if (cmd == RespCommand.GETIFNOTMATCH)
-            {
-                if (handleGetIfNotMatch(in srcLogRecord, ref input, ref output, ref readInfo))
-                    return true;
-            }
-            else if (cmd > RespCommandExtensions.LastValidCommand)
+            var metaCmd = input.metaCommandInfo.MetaCommand;
+            if (metaCmd != RespMetaCommand.None || srcLogRecord.Info.HasETag)
```
TedHartMS: Should this check .IsETagCommand instead of just not .None?

libs/server/MetaCommandInfo.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// <param name="updatedEtag">Etag value that should be assigned to the record, should the operation run and succeed</param>
+        /// <param name="initContext">True if method called from initial updater context</param>
+        /// <param name="readOnlyContext">True if method called from read-only context</param>
+        /// <returns>True if operation should execute</returns>
+        public bool CheckConditionalExecution(long currEtag, out long updatedEtag, bool initContext = false, bool readOnlyContext = false)
+        {
+            updatedEtag = currEtag;
+
+            // If there is no meta-command or current record does not have an etag - nothing to check
+            if (MetaCommand == RespMetaCommand.None && currEtag == LogRecord.NoETag)
```
TedHartMS: Should this be checking "is not ETag metacommand"?

badrishc:
To be reconsidered at a future point, closing for now to focus on minimal API per https://github.com/microsoft/garnet/pull/1739

Issue #1477, opened 2025-12-19
Exception received at AutoCheckpointTask


wwbgo:
### Describe the bug

```
Exception received at AutoCheckpointTask
System.NullReferenceException: Object reference not set to an instance of an object.
at Garnet.server.SingleDatabaseManager.TaskCheckpointBasedOnAofSizeLimitAsync(Int64 aofSizeLimit, CancellationToken token, ILogger logger) in /_/libs/server/Databases/SingleDatabaseManager.cs:line 225
at Garnet.server.StoreWrapper.AutoCheckpointBasedOnAofSizeLimit(Int64 aofSizeLimit, CancellationToken token, ILogger logger) in /_/libs/server/StoreWrapper.cs:line 648
```

### Steps to reproduce the bug

```
"EnableStorageTier": true,
"Recover": true,
"EnableAOF": true,
"AofSizeLimit" : "2g",
"CompactionFrequencySecs": 86400,
"CompactionType": "Lookup",
"EnableIncrementalSnapshots": false,
"MemorySize": "512m",
"IndexSize": "128m",
"ObjectStoreLogMemorySize": "32m",
"ObjectStoreIndexSize": "16m",
"ObjectStorePageSize" : "1m",
"MaxDatabases": 1,
```

### Release version

1.0.91

### OS version

in docker (mcr.microsoft.com/dotnet/aspnet:9.0)

TalZaccai:
Thank you! Fixed in PR #1491 :)

PR #1491, opened 2026-01-07
Bugfix in SingleDatabaseManager

TalZaccai:
Fixing access to clusterProvider without checking if cluster mode enabled + removing unnecessary duplicate storeWrapper field (Fixes issue #1477).
FYI @vazois

### DIFF
--- a/libs/server/Databases/SingleDatabaseManager.cs
+++ b/libs/server/Databases/SingleDatabaseManager.cs
@@ -26,14 +26,10 @@ namespace Garnet.server
 
         readonly GarnetDatabase defaultDatabase;
 
-        readonly StoreWrapper storeWrapper;
-
-
         public SingleDatabaseManager(StoreWrapper.DatabaseCreatorDelegate createDatabaseDelegate, StoreWrapper storeWrapper, bool createDefaultDatabase = true) :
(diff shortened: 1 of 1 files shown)


libs/server/Databases/SingleDatabaseManager.cs
```diff
@@ -222,7 +218,7 @@ public override async Task TaskCheckpointBasedOnAofSizeLimitAsync(long aofSizeLi
             try
             {
                 // Checkpoint will be triggered from AOF replay
-                if (storeWrapper.clusterProvider.IsReplica())
+                if (StoreWrapper.serverOptions.EnableCluster && StoreWrapper.clusterProvider.IsReplica())
```
vazois, line 221: Why would we be calling the SingleDatabaseManager implementation of TaskCheckpointBasedOnAofSizeLimitAsync, when the cluster is disabled? Shouldn't we be calling the MultiDatabaseManager implementation instead?

TalZaccai: Becuase MultiDatabaseManager's implementation consists of unnecessary extra logic when dealing with a single database...

vazois (approved):

Issue #1446, opened 2025-11-26
Support `--cluster-announce-hostname` for DNS based clustering


darwin67:
### Feature request type

enhancement

### Is your feature request related to a problem? Please describe

Hi folks,

Currently Garnet cluster nodes have a problem when running on Kubernetes.
When using a statefulset to configure the cluster, a restart/termination of a pod will result in a stale node in the cluster without `CleanClusterConfig` enabled.

This is because pods in k8s are not guaranteed of an IP, and when cluster data is preserved via PVs, the reused node information will not match, causing the cluster to go into a weird state, where the restarted node is not actually recognized.

~I believe Redis' gossip protocol will still resolve that over time when the IP is corrected, but this doesn't happen on Garnet, which is a bug of its own.~

### Describe the solution you'd like

Provide additional cluster announcement options as redis/valkey to allow using the hostname instead of the IP.
This will make operations in Kubernetes a lot easier since statefulsets guarantees hostname consistency, but not IPs.

Reference: https://raw.githubusercontent.com/redis/redis/7.0/redis.conf
```
# Clusters can configure their announced hostname using this config. This is a common use case for 
# applications that need to use TLS Server Name Indication (SNI) or dealing with DNS based
# routing. By default this value is only shown as additional metadata in the CLUSTER SLOTS
# command, but can be changed using 'cluster-preferred-endpoint-type' config. This value is 
# communicated along the clusterbus to all nodes, setting it to an empty string will remove 
# the hostname and also propagate the removal.
#
# cluster-announce-hostname ""

# Clusters can advertise how clients should connect to them using either their IP address,
# a user defined hostname, or by declaring they have no endpoint. Which endpoint is
# shown as the preferred endpoint is set by using the cluster-preferred-endpoint-type
# config with values 'ip', 'hostname', or 'unknown-endpoint'. This value controls how
# the endpoint returned for MOVED/ASKING requests as well as the first field of CLUSTER SLOTS. 
# If the preferred endpoint type is set to hostname, but no announced hostname is set, a '?' 
# will be returned instead.
#
# When a cluster advertises itself as having an unknown endpoint, it's indicating that
# the server doesn't know how clients can reach the cluster. This can happen in certain 
# networking situations where there are multiple possible routes to the node, and the 
# server doesn't know which one the client took. In this case, the server is expecting
# the client to reach out on the same endpoint it used for making the last request, but use
# the port provided in the response.
#
# cluster-preferred-endpoint-type ip
```

[...]

PR #1495, opened 2026-01-10
Support DNS based clustering

bandalgomsu:
Add `cluster-announce-hostname` and `cluster-preferred-endpoint-type` options to support DNS-based clustering.
Implemented with reference to the behavior of redis.

thanks 👍 

address : #1446

### DIFF
--- a/libs/cluster/Server/Replication/ReplicaOps/ReplicaDisklessSync.cs
+++ b/libs/cluster/Server/Replication/ReplicaOps/ReplicaDisklessSync.cs
@@ -36,7 +36,7 @@ namespace Garnet.cluster
                     return false;
 
                 // Wait for threads to agree configuration change of this node
-                session.UnsafeBumpAndWaitForEpochTransition();
+                session?.UnsafeBumpAndWaitForEpochTransition();
                 if (options.Background)
                     _ = Task.Run(() => TryBeginReplicaSync(options.UpgradeLock));
                 else
@@ -52,7 +52,9 @@ namespace Garnet.cluster
             catch (Exception ex)
             {
                 logger?.LogError(ex, $"{nameof(TryReplicateDisklessSync)}");
+                return false;
             }
+
             return true;
 
             async Task<string> TryBeginReplicaSync(bool downgradeLock)
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -6487,9 +6487,9 @@ lodash.uniq@^4.5.0:
   integrity sha512-xfBaXQd9ryd9dlSDvnvI0lvxfLJlYAZzXomUYzLKtUeOQvOP5piqAWuGtrhWeqaXK9hhoM/iyJc5AV+XfsX3HQ==
 
 lodash@^4.17.20, lodash@^4.17.21:
-  version "4.17.21"
-  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.21.tgz#679591c564c3bffaae8454cf0b3df370c3d6911c"
-  integrity sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==
+  version "4.17.23"
+  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.23.tgz#f113b0378386103be4f6893388c73d0bde7f2c5a"
+  integrity sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==
 
 longest-streak@^3.0.0:
   version "3.1.0"


bandalgomsu:

libs/host/Configuration/Options.cs
```diff
@@ -53,6 +53,12 @@ internal sealed class Options : ICloneable
         [Option("cluster-announce-ip", Required = false, HelpText = "IP address that this node advertises to other nodes to connect to for gossiping.")]
         public string ClusterAnnounceIp { get; set; }
 
+        [Option("cluster-announce-hostname", Required = false, HelpText = "Hostname that this node advertises to other nodes to connect to for gossiping.")]
+        public string ClusterAnnounceHostname { get; set; }
+
+        [Option("cluster-preferred-endpoint-type", Required = false, HelpText = "Determines the endpoint type to be advertised to other nodes. (value options: ip, hostname, unknown-endpoint)")]
+        public string ClusterPreferredEndpointType { get; set; }
```
vazois: nit: Can you make this an enum type? See DeviceType option for an example

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -709,32 +722,93 @@ string CreateFormattedNodeInfo(int workerId, bool connected)
...
+                nodeInfo += $"${k.Length}\r\n{k}\r\n";
+                nodeInfo += FormatValueOrNull(v);
+            }
+
+            return nodeInfo;
+        }
+
+        private static string FormatValueOrNull(string value)
+        {
+            if (string.IsNullOrEmpty(value)) return "$-1\r\n";
```
vazois: Add "$-1\r\n" to CmdStrings.cs

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -709,32 +722,93 @@ string CreateFormattedNodeInfo(int workerId, bool connected)
...
-
-                rangeInfo += $"*4\r\n${replicaAddress.Length}\r\n{replicaAddress}\r\n:{replicaPort}\r\n${replicaId.Length}\r\n{replicaId}\r\n";
-                rangeInfo += $"*2\r\n$8\r\nhostname\r\n${replicaHostname.Length}\r\n{replicaHostname}\r\n";
+                rangeInfo += CreateNodeNetworkingInfo(replicaIp, replicaPort, replicaId, replicaHostname, preferredEndpointType);
             }
+
             return rangeInfo;
         }
 
+        private string CreateNodeNetworkingInfo(
```
vazois, line 754: Can you use StringBuilder with this class?
Also, could you simplify this seems overly complicated. What is the RESP format expected here?

Maybe something like this

``` suggestion
            switch (preferredEndpointType)
            {
                case ClusterPreferredEndpointType.Ip:
                    FormatValueOrNull(ipAddress).Append(':').Append(port).Append('$').Append(nodeid.Length).Append("\r\n").Append(nodeid).Append("\r\n");
                    break;
                // Rest options here....
            }

            return nodeInfo;

            StringBuilder FormatValueOrNull(string value)
            {
                if (string.IsNullOrEmpty(value))
                    return stringBuilder.Append("$-1\r\n");
                else
                    return stringBuilder.Append('$').Append(value.Length).Append("\r\n").Append(value).Append("\r\n");
            }
```

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -709,32 +722,93 @@ string CreateFormattedNodeInfo(int workerId, bool connected)
...
+            foreach (var (k, v) in metaPairs)
+            {
+                nodeInfo += $"${k.Length}\r\n{k}\r\n";
+                nodeInfo += FormatValueOrNull(v);
+            }
+
+            return nodeInfo;
+        }
+
+        private static string FormatValueOrNull(string value)
```
vazois: Can you make this a local method? It will be better not to pollute the scope with locally used methods.

vazois (changes requested):
Does CLUSTER SHARDS command need update?

bandalgomsu:
The implementation of CLUSTER SHARDS has been postponed due to differences in the response format compared to Redis.

https://redis.io/docs/latest/commands/cluster-shards/

If necessary, i will implement it in the same format as Redis !

modify address -> ip
add endpoint

libs/host/Configuration/Options.cs
bandalgomsu: I want to keep the `unknown-endpoint` name as input (because of compatibility with redis)
, but the Parser inside 

TryParseCommandLineArguments doesn't work well for the kebab case, so I implemented it like this. Is there a better way?

libs/cluster/Server/ClusterConfig.cs
bandalgomsu: It expects the following format !

1) string: preferred endpoint
   - Ip: ip address
   - Hostname: hostname, or "?" if hostname is null or empty
   - UnknownEndpoint: null
2) integer: port
3) string: node id
4) metadata: map-like key/value pairs
   - "ip" is included if preferred endpoint type != Ip
   - "hostname" is included if preferred endpoint type != Hostname AND hostname is not null or empty

libs/host/Configuration/Options.cs
vazois: you can rename it unknown and if it is needed for the resp reply you should be able to inline it.

libs/cluster/Server/ClusterConfig.cs
vazois: ok try to simplify it and group it per type (as in the example I shared) even if there is some minor duplication. It will be more maintainable I feel

bandalgomsu: thanks for feedback ! 

i completed modifying

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -468,10 +470,21 @@ public string GetHostNameFromNodeId(string nodeId)
...
+
+            return (GetEndpointByPreferredType(workerId, type), workers[workerId].Port);
+        }
+
+        private string GetEndpointByPreferredType(int workerId, ClusterPreferredEndpointType type)
+        {
+            return type switch
+            {
+                ClusterPreferredEndpointType.Ip => workers[workerId].Address,
+                ClusterPreferredEndpointType.Hostname => string.IsNullOrEmpty(workers[workerId].hostname) ? "?" : workers[workerId].hostname,
```
vazois, line 485: Please add explicit handling for the unknown type.

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -709,32 +722,118 @@ string CreateFormattedNodeInfo(int workerId, bool connected)
...
+                        .Append(FormatValueOrNull(ipAddress));
+
+                    if (!isNullOrEmptyHostname)
+                    {
+                        sb.Append("$8\r\nhostname\r\n");
+                        sb.Append(FormatValueOrNull(hostname));
+                    }
+
+                    return sb.ToString();
+            }
```
vazois, line 817:
```suggestion
            }

            static string FormatValueOrNull(string value)
            {
                if (string.IsNullOrEmpty(value)) return CmdStrings.GenericNullValue; // "$-1\r\n"
                return $"${value.Length}\r\n{value}\r\n";
            }
```

libs/cluster/Server/ClusterManager.cs
```diff
@@ -205,7 +206,7 @@ private void InitLocal(string address, int port, bool recoverConfig)
                     configEpoch: 0,
                     NodeRole.PRIMARY,
                     null,
-                    Format.GetHostName());
+                    hostname: string.IsNullOrEmpty(hostname) ? "" : hostname);
```
vazois: We need to support the legacy implementation because services that have been using Garnet cluster for a long time, they might be relying on that.
Maybe you can add an extra option to force the hostname to be the empty string, otherwise by default it should call Format.GetHostname().
This should apply to the above also.

libs/cluster/Server/ClusterManagerWorkerState.cs
```diff
@@ -130,7 +131,7 @@ public ReadOnlySpan<byte> TryReset(bool soft, int expirySeconds = 60)
                         configEpoch: configEpoch,
                         role: NodeRole.PRIMARY,
                         replicaOfNodeId: null,
-                        Format.GetHostName());
+                        hostname: string.IsNullOrEmpty(hostname) ? "" : hostname);
```
vazois, line 134: Same here, the legacy option where we compute the hostname ourselves should still be available and work as default.

libs/cluster/Session/SlotVerification/RespClusterSlotVerify.cs
```diff
@@ -18,12 +18,12 @@ internal sealed unsafe partial class ClusterSession : IClusterSession
         /// <param name="slot"></param>
         /// <param name="config"></param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private void Redirect(ushort slot, ClusterConfig config)
+        private void Redirect(ushort slot, ClusterConfig config, ClusterPreferredEndpointType type)
         {
-            var (address, port) = config.GetEndpointFromSlot(slot);
+            var (endpoint, port) = config.GetEndpointFromSlot(slot, type);
```
vazois, line 24: nit; we don't need to pass the Endpoint type here since Redirect has access to serverOptions through the clusterProvider

vazois (changes requested):
Some more changes needed

libs/cluster/Server/ClusterConfig.cs
vazois: my mistake, can you declare this as a local function within CreateNodeNetworkingInfo

PR #1498, opened 2026-01-13
Implement TaskManager

vazois:
This PR implements a TaskManager to manage and categorize the background system tasks of Garnet.
This was needed to fix an issue related to AOF divergence caused by the ObjectCollectTask background task which was not supposed to be running when the Garnet node is a replica.
In addition, it consolidates all background task running and management into a single location to ensure robustness.

### DIFF
--- a/test/Garnet.test/TaskManagerTests.cs
+++ b/test/Garnet.test/TaskManagerTests.cs
@@ -4,14 +4,16 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Allure.NUnit;
 using Garnet.server;
 using NUnit.Framework;
 using NUnit.Framework.Legacy;
 
 namespace Garnet.test
 {
+    [AllureNUnit]
     [TestFixture]
-    internal class TaskManagerTests
+    internal class TaskManagerTests : AllureTestBase
     {
         [Test]
         public void TestBasicRegisterAndRun()


libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        /// <summary>
+        /// Register and start new task using the provider taskType and taskFactory
+        /// </summary>
+        /// <param name="taskType"></param>
+        /// <param name="taskFactory"></param>
+        /// <param name="taskCategory"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        public void Register(TaskType taskType, TaskCategory taskCategory, Func<CancellationToken, Task> taskFactory)
```
TalZaccai: naming: `Register` might be a bit confusing, maybe `RegisterAndRun`?

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                    linkedCts.Dispose();
+                    throw new Exception("Failed to add task to registry!");
+                }
+
+                // Finally run the task and update the registry
+                var task = Task.Run(async () => await taskFactory(linkedCts.Token), linkedCts.Token);
+                if (!tasks.TryUpdate(taskType, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = task }, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = null }))
+                {
+                    linkedCts.Cancel();
+                    task.Wait();
```
TalZaccai: Why close the 2 Copilot comments here? They have merit...

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                if (!tasks.TryAdd(taskType, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = null }))
+                {
+                    linkedCts.Cancel();
+                    linkedCts.Dispose();
+                    throw new Exception("Failed to add task to registry!");
+                }
+
+                // Finally run the task and update the registry
+                var task = Task.Run(async () => await taskFactory(linkedCts.Token), linkedCts.Token);
+                if (!tasks.TryUpdate(taskType, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = task }, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = null }))
```
TalZaccai: You might want to implement `IEquatable<TaskInfo>`.
Also, you can define the "null task struct" once for both the initialization and for comparison.

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <summary>
+        /// Cancel task
+        /// </summary>
+        /// <param name="taskType"></param>
+        /// <param name="taskCategory"></param>
+        /// <returns></returns>
+        public async Task TryCancelTask(TaskType taskType, TaskCategory taskCategory)
+        {
+            // TryGetValue task from registry
+            if (tasks.TryGetValue(taskType, out var taskInfo) && (taskInfo.taskCategory == taskCategory || taskCategory == TaskCategory.All) && tasks.TryRemove(taskType, out taskInfo))
```
TalZaccai: nit: Can make `TaskCategory` a flag so you don't have to check both `taskCategory` and `All`

libs/server/TaskManager.cs
TalZaccai: Need to pass `cts.Token` to the `Wait()` also?

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                taskInfo.cts.Cancel();
```
TalZaccai: nit: can call `await taskInfo.cts.CancelAsync();` here

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        /// <summary>
+        /// Try cancel all background tasks
+        /// </summary>
+        /// <returns></returns>
+        /// <param name="taskCategory"></param>
+        public async Task TryCancelAllTasks(TaskCategory taskCategory = TaskCategory.All)
+        {
+            foreach (var taskType in tasks.Keys)
```
TalZaccai: Can't we just cancel `cts` and wait for all tasks to finish?

TalZaccai (changes requested):
Added some comments

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        CommitTask,
+        CompactionTask,
+        ObjectCollectTask,
+        ExpiredKeyDeletionTask,
+        IndexAutoGrowTask,
+    }
+
+    public enum TaskCategory
+    {
+        All,
```
kevin-montrose: What's the distinction between `All` and `Generic`?

More broadly - I wonder if a `[Flags]` enum (something like `TaskNodePlacement` maybe) with something like:
```csharp
{
   None = 0,
   Primary = 1 << 0,
   Replica = 1 << 1,
   All = Primary | Replica,
}
```

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                    return;
+                }
+
+                // Create a linked cts
+                var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cts.Token);
+
+                // Try to add task to registry
+                if (!tasks.TryAdd(taskType, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = null }))
+                {
+                    linkedCts.Cancel();
```
kevin-montrose: You don't need to cancel since nothing could have subscribed at this point, just dispose.

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                    linkedCts.Dispose();
+                    throw new Exception("Failed to add task to registry!");
```
kevin-montrose: Throw something more specific than `Exception`.

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+
+                // Finally run the task and update the registry
+                var task = Task.Run(async () => await taskFactory(linkedCts.Token), linkedCts.Token);
```
kevin-montrose: `async` and `await` are unnecessary here.

Honestly, since we control all the registrations we don't even need the `Task.Run`, all that's doing is protecting us against the case that `taskFactory` blocks - which is an error anyway.

I suggest just `var task = taskFactory(linkedCts.Token);`.

libs/server/TaskManager.cs
kevin-montrose: What's the benefit of this TryAdd/TryUpdate dance versus making `TaskInfo` a sealed class and just setting `TaskInfo.task`?

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                if (!tasks.TryUpdate(taskType, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = task }, new TaskInfo() { cts = linkedCts, taskCategory = taskCategory, task = null }))
+                {
+                    linkedCts.Cancel();
+                    task.Wait();
+                    linkedCts.Dispose();
+                    throw new Exception("Failed to update registry with running task!");
```
kevin-montrose: More specific exception type please.

libs/server/TaskManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Cancel task
+        /// </summary>
+        /// <param name="taskType"></param>
+        /// <param name="taskCategory"></param>
+        /// <returns></returns>
+        public async Task TryCancelTask(TaskType taskType, TaskCategory taskCategory)
```
kevin-montrose: `TryXXX` implies a boolean return.  Either make this return `Task<bool>` or rename `CancelTask`.

It's also a bit strange to me that this takes `TaskCategory` at all - only one task can be registered under a given `TaskType` at any time.

libs/server/TaskManager.cs
kevin-montrose: Minor, but you can be clever with `TryRemove` instead of calling `TryGetValue` and `TryRemove`.

The [`TryRemove(KeyValuePair<...>)`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryremove?view=net-8.0#system-collections-concurrent-concurrentdictionary-2-tryremove(system-collections-generic-keyvaluepair((-0-1)))) override enforces that both key and value removed match expectations, so a clever definition of `IEquatable<TaskInfo>` on `TaskInfo` will let you do this in one call.

libs/server/TaskManager.cs
vazois: done!

libs/server/TaskManager.cs
vazois: well, if the task fails and exception is thrown the if clause will fail and it will never run.
If the update was to fail the cancel should propagate to the task and the task should complete right? Ideally, we don't want to timeout and let a task tangling stuck. So, it might be better to hang if the task does not complete even with cancellation, at least we can identify the offending background task and implement it such it completes.
I will consider this as resolved because I have made more changes to simplify this part of the code

libs/server/TaskManager.cs
vazois: not longer relevant because I removed this from taskInfo but good advice!

libs/server/TaskManager.cs
vazois: I tried to simplify this by adding a mapping between TaskType and its placement category. Thanks!

libs/server/TaskManager.cs
vazois: I renamed this

libs/server/TaskManager.cs
vazois: I tried to simplify this. We need to add an empty entry and then initialize the cts and run the task.
This should prevent the task from running if it failed to register. 
https://github.com/microsoft/garnet/commit/cb3c427327d35558126d38aebce067328958fdd5

libs/server/TaskManager.cs
vazois: yeah, it was too complicated because of the flag situation. Since I added a mapping from PlacementCategory to TaskType I don't need to TryGet I can do TryRemove directly.
It should be more clean now.

libs/server/TaskManager.cs
vazois: I am not using TaskInfo as a key, so I am not sure is relevant to implement IEquatable.

libs/server/TaskManager.cs
vazois: done!

libs/server/TaskManager.cs
vazois: I switched them to flags but did not add everything. New categories should be added as the need arises.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        public void RegisterAndRun(TaskType taskType, Func<CancellationToken, Task> taskFactory)
+        {
+            if (!dispose.TryReadLock())
+                return;
+
+            var failed = false;
+            TaskMetadata taskInfo = null;
+            try
```
TalZaccai, line 79: Can probably move the `try` to before calling `taskFactory(...)`... No reason why anything should fail before that, right?

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+                failed = true;
+            }
+            finally
+            {
+                if (failed)
+                {
+                    try
+                    {
+                        taskInfo?.cts.Cancel();
+                        taskInfo?.task.Wait(disposed ? default : cts.Token);
```
TalZaccai: It's possible that `taskInfo.task` is null here, if `taskFactory` failed

libs/server/TaskManager/TaskInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Represents metadata and control objects for a background maintenance task.
+    /// </summary>
+    public sealed class TaskMetadata
+    {
+        /// <summary>
+        /// Cancellation token source associated with this specific task
+        /// </summary>
+        public CancellationTokenSource cts;
```
TalZaccai: naming: public class fields should be capitalized

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        /// <summary>
+        /// Cancel tasks associated with the provided TaskPlacementCategory
+        /// </summary>
+        /// <param name="taskPlacementCategory"></param>
+        /// <returns></returns>
+        public async Task CancelTasks(TaskPlacementCategory taskPlacementCategory)
+        {
+            foreach (var taskType in TaskTypeExtensions.GetTaskTypes(taskPlacementCategory))
```
TalZaccai, line 151: You can also add a linked CTS for every task type, that way you can only cancel that one to cancel all the tasks of the same type.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Wait for task associated with the provided TaskType to complete.
+        /// </summary>
+        /// <param name="taskType"></param>
+        /// <param name="token"></param>
+        /// <returns></returns>
+        public bool Wait(TaskType taskType, CancellationToken token = default)
+        {
+            if (registry.TryGetValue(taskType, out var taskInfo))
```
TalZaccai: Duplicated code, could just do: 
```cs
public bool Wait(TaskType taskType, CancellationToken token = default) => 
    WaitAsync(taskType, token).Result;
```

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// </summary>
+        /// <param name="taskType"></param>
+        /// <returns></returns>
+        public async Task CancelTask(TaskType taskType)
+        {
+            if (registry.TryRemove(taskType, out var taskInfo))
+            {
+                try
+                {
+                    await taskInfo?.cts.CancelAsync();
```
TalZaccai: It's possible that this is called before `taskInfo` is populated? Needs a null check for `taskInfo.cts` + `taskInfo.task`

libs/server/TaskManager/TaskType.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Collections.Generic;
+
+namespace Garnet.server
+{
+    /// <summary>
+    /// Defines the types of background maintenance tasks that can be managed by the TaskManager.
+    /// These tasks represent various server operations that run asynchronously to maintain
+    /// system health, performance, and data integrity.
+    /// </summary>
+    public enum TaskType
```
TalZaccai: nit: can be a `byte`, I doubt that we'll have more than 256 tasks :)

libs/server/TaskManager/TaskType.cs
```diff
@@ -0,0 +1,113 @@
...
+            TaskPlacementMapping[(int)TaskType.ExpiredKeyDeletionTask] = TaskPlacementCategory.Primary;
+            TaskPlacementMapping[(int)TaskType.IndexAutoGrowTask] = TaskPlacementCategory.All;
+        }
+
+        /// <summary>
+        /// Retrieves task types for a placement category.
+        /// </summary>
+        /// <param name="lookupPlacementCategory">The placement category to filter by.</param>
+        /// <returns>An enumerable of task types that match the placement category.</returns>
+        public static IEnumerable<TaskType> GetTaskTypes(TaskPlacementCategory lookupPlacementCategory)
```
TalZaccai, line 81: If the only matching you need is `TaskPlacementCategory` -> `TaskType`, why not invert the map?

TalZaccai (changes requested):
Added some more comments...

libs/server/TaskManager/TaskType.cs
vazois: Because I will need to intersect the maps if a task belongs to one or more than one placement categories in the future and that will result in more allocations.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+                taskInfo.task = taskFactory(taskInfo.cts.Token);
+            }
+            catch (Exception ex)
+            {
+                logger?.LogError(ex, "Failed starting task {taskType} with {method}", taskType, nameof(RegisterAndRun));
+                failed = true;
+            }
+            finally
+            {
+                if (failed)
```
kevin-montrose: `failed` is only set in the `catch` path... so why not put the cleanup code in there instead of a separate `finally` block?

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+                {
+                    try
+                    {
+                        taskInfo?.cts.Cancel();
```
kevin-montrose: You can do `using(taskInfo?.cts)` rather than dispose in `finally`.

libs/server/TaskManager/TaskManager.cs
kevin-montrose: It's not legal to await a null value, null coalescing will not skip it.

libs/server/TaskManager/TaskManager.cs
vazois: This is completely different design. I am not sure what is the overhead of that because I will still need to find the task that was cancelled and wait on it. I am not sure there is an actual benefit other than cancelling in unison. I will still need to find which are those tasks that got cancelled and wait for them to complete. So, in effect I am increasing the space overhead because of the additional cancellation tokens needed.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <param name="taskType"></param>
+        /// <returns></returns>
+        public async Task CancelTask(TaskType taskType)
+        {
+            if (registry.TryRemove(taskType, out var taskInfo))
+            {
+                try
+                {
+                    await taskInfo?.cts.CancelAsync();
+                    await taskInfo?.task.WaitAsync(disposed ? default : cts.Token);
```
kevin-montrose: Same note, await'ing null will throw.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+                }
+                catch (Exception ex)
+                {
+                    logger?.LogCritical(ex, "Unknown exception received for {CancelTask}.", nameof(CancelTask));
+                }
+                finally
+                {
+                    taskInfo?.cts.Dispose();
```
kevin-montrose: Same pattern note, you can use `using(taskInfo?.cts)` to simplify this a bit.

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+            if (registry.TryRemove(taskType, out var taskInfo))
+            {
+                try
+                {
+                    await taskInfo?.cts.CancelAsync();
+                    await taskInfo?.task.WaitAsync(disposed ? default : cts.Token);
+                }
+                catch (Exception ex)
+                {
+                    logger?.LogCritical(ex, "Unknown exception received for {CancelTask}.", nameof(CancelTask));
```
kevin-montrose: Maybe also log the `TaskType`?

libs/server/TaskManager/TaskManager.cs
```diff
@@ -0,0 +1,186 @@
...
+                if (failed)
+                {
+                    try
+                    {
+                        taskInfo?.cts.Cancel();
+                        taskInfo?.task.Wait(disposed ? default : cts.Token);
+                    }
+                    catch (Exception ex)
+                    {
+                        logger?.LogCritical(ex, "Unknown exception received for {RegisterAndRun} at finally.", nameof(RegisterAndRun));
```
kevin-montrose: Maybe also log the `TaskType`?

kevin-montrose (changes requested):

libs/server/TaskManager/TaskManager.cs
vazois: I can do that but then I have to release the lock if the registration fails.
This should also happen when disposed. I am thinking I should put everything in the try so I can execute the finally and release the lock whatever happens.

PR #1508, opened 2026-01-23
Add post-operation hooks and epoch-safe AOF logging

badrishc:
* Introduce `PostUpsertOperation`, `PostRMWOperation`, and `PostDeleteOperation` hooks to `ISessionFunctions` in TsavoriteKV, enabling actions after upsert, RMW, and delete but before record unlock. Critically, these hooks are invoked at a point where it is safe to pulse the TsavoriteKV epoch (no protected memory is accessed here/after this point).
* Add `IEpochAccessor` interface for epoch management by Post*Operation hooks.
* Refactor `TsavoriteLog.Enqueue` to accept an epoch accessor from caller (e.g., TsavoriteKV ISF callbacks in our case), ensuring we correctly suspend/resume the outer epoch during blocking log appends.
* Move AOF write logic in Garnet server to these post-operation hooks, using a `UserData` flag to defer logging until after the main operation (but before unlocking).
* Make `UpsertInfo`, `RMWInfo`, and `DeleteInfo` structs carry a `UserData` byte for application-specific operation flags.
* Update existing `ISessionFunctions` implementations with no-op post-operation hooks.
* Enums `UpsertAction`, `RMWAction`, and `DeleteAction` are now byte-sized to save a bit of stack space.

These changes ensure that if we block waiting for `AllocateBlock` during AOF logging in TsavoriteLog, we will not be holding the outer (TsavoriteKV-level) epoch, thereby preventing a epoch table full situation for other normal user operation (GET/SET) threads under very heavy workloads.

### DIFF
--- a/libs/cluster/Server/Failover/ReplicaFailoverSession.cs
+++ b/libs/cluster/Server/Failover/ReplicaFailoverSession.cs
@@ -145,6 +145,9 @@ namespace Garnet.cluster
                 if (!clusterProvider.replicationManager.InitializeCheckpointStore())
                     logger?.LogWarning("Failed acquiring latest memory checkpoint metadata at {method}", nameof(TakeOverAsPrimary));
                 _ = clusterProvider.BumpAndWaitForEpochTransition();
+
+                // Resume all background maintenance that were possibly shutdown when this node became a replica
+                clusterProvider.storeWrapper.StartPrimaryTasks();
             }
             finally
             {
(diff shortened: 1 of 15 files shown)


libs/server/Storage/Functions/MainStore/DeleteMethods.cs
```diff
@@ -22,7 +22,7 @@ public bool SingleDeleter(ref SpanByte key, ref SpanByte value, ref DeleteInfo d
         public void PostSingleDeleter(ref SpanByte key, ref DeleteInfo deleteInfo)
         {
             if (functionsState.appendOnlyFile != null)
-                WriteLogDelete(ref key, deleteInfo.Version, deleteInfo.SessionID);
+                deleteInfo.UserData |= 0x1; // Mark that we need to write to AOF
```
TedHartMS: Make this a named constant rather than a magic number

PR #1513, opened 2026-01-26
Create Grid tests for DiskANN

tiagonapoli:
- Create vectors on a grid and do sanity checks on VSIM results and expected recalls

### DIFF
--- a/test/Garnet.test/DiskANN/DiskANNGridTests.cs
+++ b/test/Garnet.test/DiskANN/DiskANNGridTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Garnet.server;
 using NUnit.Framework;
 using NUnit.Framework.Legacy;
 using StackExchange.Redis;
@@ -11,9 +12,9 @@ using StackExchange.Redis;
 namespace Garnet.test.DiskANN
 {
     [TestFixture]
-    class DiskANNGridTests
+    public class DiskANNGridTests
     {
-        GarnetServer server;
+        private GarnetServer server;
 
         [SetUp]
         public void Setup()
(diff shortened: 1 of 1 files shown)


test/Garnet.test/DiskANN/DiskANNGridTests.cs
```diff
@@ -0,0 +1,259 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test.DiskANN
+{
+    [TestFixture]
+    class DiskANNGridTests
```
kevin-montrose: style nit: we're inconsistent in Garnet, but I would prefer if we had visibility modifiers on all these members.

So `public class DiskANNGridTests`, `private GarnetServer server;`, `private List<GridVector> GenerateGridVectors(int dimensions, int gridSize)` et. al., `private sealed class GridVector`, etc.

test/Garnet.test/DiskANN/DiskANNGridTests.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        [TearDown]
+        public void TearDown()
+        {
+            server.Dispose();
+            TestUtils.DeleteDirectory(TestUtils.MethodTestDir);
+        }
+
+        [Test]
+        [TestCase(100, 1, "NOQUANT")]
```
kevin-montrose: nit: I'd prefer using `VectorQuantType.NoQuant`, `Bin`, `Q8`, etc.  Makes it easier to confirm we've tested all quantizers.

test/Garnet.test/DiskANN/DiskANNGridTests.cs
```diff
@@ -0,0 +1,259 @@
...
+        [TestCase(10, 2, "NOQUANT")]
```
kevin-montrose: `NOQUANT` isn't implemented right, and defaults to `XPREQ8`... are these tests quantizer agnostic?

test/Garnet.test/DiskANN/DiskANNGridTests.cs
```diff
@@ -0,0 +1,259 @@
...
+                ClassicAssert.AreEqual(25, res.Length);
+                var vsimIds = GetVectorIdsForVsimResults((RedisResult[])res);
+                var expectedNN = BruteForceNearestNeighbors(addedGridVectors, addedGridVector.Vector, 25);
+                var intersectionCount = CalculateDistanceCountsIntersection(addedGridVectors, addedGridVector.Vector, expectedNN, vsimIds);
+                var recall = (double)intersectionCount / expectedNN.Count;
+                ClassicAssert.GreaterOrEqual(recall, 0.99, $"Recall too low: {recall} for vector ID {addedGridVector.Id}");
+            }
+        }
+
+        List<GridVector> GenerateGridVectors(int dimensions, int gridSize)
```
kevin-montrose: nit: some of these can be `static` right?

test/Garnet.test/DiskANN/DiskANNGridTests.cs
tiagonapoli: Grid tests should work for other quantizers as well - Let me change to XPREQ8 for now then, and in the future we add tests cases for other quantizers as we implement

test/Garnet.test/DiskANN/DiskANNGridTests.cs
tiagonapoli: Good point - fixed

test/Garnet.test/DiskANN/DiskANNGridTests.cs
tiagonapoli: Good point - fixed

test/Garnet.test/DiskANN/DiskANNGridTests.cs
tiagonapoli: Fixed

PR #1543, opened 2026-02-05
Replace spin-wait with semaphore-based backoff for epoch table exhaustion

badrishc:
When hundreds of threads compete for epoch table entries, the previous spin-wait loop in `ReserveEntry` caused 100% CPU utilization due to tight spinning with `Thread.Yield()`.

Changes:
   - Add `SemaphoreSlim`-based wait mechanism for threads when epoch table is full
   - Split `ReserveEntry` into fast path (`TryAcquireEntry`) and slow path (`ReserveEntryWait`)
   - Fast path: probes `startOffset1`, `startOffset2`, then circles table twice - fully inlinable
   - Slow path: uses try/finally with semaphore wait - marked `NoInlining` since kernel wait dominates cost anyway
   - `Release()` signals one waiting thread via volatile `waiterCount` check (nearly zero overhead when no waiters)
   - Double-check pattern in `ReserveEntryWait` prevents lost wakeups: increment `waiterCount`, re-check for slots, then wait
   - `SemaphoreSlim` uses `Monitor.Pulse` internally which provides FIFO wake-up order, preventing starvation
 
LightEpoch Change:

We make `LightEpoch` keep a separate epoch table and thread allocation per instance instead of using a static thread index table. This ensure no deadlock in case we are holding one epoch (e.g. KV) while blocked on another (e.g. AOF) because the AOF epoch is guaranteed to progress as long as other AOF threads are able to find slots in their dedicated thread table.

Performance characteristics:
   - No contention: unchanged - fast path acquires entry with same probing logic
   - Table full: threads wait efficiently instead of burning CPU
   - Release hot path: single volatile read of `waiterCount` when no waiters

If lock contention in `SemaphoreSlim` becomes a problem, a possible optimization would be a batched signaling approach, i.e., we wait for some number of entries to be released before signaling that many number of waiters.


libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -185,16 +274,35 @@ public bool TrySuspend()
...
 
+            // Max epoch across all threads may have advanced, so check for pending drain actions to process
             if (drainCount > 0)
             {
                 Drain((*(tableAligned + entry)).localCurrentEpoch);
             }
+
+            if (waiterCount > 0)
+            {
+                SuspendResume();
```
TedHartMS, line 305: Are we sure this is safe for all callers?

TedHartMS (approved):

TedHartMS (approved):

TedHartMS (approved):

TedHartMS (approved):

PR #1568, opened 2026-02-18
Vector Set Test Hardening

kevin-montrose:
Since merge of the Vector Set preview, test pipelines have been flaky.

Having setup a separate VM for stressing outside of GH, this PR contains changes that may fix that flakiness:
 - Increasing AOF memory size to make IOPS exhaustion less likely
 - DRY up some common creation ops
 - Reduce noisy logging in some tests
 - Update unit test working directory creation to include arguments
 - Harden `TearDown` methods to do more of a reset even in the face of error so subsequent tests can still make progress
 - Modify `MigrateVectorStressAsync` test to be more reliable in the face of connection state lagging cluster state
 - Fix deadlock where disposal of `MigrateSession` could re-enter `MigrateSessionTaskStore` under a write lock
 - Move `RespServerSession` allocation onto `Task` during `VADD` replication
 - Fix a use after free bug in `ArgSliceVector` which could corrupt arbitrary key-value pairs during a migration
 - Assert `ArgSliceVector` is not being enumerated when modified, and that only a single enumeration occurs at once

### DIFF
--- a/libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
+++ b/libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Threading;
 using Garnet.common;
 using Garnet.server;
 using Microsoft.Extensions.Logging;
@@ -24,20 +23,25 @@ namespace Garnet.cluster
             this.logger = logger;
         }
 
+        /// <inheritdoc/>
         public void Dispose()
         {
-            try
+            _lock.WriteLock();
+            var skipDispose = _disposed;
+
+            _disposed = true;
+            _lock.WriteUnlock();
+
+            if (skipDispose)
             {
-                _lock.WriteLock();
-                _disposed = true;
-                for (var i = 0; i < sessions.Length; i++)
-                    sessions[i]?.Dispose();
-                Array.Clear(sessions);
+                return;
             }
-            finally
+
+            for (var i = 0; i < sessions.Length; i++)
             {
-                _lock.WriteUnlock();
+                sessions[i]?.Dispose();
             }
+            Array.Clear(sessions);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


libs/server/ArgSlice/ArgSliceVector.cs
```diff
@@ -71,8 +76,20 @@ public void Clear()
 
         public IEnumerator<SpanByte> GetEnumerator()
         {
-            foreach (var item in items)
-                yield return item;
+            var full = bufferManager.ViewFullArgSlice();
```
badrishc, line 88: Per the semantics of [ScratchBufferBuilder](https://github.com/microsoft/garnet/blob/main/libs/server/ArgSlice/ScratchBufferBuilder.cs), once you start returning ArgSlices to callers, you need to "seal" the data structure (before enumeration) to prevent any more allocations, in order to ensure validity of the ArgSlices returned by enumeration. It may be useful to enforce it (i.e., sealing) for the `ArgSliceVector` data structure.

The other option, as discussed offline, is to use [ScratchBufferAllocator](https://github.com/microsoft/garnet/blob/main/libs/server/ArgSlice/ScratchBufferAllocator.cs), which does not have such a restriction.

badrishc (approved):

libs/cluster/Server/Migration/MigrateSessionTaskStore.cs
```diff
@@ -23,20 +23,25 @@ public MigrateSessionTaskStore(ILogger logger = null)
...
-                    sessions[i]?.Dispose();
-                Array.Clear(sessions);
+                return;
             }
-            finally
+
+            for (var i = 0; i < sessions.Length; i++)
             {
-                _lock.WriteUnlock();
+                sessions[i]?.Dispose();
```
vazois: Add a try catch here to continue disposing all session if an exception is triggered.

vazois (approved):
Nice work!

libs/server/ArgSlice/ArgSliceVector.cs
kevin-montrose: I've added some asserts modification while enumerating and more-than-one-enumerator will fail in DEBUG builds.

vazois (approved):

PR #1570, opened 2026-02-20
Add filter support for VSIM vector search

hailangx:
Adds post-filter support for VSIM vector search results by introducing a JSON-attribute filter expression engine and integrating it into VectorManager after similarity search.

Changes:

- Introduces a filter module to evaluate over attributes  
- Integrates post-filtering into VectorManager for both value-based and element-based similarity search paths.
- Adds unit tests for the filter engine and RESP integration tests for VSIM ... FILTER ....
- Added a micro bench for the filter module to bench the memory allocation

Supported syntax documented at   Vector Filter Expressions (`VSIM ... FILTER`)
[website/docs/dev/vector-sets.md](https://github.com/microsoft/garnet/blob/58147041cf081dc823e55bba0d6e24fbf1cbfffb/website/docs/dev/vector-sets.md)
 
 Key design constraint: Zero heap allocation in the per-candidate evaluation loop. All buffers are borrowed from the session-local ScratchBufferBuilder (~9 KB), a pinned byte[] that persists for the session's lifetime. After the first VSIM FILTER query, the buffer is already large enough — subsequent calls have zero allocation cost and zero GC pressure.
 
 design doc [website/docs/dev/post_filter_design.md](https://github.com/microsoft/garnet/blob/58147041cf081dc823e55bba0d6e24fbf1cbfffb/website/docs/dev/post_filter_design.md)

Follow ups:

- Use SIMD json way to attract fields 
- Improve recall for the low selectivity case: integrate the beta filter search algo and skip copy the non-match results during the graph traverse with benchmark to see the outcome.


libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +910,108 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+        private static bool EvaluateFilter(Expr filterExpr, ReadOnlySpan<byte> attributeJson)
+        {
+            try
+            {
+                using var jsonDoc = JsonDocument.Parse(attributeJson.ToArray());
+                var root = jsonDoc.RootElement;
+                var result = VectorFilterEvaluator.EvaluateExpression(filterExpr, root);
+
+                return VectorFilterEvaluator.IsTruthy(result);
+            }
```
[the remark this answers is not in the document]
hailangx: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/garnet/pull/1570#discussion_r2830870604)

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -564,6 +566,19 @@ out var continuation
...
+            // Apply post-filtering if filter is specified
+            if (!filter.IsEmpty)
+            {
+                if (!includeAttributes)
+                {
+                    // Filters currently rely on attributes; reject inconsistent request instead of silently ignoring filter
+                    logger?.LogWarning("Filter expression was provided, but includeAttributes is false. Post-filtering requires attributes.");
+                    outputIdFormat = VectorIdFormat.Invalid;
+                    return VectorManagerResult.BadParams;
+                }
```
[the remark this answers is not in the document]
hailangx: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/garnet/pull/1570#discussion_r2830870590) , fetching attributes internally for filtering even when not returning them

harsha-simhadri:
CAn you link the specification of expression syntax you are implementing here

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
```diff
@@ -0,0 +1,123 @@
...
+                    && (tokens.Count == 0
+                        || tokens[^1].Type == TokenType.Operator
+                        || tokens[^1].Type == TokenType.Keyword
+                        || (tokens[^1].Type == TokenType.Delimiter && tokens[^1].Value == "("));
+
+                if (char.IsDigit(input[i]) || isNegativeNumber)
+                {
+                    var start = i;
+                    if (input[i] == '-') i++;
+                    while (i < input.Length && (char.IsDigit(input[i]) || input[i] == '.'))
```
tiagonapoli: nit: Is it ok to tokenize a number like `123.....23` (multiple dots?) or `1.2.3` - I guess in upper layer it will see that it's invalid right? Maybe worth having a test to confirm it will reject?

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
```diff
@@ -0,0 +1,123 @@
...
+                        i++;
+                    tokens.Add(new Token { Type = TokenType.Number, Value = input.Substring(start, i - start) });
```
tiagonapoli: @hailangx @kevin-montrose  Do we want to reduce allocations for the tokenizer? Like, instead of substring use a span and keep the input memory around? Or for this initial version a simpler implementaiton is ok?

hailangx: or is there any suggestion on reusing an existing one in the repo?

hailangx:

hailangx:
> CAn you link the specification of expression syntax you are implementing here
added into the vector set document

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
hailangx: Short answer: yes, it will rejected at parse/eval stage, but it gets into a number token.  malformed store attribute data is ignored per result, but malformed filter syntax errors the query.

Bad stored attribute JSON is handled in the per-item evaluator and returns false for that item, so that candidate is excluded, not fatal: VectorManager.cs:1035-1050
Missing fields also behave like non-match (evaluate to null/falsy), so those items are filtered out.
Bad filter expression text (for example malformed number token usage) is caught in post-filter parse/eval and rethrown as Invalid filter expression, which fails the VSIM request: VectorManager.cs:1025-1028

so, should we enforce a valid Json when set attribute?  if the answer is no, then during post processing we will just ignore items with invalid json.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+            ref SpanByteAndMemory outputIds,
+            ref SpanByteAndMemory outputDistances,
+            ref SpanByteAndMemory outputAttributes)
+        {
+            if (numResults == 0)
+            {
+                return numResults;
+            }
+
+            var filterStr = Encoding.UTF8.GetString(filter);
```
kevin-montrose: We definitely do not want to work in terms of strings here, that's some expensive validation plus an allocation in a hot path.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+
+            try
+            {
+                var filteredCount = 0;
+
+                // Parse the filter expression once, then evaluate per result
+                var tokens = VectorFilterTokenizer.Tokenize(filterStr);
```
kevin-montrose: Ideally we would be able to do this tokenize and parse in one pass - failing that the definitely shouldn't be allocating a new List on each run.  Tokenize into offsets and lengths, try and keep it on stack if we can and promote to a heap allocated array only when we exceed some reasonable maximum (512 bytes is probably fine there).

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+                var filterExpr = VectorFilterParser.ParseExpression(tokens, 0, out var endIndex);
+
+                // Ensure the entire token stream was consumed by the parser
+                if (endIndex != tokens.Count)
+                {
+                    throw new ArgumentException("Invalid filter expression: unexpected tokens after end of expression.", nameof(filter));
```
kevin-montrose: We cannot use exceptions as control flow, it's too expensive.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+                }
+
+                // Update lengths
+                outputIds.Length = idWritePos;
+                outputDistances.Length = distWritePos * sizeof(float);
+                outputAttributes.Length = attrWritePos;
+
+                return filteredCount;
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException || ex is InvalidOperationException)
```
kevin-montrose: See above comment, we cannot use exceptions as control flow - use returns.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+                var distancesSpan = MemoryMarshal.Cast<byte, float>(outputDistances.AsSpan());
+                var attributesSpan = outputAttributes.AsSpan();
+
+                var idReadPos = 0;
+                var attrReadPos = 0;
+                var idWritePos = 0;
+                var distWritePos = 0;
+                var attrWritePos = 0;
+
+                for (var i = 0; i < numResults; i++)
```
kevin-montrose: This will copy many id, attribute, and distance value if filters exclude anything - which we'd expect to be common.  For large results sets (or large attributes) that a bunch of work.

A better approach would be to have the `ValueSimilarity` and `ElementSimilarity` calls indicate a match count and a set of passing elements (probably actually a bitmap) and then let `NetworkVSIM` handle skipping elements.  Then we have no extra copying.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+        }
+
+        /// <summary>
+        /// Evaluate a pre-parsed filter expression against attribute data.
+        /// </summary>
+        private static bool EvaluateFilter(Expr filterExpr, ReadOnlySpan<byte> attributeJson)
+        {
+            try
+            {
+                var reader = new Utf8JsonReader(attributeJson);
```
kevin-montrose: Actually parsing the JSON is both too expensive, and subtlty wrong.

It's too expensive mostly because of allocations - but a lot of the work is wasted since filters only apply to top level fields.

It's wrong because Vector Set attributes don't have to be JSON - they are treated as non-matching.  This is expected behavior, so using exceptions to detect it is way too expensive.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -883,6 +933,123 @@ internal static uint CalculateValueDimensions(VectorValueType valueType, ReadOnl
...
+                using var jsonDoc = JsonDocument.ParseValue(ref reader);
+                var root = jsonDoc.RootElement;
+                var result = VectorFilterEvaluator.EvaluateExpression(filterExpr, root);
+
+                return VectorFilterEvaluator.IsTruthy(result);
+            }
+            catch (Exception ex) when (ex is JsonException or InvalidOperationException)
+            {
+                // If filter evaluation fails (malformed JSON or invalid expression), exclude the result
+                Trace.TraceWarning("Vector filter evaluation failed: {0}", ex);
```
kevin-montrose: As noted above, this is wrong as it will fail the search if any attribute is not JSON - when just that vector should be filtered out.

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        private const string OpEqualEqual = "==";
+        private const string OpBangEqual = "!=";
+        private const string OpGreaterEqual = ">=";
+        private const string OpLessEqual = "<=";
+        private const string OpAmpAmp = "&&";
+        private const string OpPipePipe = "||";
+        private const string OpStarStar = "**";
+
+        public static List<Token> Tokenize(string input)
```
kevin-montrose: A general note - leaning on `char` is kinda wrong here.  Because we want to match Redis behavior, we have to think hard about what it actually validates.

For example, using `char.IsWhiteSpace` will skip "&#x202F;" (U+202F), which Redis would error on.
Similarly, `char.IsDigit` will accept "٤" (U+0664) which should also be errored on.

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
```diff
@@ -0,0 +1,195 @@
...
+                    continue;
+                }
+
+                // Identifiers and keywords (field names starting with .)
+                if (input[i] == '.' || char.IsLetter(input[i]) || input[i] == '_')
+                {
+                    var start = i;
+                    while (i < input.Length && (char.IsLetterOrDigit(input[i]) || input[i] == '_' || input[i] == '.'))
+                        i++;
+                    var value = input.Substring(start, i - start);
```
kevin-montrose: I've noted elsewhere allocations are a problem with this approach, but this is a particularly scary one.  Even though elsewhere we try to reuse allocations for common keywords, we'll always allocated a string for each token.

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        public TokenType Type { get; }
+        public string Value { get; }
+    }
+
+    /// <summary>
+    /// Tokenizer for vector filter expressions.
+    /// Converts filter strings into tokens for parsing.
+    /// </summary>
+    internal static class VectorFilterTokenizer
```
kevin-montrose: As noted elsewhere, perhaps a one pass approach is possible and preferrable.

Something like a [SIMDJson approach](https://github.com/simdjson/simdjson?tab=readme-ov-file#about-simdjson) to produce addresses of delimiters is an attractive option if we still have to do a two pass tokenize-then-parse.

libs/server/Resp/Vector/Filter/VectorFilterExpression.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Runtime.InteropServices;
+using System.Text.Json;
+
+namespace Garnet.server.Vector.Filter
+{
+    /// <summary>
+    /// Discriminated union value type to eliminate boxing of doubles/strings
+    /// throughout the filter evaluation pipeline.
+    /// </summary>
+    [StructLayout(LayoutKind.Auto)]
```
kevin-montrose: nit: this is unnecessary

libs/server/Resp/Vector/Filter/VectorFilterParser.cs
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+
+            return ParsePrimary(tokens, start, out end);
+        }
+
+        private static Expr ParsePrimary(List<Token> tokens, int start, out int end)
+        {
+            if (start >= tokens.Count)
+                throw new InvalidOperationException("Unexpected end of expression");
```
kevin-montrose: Exceptions as control flow are too expensive, this should work in a `TryParse` style.

libs/server/Resp/Vector/Filter/VectorFilterParser.cs
```diff
@@ -0,0 +1,228 @@
...
+
+namespace Garnet.server.Vector.Filter
+{
+    /// <summary>
+    /// Recursive descent parser for vector filter expressions.
+    /// Supports arithmetic, comparison, logical operators, containment, and grouping.
+    /// </summary>
+    internal static class VectorFilterParser
+    {
+        public static Expr ParseExpression(List<Token> tokens, int start, out int end)
```
kevin-montrose: If I'm reading this correctly, it's extremely recursive - which is very scary.  Is a stack overflow possible with reasonable filters here?  And how complex a filter can we test with before something breaks?

One complication is that Windows and Linux tend to use different default stack sizes - we mostly dev on Windows, but deployments are more commonly Linux.

libs/server/Resp/Vector/Filter/VectorFilterEvaluator.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// Evaluates parsed expression trees against JSON attribute data.
+    /// Returns FilterValue (a struct) to avoid boxing allocations on every evaluation.
+    /// </summary>
+    internal static class VectorFilterEvaluator
+    {
+        /// <summary>
+        /// Evaluate a filter expression against a JSON element.
+        /// Returns a FilterValue (struct) — no boxing occurs for numeric results.
+        /// </summary>
+        public static FilterValue EvaluateExpression(Expr expr, JsonElement root)
```
kevin-montrose: It's weird to me to return something other than `bool` here.

libs/server/Resp/Vector/Filter/VectorFilterEvaluator.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Text.Json;
+
+namespace Garnet.server.Vector.Filter
+{
+    /// <summary>
+    /// Evaluator for vector filter expressions.
+    /// Evaluates parsed expression trees against JSON attribute data.
+    /// Returns FilterValue (a struct) to avoid boxing allocations on every evaluation.
+    /// </summary>
+    internal static class VectorFilterEvaluator
```
kevin-montrose: As a general note, this feels overly complicated.

We're taking a filter that we've parsed _and_ a whole JSON document.  But the filter only applies over top level elements of that document... so really, this is operating over the filter and a dictionary (a dictionary that contains no other dictionaries at that).

I've noted elsewhere we should remove most of these allocations - a natural-ish approach would be to a filter, the attribute span, and a collection of (length, offset) pairs to top level attributes.

kevin-montrose (dismissed):
I think some fundamental reworking is needed here, exceptions and allocations need to go - I've left a bunch of guideline comments to get us pointed in the right direction.  It's not quite an exhaustive review - there are minor optimizations and style things we can revisit latter when we're closer to mergeable.

libs/server/Resp/Vector/Filter/VectorFilterTokenizer.cs
hailangx: refactor it to one pass postfix (reverse-Polish) design which  redis is using

libs/server/Resp/Vector/Filter/VectorFilterParser.cs
hailangx: refactor it to one pass postfix (reverse-Polish) design which redis is using, so only a bound stack. will be used.

libs/server/Resp/Vector/Filter/VectorFilterEvaluator.cs
hailangx: Thanks for the review and suggestion, I have move away from the whole json paring and adapt what redis is doing to match the behavior

libs/server/Resp/Vector/VectorManager.cs
hailangx: move away from it

libs/server/Resp/Vector/VectorManager.cs
hailangx: refactored to a one passing method

libs/server/Resp/Vector/VectorManager.cs
hailangx: can we defer this to next round optimization as we need to have the bitmap optimization when index is there and even with the index, we need some routing to fallback to pure post processing or prefiltering given the query's selectivity.

libs/server/Resp/Vector/VectorManager.cs
hailangx: fixed.

kevin-montrose:
For future reference, a slightly cleaned up version of the plan we discussed offline:

---

`FILTER` is structured kind of like a pipeline:

 1. Take in `FILTER`, extract relevant fields
 2. After "normal" part of search, we have all the attribute data in a contiguous blob (that's what [FetchVectorElementAttributes](https://github.com/microsoft/garnet/blob/280539371a3d445551bfd3b0714ad293ef4d3dbc/libs/server/Resp/Vector/VectorManager.cs#L711) produces)
    * [Matching Redis](https://redis.io/docs/latest/develop/data-types/vector-sets/filtered-search/#filter-ef), we over select when we're going to post-filter (using the `FILTER-EF` if specified, and `COUNT * 100` fi not)
 3. We "extract" the whole blob to produce a set of `(count fields): [(field offset) (field length)] ...` data; we only include fields we found to care about in #1, and malformed attributes produce some special count (presumably `-1`)
    * We can right-size the buffer here because we know each attribute has AT MOST one instance each of the relevant fields
    * In the case the same field is specified multiple times (and thus we overflow) we can ignore the attribute
    * This part is amenable to SimdJSON-esque parsing, where we track delimiters in bitmaps and process many characters at once rather than going char-by-char like Redis does
 4. We apply the filter built in #1 against the result of #3, producing a bitmap of matching elements
 5. We use that bitmap in `RespServerSessionVectors` to write the final results out
    * Importantly, we don't compact or otherwise move element or attribute data once it lands in the result buffers 

While generally we do not want early-exits in this code (batching and SIMD want very little control flow to be beneficial), it makes sense to check after every `COUNT` elements have been compared against the filter to see if we're done.

---

Minor thoughts after a quick glance before vacaction:
 - Probably not worth putting FILTER code in a separate folder / namespace; put it all int eh same level as VectorManager, and hide any FILTER specific types/code in a new VectorManager partial
 - Avoid static constructors, they have gnarly runtime implications
 - Probably want a union-ish thing for tokens rather than a fat structure, and don't want any cases where we allocate strings or arrays - structure so that stuff can be punned spans
 - Similar for extracted fields, we can work in terms of offsets into the input FILTER rather than allocating strings

harsha-simhadri:
Haiyang, let us plan to paginate diskann when the first round of post processing does not generate enough results, to mitigate low recall for selective predicates.

libs/server/Resp/Vector/VectorManager.Filter.cs
```diff
@@ -0,0 +1,218 @@
...
+                var instrBuf = span.Slice(offset, MaxInstructions); offset += MaxInstructions;
+                var tuplePoolBuf = span.Slice(offset, MaxTuplePool); offset += MaxTuplePool;
+                var tokensBuf = span.Slice(offset, MaxInstructions); offset += MaxInstructions;
+                var opsStackBuf = span.Slice(offset, MaxInstructions); offset += MaxInstructions;
+                var runtimePoolBuf = span.Slice(offset, MaxRuntimePool); offset += MaxRuntimePool;
+                var extractedFields = span.Slice(offset, MaxSelectors); offset += MaxSelectors;
+                var stackBuf = span.Slice(offset, StackCapacity);
+
+                // ── Compile ────────────────────────────────────────────────
+                var instrCount = ExprCompiler.TryCompile(filter, instrBuf, tuplePoolBuf, tokensBuf, opsStackBuf, out var tupleCount, out _);
```
badrishc, line 130: isnt TryCompile expensive? why not cache the compilation result for a filter and reuse for subsequent calls.

hailangx: it's reused for different candidates, but it can only be reused for subsequent calls when they share the same filters expression, we could optimize it with a proper cache for same filter expression.

hailangx: One-time cost per VSIM query. Compiles filter string to postfix program.

| Filter expression | .NET 8 | .NET 10 | Allocated |
|-------------------|-------:|--------:|----------:|
| `.year > N` (comparison) | 248 ns | 284 ns | - |
| `.year > N and .rating >= M` (logical AND) | 517 ns | 593 ns | - |
| `.genre == "action"` (string equality) | 256 ns | 281 ns | - |
| `(.year - 2000) ** 2 < 100` (arithmetic + power) | 615 ns | 735 ns | - |
| `"classic" in .tags` (containment) | 235 ns | 266 ns | - |
| Combined (all ops) | 1,151 ns | 1,332 ns | - |

hailangx: I think we can avoid cache for now ?  it's not the biggest concern, as I have some bench on the SIMDJson, I think it's much worth to improve, but I would defer it to a separate PR

*ShortRun, AMD EPYC 7763 (Zen 3, AVX2), .NET 8.0:*

### Number·Small (~26 B, first field)

| Method | Mean | Ratio | Allocated |
|---|---|---|---|
| **SimdOnDemand** | **58.5 ns** | **0.45×** | **0 B** |
| SimdJsonNative | 99.8 ns | 0.76× | 0 B |
| SimdTwoPass | 103.7 ns | 0.79× | 0 B |
| Current | 131.2 ns | 1.00× | 0 B |
| SimdJson | 6,583 ns | 50.16× | 48 B |

### String·Medium (~100 B, third field)

| Method | Mean | Ratio | Allocated |
|---|---|---|---|
| **SimdOnDemand** | **61.5 ns** | **0.23×** | **0 B** |
| SimdTwoPass | ~105 ns | ~0.39× | 0 B |
| SimdJsonNative | 129.4 ns | 0.48× | 40 B |
| Current | 271.1 ns | 1.00× | 0 B |
| SimdJson | 9,757 ns | 36.00× | 88 B |

badrishc (approved):

PR #1573, opened 2026-02-20
bump  diskann-garnet 1.0.23 with crash fix

hailangx:
bumps the diskann-garnet package from version 1.0.20 to 1.0.23 and includes a crash fix along with changes to default vector set parameters. 

The crash fix adds validation to prevent using an invalid index when CreateIndex fails, while the default parameter changes update the quantization type from Q8 to NoQuant and the distance metric from L2 to Cosine.


kevin-montrose:
`main` has been updated to .23 by other work, superseding this PR.

PR #1578, opened 2026-02-26
Cherry-Picking commits from main into dev

TalZaccai:
This PR contains 10 cherry-picked commits from main that don't include latest vector-related work.
Included commits shown below.


libs/server/GarnetDatabase.cs
```diff
@@ -166,6 +166,7 @@ public void Dispose()
                         Thread.Yield();
                 }
             }
+            disposed = true;
```
badrishc: this change is not needed as dev sets disposed to true earlier on line 146

badrishc (changes requested):

badrishc (approved):

libs/client/GarnetClient.cs
```diff
@@ -151,6 +153,7 @@ public GarnetClient(
             bool recordLatency = false,
             bool useTimeoutChecker = true,
             int networkSendThrottleMax = 8,
+            LightEpoch epoch = null,
```
vazois, line 157: nit; add epoch parameter to method summary.

vazois (approved):

TedHartMS (approved):

PR #1582, opened 2026-02-27
TKey and an interface for key bytes and namespaces

kevin-montrose:
- [x] Plumb TKey everywhere
- [x] .NET 10 and .NET 8 implementations
- [x] Test passing parity(?)
- [ ] Add namespaces to IKey, wire up TSourceLogRecord accordingly
- [ ] Tests with actually setting namespaces


libs/common/SpanByteKey.cs
```diff
@@ -0,0 +1,88 @@
...
+            , allows ref struct
+#endif
+            => KeyBytes.SequenceEqual(other.KeyBytes);
+
+        /// <summary>
+        /// Explicit conversion from <see cref="ReadOnlySpan{T}"/> of bytes.
+        /// 
+        /// TODO: We could maybe obviate this with a set of extension methods on Tsavorite context?
+        /// </summary>
+        public static explicit operator SpanByteKey(ReadOnlySpan<byte> key) => new(key);
```
badrishc, line 86: this would be correct only if we were told that the given ReadOnlySpan key is already pinned by caller.

libs/common/SpanByteKey.cs
```diff
@@ -0,0 +1,88 @@
...
+                    return new(ptr, len);
+                }
+#endif
+            }
+        }
+
+        /// <summary>
+        /// Create a new SpanByteKey wrapping the given bytes.
+        /// </summary>
+        private SpanByteKey(ReadOnlySpan<byte> key)
```
badrishc, line 50: Maybe rename to pinnedKey to make it clear it needs to be pinned.

PR #1610, opened 2026-03-05
Harden Offset Validation for Bitmap Operations

vazois:
This PR add out-of-bound validation checks for offset parameter used in bitmap operations.

### DIFF
--- a/libs/server/Resp/Bitmap/BitmapCommands.cs
+++ b/libs/server/Resp/Bitmap/BitmapCommands.cs
@@ -205,7 +205,7 @@ namespace Garnet.server
             where TGarnetApi : IGarnetApi
         {
             var count = parseState.Count;
-            if (count < 1 || count > 4)
+            if (count is not 1 and not 3 and not 4)
             {
                 return AbortWithWrongNumberOfArguments(nameof(RespCommand.BITCOUNT));
             }
@@ -222,20 +222,17 @@ namespace Garnet.server
                     return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
                 }
 
-                if (count > 2)
+                if (!parseState.TryGetLong(2, out _))
                 {
-                    if (!parseState.TryGetLong(2, out _))
-                    {
-                        return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
-                    }
+                    return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
+                }
 
-                    if (count > 3)
+                if (count > 3)
+                {
+                    var spanOffsetType = parseState.GetArgSliceByRef(3).ReadOnlySpan;
+                    if (!spanOffsetType.EqualsUpperCaseSpanIgnoringCase("BIT"u8) && !spanOffsetType.EqualsUpperCaseSpanIgnoringCase("BYTE"u8))
                     {
-                        var spanOffsetType = parseState.GetArgSliceByRef(3).ReadOnlySpan;
-                        if (!spanOffsetType.EqualsUpperCaseSpanIgnoringCase("BIT"u8) && !spanOffsetType.EqualsUpperCaseSpanIgnoringCase("BYTE"u8))
-                        {
-                            return AbortWithErrorMessage(CmdStrings.RESP_SYNTAX_ERROR);
-                        }
+                        return AbortWithErrorMessage(CmdStrings.RESP_SYNTAX_ERROR);
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -138,7 +138,7 @@ private bool NetworkStringSetBit<TGarnetApi>(ref TGarnetApi storageApi)
             var sbKey = parseState.GetArgSliceByRef(0).SpanByte;
 
             // Validate offset
-            if (!parseState.TryGetLong(1, out var offset) || (offset < 0))
+            if (!parseState.TryGetLong(1, out var offset) || (offset < 0) || !BitmapManager.IsValidBitOffset(offset))
```
TalZaccai, line 141: Perhaps we need a different error message for `!IsValidBitOffset`? Might be confusing to the user. What does Redis return?

libs/server/Resp/Bitmap/BitmapCommands.cs
```diff
@@ -205,29 +205,41 @@ private bool NetworkStringBitCount<TGarnetApi>(ref TGarnetApi storageApi)
...
+            // start, end, [BIT|BYTE]
+            if (count > 1)
             {
-                if (!parseState.TryGetInt(1, out _) || (parseState.Count > 2 && !parseState.TryGetInt(2, out _)))
+                if (!parseState.TryGetLong(1, out _))
                 {
                     return AbortWithErrorMessage(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
                 }
+
+                if (!parseState.TryGetLong(2, out _))
```
TalZaccai: Why not `||` this with the previous `if`?

libs/server/Resp/Bitmap/BitmapManagerBitfield.cs
```diff
@@ -56,7 +56,9 @@ public static int LengthFromType(BitFieldCmdArgs args)
         {
             var offset = args.offset;
             var bitCount = (byte)(args.typeInfo & 0x7F);
-            return LengthInBytes(offset + bitCount - 1);
+            if (!TryValidateBitfieldOffset(offset, bitCount, multiplyOffset: false, out _, out var endOffset))
```
TalZaccai: Shouldn't it already be validated at this point? Are we duplicating this logic?

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
```diff
@@ -150,20 +150,24 @@ void CopyRespToWithInput(ref RawStringInput input, ref SpanByte value, ref SpanB
...
 
                 case RespCommand.BITCOUNT:
-                    var bcStartOffset = 0;
-                    var bcEndOffset = -1;
+                    long bcStartOffset = 0;
+                    long bcEndOffset = -1;
                     byte bcOffsetType = 0x0;
 
-                    if (input.parseState.Count > 1)
+                    if (input.parseState.Count > 0 && input.parseState.TryGetLong(0, out var parsedBcStartOffset))
```
TalZaccai: You can use GetLong in here, as we already validated the input at this point

libs/server/Storage/Functions/MainStore/PrivateMethods.cs
```diff
@@ -150,20 +150,24 @@ void CopyRespToWithInput(ref RawStringInput input, ref SpanByte value, ref SpanB
...
-                        }
+                    if (input.parseState.Count > 1 && input.parseState.TryGetLong(1, out var parsedBcEndOffset))
+                    {
+                        bcEndOffset = parsedBcEndOffset;
+                    }
+
+                    if (input.parseState.Count > 2)
+                    {
+                        var spanOffsetType = input.parseState.GetArgSliceByRef(2).ReadOnlySpan;
+                        bcOffsetType = spanOffsetType.EqualsUpperCaseSpanIgnoringCase("BIT"u8) ? (byte)0x1 : (byte)0x0;
```
TalZaccai: Might be worthwhile to encode thsi in the input's `arg1` if it's not being used

TalZaccai (changes requested):

libs/server/Resp/Bitmap/BitmapCommands.cs
vazois: good point! I checked and it seems that its the same message.

image

libs/server/Resp/Bitmap/BitmapCommands.cs
vazois: Fixed!

libs/server/Resp/Bitmap/BitmapManagerBitfield.cs
vazois: I made this Debug.Assert just to prevent improper usage if BitmapManagerBitfield.cs methods are called directly for some reason.

TalZaccai (approved):

PR #1613, opened 2026-03-06
Add Range Index data type based on Bf-Tree (single instance)

badrishc:
### Description of Change

Adds Range Index as a new Garnet data type, backed by [Bf-Tree](https://github.com/microsoft/bf-tree) — a high-performance B-tree for ordered key-value storage with range scan support. Gated behind `--enable-range-index-preview`.

   **Commands implemented (9):**
   - `RI.CREATE` — Create index with DISK or MEMORY backend
   - `RI.SET` / `RI.GET` / `RI.DEL` — Point operations on entries
   - `RI.SCAN` / `RI.RANGE` — Range queries (DISK backend only)
   - `RI.EXISTS` / `RI.CONFIG` / `RI.METRICS` — Utility commands
   - `TYPE` returns `"rangeindex"` for RI keys
   - `DEL` / `UNLINK` frees the underlying BfTree

   **Lifecycle infrastructure (Tsavorite IRecordTriggers):**
   - `OnFlush` — Snapshot BfTree to flush.bftree, set FlagFlushed for lazy promote
   - `OnEvict` — Free BfTree under exclusive lock
   - `OnDiskRead` — Zero stale TreeHandle
   - `OnCheckpoint` — VersionShift (barrier), FlushBegin (snapshot), CheckpointCompleted (cleanup)
   - `OnRecovery` / `OnRecoverySnapshotRead` — Set recovered token / FlagRecovered

   **Checkpoint consistency:**
   - Trees with `SnapshotPending=1` at barrier time are snapshotted; v+1 trees are skipped
   - Concurrent restores serialized via exclusive lock
   - Old checkpoint snapshots purged at CheckpointCompleted (gated on removeOutdated)
   - RIPROMOTE clears source handle in PostCopyUpdater (after CAS success, not before)
   - Restored trees use data.bftree as working file (not snapshot artifacts)

   **AOF logging:**
   - RI.SET/RI.DEL logged via synthetic no-op RMW (VectorManager pattern)
   - RI.CREATE logged with stub bytes; replay creates fresh BfTree
   - AOF-only recovery (no checkpoint) supported

   **WRONGTYPE safety:**
   - Read, RMW, and Upsert paths block cross-type access
   - GET/SET on RI keys → WRONGTYPE error
   - RI commands on non-RI keys → WRONGTYPE error
   - UpsertAction.WrongType added to Tsavorite

   **55 tests** covering CRUD, scan, range, lifecycle (flush/evict/promote/restore), checkpoint recovery, AOF replay, AOF-only recovery, WRONGTYPE,
  concurrent stress (4 threads + blocking SAVE with strict prefix verification), and utility commands.

   **Documentation:**
   - Dedicated Range Index commands page (`website/docs/commands/range-index.md`)
   - API compatibility listing
   - Design doc updated with implementation status

   ### Not in this PR (deferred)
   - Batch commands (RI.MSET/MGET/MDEL)
   - Cluster replication/migration
   - Transaction (MULTI/EXEC) support
   - Memory-only tree snapshot (blocked on upstream bf-tree)


website/docs/dev/range-index-resp-api.md
```diff
@@ -0,0 +1,2323 @@
...
+All commands follow the `RI.*` prefix convention (short for **R**ange**I**ndex).
+
+---
+
+### 1. Lifecycle / Management Commands
+
+| Command | Syntax | Description | Maps to |
+|---|---|---|---|
+| **RI.CREATE** | `RI.CREATE key [options...]` | Create a new RangeIndex. Options allow tuning the underlying BfTree config | `BfTree::new()` / `BfTree::with_config()` |
+| **RI.DROP** | `RI.DROP key` | Destroy a RangeIndex and free its resources | `drop(BfTree)` |
```
kevin-montrose: Having a separate delete operation is a little weird - not unprecedented (FT.* has a custom delete), but worth thinking about.

Some of this stuff is still in the air for Vector Sets too, so there's an opportunity to unify a bit.

To my eye we need to account for:
 - explicit deletes (`DEL`, `DELEX`, 'UNLINK', `RI.DROP`, etc.)
 - implicit deletes (key eviction, including sub-key data like Vector Set elements)
 - dropping the whole DB (`FLUSHDB`)

Today Vector Sets allow `DEL`/`UNLINK` - this is done in a two phase style where first deletion is cancelled, then the Rust bits are dropped, then the index key is deleted and element data is scheduled for cleanup.  This feels relatively clean, and will be generalizable for different record types.

`FLUSHDB` isn't handled, but I know how we would - enumerate all active contexts and drop the rust bits, then let `FLUSHDB` proceed as normal.  For Vector Sets we "know" all active Vector Sets so this is tractable, would that be true for Range Indexes?

Eviction is tricky for both.  I think for Range Index an eviction needs to trigger cleanup, whereas for Vector Sets we probably need to just... block eviction?  Or at least, block eviction of element data, something we could infer from namespaces.  That namespaces (and/or record types) are necessary infra is one of the reasons this has been punted.

---

To bring that make to this specific plan:
 - Should we eliminate `RI.DROP` and lean on `DEL/UNLINK/etc.`?
 - How does `FLUSHDB` work?  In particular, how do cleanup the rust bits (on disk seems easy enough)?
 - For eviction/expiration/etc. is there a common design we can adopt for Vector Sets and Range Indexes?

website/docs/dev/range-index-resp-api.md
```diff
@@ -0,0 +1,2323 @@
...
+{
+    Span<byte> hash = stackalloc byte[16];
+    System.IO.Hashing.XxHash128.Hash(keyBytes, hash);
+    return new Guid(hash);
+}
+```
+
+---
+
+### A. Page Flush (Hybrid Log Eviction)
```
kevin-montrose: This will be useful infra once Vector Sets are not bound by the number of namespaces we can fit in a single byte.

website/docs/dev/range-index-resp-api.md
```diff
@@ -0,0 +1,2323 @@
...
+
+---
+
+### C. Recovery
+
+> **Reference:** `libs/server/Databases/SingleDatabaseManager.cs` — `RecoverCheckpoint()`
+> restores the store from a checkpoint. After Tsavorite recovery completes,
+> `RangeIndexManager.ResumePostRecovery()` should be called to restore BfTrees.
+
+**Post-recovery scan:** After Tsavorite recovery loads all records from the checkpoint,
```
kevin-montrose: This implies scanning the whole log and initializing all range indexes on process restart right?  That feels like that scan could be kinda expensive for very large logs.

Vector Sets defer recreation until first use, but then again Vector Set recreation is very cheap - is it important to do the disk bits of bftree at process start, or is a similar lazy recreation tractable?

kevin-montrose (commented):
Generally looks good to me - couple questions to poke at, but no blockers that I see.

website/docs/dev/range-index-resp-api.md
badrishc: The question is whether it is feasible to assume that we keep all RangeIndex instances in memory (by memory here, I mean just the memory portion of the whole b-tree). If these were used like SortedSets, there could be too many instances of these (many dormant only on disk). If we could keep all in memory, that makes the design simpler (we would just need to pin them in memory to prevent escaping to disk/null device, e.g., the stub gets copied to tail when it reaches the read-only region).

This document currently assumes we cannot afford to keep all RangeIndex instances in memory.

* Explicit delete (DEL/DROP) for RangeIndex could be modeled as an application-level lock, then `Read` followed by `Delete`. Or if made native, then `RMW` (with `NeedCopyUpdate` false and `InPlaceUpdate` setting the tombstone bit). This lets us retrieve the stub from disk in order to delete the disk files.

* FLUSHDB would probably need to find all active in-memory RangeIndex instances (perhaps with a scan of in-memory) and purge the disk folder.
 
Eviction triggers the cleanup of all in-memory state for the instance being evicted. If the stub tiers down to real storage, we have to snapshot the corresponding RangeIndex to disk and then free it (and the tree goes dormant on disk). If we are backed by a null device, we can simply free the RangeIndex and delete its files from disk.

website/docs/dev/range-index-resp-api.md
badrishc: Yeah this needs to be lazy, i will fix.

website/docs/dev/range-index-resp-api.md
badrishc: agreed. this will help streamline eviction and rehydration of individual object instances (where object is either vector set or range index).

libs/native/bftree-garnet/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "stable"
```
tiagonapoli: nit: thoughts on using a hardcoded version like diskann repo does? 
https://github.com/microsoft/DiskANN/blob/main/rust-toolchain.toml

That way we can have reproducibility between builds/releases that don't change rust version

.azure/pipelines/azure-pipelines-external-release.yml
```diff
@@ -23,9 +23,28 @@ pool:
...
+  displayName: Build bftree native (Linux)
+  pool:
+    vmImage: 'ubuntu-latest'
+  steps:
+  - checkout: self
+    clean: False
+  - bash: |
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+      source $HOME/.cargo/env
+      cargo build --release --manifest-path libs/native/bftree-garnet/Cargo.toml
```
tiagonapoli: nit: should we add `--locked` so `Cargo.lock` file is used and then we have more reproducible builds?

libs/native/bftree-garnet/src/lib.rs
```diff
@@ -0,0 +1,447 @@
...
+/// own data file.
+///
+/// Returns 0 on success, -1 on failure.
+///
+/// # Safety
+/// `tree` must be a valid BfTree pointer for a disk-backed tree.
+#[no_mangle]
+pub unsafe extern "C" fn bftree_snapshot(tree: *mut BfTree) -> i32 {
+    let tree = &*tree;
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
tiagonapoli, line 348: Any concerns that a panic could leave a particular bftree's state invalid after it tried to do a snapshot and then other operations on that bftree would fail or silently continue operating on an invalid bftree structure?

tiagonapoli, line 348: nit: maybe for follow up work we could implement panic hooks to ensure we log panic backtraces?
DiskANN has it for CosmosDB integration, something like this:
```

/// Pass formatted panic hooks to `tracing.
///
/// This uses the default formatting available in both `PanicHookInfo` and
/// `std::backtrace::Backtrace`.
fn panic_hook(info: &PanicHookInfo<'_>) {
    error!("{}\n{}", info, Backtrace::force_capture());
}

pub(crate) fn set_tracing_panic_hook() {
    // Take the current panic hook.
    let current = take_hook();
    let next = move |info: &PanicHookInfo<'_>| {
        panic_hook(info);
        current(info);
    };
    set_hook(Box::new(next));
}

static INITIALIZED_PANIC_HOOK: Once = Once::new();

/// Initialize tokio runtime and return a handle to it
///
/// This function also establishes a panic hook that logs panics.
#[no_mangle]
pub extern "C" fn initialize_rust_runtime(threadcount: u32) -> DiskANNResult {
    let result = Builder::new_multi_thread()
        .worker_threads(threadcount as usize)
        .build();

    let mut diskannresult: DiskANNResult = DiskANNResult::default();
    match result {
        Ok(resultp) => {
            // If we are successful, install the panic hook.
            INITIALIZED_PANIC_HOOK.call_once(panic::set_tracing_panic_hook);
            diskannresult.result = Box::into_raw(Box::new(resultp)) as *const c_void;
        }
        Err(error) => {
            let error = ANNError::log_async_runtime_error(error.to_string());
            diskannresult.set_error(&error);
        }
    };
    diskannresult
}

File: libs/native/bftree-garnet/BfTreeService.cs [hunk 7]
```diff
@@ -0,0 +1,658 @@
...
+
+            if (_storageBackend == StorageBackendType.Disk)
+            {
+                // Snapshot in-place first (drains circular buffer to data file)
+                int result = NativeBfTreeMethods.bftree_snapshot(_tree);
+                if (result != 0)
+                    throw new InvalidOperationException("Failed to snapshot disk-backed BfTree before file copy.");
+
+                // Copy the data file to the target path
+                System.IO.File.Copy(_filePath, targetPath, overwrite: true);
```
tiagonapoli, line 523: Maybe for better durability guarantees should we consider the following?
1. Looking at File.Copy in dotnet runtime it doesn't ensure fsync to flush OS disk buffers - Should we manully flush here?
2. Any concerns if there's a process exit while copying and leaving the file in an invalid state? Should we copy to a tmp file and then rename it once the full copy is complete instead?

libs/native/bftree-garnet/BfTreeService.cs
```diff
@@ -0,0 +1,658 @@
...
+        private static List<ScanRecord> DrainScanIteratorToList(nint handle, ScanReturnField returnField)
+        {
+            var results = new List<ScanRecord>();
+            Span<byte> buffer = stackalloc byte[8192];
+            while (true)
+            {
+                int keyLen = 0, valueLen = 0;
+                int hasNext;
+                fixed (byte* bp = buffer)
+                    hasNext = NativeBfTreeMethods.bftree_scan_next(
```
tiagonapoli, line 626: For my understanding, what happens if the buffer provided to `bftree_scan_next` is not enough for the key/value? Will it panic?

libs/native/bftree-garnet/BfTreeService.cs
```diff
@@ -0,0 +1,658 @@
...
+                };
+                results.Add(record);
+            }
+            return results;
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            if (!_disposed)
```
tiagonapoli: super nit: Just to be safe should we InterlockedExchange `_disposed` and ensure this always is disposed only once? Since we already have the `if (!_disposed)`, we could do ensure that this works in multi-threaded scenarios as well

libs/native/bftree-garnet/rust-toolchain.toml
badrishc: hardcoded to 1.94 in latest commit

.azure/pipelines/azure-pipelines-external-release.yml
badrishc: done

libs/native/bftree-garnet/src/lib.rs
badrishc: Agreed. Bf-tree snapshot is being reworked so we can consider this after that is done, or separately.

badrishc: Good idea for follow up, thanks!

libs/native/bftree-garnet/BfTreeService.cs
badrishc: done

libs/native/bftree-garnet/BfTreeService.cs
badrishc: file copy is a bit flaky, bf-tree will soon have direct snapshot to specified file

libs/native/bftree-garnet/BfTreeService.cs
badrishc: not sure, but good question. have to try and fix as needed, separately.

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -0,0 +1,614 @@
...
+        /// checkpoint recovery), the RMW returns <c>InPlaceUpdated</c> and the fresh
+        /// BfTree is disposed.
+        /// </remarks>
+        /// <param name="session">The storage session for issuing the RMW.</param>
+        /// <param name="key">The Garnet key being created.</param>
+        /// <param name="input">The RMW input containing the stub bytes in parseState.</param>
+        internal unsafe void HandleRangeIndexCreateReplay(StorageSession session, ReadOnlySpan<byte> key, ref StringInput input)
+        {
+            var stubSpan = input.parseState.GetArgSliceByRef(0).Span;
+            if (stubSpan.Length < IndexSizeBytes)
```
tiagonapoli: nit: are there legitimate code paths where this could happen? Should we instead log an error and assert/throw?

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -0,0 +1,614 @@
...
+            try
+            {
+                bfTree = CreateBfTree(
+                    (StorageBackendType)stub.StorageBackend, key,
+                    stub.CacheSize, stub.MinRecordSize, stub.MaxRecordSize,
+                    stub.MaxKeyLen, stub.LeafPageSize);
+            }
+            catch
+            {
+                return;
```
tiagonapoli, line 539: nit: Should we log error?

libs/server/Resp/RangeIndex/RangeIndexManager.Index.cs
```diff
@@ -0,0 +1,255 @@
...
+            stub.TreeHandle = nint.Zero;
+        }
+
+        /// <summary>
+        /// Dispose the BfTree instance referenced by a stub, if the pointer is valid
+        /// and registered with this manager. Called during delete and page eviction.
+        /// If the tree handle is zero (already freed or never set), this is a no-op.
+        /// </summary>
+        /// <param name="valueSpan">The store value span containing the stub.</param>
+        internal void DisposeTreeIfOwned(ReadOnlySpan<byte> valueSpan)
```
tiagonapoli: nit: looks like this method is unused and could be removed (`DisposeTreeUnderLock` should be the one used right?)

Also, if this method `DisposeTreeIfOwned` is removed, `UnregisterIndex` will be called only by `DisposeTreeUnderLock` and then we could move the code there instead for simplicity, instead of the indirect requirement on the method comment?
```
        /// Unregister and dispose a BfTreeService. The caller must already hold
        /// an exclusive lock for the corresponding key hash.
```

libs/server/Resp/RangeIndex/RangeIndexManager.Index.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        /// <summary>
+        /// Set the Flushed flag on the in-memory stub. Called from <see cref="GarnetRecordTriggers.OnFlush"/>
+        /// on the original (not copied) record, so the next data operation detects the flag
+        /// and promotes the stub to the mutable tail region via <see cref="PromoteToTail"/>.
+        /// </summary>
+        /// <param name="valueSpan">The store value span containing the stub (writable even though declared ReadOnlySpan, via Unsafe.As).</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void SetFlushedFlag(ReadOnlySpan<byte> valueSpan)
```
tiagonapoli: nit: looking at the callsites looks like we could use `Span<byte>` here instead of `ReadOnlySpan<byte>`

libs/server/Resp/RangeIndex/RangeIndexManager.Locking.cs
```diff
@@ -0,0 +1,345 @@
...
+                return default;
+            }
+
+            var outputSpan = output.SpanByteAndMemory.IsSpanByte
+                ? output.SpanByteAndMemory.SpanByte.ReadOnlySpan
+                : output.SpanByteAndMemory.MemorySpan;
+
+            if (outputSpan.Length < IndexSizeBytes)
+            {
+                status = GarnetStatus.NOTFOUND;
```
tiagonapoli: nit: should this be an assert or throw instead? That would be highly unexpected, right?

libs/server/Resp/RangeIndex/RangeIndexManager.cs
badrishc: Fixed

PR #1631, opened 2026-03-16
Fix dispose hang in network handler and buffer pool cleanup

vazois:
## Problem

`ClusterResetHardDuringDisklessReplicationAttach` test times out during TearDown because `LimitedFixedBufferPool.Dispose()` hangs indefinitely.

### 1. `TcpNetworkHandlerBase.Dispose()` missing `DisposeImpl()` call

The public `Dispose()` override closes the socket but never calls `DisposeImpl()`. When a handler thread is blocked synchronously (e.g., in `TryBeginDisklessSync`), no SAEA completion callback fires, so:
- The CTS is never cancelled (blocked thread never wakes up)
- `DisposeMessageConsumer` is never called (`activeHandlerCount` never decrements)
- `DisposeActiveHandlers()` spins forever

**Fix:** Added `DisposeImpl()` call at the end of public `Dispose()`. This is safe because the existing `disposeCount` Interlocked guard prevents double execution.

### 2. `GarnetTcpNetworkSender.DisposeNetworkSender()` leaks `responseObject`

`DisposeNetworkSender()` disposes the `saeaStack` but not the current `responseObject` (a `GarnetSaeaBuffer` holding a `PoolEntry`). When a handler is disposed mid-operation, the `PoolEntry` is never returned to the pool, causing `LimitedFixedBufferPool.Dispose()` to spin forever on `totalReferences > 0`.

**Fix:** Added `responseObject?.Dispose(); responseObject = null;` before `saeaStack.Dispose()`.

## Additional: Buffer Pool Tracking Infrastructure

Added `PoolEntryBufferType` and `PoolOwnerType` enums with byte-packed `source` field on `PoolEntry` to track buffer origin. Under `#if DEBUG`, `LimitedFixedBufferPool` tracks outstanding entries and logs unreturned buffer details (owner type, buffer type, size) after a 5-second timeout during disposal.

## Testing

`ClusterResetHardDuringDisklessReplicationAttach` now passes reliably (~1s teardown vs infinite hang).

### DIFF
--- a/libs/client/ClientSession/GarnetClientSession.cs
+++ b/libs/client/ClientSession/GarnetClientSession.cs
@@ -127,7 +127,7 @@ namespace Garnet.client
 
             this.usingManagedNetworkPool = networkPool != null;
             this.networkBufferSettings = networkBufferSettings;
-            this.networkPool = networkPool ?? networkBufferSettings.CreateBufferPool();
+            this.networkPool = networkPool ?? networkBufferSettings.CreateBufferPool(ownerType: PoolOwnerType.ClientSession);
             this.bufferSizeDigits = NumUtils.CountDigits(this.networkBufferSettings.sendBufferSize);
 
             this.logger = logger;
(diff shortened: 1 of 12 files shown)


vazois:
Closing in favor of #1632 (targeting dev). Will merge into vazois/mmrt-dev from dev after approval.

PR #1632, opened 2026-03-16
Add LFBP Instrumentation, Server Shutdown Ordering, and Test TearDown Hardening

vazois:
## Summary

Adds instrumentation for tracking owners and type of buffers acquired from `LimitedFixedBufferPool` (LFBP) to diagnose dispose issues and buffer leaks. Improves server shutdown ordering so listening sockets are closed immediately (freeing ports), and hardens cluster test teardown so cleanup always runs even when tests fail.

### Buffer Pool Tracking Instrumentation

- **New enums** (`PoolEntryTypes.cs`): `PoolEntryBufferType` (identifies buffer role, e.g. `NetworkReceiveBuffer`, `SaeaSendBuffer`) and `PoolOwnerType` (identifies pool creator, e.g. `ServerNetwork`, `Replication`, `ClientSession`).
- **`PoolEntry.source`** field: packed int where the low byte stores `PoolEntryBufferType` and byte 1 stores `PoolOwnerType`, set when the entry is acquired via `LimitedFixedBufferPool.Get()`.
- **`LimitedFixedBufferPool`** now accepts `PoolOwnerType` at construction and `PoolEntryBufferType` in `Get()`. Under `#if DEBUG`, a `ConcurrentDictionary` tracks all outstanding (checked-out) entries.
- **Dispose diagnostics** (DEBUG only): after a 5-second timeout, `LimitedFixedBufferPool.Dispose()` logs all unreturned buffer details (owner type, buffer type, size) to aid leak diagnosis.
- **All call sites updated** to pass appropriate `PoolOwnerType` and `PoolEntryBufferType` values (`GarnetClientSession`, `ReplicationManager`, `GarnetServerTcp`, `NetworkHandler`, `TcpNetworkHandlerBase`, `GarnetSaeaBuffer`, `LightClient`, `MigrationManager`, `GarnetClient`).

### Server Shutdown Ordering & Port Reuse

- **`IGarnetServer.Close()`**: new interface method that stops listening without waiting for active connections to drain.
- **`GarnetServerTcp.Close()`**: closes the listening socket to free the port immediately.
- **`GarnetServer.InternalDispose`**: three-phase shutdown — (1) `Close()` all servers to free ports, (2) dispose the provider (storage engine), (3) `Dispose()` servers to drain active handlers.
- **`GarnetServerTcp` constructor**: sets `SO_REUSEADDR` on TCP sockets before `Bind` to handle `TIME_WAIT` states. UDS initialization refactored for clarity.
- **`GarnetServerTcp.Dispose`**: reordered to close the listening socket before calling `base.Dispose()` (which drains active handlers).

### ExceptionInjectionHelper Improvements

- Replaced busy `Task.Yield()` loops with a `TaskCompletionSource<bool>` notification pattern, eliminating spin-waiting when tests wait for exception injection points.
- Renamed `WaitOnSet` → `ResetAndWaitAsync` for clarity.
- Updated `WaitOnClearAsync` to use the same TCS-based notification.
- All callers updated: `MigrateSessionSlots`, `ReplicaSyncSession`, `ReplicaDisklessSync`, `ReplicaReceiveCheckpoint`.

### LocalStorageDevice Logging

- Added `ILogger` parameter to `LocalStorageDevice` constructors.
- `ReadAsync` / `WriteAsync` now log caught exceptions at `Critical` level instead of silently swallowing them.
- `Devices.cs` passes the logger through on Windows.

### Cluster Test TearDown Hardening

- **`ClusterTestContext.TearDown()`**: captures `TestContext.CurrentContext.Result.Outcome` before cleanup begins. All cleanup phases (DisposeCluster, logger factory, delete directory, `OnTearDown`) now run unconditionally; `Assert.Fail` is deferred to the end so resources are always released.
- Increased teardown timeout from 5 → 30 seconds.
- **`SimplePrimaryReplicaSetup()`**: new helper that configures a one-primary-one-replica cluster with slot assignment and `MEET`.
- **`AttachAndWaitForSync`**: removed redundant `Meet` and `BumpEpoch` calls (now handled by `SimplePrimaryReplicaSetup`).
- **`ReplicaSyncSession`**: added `cts.Token.ThrowIfCancellationRequested()` at the top of the `AcquireCheckpointEntry` loop.

### Misc

- Upgraded `DisposeActiveHandlers()` diagnostics from `HANGDETECT` to `DEBUG` with `Stopwatch`-based timeout logging of stuck handlers.
- Fixed typo: "Dipose" → "Dispose" in `LimitedFixedBufferPool` doc comment.


libs/common/Networking/GarnetTcpNetworkSender.cs
```diff
@@ -252,6 +252,12 @@ public override void DisposeNetworkSender(bool waitForSendCompletion)
             // Wait for ongoing sends to complete
             while (throttleCount >= 0 && Interlocked.CompareExchange(ref throttleCount, int.MinValue, 0) != 0) Thread.Yield();
 
+            // Atomically take ownership of the current response object so it is
+            // disposed together with the rest of the stack entries below.
+            var obj = Interlocked.Exchange(ref responseObject, null);
```
badrishc: There could be an ongoing IO whose ReturnResponseObject is tasked with returning this response object. That is why we did not return it here earlier.

badrishc (approved):

PR #1637, opened 2026-03-17
[Docs] Add Copilot skill for implementing new built-in RESP commands

badrishc:
### Description of Change

   Adds a new Copilot skill definition (`.github/skills/add-garnet-command/SKILL.md`)
   that provides step-by-step guidance for implementing new built-in RESP commands in
   Garnet. The skill covers:

   - RespCommand enum registration and ordering rules
   - Command parsing (fast path vs slow path)
   - Dispatch, RESP handler, API surface, and storage session wiring
   - RMW/Read callback implementation
   - Command metadata JSON generation via CommandInfoUpdater
   - ACL test requirements and integration test patterns
   - Common caveats (dot-prefixed commands, VarLenInputMethods, NeedCopyUpdate, etc.)

   This is a documentation-only change — no runtime code is modified.


.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+Adding a single new command touches **at minimum** these areas:
+
+| # | Area | Files | Required? |
+|---|------|-------|-----------|
+| 1 | RespCommand enum | `libs/server/Resp/Parser/RespCommand.cs` | ✅ Always |
+| 2 | Command parsing | `libs/server/Resp/Parser/RespCommand.cs` | ✅ Always |
+| 3 | Command dispatch | `libs/server/Resp/RespServerSession.cs` | ✅ Always |
+| 4 | RESP handler | `libs/server/Resp/<Category>/*.cs` | ✅ Always |
+| 5 | API interface | `libs/server/API/IGarnetApi.cs` | If key-value command (not blocking/admin) |
+| 6 | API delegation | `libs/server/API/GarnetApi.cs` | If key-value command (not blocking/admin) |
```
TalZaccai: This is always true for the `GarnetApi` class, but the implementation may reside in `GarnetApiUnifiedCommands.cs` or `GarnetApiObjectCommands`.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+| 7 | Storage session | `libs/server/Storage/Session/` | If key-value command (not blocking/admin) |
```
TalZaccai: Might be confusing as this is not a file but a directory? Could we do something like `libs/server/Storage/Session/[Main|Object|Unified]Store/*Ops.cs`?

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+| 8 | RMW/Read callbacks | `libs/server/Storage/Functions/MainStore/` | If string command using RMW/Read |
```
TalZaccai: Same comment here, files should be `[Read|RMW]Methods.cs`

TalZaccai: For read commands implementation in MainStore, the necessary implementation change is actually in `PrivateMethods.cs` (in the `CopyRespToWithInput` method)

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+| 9 | VarLen methods | `libs/server/Storage/Functions/MainStore/VarLenInputMethods.cs` | If string command using RMW |
```
TalZaccai: Relevant also for unified store

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+---
+
+## Step 2: Add Command Parsing
+
+**File:** `libs/server/Resp/Parser/RespCommand.cs`
+
+Two parsing paths exist:
+
+### Fast path: `FastParseCommand()` / `FastParseArrayCommand()`
+For commands with short, fixed-length names (≤8 chars, no dots). Uses `ulong` pointer comparisons on `(count << 4) | length` patterns. Only add here if the command name is a simple word.
```
TalZaccai: Could be worthwhile to point the differences between the two methods (e.g. `FastParseCommand` mostly for fixed # of args & command name up to 9 chars, `FastParseArrayCommand` for variable # of args & command name up to 16 chars)

TalZaccai (commented):
Added some comments, still reading

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+Command handlers are methods on the `RespServerSession` partial class:
+
+```csharp
+private bool NetworkMYCMD<TGarnetApi>(ref TGarnetApi storageApi)
+    where TGarnetApi : IGarnetApi
+{
+    // 1. Validate argument count
+    if (parseState.Count != N)
+        return AbortWithWrongNumberOfArguments(nameof(RespCommand.MYCMD));
+
```
TalZaccai, line 163: Might do other input validations here to short-circuit going to storage in case of an input error

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+    // 2. Parse arguments
+    var key = parseState.GetArgSliceByRef(0);
+    var value = parseState.GetArgSliceByRef(1);
+
```
TalZaccai, line 172: For the common case, we need to instantiate the input and output structs here.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+    // 3. Call storage API
+    var status = storageApi.MyOperation(key, value, out var result);
+
+    // 4. Write RESP response
+    if (status == GarnetStatus.OK)
+    {
+        while (!RespWriteUtils.WriteDirect(CmdStrings.RESP_OK, ref dcurr, dend))
```
TalZaccai: For the common case we use `ProcessOutput()`, unless operation specifically does not write to output

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+            SendAndReset();
+    }
+    else
+    {
+        while (!RespWriteUtils.WriteError("ERR message"u8, ref dcurr, dend))
```
TalZaccai: a) Better to use `WriteError` extension method instead.
b) Error strings should be stored as u8 literals in `CmdStrings`.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+            SendAndReset();
+    }
+
+    return true;
+}
+```
+
+**Key patterns:**
+- Arguments: `parseState.GetArgSliceByRef(i)` returns `ref PinnedSpanByte`
+- Response: Use `RespWriteUtils.WriteDirect`, `WriteError`, `TryWriteInt32`, `WriteBulkString`, etc.
```
TalZaccai: Use `ProcessOutput()` in the common happy-flow case, otherwise use `RespServerSession` extension methods (e.g. `WriteError`, `WriteDirect`, `WriteInt64` etc.)

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+- Always call `SendAndReset()` in the while-not-written loop
```
TalZaccai: Can remove this line, as it is internally handled in the extension methods

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+- Return `true` (command handled) or `false` (need more data)
```
TalZaccai: We always return `true`, there are no more partial executions

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+
+### Object command RESP handler pattern
+
+Object commands (Hash, List, Set, SortedSet) follow a different pattern. The RESP handler builds an `ObjectInput` with the appropriate operation enum and delegates to the storage API:
```
TalZaccai: I don't think this is true anymore, now that have `StringOutput`, these handlers are quite similar.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+            ProcessOutput(output.SpanByteAndMemory);
+            break;
+    }
+
+    return true;
+}
+```
+
+**Key differences from string commands:**
+- Uses `ObjectInput` (not `StringInput`) with a `RespInputHeader(GarnetObjectType.XXX)` and an operation enum (e.g., `SortedSetOperation.ZADD`)
```
TalZaccai: Note that there was no mention of `StringInput` in the `Implement the RESP Handler` section

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+- Must handle `GarnetStatus.WRONGTYPE` — object commands can fail if the key holds a different object type
+- The actual data operation logic lives in `libs/server/Objects/[ObjectName]/[ObjectName]ObjectImpl.cs`, dispatched via the operation enum
+
+For new object sub-operations, also add a value to the `[ObjectName]Operation` enum in `libs/server/Objects/[ObjectName]/[ObjectName]Object.cs` and handle it in the `Operate` method's switch statement.
```
TalZaccai: Not the right place for this comment

.github/skills/add-garnet-command/SKILL.md
TalZaccai: Also, there is no mention of unified commands handling (which is similar to both string and object)

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+> **Note on context types:** The unified single-store has three context types: **string context** (for raw string commands like GET/SET), **object context** (for collection commands like ZADD/LPUSH), and **unified context** (for type-agnostic commands like EXISTS/DELETE/TTL/EXPIRE). Most new commands use either the string or object context — the unified context is only for commands that must work across both value types.
+
+## Step 5: Add API Interface Method
+
+**File:** `libs/server/API/IGarnetApi.cs`
+
+Add method signature to `IGarnetApi` (read-write) or `IGarnetReadApi` (read-only):
+
+```csharp
+GarnetStatus MyOperation(PinnedSpanByte key, PinnedSpanByte value, out MyResult result);
```
TalZaccai: Common case should be: 
```csharp
GarnetStatus MyOperation(PinnedSpanByte key, ref [String|Unified|Object]Input, ref [String|Unified|Object]Output);
```

TalZaccai: We can also add that there can be a different overload for programmatic callers that avoid forcing the caller to create the Input/Output structs.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+    => storageSession.MyOperation(key, value, out result);
+```
+
+**⚠️ Caveat:** `GarnetApi` is a generic partial struct: `GarnetApi<TStringContext, TObjectContext, TUnifiedContext>`. Method implementations go in the appropriate partial file (`GarnetApi.cs` for string ops, `GarnetApiObjectCommands.cs` for object ops, etc.).
+
+---
+
+## Step 6: Implement Storage Session Layer
+
+**File:** New or existing file in `libs/server/Storage/Session/MainStore/` (for string-context ops) or `libs/server/Storage/Session/ObjectStore/` (for object-context ops)
```
TalZaccai: And `UnifiedStore`...

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+
+### String command pattern
+
+This layer wraps Tsavorite API calls:
+
+```csharp
+public GarnetStatus MyOperation(PinnedSpanByte key, PinnedSpanByte value, out MyResult result)
```
TalZaccai: Again, this is the case for a programmatic call, in the "regular" network path it'll be something like:
```csharp
public GarnetStatus MyOperation<T[String|Unified|Object]Context>(PinnedSpanByte key, ref [String|Unified|Object]Input input, ref [String|Unified|Object]Output output, ref T[String|Unified|Object]Context context)
  where T[String|Unified|Object]Context : ITsavoriteContext...
    => RMW[String|Unified|Object]StoreOperation(key.ReadOnlySpan, ref input, ref objectContext, ref output);
```

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+    if (status.IsPending)
+        CompletePendingForSession(ref status, ref output, ref stringBasicContext);
+
+    // Interpret result
+    result = status.Found ? MyResult.OK : MyResult.NotFound;
+    return GarnetStatus.OK;
+}
+```
+
+### Object command pattern
```
TalZaccai: Again, I don't think that there is much different between string and object implementation here.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+new("DEBUG", RespCommand.DEBUG),
+```
+
+`StoreType` values: `Main` (string store), `Object` (object store), `All` (both), `None` (no keys).
+
+### 8b. Add to GarnetCommandsInfo.json (Garnet-only commands)
+
+**File:** `playground/CommandInfoUpdater/GarnetCommandsInfo.json`
+
+Only needed for commands that don't exist in standard Redis (e.g., `DELIFGREATER`, `SETIFMATCH`). Standard Redis commands (e.g., `DEBUG`, `GETDEL`) get their metadata from a running Redis server automatically via the CommandInfoUpdater tool — skip this step and Step 8c for those.
```
TalZaccai: Could also be a standard Redis command that for some reason we need to override its info

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+- `Flags`: `ReadOnly`, `Write`, `DenyOom`, `Fast`, `Admin`, `NoAuth`, `Module`, etc.
+- `AclCategories`: Used for ACL permission checks. Use `Garnet` for Garnet-specific commands
+- `KeySpecifications`: Drives automatic transaction key locking — no per-command switch needed
+- `KeySpec.Flags`: `RO` (read-only), `RW` (read-write), `Access`, `Insert`, `Update`, `Delete`
+
+### 8c. Add to GarnetCommandsDocs.json (Garnet-only commands)
+
+**File:** `playground/CommandInfoUpdater/GarnetCommandsDocs.json`
+
+Add documentation entry:
```
TalZaccai, line 492: This is not necessary for internal commands. The main usage of the docs is to enable client auto-complete for the command.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+`Bitmap`, `Cluster`, `Connection`, `Generic`, `Geo`, `Hash`, `HyperLogLog`, `List`, `Module`, `PubSub`, `Scripting`, `Sentinel`, `Server`, `Set`, `SortedSet`, `Stream`, `String`, `Transactions`
+
+Do NOT invent new group names — the JSON deserializer will fail.
+
+### 8d. Generate the resource JSON files
+
+**⚠️ CRITICAL: Never edit `libs/resources/RespCommandsInfo.json` or `libs/resources/RespCommandsDocs.json` directly.** These are generated by the CommandInfoUpdater tool.
+
+**Steps:**
+1. Build the server: `dotnet build main/GarnetServer/GarnetServer.csproj -c Debug -f net10.0`
```
TalZaccai: This is incorrect. The user should start a local RESP server (not Garnet), such as `valkey` so the `CommandInfoUpdater` could get the new command info from it.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+2. Start a local Garnet server: `dotnet run --project main/GarnetServer/GarnetServer.csproj -c Debug -f net10.0 --no-build -- --port 6399 --logger-level Warning`
+3. Build and run the tool:
+   ```bash
+   cd playground/CommandInfoUpdater
+   dotnet build -f net10.0
+   dotnet run -f net10.0 --no-build -- --port 6399 --output ../../libs/resources
```
TalZaccai, line 531: The port here should match the local resp server port

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -0,0 +1,756 @@
...
+}
+```
+
+**⚠️ Caveat — Idempotency:** The ACL framework calls your command multiple times (for different user/permission combinations). If your command is NOT idempotent (e.g., `RI.CREATE` fails on duplicate), use a counter to generate unique keys per invocation (see the SADD ACL test pattern).
+
+---
+
+## Step 10: Add Integration Tests
+
+**New file:** `test/Garnet.test/Resp<Feature>Tests.cs`
```
TalZaccai: A new file is not necessarily required. String / Unified command tests usually go in `RespTests.cs`, Object command tests usually go in `<ObjectName>RespTests.cs`

TalZaccai (changes requested):
Added more comments

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -156,23 +161,23 @@ private bool NetworkMYCMD<TGarnetApi>(ref TGarnetApi storageApi)
...
+    // 2. Validate other inputs (short-circuit before going to storage)
     var key = parseState.GetArgSliceByRef(0);
-    var value = parseState.GetArgSliceByRef(1);
+    // e.g., parse and validate optional flags, numeric arguments, etc.
 
-    // 3. Call storage API
-    var status = storageApi.MyOperation(key, value, out var result);
+    // 3. Build input/output and call storage API
+    var input = new StringInput(RespCommand.MYCMD, ref parseState, startIdx: 1);
+    var output = new StringOutput();
```
TalZaccai: Should be `GetStringOutput()`, unless operation does not write to the output.

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -156,23 +161,23 @@ private bool NetworkMYCMD<TGarnetApi>(ref TGarnetApi storageApi)
...
     {
-        while (!RespWriteUtils.WriteDirect(CmdStrings.RESP_OK, ref dcurr, dend))
-            SendAndReset();
+        ProcessOutput(output);
     }
     else
     {
-        while (!RespWriteUtils.WriteError("ERR message"u8, ref dcurr, dend))
-            SendAndReset();
+        this.WriteError(CmdStrings.RESP_ERR_MY_MESSAGE);
```
TalZaccai: nit: Can remove `this.`

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -156,23 +161,23 @@ private bool NetworkMYCMD<TGarnetApi>(ref TGarnetApi storageApi)
     if (parseState.Count != N)
         return AbortWithWrongNumberOfArguments(nameof(RespCommand.MYCMD));
 
-    // 2. Parse arguments
+    // 2. Validate other inputs (short-circuit before going to storage)
     var key = parseState.GetArgSliceByRef(0);
-    var value = parseState.GetArgSliceByRef(1);
+    // e.g., parse and validate optional flags, numeric arguments, etc.
```
TalZaccai, line 166: Could add an example here, such as:
```csharp
if (!parseState.TryGetInt(1, out var _)
{
    WriteError(CmdStrings.RESP_ERR_GENERIC_VALUE_IS_NOT_INTEGER);
    return true;
}
```

TalZaccai, line 166: We can also note that sometimes in order to not have to double-parse a parameter, we may pass it in the input struct's auxilary arguments (`arg1`)

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -222,11 +229,37 @@ private unsafe bool SortedSetAdd<TGarnetApi>(ref TGarnetApi storageApi)
...
-- Uses `ObjectInput` (not `StringInput`) with a `RespInputHeader(GarnetObjectType.XXX)` and an operation enum (e.g., `SortedSetOperation.ZADD`)
+- Uses `ObjectInput` with a `RespInputHeader(GarnetObjectType.XXX)` and an operation enum (e.g., `SortedSetOperation.ZADD`)
 - Must handle `GarnetStatus.WRONGTYPE` — object commands can fail if the key holds a different object type
 - The actual data operation logic lives in `libs/server/Objects/[ObjectName]/[ObjectName]ObjectImpl.cs`, dispatched via the operation enum
 
-For new object sub-operations, also add a value to the `[ObjectName]Operation` enum in `libs/server/Objects/[ObjectName]/[ObjectName]Object.cs` and handle it in the `Operate` method's switch statement.
+**For new object sub-operations:**
+Add a value to the `[ObjectName]Operation` enum in `libs/server/Objects/[ObjectName]/[ObjectName]Object.cs` and handle it in the `Operate` method's switch statement.
+
+### Unified command RESP handler pattern
```
TalZaccai: This is also similar to both string and object handlers, not sure this requires a separate section

.github/skills/add-garnet-command/SKILL.md
```diff
@@ -282,44 +315,48 @@ private unsafe bool SortedSetBlockingPop(RespCommand command)
...
-**File:** New or existing file in `libs/server/Storage/Session/MainStore/` (for string-context ops) or `libs/server/Storage/Session/ObjectStore/` (for object-context ops)
+**File:** New or existing file in `libs/server/Storage/Session/MainStore/` (for string-context ops), `libs/server/Storage/Session/ObjectStore/` (for object-context ops), or `libs/server/Storage/Session/UnifiedStore/` (for unified-context ops)
 
 ### String command pattern
 
 This layer wraps Tsavorite API calls:
 
 ```csharp
-public GarnetStatus MyOperation(PinnedSpanByte key, PinnedSpanByte value, out MyResult result)
+public GarnetStatus MyOperation(PinnedSpanByte key, ref StringInput input, ref StringOutput output)
```
TalZaccai: It is worthwhile mentioning that we can add multiple overloads to this method. This is the basic signature used by the network flow of Garnet, but there might be programatic calls that would have a simpler signature. For instance: 
```csharp
public GarnetStatus MyOperation(PinnedSpanByte key, double val, out double output)
```

This method will generate the appropriate input and output and only return the desired value to the caller instead of writing anything to the output buffer.

TalZaccai (commented):
Added some more comments

TalZaccai (approved):

PR #1640, opened 2026-03-19
Use simplified key specs in cluster slot verification

TalZaccai:
**Problem:**

   - `ClusterSlotVerificationInput` currently stores pre-computed key search fields (`firstKey`, `lastKey`, `step`, `keyNumOffset`) populated by `ClusterProvider.ExtractKeySpecs` - this logic may ignore some scenarios such as multiple key specifications per command.
   - `ExtractKeySpecs` duplicated key spec resolution logic that already existed in `TryGetKeySearchArgsFromSimpleKeySpec`, with its own implementation handling multiple key spec types (`BeginSearchIndex`, `FindKeysRange`, `FindKeysKeyNum`).
   - The `keyNumOffset` field was a workaround for 2-spec commands where a count argument sits between keys — `TryGetKeySearchArgsFromSimpleKeySpec` handles this naturally by resolving each spec independently

**Changes:**

   - `ClusterSlotVerificationInput` — Replaced `firstKey`, `lastKey`, `step`, `keyNumOffset` with `SimpleRespKeySpec[] keySpecs` and `bool isSubCommand`
   - `MultiKeySlotVerify` — Now iterates over keySpecs, resolves each via `TryGetKeySearchArgsFromSimpleKeySpec`, and verifies keys per-spec; eliminates `keyNumOffset` skip hack
   - `CanServeSlot` — Uses `TryGetSimpleRespCommandInfo` instead of `TryFastGetRespCommandInfo` + `ExtractKeySpecs`
   - `ClusterSession.ProcessClusterCommands` — Same pattern change; removed unused `commandInfo` variable
   - `ClusterProvider.ExtractKeySpecs` — Deleted entirely (65 lines)
   - `IClusterProvider` — Removed ExtractKeySpecs from interface
   - `TryGetKeySearchArgsFromSimpleKeySpec` — Changed from internal to public for cross-project access

  **Result:**

   - Net -73 lines (42 added, 115 removed)
   - Single code path for key spec resolution across both transaction locking and slot verification


libs/cluster/Session/SlotVerification/ClusterSlotVerify.cs
```diff
@@ -143,26 +143,36 @@ ClusterSlotVerificationResult MultiKeySlotVerify(ClusterConfig config, ref Span<
...
+                for (var i = searchArgs.firstIdx; i <= searchArgs.lastIdx; i += searchArgs.step)
+                {
+                    ref var key = ref parseState.GetArgSliceByRef(i);
+                    var slot = HashSlotUtils.HashSlot(key);
+                    var result = SingleKeySlotVerify(ref config, ref key, csvi.readOnly, csvi.sessionAsking, slot);
 
-                // Check if any key might have moved
-                if (_verifyResult.state != verifyResult.state)
-                    return new(SlotVerifiedState.TRYAGAIN, slot);
+                    if (isFirstKey)
```
vazois: Is it possible to initialize for the firstKey outside the for-loop to avoid the additional check at every iteration?

vazois (approved):

PR #1642, opened 2026-03-19
[Bugfix] SortedSet Race Condition - Mutating under Shared Lock

hamdaankhalid:
TLDR:
SortedSetTimeToLive/ZTTL uses the Read Tsavorite method which uses Shared Lock, and then in SortedSetObjectImpl layer it was mutating shared data structure.  

Reasoning:
As I understand, we handle thread safety for SortedSet data structures by depending on the state mutating methods like DeleteExpiredItem to be called for commands under the ExclusiveLock commands using Tsavorite Upsert or RMW methods.
If a SortedSet command is using the Read methods and then it mutates state, that breaks the invariant.

image


test/Garnet.test/SortedSetShallowCloneRaceTests.cs
hamdaankhalid, line 1: incorrect file, deleting

badrishc (approved):

PR #1646, opened 2026-03-24
Bugfix: Accept Loop socket error handling

hamdaankhalid:
Prevent Accept loop from breaking from all errors in socket acceptance portion.

### DIFF
--- a/libs/server/Servers/GarnetServerTcp.cs
+++ b/libs/server/Servers/GarnetServerTcp.cs
@@ -134,13 +134,11 @@ namespace Garnet.server
             {
                 do
                 {
-                    /*
-                    When the while condition exits normally(AcceptAsync returned true), an
-                    accept is already pending on IOCP — the callback will fire on next connection.But when
-                    HandleNewConnection returns false, the break skips the while entirely — no AcceptAsync is issued,
-                    nothing is pending.
-                    Without ScheduleAcceptRetry calling AcceptAsync after a delay, the accept loop would be permanently dead.
-                    */
+                    // When the while condition exits normally(AcceptAsync returned true), an
+                    // accept is already pending on IOCP — the callback will fire on next connection.But when
+                    // HandleNewConnection returns false, the break skips the while entirely — no AcceptAsync is issued,
+                    // nothing is pending.
+                    // Without ScheduleAcceptRetry calling AcceptAsync after a delay, the accept loop would be permanently dead.
                     if (!HandleNewConnection(e)) break;
                     e.AcceptSocket = null;
                 } while (!listenSocket.AcceptAsync(e));
--- a/libs/server/Servers/GarnetServerTcp.cs
+++ b/libs/server/Servers/GarnetServerTcp.cs
@@ -135,7 +135,7 @@ namespace Garnet.server
                 do
                 {
                     /*
-                    Wwhen the while condition exits normally(AcceptAsync returned true), an
+                    When the while condition exits normally(AcceptAsync returned true), an
                     accept is already pending on IOCP — the callback will fire on next connection.But when
                     HandleNewConnection returns false, the break skips the while entirely — no AcceptAsync is issued,
                     nothing is pending.
@@ -159,7 +159,10 @@ namespace Garnet.server
                 case SocketError.Shutdown:
                 case SocketError.NotInitialized:
                 case SocketError.VersionNotSupported:
-                    logger?.LogCritical("Fatal accept error, stopping accept loop: {error}", e.SocketError);
+                    if (!Disposed || e.SocketError != SocketError.OperationAborted)
+                    {
+                        logger?.LogCritical("Fatal accept error, stopping accept loop: {error}", e.SocketError);
+                    }
                     e.Dispose();
                     return false;
 
@@ -187,9 +190,15 @@ namespace Garnet.server
             acceptRetryTimer?.Dispose();
             acceptRetryTimer = new Timer(_ =>
             {
-                e.AcceptSocket = null;
+                // Best effort guard against race with Dispose(): Timer.Dispose() does not
+                // guarantee an in-flight callback has finished, so this callback
+                // can fire after the server has started disposing.
+                // if disposed happens after check then try catch handles it.
+                if (Disposed) return;
+
                 try
                 {
+                    e.AcceptSocket = null;
                     if (!listenSocket.AcceptAsync(e))
                         AcceptEventArg_Completed(null, e);
                 }
--- a/libs/server/Servers/GarnetServerTcp.cs
+++ b/libs/server/Servers/GarnetServerTcp.cs
@@ -29,6 +29,12 @@ namespace Garnet.server
         readonly string unixSocketPath;
         readonly UnixFileMode unixSocketPermission;
 
+        // Accept loop backoff state for resource pressure errors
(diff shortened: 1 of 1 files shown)


libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -135,14 +149,70 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
...
+            switch (e.SocketError)
+            {
+                // Tier 1 — Fatal: listen socket is dead, stop accepting
+                case SocketError.OperationAborted:
+                case SocketError.NotSocket:
+                case SocketError.Shutdown:
+                case SocketError.NotInitialized:
+                case SocketError.VersionNotSupported:
+                    logger?.LogCritical("Fatal accept error, stopping accept loop: {error}", e.SocketError);
+                    e.Dispose();
```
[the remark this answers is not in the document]
hamdaankhalid: The listen socket cleanup is Dispose()'s responsibilit

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -135,14 +149,84 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
...
+                case SocketError.NetworkDown:
+                case SocketError.SystemNotReady:
+                case SocketError.ProcessLimit:
+                    logger?.LogWarning("Accept backoff ({backoffMs}ms) due to resource pressure: {error}", acceptBackoffMs, e.SocketError);
+                    ScheduleAcceptRetry(e);
+                    return false;
+
+                // Tier 3 — Client-caused transient or irrelevant: log and continue
+                default:
+                    logger?.LogDebug("Transient accept error, continuing: {error}", e.SocketError);
```
kevin-montrose, line 214: This seems wrong - we neither retry, nor dispose on an unexpected error type?

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -127,6 +134,13 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
             {
                 do
                 {
+                    /*
```
kevin-montrose: nit: prefer `//` comments - jarring to mix multiline `//` and `/*` in the same file.

libs/server/Servers/GarnetServerTcp.cs
hamdaankhalid: we disposed at the top of the method

hamdaankhalid: and if it is a client caused error, the retry should come from the client imo

kevin-montrose: `e.Dispose()` is in the old code unconditionally - in this code we `e.Dispose()` in Tier 1 path, retry in Tier 2, but do nothing in Tier 3?

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -135,14 +149,84 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
...
+        private void ScheduleAcceptRetry(SocketAsyncEventArgs e)
+        {
+            // Schedules the accept loop to resume after a backoff delay.
+            // Use a Timer so we don't block the IOCP thread pool thread.
+            //
+            // No concurrent callback race: ScheduleAcceptRetry is only reachable through
+            // the serialized accept path (single SAEA), so a new timer can't be scheduled
+            // until the previous callback has completed, called AcceptAsync, and that
+            // accept has completed with another Tier 2 error.
+            acceptRetryTimer?.Dispose();
```
kevin-montrose: I think `e` is always `== acceptEventArg` here, if so there's no reason to dispose and recreate the timer.

Just call `Change(acceptBackoffMs, Timeout.Infinite)`.

For creating the timer we can just do that unconditionally, we don't make that many `GarnetServerTcp` instances.

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -135,14 +149,84 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
...
+            acceptRetryTimer = new Timer(_ =>
```
kevin-montrose: More a style thing, but I suspect this callback can be `static`.

Pass `this` and use `acceptEventArg ` for the `SocketAsyncEventArgs`.  That does assume I'm correct in my `e == acceptEventArg` reading.

Would save a couple allocations, minimal impact unless something unusual (like a connection storm) is happening.

libs/server/Servers/GarnetServerTcp.cs
hamdaankhalid: I dispose the accept socket unconditionally. If I plan on reconnecting via timer call (Tier 2) I shouldn't be disposing the SAEA. 

SAEA is to my understanding a container with only 2 sockets the connect and accept socket references.

If I intend to reconnect such as in Tier 2, and Tier 3. Then I can reuse the same SAEA container. I just need to handle the dispose and setting to null of the accept socket only. This is essentially how we deal with new connections today too.

Tier 3 we do nothing because we have already disposed the accept socket, and the client can retry. In the old code, we would unconditionally dispose the SAEA itself, I don't see why we need to do that, unless the server should be shutting down entirely.

libs/server/Servers/GarnetServerTcp.cs
hamdaankhalid: Now that we are not allocating repeatedly, perf tradeoff is not there. However, I agree explicit state capture is better than implict. I will make this change.

libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -135,14 +145,61 @@ private void AcceptEventArg_Completed(object sender, SocketAsyncEventArgs e)
...
+                case SocketError.TooManyOpenSockets:
+                case SocketError.NoBufferSpaceAvailable:
+                case SocketError.NetworkDown:
+                case SocketError.SystemNotReady:
+                case SocketError.ProcessLimit:
+                    logger?.LogWarning("Accept backoff ({backoffMs}ms) due to resource pressure: {error}", acceptBackoffMs, e.SocketError);
+                    // This will hold the IOCP thread hostage, but anyway we hold it hostage in the do while accept loop. So this is not really deflecting from design.
+                    // If this is ever a concern in the future we can use a timer to schedule the next attempt instead of blocking the thread.
+                    Thread.Sleep(acceptBackoffMs);
+                    acceptBackoffMs = Math.Min(acceptBackoffMs * 2, MaxAcceptBackoffMs);
```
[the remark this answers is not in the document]
hamdaankhalid: comment gives reasoning of why we do this.

vazois (approved):

libs/server/Servers/GarnetServerTcp.cs
hamdaankhalid: Changed the code to not use timer, since it's fine to hold that single IOCP thread. However, the e.Dispose logic still remains. I removed calling e.Dispose from here since the dispose method will do that anyway for legit shutdown,  for tier 2 and tier 3 we don't need it at all based on the same logic as before. Just disposing accept socket.

badrishc (approved):

kevin-montrose (approved):

PR #1648, opened 2026-03-27
[Bugfix] Handle SocketException on accepted socket configuration

hamdaankhalid:
Wrap the post-accept socket configuration in a try/catch, dispose the dead socket, and continue the accept loop. There is small gap where bad timing can still kill accept loop.

### DIFF
--- a/libs/server/Servers/GarnetServerTcp.cs
+++ b/libs/server/Servers/GarnetServerTcp.cs
@@ -226,21 +226,24 @@ namespace Garnet.server
             // Reset backoff on successful accept
             acceptBackoffMs = InitialAcceptBackoffMs;
 
-            if (e.AcceptSocket.LocalEndPoint is not UnixDomainSocketEndPoint)
-                e.AcceptSocket.NoDelay = true;
-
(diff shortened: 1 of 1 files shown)


libs/server/Servers/GarnetServerTcp.cs
```diff
@@ -226,36 +226,53 @@ private unsafe bool HandleNewConnection(SocketAsyncEventArgs e)
...
+                        if (handler != null)
+                        {
+                            logger?.LogError(ex, "Error creating and registering network handler");
+                            // We did not start the handler, so we need to call DisposeResources() to clean up resources.
+                            handler.DisposeResources();
+                            handler.Dispose();
+                        }
+                        else
+                        {
+                            if (ex is SocketException se)
```
hamdaankhalid, line 265: I know it would be cleaner to catch this exception, however I am not sure of what other exceptions are possible with the 2 lines I moved in the try catch block. So for now I am adding this if-else here as well.

kevin-montrose (approved):

badrishc (approved):

PR #1652, opened 2026-03-30
Pull Vector Set implementation into dev

kevin-montrose:
Brings Vector Set implementation into dev, moving to Namespace and RecordType in store v2 instead of the hackery on top of store v1.

Meaningful differences:
 - [`VectorElementKey`](https://github.com/microsoft/garnet/pull/1652/changes?mode=single#diff-34dc0c6b4c6474c30e6b4fa7251cf89ad6b2e23dfe10ff4f73eb4e504ba961b8)
 - [Setting `RecordType`](https://github.com/microsoft/garnet/pull/1652/changes?mode=single#diff-8bfd23aa501a932f28993e6e5cefa7f2bf7d0c041ade143b6e90b201ae0b74e2R267)
 - A prefix byte [to indicate replication type, since shipping `LogRecord` isn't sufficient](https://github.com/microsoft/garnet/pull/1652/changes?mode=single#diff-2d25544a87056a6c029449ba185735d36e4d6b5d961db8f5d043f38bf8b4529a)

Store V2 makes unaligned values a possibility, which violates some guarantees Vector Sets made to DiskANN.  I paper over this by over-allocating some padding and adjusting things before calling into DiskANN.  This does complicate `VectorSessionFunctions` and replication, as those padding bytes must be accounted for.  Eventually we can make alignment a Tsavorite thing, and then all this disappears.


libs/common/ReadOptimizedLock.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// 
+        /// Walking <paramref name="currentProcessorHint"/> from 0 to (<see cref="coreSelectionMask"/> + 1) [exclusive] will return
+        /// all possible counts for a given hash.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public readonly int CalculateIndex(long hashLong, int currentProcessorHint)
+        {
+            // Throw away half the top half of the hash
+            //
+            // This set of locks will be small enough that the extra bits shoulnd't matter
```
badrishc: nit: shoulnd't -> shouldn't

badrishc (approved):

libs/common/Synchronization/ReadOptimizedLock.cs
```diff
@@ -0,0 +1,408 @@
...
+    ///   We access array elements via reference, to avoid thrashing cache lines due to length checks
+    ///   Each shard is placed, in so much as is possible, into a different cache line rather than grouping a hash's counts physically near each other
+    ///     This will tend to allow a core to retain ownership of the same cache lines even as it moves between different hashes
+    ///     
+    /// Experimentally (using some rough microbenchmarks) various optimizations are worth (on either shared or exclusive acquisiton paths):
+    ///   - Split shards across cache lines   : 7x (read path), 2.5x (write path)
+    ///   - Fast math instead of mod and mult : 50% (read path), 20% (write path)
+    ///   - Unsafe ref instead of array access:  0% (read path), 10% (write path)
+    /// </remarks>
+    public struct ReadOptimizedLock
```
vazois, line 46: nit; can we move this under Synchronization folder in common?

libs/common/Synchronization/CountingEventSlim.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Garnet.common
+{
+    /// <summary>
+    /// A <see cref="ManualResetEventSlim"/> that is triggered based on a count hitting or advancing above 0 rather than
+    /// explicit Set and Reset calls.
+    /// </summary>
+    /// <remarks>
+    /// Akin to a <see cref="CountdownEvent"/>, but allows for count to go back up after hitting 0.
+    /// </remarks>
+    public struct CountingEventSlim : IDisposable
```
vazois, line 17: nit; can we move this under Synchronization folder in common?
PS: Along with SingleWriterMultiReaderLock.cs

libs/common/Synchronization/ReadOptimizedLock.cs
kevin-montrose: To be clear, that folder doesn't exist right now - so create a new folder and move these two files?

vazois: Yes, please do. I have it in my PR but since we keep adding synchronization primitives it might be good to put all of them under the same folder.

PR #1653, opened 2026-03-31
Migrate: Improved Error Handling on Failed SetSlotRange

kevinmichaelbowersox:
### Summary

Refactors `MigrateSession.TrySetSlotRanges` to replace  `.ContinueWith(...).WaitAsync().Result` async pattern with a more traceable and debuggable implementation. Adds explicit timeout/cancellation handling and ensures `MigrateState.FAIL` is consistently set on all error paths.

### Motivation

The previous implementation used `ContinueWith(TaskContinuationOptions.OnlyOnRanToCompletion)` chained with `.WaitAsync().Result`. This pattern had several issues:

- **Silent failures**: If the task faulted or was canceled, the `OnlyOnRanToCompletion` continuation never ran, and the resulting `TaskCanceledException` was caught by the generic `catch (Exception)` block — which did **not** set `Status = MigrateState.FAIL`, leaving the migration in an indeterminate state.
- **Poor diagnostics**: The generic catch logged `"An error occurred"` with no context about which slots were affected or whether the failure was a timeout vs. an unexpected error.
- **Unnecessary indirection**: Wrapping the result check in a continuation added complexity without benefit.

### Changes

**MigrateSession.cs**

- Replace `.ContinueWith(...).WaitAsync(_timeout, _cts.Token).Result` with direct `task.WaitAsync(_timeout, _cts.Token).GetAwaiter().GetResult()`
- Add dedicated `catch (TaskCanceledException)` handler for timeout/cancellation scenarios
- Add `catch (AggregateException aex) when (aex.InnerException is TaskCanceledException)` for wrapped timeout exceptions
- Set `Status = MigrateState.FAIL` in **all** error paths (the old generic `catch` missed this)
- Add structured log messages that include the affected slot ranges and timeout values
- Add trace logging before the `SETSLOTRANGE` call for better migration observability

**ClusterMigrateTests.cs**

- Add `ClusterMigrateSetSlotRangeResilience` test that exercises the `TrySetSlotRanges` code path through a full slot migration:
  - Sets up a 2-node cluster and creates 50 keys in a slot
  - Migrates the slot from source to target (internally calls `TrySetSlotRanges` for IMPORTING and NODE states)
  - Verifies key count and data integrity on the target node
  - Verifies slot ownership transferred correctly

### Testing

- New test `ClusterMigrateSetSlotRangeResilience` passes
- Existing migration tests remain green

Fixes #1655 1655

### DIFF
--- a/libs/cluster/Server/Migration/MigrateSession.cs
+++ b/libs/cluster/Server/Migration/MigrateSession.cs
@@ -249,61 +249,7 @@ namespace Garnet.cluster
             return slotRanges;
         }
 
-        /// <summary>
-        /// Change remote slot state
-        /// </summary>
-        /// <param name="nodeid"></param>
-        /// <param name="state"></param>
-        /// <returns></returns>
-        public bool TrySetSlotRanges(string nodeid, MigrateState state)
-        {
-            var client = migrateOperation[0].Client;
-            try
-            {
-                if (!CheckConnection(client))
-                {
-                    Status = MigrateState.FAIL;
-                    return false;
-                }
 
-                var stateBytes = state switch
-                {
-                    MigrateState.IMPORT => IMPORTING,
-                    MigrateState.STABLE => STABLE,
-                    MigrateState.NODE => NODE,
-                    _ => throw new Exception("Invalid SETSLOT Operation"),
-                };
-
-                logger?.LogTrace("Sending CLUSTER SETSLOTRANGE {state} {nodeid} {slots}", state, nodeid ?? "null", ClusterManager.GetRange([.. _sslots]));
-
-                // Get the task and wait for it with timeout
-                var task = client.SetSlotRange(stateBytes, nodeid, _slotRanges);
-                var result = task.WaitAsync(_timeout, _cts.Token).GetAwaiter().GetResult();
-
-                // Check if setslotsrange executed correctly
-                if (!result.Equals("OK", StringComparison.Ordinal))
-                {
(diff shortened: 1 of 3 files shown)


libs/cluster/Server/Migration/MigrateSession.cs
```diff
@@ -271,27 +271,41 @@ public bool TrySetSlotRanges(string nodeid, MigrateState state)
                     _ => throw new Exception("Invalid SETSLOT Operation"),
                 };
 
-                status = client.SetSlotRange(stateBytes, nodeid, _slotRanges).ContinueWith(resp =>
+                logger?.LogTrace("Sending CLUSTER SETSLOTRANGE {state} {nodeid} {slots}", state, nodeid ?? "null", ClusterManager.GetRange([.. _sslots]));
+
+                // Get the task and wait for it with timeout
+                var task = client.SetSlotRange(stateBytes, nodeid, _slotRanges);
+                var result = task.WaitAsync(_timeout, _cts.Token).GetAwaiter().GetResult();
```
vazois: The caller of TrySetSlotRanges should be an async task. In that case, can we use await instead of synchronous blocking here? Also ensure to wrap it with ConfigureAwait(false).

kevinmichaelbowersox: Switched `TrySetSlotRanges` over to async along with `TryRecoverFromFailure`

vazois (approved):

PR #1656, opened 2026-03-31
Kill AOF BulkConsume When Connection is Dead

vazois:
This PR introduces a stopConsume callback that checks the health of the client connection when streaming the AOF.
I used a callback here to avoid maintaining a second background task that monitors the connection and kills the AOF BulkConsume loop.


libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs
```diff
@@ -90,6 +90,9 @@ public unsafe void Consume(byte* payloadPtr, int payloadLength, long currentAddr
 
         public void Throttle()
         {
+            if (!garnetClient.IsConnected)
+                throw new GarnetException("AOF stream client disconnected!");
```
badrishc: use ThrowException helper method that is marked NoInline and DoesNotReturn (this ensures inlining opportunities are retained for Throttle).

badrishc (approved):

PR #1662, opened 2026-04-01
Tedhar/dev fixes

TedHartMS:
This PR has a few test fixes but primarily focuses on performance improvements. The biggest gains are in IPU (InPlaceUpdater, or InPlaceWriter), creating RecordSizeInfo only when needed. Also adds improvements to BDN tests, particularly Tsavorite's.


badrishc (approved):
We are merging this as is into dev, so it does not have to wait until Ted is back.

PR #1666, opened 2026-04-02
Ensure CompletedSemaphore is Always Released

vazois:
This PR attempts to fix checkpoint hanging due to uncaught exception during AsyncFlushPagesForSnapshot.


libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -338,9 +338,15 @@ void FlushRunner()
                         epoch.Suspend();
                 }
 
-                if (destOffset > 0)
-                    deltaLog.Seal(destOffset);
-                _completedSemaphore.Release();
+                try
```
badrishc, line 347: why is only this covered by try/finally? is there reason to believe only this will throw exception and we need to anyway release semaphore, and that does not apply to the code that exists earlier up?

libs/storage/Tsavorite/cs/src/core/Index/Checkpointing/StateMachineDriver.cs
```diff
@@ -66,6 +66,12 @@ void DecrementActiveTransactions(long txnVersion)
 
         internal void TrackLastVersion(long version)
         {
+            // Only create and enqueue one semaphore per version, if we create a
```
badrishc: as discussed, we can revert this unless proven to be a problem.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
vazois: I added a catch to the above to ensure that the semaphore is released.

PR #1672, opened 2026-04-06
[Storage] Fix use-after-free in ProcessResp*Output; use ScratchBufferAllocator for API-returned PinnedSpanByte

badrishc:
### Root Cause

`ProcessResp2ArrayOutput`, `ProcessResp3ArrayOutput`, `ProcessRespArrayOutputAsPairs`, and `ProcessRespSingleTokenOutput` in `StorageSession` return `PinnedSpanByte` values pointing into `IMemoryOwner<byte>` buffers from `MemoryPool<byte>.Shared`. These buffers are **not pinned** — `RespMemoryWriter` pins them temporarily via `MemoryHandle` during write, but unpins on dispose. The `ProcessResp*` methods then dispose the `IMemoryOwner` in their `finally` blocks, returning dangling pointers into freed, unpinned memory.

This only affects the `new ObjectOutput()` code path (SpanByte invalidated, forcing IMemory allocation). The RESP handler path uses `GetObjectOutput()` backed by the pinned session network buffer, so it is unaffected.

Same bug exists on `main` with `ArgSlice`/`GarnetObjectStoreOutput` types.

### Affected Commands

ZMPOP, LMPOP, BLMPOP, BZMPOP, and any custom procedures or Lua scripts calling `HashGet`, `HashGetAll`, `HashGetMultiple`, `HashRandomField`, `SetMembers`, `SetPop`, `SortedSetPop`, `SortedSetRange`, `SortedSetIncrement`, `GET`, or `ObjectScan` through `IGarnetApi`.

### Description of Change

Use existing `ScratchBufferAllocator` (SBA) alongside existing `ScratchBufferBuilder` (SBB) in `StorageSession`. SBA uses pinned arrays (`GC.AllocateArray(_, true)`) and keeps old buffers rooted, so returned `PinnedSpanByte` values remain valid across multiple API calls within a batch.

**Key changes:**
- **Core fix** (`Common.cs`): Copy parsed output to SBA inside `fixed` block before `IMemoryOwner` disposal
- **GET fix** (`MainStoreOps.cs`): Use SBA `ViewRemainingArgSlice` for zero-copy common case; copy on IMemory overflow
- **Custom command fix** (`CustomRespCommands.cs`): All returned `PinnedSpanByte` output uses SBA + proper Memory disposal
- **SBA wiring**: Added to `StorageSession`, `DatabaseManagerBase`, `RespServerSession`, `LocalServerSession`
- **API cleanup** (`IGarnetApi`): Removed `GetScratchBufferOffset()` and `ResetScratchBuffer(int)`, added `ResetScratchBuffer()` (full SBA reset)
- **SBB debug safety** (`ScratchBufferBuilder.cs`): Debug-only `outstandingSlices` counter asserts no more than one live `PinnedSpanByte` at a time
- **SBB safety fixes**: Added missing rewinds in `SortedSetIncrement`, `KeyAdminCommands`, `CollectionItemBroker`; single contiguous allocation for `SortedSetRange LIMIT`; create+use+rewind for Lua `AddKey`
- **New API**: `CreateArgSliceAsOffset` returns `(Offset, Length)` instead of `PinnedSpanByte` — safe for multi-alloc (used by `ArgSliceVector`)
- **Visibility**: `scratchBufferBuilder` and `scratchBufferAllocator` marked `internal` on `StorageSession`
- **Cleanup**: Removed unused `FormatScratchAsResp`, `GetSliceFromTail`, `ResetScratchBuffer(int)` from SBB

### Key Technical Details

**SBB vs SBA roles:**
- `ScratchBufferBuilder` — contiguous temporary workspace (Lua serialization, command input building with rewind). Single buffer, reallocates in place.
- `ScratchBufferAllocator` — persistent output storage for API-returned `PinnedSpanByte`. Keeps old buffers rooted in a stack. Reset between batches.

**Affected types:**
- `StorageSession` — new `scratchBufferAllocator` field
- `IGarnetApi` / `GarnetApi` / `GarnetWatchApi` — simplified scratch buffer API
- `ScratchBufferBuilder` — debug counter, `CreateArgSliceAsOffset`, doc comments on view APIs, removed unused APIs

### What NOT to Do

- ❌ **Do not return `PinnedSpanByte` from `ScratchBufferBuilder`** for values that must survive across multiple API calls — use `ScratchBufferAllocator`
- ❌ **Do not call `CreateArgSlice` on SBB without rewinding** before the next `CreateArgSlice` — debug asserts enforce this
- ❌ **Do not skip `Memory.Dispose()`** after copying from `IMemoryOwner` to SBA — causes pool buffer leaks
- ❌ **Do not store `ViewFullArgSlice` / `ViewRemainingArgSlice` results** — they are immediate-use views invalidated by subsequent operations

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -205,6 +205,23 @@ To add a new Garnet server setting:
 - `LightEpoch` instances track ownership — only dispose if owned
 - In parallel tests, share a `LightEpoch` instance across `GarnetClient` instances
 
+### Scratch Buffer Conventions
+
+`StorageSession` has two scratch buffer types — use the right one:
+
+- **`ScratchBufferBuilder` (SBB)** — Single contiguous buffer for temporary workspace. Reallocates in place (old buffer is freed). Use for building command inputs, Lua serialization, or any data that is consumed immediately and then rewound. **Do not** return `PinnedSpanByte` from SBB to callers — it may be invalidated by subsequent allocations. Always rewind after use. Debug builds enforce single-outstanding-slice discipline via asserts.
+  - Key APIs: `CreateArgSlice` (returns `PinnedSpanByte`, must rewind), `CreateArgSliceAsOffset` (returns `(Offset, Length)`, safe for multi-alloc), `ViewRemainingArgSlice`/`ViewFullArgSlice` (immediate-use views, do not store), `MoveOffset`, `Reset`, `RewindScratchBuffer`.
+
+- **`ScratchBufferAllocator` (SBA)** — Multi-buffer allocator that keeps old buffers rooted (pinned arrays via `GC.AllocateArray(_, true)`). Use for `PinnedSpanByte` values returned via `out` parameters or `IGarnetApi` that callers retain across multiple API calls. Reset between batches.
+  - Key APIs: `CreateArgSlice`, `ViewRemainingArgSlice`, `Reset`.
+
+**Rules:**
+1. Any `StorageSession` or `IGarnetApi` method returning `PinnedSpanByte` via `out` must use SBA, not SBB.
+2. When using SBB's `CreateArgSlice`, always `RewindScratchBuffer` after use — debug asserts enforce at most one outstanding slice.
+3. For multiple allocations without rewind, use `CreateArgSliceAsOffset` (returns offsets that survive reallocation).
+4. When copying from `IMemoryOwner<byte>` (e.g., `ObjectOutput.SpanByteAndMemory.Memory`), always `Dispose()` after copying — do not leak pooled buffers.
+5. `ViewFullArgSlice` and `ViewRemainingArgSlice` return immediate-use views — do not store or return them.
+
 ### Releasing a New Version
 
 To update the version and make a new release, increment the `VersionPrefix` in `Version.props` at the repo root and submit a PR with that change.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -209,10 +209,10 @@ To add a new Garnet server setting:
 
 `StorageSession` has two scratch buffer types — use the right one:
 
-- **`ScratchBufferBuilder` (SBB)** — Single contiguous buffer for temporary workspace. Reallocates in place (old buffer is freed). Use for building command inputs, Lua serialization, or any data that is consumed immediately and then rewound. **Do not** return `PinnedSpanByte` from SBB to callers — it may be invalidated by subsequent allocations. Always rewind after use. Debug builds enforce single-outstanding-slice discipline via asserts.
-  - Key APIs: `CreateArgSlice` (returns `PinnedSpanByte`, must rewind), `CreateArgSliceAsOffset` (returns `(Offset, Length)`, safe for multi-alloc), `ViewRemainingArgSlice`/`ViewFullArgSlice` (immediate-use views, do not store), `MoveOffset`, `Reset`, `RewindScratchBuffer`.
+- **`ScratchBufferBuilder` (SBB)** — Single contiguous buffer for temporary workspace. All data is laid out sequentially in one buffer. On expansion, the previous data is copied into a new larger buffer and the old buffer is freed — **any existing pointers into the old buffer become invalid**. Use for building command inputs, Lua serialization, or any data that is consumed immediately and then rewound. **Do not** return `PinnedSpanByte` from SBB to callers — it may be invalidated by subsequent allocations. Always rewind after use. Debug builds enforce single-outstanding-slice discipline via asserts.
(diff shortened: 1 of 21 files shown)


libs/server/API/IGarnetApi.cs
```diff
@@ -1953,17 +1953,9 @@ public bool IterateStore<TScanFunctions>(ref TScanFunctions scanFunctions, ref l
...
-        /// <returns>Current offset</returns>
-        int GetScratchBufferOffset();
-
-        /// <summary>
-        /// Resets the scratch buffer to the given offset.
-        /// </summary>
-        /// <param name="offset">Offset to reset to</param>
-        /// <returns>True if successful, else false</returns>
-        bool ResetScratchBuffer(int offset);
+        void ResetScratchBuffer();
```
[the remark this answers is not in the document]
badrishc: This is as intended as we are moving to SBA which has different semantics

.github/copilot-instructions.md
```diff
@@ -205,6 +205,23 @@ To add a new Garnet server setting:
 - `LightEpoch` instances track ownership — only dispose if owned
 - In parallel tests, share a `LightEpoch` instance across `GarnetClient` instances
 
+### Scratch Buffer Conventions
+
+`StorageSession` has two scratch buffer types — use the right one:
+
+- **`ScratchBufferBuilder` (SBB)** — Single contiguous buffer for temporary workspace. Reallocates in place (old buffer is freed). Use for building command inputs, Lua serialization, or any data that is consumed immediately and then rewound. **Do not** return `PinnedSpanByte` from SBB to callers — it may be invalidated by subsequent allocations. Always rewind after use. Debug builds enforce single-outstanding-slice discipline via asserts.
```
vazois: nit; Maybe consider adding a note of caution, that this tries to maintain a single buffer, so every ArgSlice is layed out in sequence. Hence, Expansion will copy the previous data into a new expanded buffer hence any pointers to the buffer will be invalidated.

.github/copilot-instructions.md
```diff
@@ -205,6 +205,23 @@ To add a new Garnet server setting:
...
+  - Key APIs: `CreateArgSlice` (returns `PinnedSpanByte`, must rewind), `CreateArgSliceAsOffset` (returns `(Offset, Length)`, safe for multi-alloc), `ViewRemainingArgSlice`/`ViewFullArgSlice` (immediate-use views, do not store), `MoveOffset`, `Reset`, `RewindScratchBuffer`.
+
+- **`ScratchBufferAllocator` (SBA)** — Multi-buffer allocator that keeps old buffers rooted (pinned arrays via `GC.AllocateArray(_, true)`). Use for `PinnedSpanByte` values returned via `out` parameters or `IGarnetApi` that callers retain across multiple API calls. Reset between batches.
```
vazois: nit;Maybe consider adding a note, that this maintains a collection of (fragmented) buffers as more items are added through CreateArgSlice.

libs/server/ArgSlice/ScratchBufferAllocator.cs
```diff
@@ -254,6 +254,20 @@ public PinnedSpanByte CreateArgSlice(int length)
             return retVal;
         }
 
+        /// <summary>
+        /// View remaining scratch space (of specified minimum length) as a <see cref="PinnedSpanByte"/>.
+        /// Does NOT move the offset forward.
+        /// </summary>
+        /// <param name="minLength">Minimum length of remaining space</param>
+        /// <returns>A <see cref="PinnedSpanByte"/> covering the remaining space</returns>
+        public PinnedSpanByte ViewRemainingArgSlice(int minLength = 0)
```
vazois, line 263: This seems like a readonly operation. Why do we expand the ScratchAllocatorBuffer?

badrishc: We use ViewRemainingArgSlice to be able to acquire "potential" inline space of a specified minimum length, e.g. we might be doing a GET and we want at least a buffer of 1K to hold the result inline. We might not use the whole thing, but we want the ability to. This is why it takes a minLength.

libs/server/Objects/ItemBroker/CollectionItemBroker.cs
```diff
@@ -668,6 +668,7 @@ private unsafe bool TryGetResult(byte[] key, StorageSession storageSession, Resp
             }
             finally
             {
+                storageSession.scratchBufferBuilder.RewindScratchBuffer(asKey);
```
vazois, line 671: do we need to put everything between 
525: var asKey = storageSession.scratchBufferBuilder.CreateArgSlice(key); and the beginning of try{} inside the try so this triggers always?

vazois (approved):

libs/server/Objects/ItemBroker/CollectionItemBroker.cs
badrishc: The exception propagates up to TryConsumeMessages where the finally will reset the buffers.

PR #1679, opened 2026-04-07
Fix snapshot flush and checkpoint exception propagation [dev]

badrishc:
### Fix snapshot flush ObjectIdMap use-after-free and checkpoint exception propagation

Root cause: During snapshot checkpoint flush, ObjectAllocatorImpl.WriteAsync accessed the page's ObjectIdMap after releasing epoch protection. Under low memory with small segments, HeadAddress could advance during epoch gaps, causing page eviction which clears the ObjectIdMap (tail = -1). Subsequent objectIdMap.Get() hit a Debug.Assert failure, throwing an unhandled exception in FlushRunner. Since FlushRunner ran via Task.Run, the exception was unobserved, the completion semaphore was never released, and the checkpoint hung indefinitely.

Fixes:
- ObjectAllocatorImpl: Capture objectIdMap under epoch protection with a HeadAddress guard. Skip objectIdMap dereference (KeyOverflow, ValueOverflow, ValueObject) when HeadAddress advances past the current record, marking the record invalid instead. All objectIdMap accesses now occur strictly under epoch protection.
- AllocatorBase: Change AsyncFlushPagesForSnapshot and AsyncFlushDeltaToDevice to return Task instead of SemaphoreSlim, so exceptions in FlushRunner propagate naturally through the Task to the state machine.
- StateMachineDriver: Unify waiting list as List<(StateMachineTaskType, Task)>. Add StateMachineTaskType enum for identifying faulting tasks in error logs. ProcessWaitingListAsync logs the task type before re-throwing exceptions.
- AsyncFlushPageForSnapshotCallback: Move result.Release() into finally block to ensure release even if dirty-bit clearing throws.

Test fixes:
- ClusterTestUtils.Checkpoint: Retry on 'checkpoint already in progress' (race with on-demand checkpoint from replication).
- ClusterTestContext.AttachAndWaitForSync: Fix loop bounds bug (was i < replica_count, should be i < replicaStartIndex + replicaCount). Improve parameter names for clarity.
- ClusterReplicationBaseTests: Use Task.WhenAll for parallel task waiting with diagnostic status on timeout.


libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -1573,7 +1581,7 @@ protected void ShiftFlushedUntilAddress()
                     flushEvent.Set();
 
                     if ((oldFlushedUntilAddress < notifyFlushedUntilAddress) && (currentFlushedUntilAddress >= notifyFlushedUntilAddress))
-                        _ = notifyFlushedUntilAddressSemaphore.Release();
+                        _ = notifyFlushedUntilAddressTcs?.TrySetResult(true);
```
vazois, line 1585: Is this ensured to be triggered at most once? I am wondering why it is gated through the clause. I am worried that there is no error, but this never triggers because notifyFlushedUntilAddress is outside the range.

vazois (approved):
LGTM!

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
badrishc: TrySetResult is idempotent - the first call wins, subsequent ones return false, so the semantics should be the same as before.

PR #1691, opened 2026-04-10
Fixes for migration reliability

kevin-montrose:
Five separate fixes here, aimed at improving Vector Set migration test reliability.

## 1. Async all migration tasks

Rarely (but not so rarely it doesn't happen in CIs) migrations can deadlock on [this `.Result`](https://github.com/microsoft/garnet/blob/037e4a78be93968724357cf80d2d9445ceaf5569/libs/cluster/Server/Migration/MigrateSessionCommonUtils.cs#L121-L122).

I suspect, but cannot prove, this is because we block an IOCP thread that would serve the _read_ side of the request (where `MIGRATE ...` would be executed) on the destination node.  If I'm right, this means we only suffer this in tests where we host all the Garnet nodes in the same process.

The fix is to punch async/await all the way down so we don't have to `.Result` there.

## 2. Handle `TransmitSlot` failures

There are a number of places where we ignored the result of a `Task<bool>` or `Task<bool>[]` in migrations.  This is unlikely to cause test flakiness directly (they'd still fail if the task failed) but made debugging very painful.

## 3.  Pulled in a separate fix to `SessionParseState`

I never this directly cause an issue, but it's a issue reported separately in the same general area of migration/replication.

## 4. Removed pointers from `LightEpoch.InstanceIndexBuffer`

If you'd asked me last week if [the code in `InstanceIndexBuffer.GetRef`](https://github.com/microsoft/garnet/blob/037e4a78be93968724357cf80d2d9445ceaf5569/libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs#L40) is safe I'd have said yes.  I would expect `[ThreadStatic]` `struct` to be implicitly pinned.  But, rarely, I have seen some evidence that it was relocated.  Nothing definitive but out of an abundance of caution I've redone it in equivalent but safe-ish code.

Safe-ish because I'm still doing a `Unsafe.Add`, but we never have a pointer in a local so any movement between `ref field0` and `Unsafe.Add` will be correctly updated by the GC.

At least in .NET 8 the codegen for the two versions is identical -the only difference is the safe-ish version has no GC hole.

## 5. Pulled in a separate fix to `GarnetServer.Dispose`

Another (and more likely) explanation for the `LightEpoch` issue has to do with incorrect dispose ordering in `GarnetServer`.  Pulled that fix in as well.

## TODOs

 - [ ] Stress harder
 - [ ] Remove debugger/logging/etc.
 - [ ] Audit async/await changes for places to remove allocations


libs/client/LightEpoch.cs
```diff
@@ -19,7 +20,7 @@ public sealed unsafe class LightEpoch
         /// (1) in AssignInstance, to assign a unique instanceId to each LightEpoch instance, and
         /// (2) in Metadata, to track per-thread epoch table entries for each LightEpoch instance.
         /// </summary>
-        [StructLayout(LayoutKind.Explicit, Size = MaxInstances * sizeof(int))]
+        [InlineArray(MaxInstances)]
```
badrishc, line 23: If this is not proven to be necessary, we can retain the prior version. I initially did InlineArray but noted some inefficiencies which led to the current struct design.

PR #1702, opened 2026-04-14
Add INFO COMMANDSTATS tracking with per-command success rates

Mathos1432:
# Why is this change being made?

Garnet currently has no way to track per-command usage statistics. Redis provides INFO COMMANDSTATS which reports how many times each command was called, how many failed, and how many were rejected. This is critical for production observability — operators need to understand command usage patterns, identify failing commands, and monitor success rates.

# What changed?

Adds Redis-compatible INFO COMMANDSTATS support, gated behind --commandstats-monitor (disabled by default, zero overhead when off). When enabled, tracks per-command calls, failed_calls, and rejected_calls.

## Key design decisions:

- Counter-only tracking — No latency (usec) on the hot path. Reports usec=0,usec_per_call=0.00 for Redis format compatibility. Latency can be added later via sampling.
- Per-session single-writer — Each session owns its own CommandStats array indexed by RespCommand enum (O(1) access). No locking on the hot path; aggregation happens on-demand when INFO COMMANDSTATS is requested.
- Error detection — Dual strategy: commandErrorWritten flag set by WriteError/AbortWithErrorMessage, plus RESP - prefix check guarded by a bufferFlushedDuringCommand flag.
- Excluded from default INFO — Only appears when explicitly requested via INFO COMMANDSTATS.

- Bug fix: GarnetServerMonitor.Dispose() deadlocked when Start() was never called (the ManualResetEvent was never signaled). This affected EmbeddedRespServer used by BDN benchmarks. Fixed by tracking a started flag.

# How was this validated?

## Unit tests (17 passing)

- 8 new RespCommandStatsTests: DisabledByDefault, CallsTracking, FailedCalls, UsecFieldsZero, MultipleCommands, SuccessRate, InfoServerShowsMonitorStatus, FormatMatchesRedisConvention
- 9 existing metrics/info tests verified unbroken

## Performance benchmarks (BenchmarkDotNet)
 - ACL — Server started with an ACL file (authentication enabled)
 - AOF — Server started with Append-Only File (persistence enabled)
 - None — Bare server, no ACL or AOF
 
### PING (100 commands/op, .NET 10, Server GC)
 | Params | Disabled | Enabled | Overhead |
 |--------|----------|---------|----------|
 | ACL    | 1.687 μs | 1.820 μs | **+7.9%** |
 | AOF    | 1.680 μs | 1.794 μs | **+6.8%** |
 | None   | 1.690 μs | 1.816 μs | **+7.5%** |
 
 PING is ~17ns/command, so ~1.3ns tracking overhead is proportionally large. This is the worst case.
 
### SET (100 commands/op, .NET 10, Server GC)
 
 | Params | Disabled | Enabled | Overhead |
 |--------|----------|---------|----------|
 | ACL    | 13.10 μs | 13.18 μs | **+0.6%** |
 | AOF    | 21.13 μs | 20.62 μs | noise    |
 | None   | 12.62 μs | 13.03 μs | **+3.2%** |
 
### GET (100 commands/op, .NET 10, Server GC)
 
 | Params | Disabled | Enabled | Overhead |
 |--------|----------|---------|----------|
 | ACL    | 14.93 μs | 14.65 μs | noise    |
 | AOF    | 14.70 μs | 14.68 μs | noise    |
 | None   | 14.24 μs | 14.82 μs | **+4.1%** |
 
 **Summary:** For real-world commands (SET/GET), overhead is **0-4%** and within measurement noise for most configurations. Disabled by default = zero overhead unless opted in.


libs/server/Resp/RespServerSession.cs
```diff
@@ -619,6 +647,21 @@ private void ProcessMessages()
...
+
+                        // Track per-command stats: calls, failures, latency
+                        if (commandStats != null)
+                        {
+                            commandStats.IncrementCalls(cmd);
+
+                            // Detect error response via flag (set by WriteError/AbortWithErrorMessage),
+                            // or check RESP '-' error prefix at response start when buffer wasn't flushed
+                            bool failed = commandErrorWritten;
+                            if (!failed && !bufferFlushedDuringCommand &&
```
vazois: I think every error that we return as a response should go through RespWriteUtils.
If so, can we increment the failure there instead of rechecking here or performing a complex check here?

libs/server/Resp/RespServerSession.cs
```diff
@@ -597,6 +617,14 @@ private void ProcessMessages()
                 {
                     var noScriptPassed = true;
 
+                    // Capture state for command stats tracking
+                    byte* responseStartBefore = commandStats != null ? dcurr : null;
+                    if (commandStats != null)
```
vazois: Maybe you can try this pattern and see if it makes a difference with the BDN penalty

internal interface IStartupConfig
{
    static abstract bool Enabled { get; }
}

internal struct CommandStatsEnabled : IStartupConfig
{
    public static bool Enabled => true;
}

internal struct CommandStatsDisabled : IStartupConfig
{
    public static bool Enabled => false;
}

libs/server/Resp/RespServerSession.cs
```diff
@@ -597,6 +617,14 @@ private void ProcessMessages()
...
+                    {
+                        commandErrorWritten = false;
```
vazois: You can potentially blindly initialize these and see if it makes a difference.
It should not affect correctness since we only use them when commandStats != null

libs/server/Metrics/GarnetServerMonitor.cs
```diff
@@ -20,7 +20,7 @@ internal enum EventType : byte
     internal sealed class GarnetServerMonitor
     {
         public readonly Dictionary<InfoMetricsType, bool>
-            resetEventFlags = GarnetInfoMetrics.DefaultInfo.ToDictionary(x => x, y => false);
+            resetEventFlags = Enum.GetValues<InfoMetricsType>().ToDictionary(x => x, y => false);
```
vazois, line 23: What is the reason behind this change?

vazois (approved):

libs/server/Metrics/GarnetServerMonitor.cs
Mathos1432: GetnInfoMetrics.DefaultInfo doesn't have InfoMetricsType.CommandStats in it because we do not want to return CommandStats by default, only when the user explicitly requests that information. This follows Redis behavior.

However, we still need to be able to reset the counter values for CommandStats, so this Dictionary needs to be updated. Using Enum.GetValues does mean we'll pull all values instead, which means these additional entries will be in the dictionary, but that seems like a negligible cost: 
- STOREHASHTABLE
- OBJECTSTOREHASHTABLE
- STOREREVIV
- OBJECTSTOREREVIV
- HLOGSCAN

vazois: What was the previous behavior? If we didn't return these options before, we should not return them now because they might be options that result in scanning the whole KV and could be expensive to compute.
I think SERedis returns sections from the INFO command, and I don't know about other clients. In that case, we could be slowing down management operations.
Maybe we need to refactor this so we can have some sections in default and then at startup enable options that the user has enabled.
This means that we might need to make this available as instance object instead of static.

libs/server/Resp/RespServerSession.cs
Mathos1432: I tried testing this approach as well, but it seemed to have no effect on the BDN penalty. If you prefer that approach over the current one, let me know and I can make the update.

libs/server/Metrics/GarnetServerMonitor.cs
Mathos1432: My understanding is that we won't start returning these options with this change, we'll only add entries for them in the resetEventFlags dictionary. This dictionary tracks is for Config ResetStat calls. 

If they aren't used, then we're just wasting the 5entries worth of memory, with no functional difference from today.

libs/server/Resp/RespServerSession.cs
Mathos1432: Modifying RespWriteUtils would require changing a large amount of the code and plumbing context that probably doesn't belong in the write utils.
I've however updated the check to be simpler(based on copilot's comment), so this no longer requires the complex check that used to be here.
Let me know if this seems acceptable.

vazois (approved):

vazois (approved):

PR #1714, opened 2026-04-17
Remove sync-over-async where possible, consolidate blocking into helpers, add analyzers

kevin-montrose:
Doing a pass through Garnet to remove blocking where possible, document where it's required, and prevent regression in the future.

There are broadly three places where blocking is still required:
 1. Startup
 2. Any of the "on the IO thread" command handling code, marked with `Network` prefix
 3. Callbacks during scan operations

Major changes:
 - All `Task`-like blocking operations in Garnet proper are moved to [`AsyncUtils`](https://github.com/microsoft/garnet/blob/fb3b59e8341fc3c653424d4a6165e1aa0fdefcb7/libs/common/AsyncUtils.cs)
 - `Task`-like returning methods conform to .NET async naming rules now
 - `.ConfigureAwait(false)` added where missing
 - Where possible, blocking operations converted to `async/await`
 - [`Microsoft.VisualStudio.Threading` analyzers](https://microsoft.github.io/vs-threading/analyzers/index.html) added to flag future violations

Tsavorite and Garnet.client were not updated as aggressively to limit breaking changes.

TODO:
 - [x] Consolidate `.GetAwaiter().GetResult()` calls into helpers that justify them
 - [x] Get builds flagging naming (that is, missing `Async` suffix) violations
 - [x] Get builds flagging blocking calls
 - [x] `dev` targeting version - it's unlikely this can be cherry-picked cleanly, so it'll be non-trivial effort

`dev` version is #1730


hosting/Windows/Garnet.worker/Garnet.worker.csproj
```diff
@@ -8,6 +8,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\libs\host\Garnet.host.csproj" />
     <ProjectReference Include="..\..\..\libs\server\Garnet.server.csproj" />
+		<PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="all" IncludeAssets="analyzers"/>
```
badrishc: nit: fix indent

libs/common/Networking/NetworkHandler.cs
```diff
@@ -204,8 +204,7 @@ async Task AuthenticateAsServerAsync(SslServerAuthenticationOptions tlsOptions,
         }
 
         /// <summary>
-        /// Begin (background) network handler (including auth). Make sure you do not send data
-        /// until authentication completes.
+        /// Begin (background) network handler (including auth).
```
badrishc, line 210: This was designed to be in the background, but it seems now it completes synchronously. We need to see what are the implications of this contract change.

kevin-montrose: Only client auth finishes synchronously - reads still don't block (note the discard in `AuthenticateAsClientAsync`).  That's not super obvious though.

IIRC this was necessary because other code assumes auth has finished (and used `StartAsync().Wait()` or something to guarantee it).  That might not be true anymore as the async-ification went in phases... lemme play with it.  I can presumably update any dependent code to wait for auth to finish instead, and restore the async nature of authentication.

kevin-montrose: Yeah, blocking is still necessary for a number of callers as they unconditionally start executing commands after calling `Start()`.

A subtlety is that we can't use `SslStream.AuthenticateAsClient()` (non-async) because it relies on `Read()` which isn't implemented.

I'll move blocking off to the callers (where it was before), since it was already there.

badrishc: > they unconditionally start executing commands after calling Start()

This was designed to be resilient to starting commands, as the auth would happen in the background and the actual data operation would block lazily:

```cs
            // Wait for SslStream async processing to complete, if any (e.g., authentication phase)
            while (readerStatus == TlsReaderStatus.Active)
                await expectingData.WaitAsync(cancellationTokenSource.Token).ConfigureAwait(false);
```

> A subtlety is that we can't use SslStream.AuthenticateAsClient() (non-async) because it relies on Read() which isn't implemented.

Correct, that is why we use the SSL async API everywhere.

badrishc (approved):

PR #1721, opened 2026-04-21
Throw exception when AofSizeLimit is set with AOF disabled

vazois:
Adds a validation check in Options to throw an exception if AofSizeLimit is configured while AOF is disabled, preventing misconfiguration.


libs/host/Configuration/Options.cs
```diff
@@ -783,6 +783,12 @@ endpoint is IPEndPoint listenEp && clusterAnnounceEndpoint[0] is IPEndPoint anno
                     throw new Exception("SlowLogThreshold must be at least 100 microseconds.");
             }
 
+            if (!EnableAOF.GetValueOrDefault())
+            {
+                if (!string.IsNullOrEmpty(AofSizeLimit))
+                    throw new Exception("AofSizeLimit cannot be enforced with disabled AOF!");
```
badrishc: GarnetException?

badrishc (approved):

PR #1722, opened 2026-04-21
Wait for Vector Set replay in tests

kevin-montrose:
When `WaitForReplicaAofSync` is called in a test, also call `VectorManager.WaitForVectorOperationsToComplete()` on relevant replica.

`dev` equivalent #1723 1723


test/Garnet.test.cluster/ClusterTestUtils.cs
```diff
@@ -3029,8 +3031,22 @@ public void WaitForReplicaAofSync(int primaryIndex, int secondaryIndex, ILogger
                 BackOff(cancellationToken: context.cts.Token, msg: $"[{endpoints[primaryIndex]}]: {primaryMainStoreVersion},{primaryReplicationOffset} != [{endpoints[secondaryIndex]}]: {replicaMainStoreVersion},{secondaryReplicationOffset1}");
             }
             logger?.LogInformation("[{primaryEndpoint}]{primaryReplicationOffset} ?? [{endpoints[secondaryEndpoint}]{secondaryReplicationOffset1}", endpoints[primaryIndex], primaryReplicationOffset, endpoints[secondaryIndex], secondaryReplicationOffset1);
+
+            WaitForVectorSetReplayQuiescence(secondaryIndex, logger);
```
badrishc: Why is this on the test side - shouldn't the replica be updating the replicated-until address only after its successfully applied to the replica state? otherwise, the same issue might hit any client who uses the addresses to determine if the primary and replica are synced up.

kevin-montrose: While vector sets make it worse, in general it is difficult for a client to know a replica is caught up to a primary - that's sort of the nature of Redis cluster.

Lemme play with some options here - I think injecting a pause in a few places might be sufficient.  If not, a pretty substantial refactor of AOF replays might be needed.

badrishc: Okay. This fix should unblock the failing tests so we can merge and consider this catch up issue separately.

kevin-montrose: I think I have an approach in latest that's not too bad - need a couple more tweaks to keep perf decent in high vector set cases, but should have that worked out shortly.

badrishc (approved):

PR #1731, opened 2026-04-23
[BFTree] Add RangeIndex cluster migration

tiagonapoli:
## Summary

Adds end-to-end cluster migration for RangeIndex keys, supporting both `MIGRATE SLOTS` and `MIGRATE KEYS` paths. RangeIndex keys are backed by a native BfTree whose on-disk state (`data.bftree`) lives outside Tsavorite — shipping just the 51-byte stub is insufficient. This change streams the entire BfTree snapshot file alongside the stub over the existing migration transport.

## Architecture

- **Serializer** (`RangeIndexChunkedSerializer`): Pure state machine over `Span<byte>` — no I/O. Takes file data as input via `MoveNext(dest, fileData, out consumed)`.
- **Migration Reader** (`RangeIndexMigrationReader`): Async wrapper that reads the snapshot file and feeds bytes to the serializer.
- **Deserializer** (`RangeIndexChunkedDeserializer`): Sync state machine that writes received file data to a temp file, validates xxHash64 checksum, recovers the native BfTree, and publishes the stub to the store.
- **Factory** (`RangeIndexManager.SnapshotRangeIndexAndCreateReader`): Snapshots the BfTree under an exclusive lock to a temp file, then creates the reader.

## Wire format

Single `MigrationRecordSpanType.SerializedRangeIndexStream` (tag 4). Stream format across chunks:

`
[4B keyLen][key bytes][8B fileSize][file bytes][8B xxHash64][4B stubLen][stub]
`

Key and file bytes may span chunks; all other elements must fit within a single chunk.

## SLOTS path

1. Scan detects RI keys via `RecordType == 2`, captures to `MigrateOperation.RangeIndexes`
2. After scan completes, `MigrateRangeIndexKeysAsync` runs a sketch-protected batch cycle:
   - Add all RI keys to sketch (INITIALIZING)
   - TRANSMITTING + epoch barrier — blocks writes during snapshot+transmit
   - Transmit each key via `TransmitRangeIndexAsync`
   - DELETING + epoch barrier — blocks reads+writes during delete
   - Delete each key
   - `finally`: clear sketch (unblocks clients)

## KEYS path

1. `GetRangeIndexKeysForMigration` discovers RI keys via RIGET
2. `TransmitKeysAsync` skips RI keys (in `rangeIndexKeysToIgnore`)
3. Each RI key transmitted via `TransmitRangeIndexAsync`, then marked in sketch for `DeleteKeysAsync`

## Bug fixes

- Remove RI+cluster startup guard (`GarnetServer.cs`)
- Fix round-trip migration: `Publish` deletes existing data file before move
- Fix `Publish` registration: accept `InPlaceUpdated`/`CopyUpdated` status
- Fix serializer empty-buffer bug in FileData phase
- `TransmitRangeIndexAsync` catches all exceptions (never throws)
- Sketch protection for RI keys in SLOTS path (previously unprotected)

## Tests

- **26 unit tests** for serializer/deserializer (round-trip, checksums, error states, buffer boundaries, chunk sizes)
- **11 cluster integration tests**: SingleBySlot, ByKeys, ManyBySlot, WhileModifying, MigrateBack, LargeTree (1KB/256KB chunks), ChunkSize variants, StressAsync (Explicit)

## TODO

- Vector Set index keys have the same sketch protection gap (documented in code)
- AOF replication of migrated trees to destination replicas


vazois:
## Missing RangeIndex feature-enabled guards

This PR doesn't consistently check whether the RangeIndex feature is enabled (i.e., whether `RangeIndexManager` is non-null) before using it. Two spots are at risk of `NullReferenceException`:

### 1. Source side — `MigrateSession.RangeIndex.cs:29`

```csharp
var rangeIndexManager = clusterProvider.storeWrapper.DefaultDatabase.RangeIndexManager;
// No null check before calling .SnapshotRangeIndexAndCreateReader(...)
```

The KEYS path correctly uses `?.GetRangeIndexKeysForMigration(...)` with a fallback to an empty dictionary, but `TransmitRangeIndexAsync` assumes the manager is always available. If the SLOTS scan somehow discovers an RI key on a node where the feature is disabled, this will crash.

### 2. Receive side — `RespClusterMigrateCommands.cs:181`

```csharp
if (!rangeIndexMigrationState.ProcessRecord(...))
```

`rangeIndexMigrationState` is set to `null` when `RangeIndexManager` is null (line 70 in `ClusterSession.cs`), but the dispatch at line 181 doesn't guard against that. If a receiving node has RangeIndex disabled and the sender transmits a `SerializedRangeIndexStream` record, this will throw.

**Suggestion:** Add a null check on the receive side (e.g., log a warning and skip/error the record), and consider a defensive null check in `TransmitRangeIndexAsync` as well.

libs/cluster/Server/Migration/MigrateOperation.cs
```diff
@@ -127,7 +135,7 @@ public async Task<bool> TransmitSlotsAsync()
                 return true;
             }
 
-            public async Task<bool> TransmitKeysAsync(Dictionary<byte[], byte[]> vectorSetKeysToIgnore)
+            public async Task<bool> TransmitKeysAsync(Dictionary<byte[], byte[]> vectorSetKeysToIgnore, Dictionary<byte[], byte[]> rangeIndexKeysToIgnore)
```
vazois: Is it possible to combine these two instead of having to maintain two dictionaries?

libs/cluster/Server/Migration/MigrateOperation.cs
```diff
@@ -285,6 +293,19 @@ public void DeleteVectorSet(PinnedSpanByte key)
...
+            /// Delete a RangeIndex after migration if _copyOption is not set.
+            /// </summary>
+            public void DeleteRangeIndex(PinnedSpanByte key)
+            {
+                if (session._copyOption)
+                    return;
+
+                var delRes = localServerSession.BasicGarnetApi.DELETE(key);
+
+                session.logger?.LogDebug("Deleting RangeIndex {key} after migration: {delRes}", System.Text.Encoding.UTF8.GetString(key), delRes);
```
vazois: nit: logging message is confusing. Should be 'Deleted..' otherwise move it before actual delete.

vazois (commented):
**libs/cluster/Server/Migration/MigrateScanFunctions.cs:53** — maybe not for this PR, but can't we just get these values from an enum?

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                foreach (var (key, _) in rangeIndexKeys)
+                {
+                    var pinnedKey = PinnedSpanByte.FromPinnedSpan(key);
+                    mo.DeleteRangeIndex(pinnedKey);
+                }
+
+                return true;
+            }
+            finally
```
vazois, line 160: maybe catch and log a possible exception here. ensure to re-throw.

libs/cluster/Server/Migration/MigrateSessionKeys.cs
```diff
@@ -35,8 +35,13 @@ private async Task<bool> MigrateKeysFromStoreAsync()
                 await WaitForConfigPropagationAsync().ConfigureAwait(false);
 
                 // Discover Vector Sets linked namespaces
+                var allKeys = migrateTask.sketch.Keys.Select(t => t.Item1.ToArray());
```
vazois: can we avoid the copy here and just iterate over the container?

vazois (commented):
**libs/cluster/Server/Migration/MigrateSessionKeys.cs:43** — seems really wasteful to maintain a separate dictionary just for skipping keys. We can potentially store the info for the key type inline and skip the key once we read it from the sketch list

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
```diff
@@ -0,0 +1,94 @@
...
+                }
+
+                Reset();
+            }
+
+            return true;
+        }
+
+        /// <summary>Reset state for the next key stream.</summary>
+        private void Reset()
```
vazois, line 119: Ensure that the reset does not race with back to back Migration calls or parallel sessions on different slots

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// <summary>
+    /// Per-<see cref="ClusterSession"/> state for receiving inbound RangeIndex migration data.
+    /// Implements a state machine: IDLE → RECEIVING → IDLE.
+    /// </summary>
+    /// <remarks>
+    /// Because the sender uses a single TCP connection, all <c>CLUSTER MIGRATE</c> commands
+    /// from one migration arrive on the same <see cref="ClusterSession"/>, guaranteeing
+    /// in-order delivery.
+    /// </remarks>
+    internal sealed class RangeIndexMigrationReceiveState : IDisposable
```
vazois, line 20: Since this implementation is specific to RangeIndex, can you rename the file to indicate that? The current filename `MigrationReceiveSession.cs` is too generic.

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
tiagonapoli: Each ClusterSession has a single RangeIndexMigrationReceiveSession - within a single ClusterSession, processing is single threaded right?

libs/cluster/Server/Migration/MigrateScanFunctions.cs
```diff
@@ -50,7 +50,12 @@ public bool Reader<TSourceLogRecord>(in TSourceLogRecord srcLogRecord, RecordMet
                     // Check if key belongs to slot that is being migrated and if it can be added to our buffer
                     if (migrateOperation.Contains(slot))
                     {
-                        if (srcLogRecord.RecordType == VectorManager.RecordType)
+                        if (srcLogRecord.RecordType == RangeIndexManager.RangeIndexRecordType)
+                        {
+                            // RangeIndex keys need out-of-band migration (snapshot + chunks)
+                            migrateOperation.EncounteredRangeIndex(key.ToArray());
```
vazois: nit; AddRangeIndexKey

test/cluster/Garnet.test.cluster.migrate.rangeindex/ClusterRangeIndexMigrateTests.cs
```diff
@@ -0,0 +1,1631 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test.cluster
+{
+    [TestFixture, NonParallelizable]
+    public class ClusterRangeIndexMigrateTests : TestBase
```
vazois, line 25: Can we move this under cluster.rangeindex project to avoid bloating the cluster tests?

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
```diff
@@ -0,0 +1,174 @@
...
+                        unsafe
+                        {
+                            fixed (byte* keyPtr = key)
+                                migrateOperation.DeleteRangeIndex(PinnedSpanByte.FromPinnedPointer(keyPtr, key.Length));
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        logger?.LogError(ex, "MigrateRangeIndexKeysAsync: failed to delete RangeIndex key {key} after migration", Encoding.UTF8.GetString(key));
+                        throw;
```
vazois, line 154: I am wondering what is the contract here? If we throw in between deleting keys then the whole operation cancels and presumably some keys have been migrated.
I think at least we should come up with a contract on how to deal with this situation, because the transmitted keys are unreachable as long as the slot is not owned by the target node.

libs/cluster/Session/ClusterSession.cs
```diff
@@ -65,6 +76,10 @@ public ClusterSession(ClusterProvider clusterProvider, TransactionManager txnMan
             this.vectorBasicContext = vectorBasicContext;
             this.networkSender = networkSender;
             this.logger = logger;
+
+            var riManager = clusterProvider.storeWrapper.DefaultDatabase.RangeIndexManager;
+            if (riManager != null)
```
vazois: nit; maybe use the EnableRangeIndexPreview option here to ensure that this path is eliminated when eventually we eliminate the preview phase.

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
vazois: What if Dispose is being called during shutdown?

libs/cluster/Session/RespClusterMigrateCommands.cs
```diff
@@ -135,6 +135,22 @@ void Process(BasicGarnetApi basicGarnetApi, byte[] input, bool replaceOption, bo
...
 
+                                // An error has occurred
+                                if (migrateState > 0)
+                                {
+                                    i++;
+                                    continue;
+                                }
+
+                                // Protocol enforcement: while receiving a RangeIndex stream, only SerializedRangeIndexStream records are valid
+                                if (rangeIndexMigrationState != null && rangeIndexMigrationState.IsReceiving && kind != MigrationRecordSpanType.SerializedRangeIndexStream)
```
vazois: nit; should we perform this check at line 180, only when if (kind == MigrationRecordSpanType.SerializedRangeIndexStream) is satisfied?

libs/cluster/Server/Migration/RangeIndex/MigrateRangeIndexMetrics.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Garnet.cluster
+{
+    /// <summary>
+    /// Metrics collected during <see cref="MigrateSession.MigrateRangeIndexKeysAsync"/>.
+    /// </summary>
+    internal struct MigrateRangeIndexMetrics
```
vazois: nit; Is it possible to consolidate all these metrics?

libs/server/Resp/RangeIndex/RangeIndexChunkedDeserializer.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            switch (state)
+            {
+                case State.Error:
+                case State.Complete:
+                case State.Disposed:
+                    return false;
+
+                case State.WaitingForKeyHeader:
+                    if (data.Length == 0)
```
vazois, line 95: nit; is this a terminating message? can you add some comments here

libs/server/Resp/RangeIndex/RangeIndexChunkedDeserializer.cs
```diff
@@ -0,0 +1,251 @@
...
+            hasher.Append(filePart);
+            fileBytesRemaining -= count;
+            data = data[count..];
+        }
+
+        private void CloseStream()
+        {
+            if (stream == null) return;
+            stream.Flush();
+            stream.Dispose();
```
vazois: make sure Dispose executes even if Flush throws

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -212,6 +218,21 @@ public RangeIndexManager(string riLogRoot, string cprDir = null,
...
+            migrationTempDir = Path.Combine(riLogRoot, "migration-tmp");
+            if (Directory.Exists(migrationTempDir))
+            {
+                try
+                {
+                    Directory.Delete(migrationTempDir, recursive: true);
+                }
+                catch (Exception ex)
+                {
+                    logger?.LogWarning(ex, "Failed to clean up migration temp directory: {Path}", migrationTempDir);
```
vazois: should we throw if we fail to cleanup? Seems that after that we attempt to create which will probably fail right?

libs/cluster/Server/Migration/RangeIndex/MigrateRangeIndexMetrics.cs
tiagonapoli: I moved them to a single file to reduce bloat, but since they are tracing different operations, if we consolidate we would have un-used fields

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
```diff
@@ -0,0 +1,174 @@
...
+            var rangeIndexManager = clusterProvider.storeWrapper.DefaultDatabase.RangeIndexManager;
+            if (rangeIndexManager == null)
+            {
+                logger?.LogError("TransmitRangeIndexAsync: RangeIndex feature is not enabled, skipping key {key}", Encoding.UTF8.GetString(keyBytes));
+                return false;
+            }
+
+            var sessionClient = migrateOperation.Client;
+            var buffer = ArrayPool<byte>.Shared.Rent(chunkSize);
+            var metrics = new TransmitRangeIndexActivityTrace { tsStart = Stopwatch.GetTimestamp() };
```
kevin-montrose: This pattern of having to set one (but not the rest) public fields feels real messy.

Maybe wrap this stuff up in methods - a static `StartTrace()` which sets `tsStart = Stopwatch.GetTimestamp()`, and then methods which set/update the other fields as necessary.  Less error prone to write, easier to read.

kevin-montrose (changes requested):

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
tiagonapoli: Agree, updated to a much better version

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
tiagonapoli: Added an atomic to ensure that we always dispose but not dispose while an operation is happening
There's a race condition possible where we're in a Networking thread that's running ProcessRecord and then if another networking thread identified the socket was closed and calls dispose on the ClusterSession, we might dipose while there's a ProcessRecord inflight. Now disposal will happen before or after

libs/cluster/Session/RespClusterMigrateCommands.cs
tiagonapoli: On this case the client should always send all `RangeIndex` chunks one after the other, without any other types in between, if we have a different type in between, that breaks protocol, for example, expectation is:

`RI_1_CHUNK_1 RI_1_CHUNK_2 RI_1_CHUNK_END LOG_RECORD`

If we receive:
`RI_1_CHUNK LOG_RECORD RI_1_CHUNK_2 RI_1_CHUNK_END`

that's unexpected protocol, since expectation is that the entire RI will be streamed first

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
tiagonapoli: Maybe as follow up work? I think that's currently a behavior that also happens for the regular keys, right?

Garnet.slnx
```diff
@@ -87,6 +87,7 @@
     <Project Path="test/standalone/Garnet.test.vectorset/Garnet.test.vectorset.csproj" />
     <Project Path="test/cluster/Garnet.test.cluster/Garnet.test.cluster.csproj" />
     <Project Path="test/cluster/Garnet.test.cluster.migrate/Garnet.test.cluster.migrate.csproj" />
+    <Project Path="test/cluster/Garnet.test.cluster.migrate.rangeindex/Garnet.test.cluster.migrate.rangeindex.csproj" />
```
vazois, line 90: Do forget to add this to the test Matrix in ci.yml

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            currentDeserializer?.Dispose();
+            currentDeserializer = null;
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            var prev = Interlocked.Or(ref state, FLAG_DISPOSED);
```
vazois: Is it possible to use some of the existing primitives for guarding against Dispose while operation is ongoing? e.g. ActiveWorkerMonitor, SingleWriterMultiReaderLock?

libs/cluster/Server/Migration/MigrateSession.RangeIndex.cs
vazois: Sure, as long as we keep track of these items

vazois (approved):

libs/cluster/Session/RangeIndexMigrationReceiveSession.cs
tiagonapoli: I'm worried about using ActiveWorkerMonitor or a Lock because if we have a ongoing processing while we Dispose we would block that thread while processing is happening - and Dispose code and Processing code both run in a Networking thread, so at that point we would block two networking threads for a bit of time - if the Processing thread only did CPU stuff I think would be fine, but it touches disk so I thought would be better to avoid

Garnet.slnx
tiagonapoli: Done — `ci.yml` line 162 now includes `'Garnet.test.cluster.migrate.rangeindex'` in the cluster test matrix, so it runs across os x {net8.0, net10.0} x {Debug, Release}. (Preserved through the latest rebase onto main.) Resolving.

test/standalone/Garnet.test/RangeIndexChunkedSerializerTests.cs
```diff
@@ -0,0 +1,1189 @@
...
+            while (!serializer.IsComplete)
+            {
+                var len = SerializerMoveNext(serializer, buffer, fs);
+                ClassicAssert.IsTrue(deserializer.ProcessChunk(buffer.AsSpan(0, len).ToArray()));
+            }
+
+            ClassicAssert.IsTrue(deserializer.IsComplete);
+
+            var tmpDir = Path.Combine(testDir, "rangeindex", ".migration-tmp");
+            var tmpFiles = Directory.GetFiles(tmpDir, "*.bftree");
```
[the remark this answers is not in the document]
tiagonapoli: Addressed: the tests no longer hard-code a path. They use `manager.DeriveTempMigrationPath()` (which resolves to `{riLogRoot}/migration-tmp/{guid}.bftree`) and assert against `Path.Combine(testDir, "migration-tmp")` — matching `RangeIndexManager` (no leading dot, no `rangeindex` subdir). Resolving.

kevin-montrose (approved):

badrishc (approved):

badrishc (approved):

kevin-montrose (approved):

Issue #86, opened 2024-03-22
security-scan found error SCS0005: Weak random number generator


springmin:
using the tools named "security-code-scan" to scan the solution ccode has found the security error SCS0005(Weak random number generator).   this cs file and position list :
1. libs\common\Generator.cs	(24,13)
2. libs\server\Objects\Set\SetObjectImpl.cs	(181,37)、(193,33):
3. libs\server\Objects\Hash\HashObjectImpl.cs	(273,46)、(275,60)、(299,33)
4. libs\server\Objects\SortedSet\SortedSetObjectImpl.cs	(683,38)、(686,52)
5. other test/benchmark/playground files.

Issue #85, opened 2024-03-22
Plans for an `IDistributedCache` implementation?


jodydonetti:
Hi all, are you planning on creating an `IDistributedCache` implementation anytime soon?

Since Garnet has been announced as being protocol-compatible with Redis I thought it would've worked right away by using the standard Redis implementation (see [here](https://github.com/dotnet/aspnetcore/blob/main/src/Caching/StackExchangeRedis/src/RedisCache.cs)), but after actually playing with it I discovered that Garnet does not currently [support LUA scripting](https://microsoft.github.io/garnet/docs/welcome/compatibility):

> 3. Garnet does not support Lua scripting. We have an experimental version, but it was noted to be too slow for realistic use so we have not added it to the project.

LUA scripting is what has been used in the aforementioned impl (see [here](https://github.com/dotnet/aspnetcore/blob/ec293ee75c0c022370951a459b188fa81ec8b7c3/src/Caching/StackExchangeRedis/src/RedisCache.cs#L206)) and therefore without it it's not possible to use Garnet this way.

Having a working impl of `IDistributedCache` for Garnet would drive the adoption a lot, since that is the standard way of using a distributed cache in .NET.

Also, that would unlock other great possibilities like being able to use Garnet in projects that use `IDistributedCache` as a _"lingua franca"_, like [FusionCache](https://github.com/ZiggyCreatures/FusionCache) (mine) which is compatible with any impl of such interface ([here](https://github.com/ZiggyCreatures/FusionCache/blob/main/docs/CacheLevels.md) for more).

All in all I think that, if the effort for such impl is not huge, that is something that you should consider to skyrocket the usage.

Thoughts?

ps: maybe @mgravell can give us some hints about such approach? Are there any hidden pitfalls that should be looked at?

mgravell:
I've created a ticket "over there ^^^" - can someone confirm whether Garnet has `MULTI`/`EXEC` support? this may influence any changes there

PaulusParssinen:
>... can someone confirm whether Garnet has `MULTI`/`EXEC` support? this may influence any changes there

Quickly looking at the code: yes, it support `MULTI`/`EXEC`.

badrishc:
More info here: https://microsoft.github.io/garnet/docs/extensions/transactions

mgravell:
I ended up just switching it to pipeline (no `multi`/`exec`) - https://github.com/dotnet/aspnetcore/pull/54689

jodydonetti:
[...]

PR #83, opened 2024-03-22
Create IHeyReally

Heyitsquoracom:

PR #1733, opened 2026-04-24
Fix 5 open Dependabot npm alerts (dompurify, uuid)

badrishc:
## Summary

Fixes all 5 open Dependabot security alerts in the website npm dependencies.

### dompurify 3.3.3 → 3.4.0 (fixes 4 alerts)
- **#82** ADD_TAGS function form bypasses FORBID_TAGS
- **#83** CVE-2026-41238: Prototype Pollution to XSS Bypass via CUSTOM_ELEMENT_HANDLING
- **#84** CVE-2026-41240: FORBID_TAGS bypassed by function-based ADD_TAGS predicate
- **#85** CVE-2026-41239: SAFE_FOR_TEMPLATES bypass in RETURN_DOM mode

### uuid 8.3.2/11.1.0 → 14.0.0 (fixes 1 alert)
- **#86** Missing buffer bounds check in v3/v5/v6 when buf is provided

### Validation
- `yarn install` succeeds
- `yarn build` succeeds — website builds cleanly

### Note on uuid 14.0.0
uuid 14 is ESM-only. The two consumers in the dependency tree are:
- **mermaid** (previously uuid ^11.1.0) — ESM, compatible
- **sockjs** (previously uuid ^8.3.2) — CJS, used only by webpack-dev-server during `docusaurus start` (dev mode). The sockjs code path only calls `uuid.v4()` which is unaffected by the vulnerability (v3/v5/v6 buffer bounds). If dev-server issues arise, the resolution can be scoped to `mermaid/uuid` only.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -40,8 +40,10 @@
     "lodash-es": "4.18.1",
     "//minimatch": "CVE-2024-21538, CVE-2023-36326, CVE-2023-34104 (ReDoS)",
     "minimatch": "3.1.5",
-    "//dompurify": "CVE-2025-26791 (XSS)",
-    "dompurify": "3.3.3"
+    "//dompurify": "CVE-2025-26791, CVE-2026-0540, CVE-2026-41238, CVE-2026-41239, CVE-2026-41240 (XSS / prototype pollution)",
+    "dompurify": "3.4.0",
+    "//uuid": "GHSA uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided",
+    "uuid": "14.0.0"
   },
   "browserslist": {
     "production": [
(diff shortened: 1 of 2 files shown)


badrishc:
Thanks for the review! I investigated the scoped `mermaid/**/uuid` approach, but Yarn v1 doesn't actually apply it — the lockfile still resolved `uuid@^11.1.0` to 11.1.0 with the scoped override.

However, the CJS compatibility concern is not an issue here because **Node.js 22+ natively supports `require()` of ESM-only packages** (the `--experimental-require-module` flag was removed and became default behavior). Since `engines` requires `node >= 22`, `sockjs`'s `require('uuid').v4` works fine with uuid 14.0.0.

I verified this:
- `node -e "const uuid = require('uuid'); console.log(uuid.v4());"` → works ✅  
- `yarn build` → succeeds ✅  
- `yarn start` (dev server with sockjs/webpack-dev-server) → starts cleanly ✅  

Updated to use the global `uuid: 14.0.0` resolution since the scoped approach doesn't work with Yarn v1 and Node.js 22+ handles the ESM/CJS interop.

vazois (approved):

PR #1741, opened 2026-04-27
Add Roaring Bitmap support (issue #1270)

### LINKED ISSUE #1270: Bitmap compression
### Feature request type

enhancement

It does not appear like garnet supports bitmap compression at the moment, but some highly efficient strategies exist that are loved by developers.

One such strategy is RoaringBitmaps https://vikramoberoi.com/posts/a-primer-on-roaring-bitmaps-what-they-are-and-how-they-work/

See also https://github.com/aviggiano/redis-roaring

### Additional context

_No response_

crprashant:
# Add Roaring Bitmap support (issue #1270)

Resolves #1270.

## Summary

Adds a Roaring Bitmaps extension to Garnet that introduces a new compressed
bitmap object type plus four `R.*` RESP commands. Implemented entirely as a
host extension in `main/GarnetServer/Extensions/RoaringBitmap/` — **zero
changes to `libs/server/`** — so this is a clean, reviewable foundation that
can be deepened in follow-up PRs.

## Why Roaring?

A naive `uint32` bitmap is 512 MiB. Roaring partitions the universe into
65 536 chunks of 65 536 bits and represents each chunk as either:

- **Array container** — sorted `ushort[]`, used while a chunk holds ≤ 4 096
  set bits (~`2·count` bytes).
- **Bitmap container** — `ulong[1024]` (8 KiB exactly), used once a chunk
  exceeds the threshold.

Empty chunks consume zero memory. Chunks promote (array → bitmap) and demote
(bitmap → array) automatically as cardinality changes.

## Commands

| Command | Description |
| --- | --- |
| `R.SETBIT key offset value` | Set bit at `offset` ∈ `[0, 2³²-1]` to `0`/`1`. Returns previous bit. |
| `R.GETBIT key offset` | Returns bit at `offset`. `0` for missing keys. |
| `R.BITCOUNT key` | Population count. `0` for missing keys. |
| `R.BITPOS key bit [from]` | First `bit` (`0`/`1`) at or after `from`. `-1` if none. |

Docs: [`website/docs/commands/roaring-bitmap.md`](website/docs/commands/roaring-bitmap.md).

## Design notes

- The data structure (`RoaringBitmap.cs`, `Containers/*`) is a pure C# library
  with no Garnet dependencies → independently unit-testable.
- `RoaringBitmapObject` (`CustomObjectBase`) wraps the structure and tracks
  size deltas via `bitmap.ByteSize` for the per-object size accounting.
- All four commands are registered as `CommandType.ReadModifyWrite`. The reads
  (`R.GETBIT` / `R.BITCOUNT` / `R.BITPOS`) do **not** mutate state, but the RMW
  path is required so that `NeedInitialUpdate` is invoked on missing keys —
  the framework's `Read` path simply returns nil otherwise. Missing-key
  responses (`0` / `-1`) are written from `NeedInitialUpdate` which then
  returns `false` to decline key creation.
- `NeedInitialUpdate` error paths use `writer.WriteError(...)` + `return false`
  rather than `AbortWithErrorMessage` (which returns `true` and would cause the
  framework to proceed into `InitialUpdater`/`Updater`, double-writing the
  response and corrupting the protocol stream).

## Tests

| Suite | Count | Status |
| --- | --- | --- |
| `RoaringBitmapDataTests` (pure data structure) | 27 | ✅ Pass |
| `RespRoaringBitmapTests` (RESP integration via SE.Redis) | 14 | ✅ Pass |

Coverage highlights:

- Empty bitmap, single bit, idempotent set/clear.
- Promotion threshold (`4096` ↔ `4097`) and demotion across both directions.
- 100 K random ops vs `HashSet<uint>` oracle.
- Boundary offsets `0`, `65535`, `65536`, `2³¹`, `2³²-1`.
- Serialize → deserialize round-trip equality (empty / sparse / dense / mixed).
- RESP-level: `R.SETBIT`/`R.GETBIT` parity with oracle, `R.BITCOUNT`,
  `R.BITPOS` (set / unset / `from` offset), large offsets, persistence across
  restart, concurrent setbits from multiple clients, error paths
  (bad offset, bad bit, bad value, wrong arity), and the two-store key
  separation property.

```text
$ dotnet test test\Garnet.test\Garnet.test.csproj -c Debug -f net8.0 \
    --filter "FullyQualifiedName~RoaringBitmap" --nologo
Passed!  - Failed: 0, Passed: 43, Skipped: 0, Total: 43
```

## Known limitations (intentional v1 scope)

- **Run container** is not implemented. Adds ~30% extra compression on
  contiguous ranges; the array/bitmap pair captures the bulk of real-world
  savings.
- **`R.BITOP AND/OR/XOR/NOT`** is not exposed. The data structure supports
  these natively; only command surface is needed.
- **Empty-key removal**: clearing the last bit leaves an empty bitmap object
  rather than removing the key. This is a property of the custom-object
  framework's tombstone path (`output.HasRemoveKey` is honoured only on the
  built-in path) and is best fixed in `libs/server/Storage/Functions/ObjectStore`
  in a separate PR.

## Files

```
main/GarnetServer/Extensions/RoaringBitmap/
  Containers/IContainer.cs
  Containers/ArrayContainer.cs
  Containers/BitmapContainer.cs
  RoaringBitmap.cs
  RoaringBitmapObject.cs
  RoaringBitmapCommands.cs
main/GarnetServer/Program.cs                    (registration only)
test/Garnet.test/RoaringBitmapDataTests.cs      (27 tests)
test/Garnet.test/RespRoaringBitmapTests.cs      (14 tests)
website/docs/commands/roaring-bitmap.md         (docs)
```


crprashant:
Thanks for the thorough review! Pushed [`65ac9f1d4`](https://github.com/crprashant/garnet/commit/65ac9f1d4) addressing each comment. Summary:

| # | File / line | Resolution |
| - | --- | --- |
| 1 | `RoaringBitmapDataTests.cs:175` — `bits[actual0]` index | **Applied.** `(int)actual0` cast plus an explicit `actual0 >= 0` guard. (Note: array indexers in C# do accept `long` per the language spec, which is why the original compiled — but the cast makes intent obvious and removes the runtime `OverflowException` risk.) |
| 2 | `RoaringBitmap.cs:189` — `ByteSize` doc/impl mismatch | **Applied.** Updated the XML doc to state that the per-entry `SortedDictionary` node overhead **is** included (the implementation was already correct; only the comment was misleading). |
| 3 | `RoaringBitmapCommands.cs:51` — `RSetBit.NeedInitialUpdate` always returns `true` | **Applied.** `NeedInitialUpdate` now validates `offset` and `value` against a copy of the `ObjectInput` (it's a `struct`, so `var validation = input;` snapshots it without disturbing what `Updater` sees). On bad input it writes the error and returns `false`, so a malformed `R.SETBIT` no longer creates an empty tombstone-style key. |
| 4 | `RoaringBitmapCommands.cs:6` — `using Garnet.common;` flagged as unused | **Disagreed (with evidence).** `RespMemoryWriter` lives in `Garnet.common`, and removing the using produces 8 × `CS0246: The type or namespace name 'RespMemoryWriter' could not be found` errors across all `NeedInitialUpdate` / `Updater` / `Reader` signatures. Restored. |
| 5 | `RoaringBitmapObject.cs:29` — inconsistent default-ctor `Size` | **Applied.** Default ctor now sets `this.Size = ObjectOverhead + bitmap.ByteSize`, matching the deserialized constructor so freshly-created and round-tripped objects report identical memory baselines and ByteSize-based mutation deltas don't double-count the empty-bitmap baseline. |

All 27 data-structure tests + 14 RESP integration tests still pass:

```
Passed!  - Failed: 0, Passed: 43, Skipped: 0, Total: 43
```

crprashant:

crprashant:

crprashant:

badrishc:
Thanks for your contribution! This extension is interesting, but instead of putting it in main, we should place it in https://github.com/microsoft/garnet/tree/main/modules (where e.g., GarnetJSON is kept) so that it is not bundled by default in the server.

badrishc:
Also, main is closed to new features, so we would request that you retarget your PR to the dev (v2) branch.

crprashant:
Thanks for the review! Both points addressed in the latest force-push:

1. **Moved extension into `modules/RoaringBitmap/`** — mirrors `GarnetJSON`: new `GarnetRoaringBitmap.csproj`, new `RoaringBitmapModule : ModuleBase` entry point that registers the factory + `R.SETBIT` / `R.GETBIT` / `R.BITCOUNT` / `R.BITPOS`, namespace renamed `Garnet.Extensions.RoaringBitmap` → `GarnetRoaringBitmap` to avoid the namespace/class collision, wired into `Garnet.slnx` and `test/Garnet.test/Garnet.test.csproj`. Nothing in `main/GarnetServer` references it anymore — it is no longer bundled by default.
2. **Retargeted PR base to `dev`.** Branch was rebased onto current `upstream/dev` and the `CustomObjectFunctions` / `CustomObjectBase` overrides updated to the new `scoped ReadOnlySpan<byte>` and `HeapMemorySize` APIs.

Local validation: all 43 RoaringBitmap tests (29 data + 14 RESP) pass on net8.0, `dotnet format` is clean.

modules/RoaringBitmap/GarnetRoaringBitmap.csproj
```diff
@@ -0,0 +1,23 @@
...
+
+  <PropertyGroup>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\libs\server\Garnet.server.csproj" />
+  </ItemGroup>
```
kevin-montrose, line 17: Let's also include our threading analyzers for consistency:
`<PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="all" IncludeAssets="analyzers"/>`

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+    /// algorithm (Chambi, Lemire, Kaser, Godin, 2014).
+    ///
+    /// The 32-bit element space is partitioned into 65,536 chunks of 65,536 elements
+    /// each, addressed by the high 16 bits of the element. Each chunk is stored as
+    /// either a sorted-array container (sparse, &lt;= 4096 elements) or a 8 KiB bitmap
+    /// container (dense). Empty chunks are not stored. Chunks are kept in
+    /// ascending order by high-key for deterministic serialization and efficient
+    /// scans (e.g., bit-position queries).
+    ///
+    /// This class is NOT thread-safe; the parent RoaringBitmapObject (added in a
```
kevin-montrose: Outdated.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            ushort hi = (ushort)(value >> 16);
+            ushort lo = (ushort)(value & 0xFFFF);
+            return chunks.TryGetValue(hi, out var c) && c.Contains(lo);
+        }
+
+        /// <summary>
+        /// Sets bit <paramref name="value"/> to 1. Returns the previous value (0 or 1).
+        /// </summary>
+        public int Add(uint value)
```
kevin-montrose: Strange to use an `int` for this, I'd switch to `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+            var updated = c.Add(lo, out bool added);
+            if (!ReferenceEquals(updated, c)) chunks[hi] = updated;
+            if (added) { totalCardinality++; return 0; }
+            return 1;
+        }
+
+        /// <summary>
+        /// Clears bit <paramref name="value"/>. Returns the previous value (0 or 1).
+        /// </summary>
+        public int Remove(uint value)
```
kevin-montrose: Same note, prefer `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+            if (!chunks.TryGetValue(hi, out var c)) return 0;
+            var updated = c.Remove(lo, out bool removed);
+            if (updated == null) { chunks.Remove(hi); }
+            else if (!ReferenceEquals(updated, c)) { chunks[hi] = updated; }
+            if (removed) { totalCardinality--; return 1; }
+            return 0;
+        }
+
+        /// <summary>Convenience wrapper used by RESP SETBIT: dispatches to <see cref="Add"/> or <see cref="Remove"/>.</summary>
+        public int SetBit(uint offset, bool set) => set ? Add(offset) : Remove(offset);
```
kevin-montrose: Same note, prefer `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        /// <summary>0 or 1 — the bit value at <paramref name="offset"/>.</summary>
+        public int GetBit(uint offset) => Contains(offset) ? 1 : 0;
```
kevin-montrose: Same note, prefer `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        /// <summary>
+        /// Position of the first bit equal to <paramref name="bit"/> at or after <paramref name="from"/>.
+        /// Returns -1 if no such bit exists in the uint32 universe at or after <paramref name="from"/>.
+        ///
+        /// For <c>bit==0</c>: scans chunks plus the gaps between/around them. Any unallocated
+        /// high-key represents a fully-unset 65,536-bit run, so the first unset bit may be
+        /// the first position in a missing high-key.
+        /// </summary>
+        public long BitPos(int bit, uint from = 0)
```
kevin-montrose: Same note, prefer `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            if (bit != 0 && bit != 1) throw new ArgumentOutOfRangeException(nameof(bit), "bit must be 0 or 1");
```
kevin-montrose: Unnecessary check if we use a `bool` here.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
+// Copyright (c) Microsoft Corporation.
```
kevin-montrose, line 1: As a general note, style here doesn't quite match the rest of Garnet.

In particular, prefer `var` (IDE0007 diagnostic) over explicit variable types.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+            ushort fromHi = (ushort)(from >> 16);
+            ushort fromLo = (ushort)(from & 0xFFFF);
+
+            if (bit == 1)
+            {
+                // Find the first chunk whose high-key >= fromHi that has a set bit at or after the low-key.
+                foreach (var kv in chunks)
```
kevin-montrose: Iterating over all chunks here feels wrong, we keep these sorted.

We should be able to find the write chunk in logarithmic time rather than linear time.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+                return -1;
+            }
+            else
+            {
+                // bit == 0: any missing high-key in [fromHi..65535] gives an instant answer.
+                // We walk in ascending high-key order. For each present chunk, check if it
+                // has an unset bit at/after the relevant low-key offset (mind chunk being full).
+                // For each *gap* (missing high-key), the first unset is at high << 16.
+                int expectedHi = fromHi;
+                foreach (var kv in chunks)
```
kevin-montrose: Same note, iterating is less than ideal here.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        /// <summary>Removes all bits.</summary>
+        public void Clear()
+        {
+            chunks.Clear();
+            totalCardinality = 0;
+        }
+
+        /// <summary>
```
kevin-montrose, line 224: Why not implement `IEnumerable<uint>` ?

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+                    yield return ((uint)hi << 16) | (uint)next;
+                    if (next == 65535) break;
+                    next = kv.Value.NextSetBit(next + 1);
+                }
+            }
+        }
+
+        public void Serialize(BinaryWriter writer)
+        {
+            if (writer == null) throw new ArgumentNullException(nameof(writer));
```
kevin-montrose: Prefer `ArgumentNullException.ThrowIfNull`.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+                writer.Write(kv.Key);
+                writer.Write((byte)kv.Value.Kind);
+                writer.Write(kv.Value.Cardinality);
+                kv.Value.SerializeBody(writer);
+            }
+        }
+
+        public static RoaringBitmap Deserialize(BinaryReader reader)
+        {
+            if (reader == null) throw new ArgumentNullException(nameof(reader));
```
kevin-montrose: Prefer `ArgumentNullException.ThrowIfNull`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        public ArrayContainer()
+        {
+            values = new ushort[InitialCapacity];
+            count = 0;
+        }
+
+        public ArrayContainer(ushort[] sortedUnique, int count)
+        {
+            if (sortedUnique == null) throw new ArgumentNullException(nameof(sortedUnique));
```
kevin-montrose: Prefer `ArgumentNullException.ThrowIfNull`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+        public int Cardinality => count;
+        public long ByteSize => 16 + 2L * values.Length; // array header ~16B + 2B per slot
+
+        public bool Contains(ushort value)
+        {
+            return BinarySearch(value) >= 0;
+        }
+
+        /// <summary>Returns the index of <paramref name="value"/>, or the bitwise complement of the insertion position.</summary>
+        private int BinarySearch(ushort value)
```
kevin-montrose: Use [`ReadOnlySpan.BinarySearch`](https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.binarysearch?view=net-10.0#system-memoryextensions-binarysearch-2(system-readonlyspan((-0))-1)), not a custom one.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+                if (cmp == 0) return mid;
+                if (cmp < 0) lo = mid + 1; else hi = mid - 1;
+            }
+            return ~lo;
+        }
+
+        public IContainer Add(ushort value, out bool added)
+        {
+            int idx = BinarySearch(value);
+            if (idx >= 0) { added = false; return this; }
```
kevin-montrose: style nit; don't put two statements on one line like this.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+            values[insert] = value;
+            count++;
+            added = true;
+            return this;
+        }
+
+        public IContainer Remove(ushort value, out bool removed)
+        {
+            int idx = BinarySearch(value);
+            if (idx < 0) { removed = false; return this; }
```
kevin-montrose: style nit; don't put two statements on one line like this.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+        public IContainer Clone()
+        {
+            var copy = new ushort[count]; // tight copy — clones don't preallocate growth
+            Array.Copy(values, copy, count);
+            return new ArrayContainer(copy, count);
+        }
+
+        public void SerializeBody(BinaryWriter writer)
+        {
+            for (int i = 0; i < count; i++)
```
kevin-montrose: Use `BinaryWriter.Write(byte[], int, int)` rather than this for loop.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+                writer.Write(values[i]);
+        }
+
+        public static ArrayContainer DeserializeBody(BinaryReader reader, int cardinality)
+        {
+            if (cardinality < 1 || cardinality > ArrayThreshold)
+                throw new InvalidDataException($"ArrayContainer cardinality out of range: {cardinality}");
+            var arr = new ushort[cardinality];
+            ushort prev = 0;
```
kevin-montrose: More efficient to use `BinaryReader.ReadExactly(MemoryMarshal.Cast<ushort, byte>(arr.AsSpan()))` (or similar) and check the monotonicity in a separate pass.

modules/RoaringBitmap/Containers/ArrayContainer.cs
```diff
@@ -0,0 +1,207 @@
...
+                if (i > 0 && v <= prev)
+                    throw new InvalidDataException("ArrayContainer values must be strictly ascending");
+                arr[i] = v;
+                prev = v;
+            }
+            return new ArrayContainer(arr, cardinality);
+        }
+
+        private void EnsureCapacity(int required)
+        {
```
kevin-montrose, line 179: This feels very confused - we never need more than 4096 (`ArrayThreshold`) elements so some of these checks feel wrong?

Something like:
```csharp

if (required <= values.Length) return;

var newCap = Math.Min(ArrayThreshold, (int)BitOperations.RoundUpToPowerOf2((uint)required));
Array.Resize(ref values, newCap);
```

Will grow iff needed, and grow in the pattern `1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096` which I think is roughly the intent here?

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        public BitmapContainer()
+        {
+            words = new ulong[WordCount];
+            cardinality = 0;
+        }
+
+        public BitmapContainer(ulong[] preExistingWords, int cardinality)
+        {
+            if (preExistingWords == null) throw new ArgumentNullException(nameof(preExistingWords));
```
kevin-montrose: Prefer `ArgumentNullException.ThrowIfNull`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        public IContainer Add(ushort value, out bool added)
+        {
+            int wordIdx = value >> 6;
+            int bitIdx = value & 0x3F;
+            ulong mask = 1UL << bitIdx;
+            ulong before = words[wordIdx];
+            ulong after = before | mask;
+            if (after == before) { added = false; return this; }
```
kevin-montrose: style nit; don't put two statements on one line like this.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+
+        public IContainer Remove(ushort value, out bool removed)
+        {
+            int wordIdx = value >> 6;
+            int bitIdx = value & 0x3F;
+            ulong mask = 1UL << bitIdx;
+            ulong before = words[wordIdx];
+            if ((before & mask) == 0) { removed = false; return this; }
```
kevin-montrose: style nit; don't put two statements on one line like this.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            words[wordIdx] = before & ~mask;
+            cardinality--;
+            removed = true;
+
+            // Demote to array container if we've crossed below the threshold.
+            // The threshold is reused but with NO hysteresis at this layer because the
+            // parent decides hysteresis when bouncing around the boundary.
+            if (cardinality == 0) return null;
```
kevin-montrose: This is impossible yeah?  We should never create a `BitmapContainer` that has 1 set bit.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (cardinality <= ArrayContainer.ArrayThreshold)
+                return ToArrayContainer();
+            return this;
+        }
+
+        public ushort First()
+        {
+            if (cardinality == 0) throw new InvalidOperationException("empty bitmap");
```
kevin-montrose: Similarly, empty `BitmapContainer`s should never occur yeah?

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            for (int i = 0; i < WordCount; i++)
```
kevin-montrose: Use `Span.IndexOfAnyExcept(0)` rather than scanning like this.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                if (words[i] != 0)
+                    return (ushort)((i << 6) + BitOperations.TrailingZeroCount(words[i]));
+            }
+            throw new InvalidOperationException("cardinality > 0 but no bits set");
+        }
+
+        public ushort Last()
+        {
+            if (cardinality == 0) throw new InvalidOperationException("empty bitmap");
```
kevin-montrose: Same note, is `BitmapContainer.cardinality==0` actually possible?

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            for (int i = WordCount - 1; i >= 0; i--)
```
kevin-montrose: Use `Span.LastIndexOfAnyExcept(0)` instead of this manual scan.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                if (words[i] != 0)
+                    return (ushort)((i << 6) + (BitsPerWord - 1 - BitOperations.LeadingZeroCount(words[i])));
+            }
+            throw new InvalidOperationException("cardinality > 0 but no bits set");
+        }
+
+        public int NextSetBit(int from)
+        {
+            if (from < 0) from = 0;
```
kevin-montrose: These two ifs cover illegal values right?  Why handle them at all?

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+                wordIdx++;
+                if (wordIdx >= WordCount) return -1;
+                word = words[wordIdx];
+            }
+        }
+
+        public int NextUnsetBit(int from)
+        {
+            if (from < 0) from = 0;
+            if (from > 65535) return -1;
```
kevin-montrose: Same note: These two ifs cover illegal values right? Why handle them at all?

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        public int NextSetBit(int from)
+        {
+            if (from < 0) from = 0;
+            if (from > 65535) return -1;
+            int wordIdx = from >> 6;
+            int bitIdx = from & 0x3F;
+            ulong word = words[wordIdx] & (~0UL << bitIdx);
+            while (true)
```
kevin-montrose, line 130: This can be written in terms of the `Span.IndexOfAnyExcept(0)` methods so we don't scan so manually.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+                }
+                wordIdx++;
+                if (wordIdx >= WordCount) return -1;
+                word = ~words[wordIdx];
+            }
+        }
+
+        public IContainer Clone()
+        {
+            var copy = new ulong[WordCount];
```
kevin-montrose: If we're just copying the whole array, just use `words.ToArray()` rather than this `BlockCopy` stuff.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+            Buffer.BlockCopy(words, 0, copy, 0, WordCount * sizeof(ulong));
+            return new BitmapContainer(copy, cardinality);
+        }
+
+        public void SerializeBody(BinaryWriter writer)
+        {
+            for (int i = 0; i < WordCount; i++)
```
kevin-montrose: Conver `words` to `Span<byte>` and use `BinaryWriter.Write(ReadOnlySpan<byte>)` instead of this loop.

modules/RoaringBitmap/Containers/BitmapContainer.cs
```diff
@@ -0,0 +1,213 @@
...
+                writer.Write(words[i]);
+        }
+
+        public static BitmapContainer DeserializeBody(BinaryReader reader, int cardinality)
+        {
+            if (cardinality < 1 || cardinality > 65536)
+                throw new InvalidDataException($"BitmapContainer cardinality out of range: {cardinality}");
+            var w = new ulong[WordCount];
```
kevin-montrose, line 170: Like in `ArrayContainer`, use `ReadExactly` after a pun to `Span<byte>` and do the count as a separate pass.

modules/RoaringBitmap/Containers/IContainer.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace GarnetRoaringBitmap.Containers
+{
+    /// <summary>
+    /// Container kinds used in serialization. Stable on-disk values — never renumber.
+    /// </summary>
+    internal enum ContainerKind : byte
+    {
+        Array = 1,
+        Bitmap = 2,
+        Run = 3,
```
kevin-montrose: `Run` isn't used anywhere?

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+        public const byte FormatVersion = 0x01;
+
+        /// <summary>
+        /// Hysteresis applied at the parent layer when demoting. A bitmap container whose
+        /// cardinality would drop to <see cref="Containers.ArrayContainer.ArrayThreshold"/>
+        /// will demote; one whose cardinality immediately re-grows past that threshold
+        /// will repromote on the next add. We accept this oscillation cost in exchange
+        /// for memory tightness — a more aggressive hysteresis could be added later.
+        /// </summary>
+        private readonly SortedDictionary<ushort, IContainer> chunks;
```
kevin-montrose: Looking at the whole thing, `SortedDictionary<,>` feels like overkill here - just a `(ushort High, IContainer Container)[]` (or two arrays so we can easily binary search the high array, which might be better for perf) seems like a better fit.

modules/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        /// <summary>Cached running total. Maintained incrementally on every mutation. Matches sum-of-chunk-cardinalities exactly.</summary>
+        private long totalCardinality;
+
+        public RoaringBitmap()
```
kevin-montrose, line 47: nit: there's no separate logic here, so `: this(new(), 0)` would be DRYer.

test/standalone/Garnet.test/RespRoaringBitmapTests.cs
```diff
@@ -0,0 +1,313 @@
+// Copyright (c) Microsoft Corporation.
```
kevin-montrose, line 1: I'd would like to see a test that does concurrent reads while writes are in progress - I see a test for concurrent writes which is not quite enough.

test/Garnet.test/RoaringBitmapDataTests.cs
```diff
@@ -0,0 +1,379 @@
...
+    [AllureNUnit]
+    [TestFixture]
+    public class RoaringBitmapDataTests : AllureTestBase
+    {
+        // ---------- Basics ----------
+
+        [Test]
+        public void EmptyBitmap_HasZeroCardinalityAndIsEmpty()
+        {
+            var rb = new global::GarnetRoaringBitmap.RoaringBitmap();
```
kevin-montrose: nit: remove all this `global::` stuff.

modules/RoaringBitmap/RoaringBitmapCommands.cs
```diff
@@ -0,0 +1,218 @@
...
+    /// Returns the bit at <c>offset</c> in the Roaring bitmap. Returns 0 if the key
+    /// does not exist or the offset has never been set. Registered as
+    /// ReadModifyWrite so that missing-key handling can route through
+    /// <see cref="NeedInitialUpdate"/> without creating an empty key.
+    /// </summary>
+    public sealed class RGetBit : CustomObjectFunctions
+    {
+        private static ReadOnlySpan<byte> ErrOffset => "ERR bit offset is not an unsigned 32-bit integer"u8;
+
+        public override bool NeedInitialUpdate(scoped ReadOnlySpan<byte> key, ref ObjectInput input, ref RespMemoryWriter writer)
```
kevin-montrose: This is more a question for @badrishc as I haven't played around with custom objects too much - is validation _expected_ to get in `NeedInitialUpdate` like this?

It's unfortunate as it forces this read command to act like a write which will hurt throughput.

modules/RoaringBitmap/RoaringBitmapCommands.cs
```diff
@@ -0,0 +1,218 @@
...
+    /// <summary>
+    /// R.BITCOUNT key
+    ///
+    /// Returns the number of bits set to 1 in the Roaring bitmap. Returns 0 if the
+    /// key does not exist. Registered as ReadModifyWrite (no actual mutation) for
+    /// the same NeedInitialUpdate-on-miss reason as <see cref="RGetBit"/>.
+    /// </summary>
+    public sealed class RBitCount : CustomObjectFunctions
+    {
+        public override bool NeedInitialUpdate(scoped ReadOnlySpan<byte> key, ref ObjectInput input, ref RespMemoryWriter writer)
```
kevin-montrose: Similar Q for @badrishc (and again in RBitPos) - using `NeedInitialUpdate` for "missing" is messy; can this be phrased as a `Reader` op instead?

website/docs/commands/roaring-bitmap.md
```diff
@@ -0,0 +1,121 @@
...
+---
+
+# Roaring Bitmap
+
+Garnet ships an extension implementing **[Roaring Bitmaps](https://roaringbitmap.org/)**:
+a compressed bitmap encoding that scales from sparse to dense `uint32` sets while
+remaining fast for membership and population-count queries.
+
+The extension lives in `main/GarnetServer/Extensions/RoaringBitmap/` and is wired
+into the default `GarnetServer` host. It introduces a new object type and four
```
kevin-montrose: This is incorrect RoaringBitmaps must be loaded manually (which is correct).

kevin-montrose (changes requested):

crprashant:
Thank you for your reviews Will address all these in next PR

test/Garnet.test/RoaringBitmapDataTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                    expected0 = from;
+                }
+                long actual1 = rb.BitPos(1, (uint)from);
+                long actual0 = rb.BitPos(0, (uint)from);
+                ClassicAssert.AreEqual(expected1, actual1, $"bitpos(1, {from})");
+                // For bitpos(0), our scan reports the first unset somewhere in the
+                // uint32 universe — past the set range any value qualifies, so
+                // accept any value >= from that's not in the bits array.
+                ClassicAssert.GreaterOrEqual(actual0, from);
+                ClassicAssert.IsTrue(actual0 == 0 || (actual0 < bits.LongLength ? !bits[actual0] : true));
```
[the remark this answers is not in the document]
crprashant: Test was rewritten — `BitPos` now returns `long` and assertions use the new `bool` API. No `bits[long]` indexing in the current code.

main/GarnetServer/Extensions/RoaringBitmap/RoaringBitmap.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        /// <summary>Estimated heap byte cost. Excludes the .NET object header overhead and the SortedDictionary node overhead.</summary>
+        public long ByteSize
+        {
+            get
+            {
+                long sum = 24; // base object overhead estimate
+                foreach (var kv in chunks)
+                {
+                    // Per-entry: key (2B), reference (8B), red-black tree node overhead (~40B), and container.
```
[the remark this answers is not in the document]
crprashant: `ByteSize` recalculated for the new parallel-array layout (no `SortedDictionary` nodes). Comment and computation now match.

main/GarnetServer/Extensions/RoaringBitmap/RoaringBitmapCommands.cs
```diff
@@ -0,0 +1,197 @@
...
+    /// Sets or clears the bit at <c>offset</c> (a 32-bit unsigned integer) in the
+    /// Roaring bitmap stored at <c>key</c>. Returns the previous bit value (0 or 1).
+    /// </summary>
+    public sealed class RSetBit : CustomObjectFunctions
+    {
+        private static ReadOnlySpan<byte> ErrOffset => "ERR bit offset is not an unsigned 32-bit integer"u8;
+        private static ReadOnlySpan<byte> ErrValue => "ERR bit value must be 0 or 1"u8;
+
+        public override bool NeedInitialUpdate(ReadOnlyMemory<byte> key, ref ObjectInput input, ref RespMemoryWriter writer) => true;
+
```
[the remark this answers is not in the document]
crprashant: Argument validation could not be moved into `NeedInitialUpdate` because parsing happens in `Updater` against the typed `ObjectInput`. Validation order is preserved and errors are returned without persisting an empty object (the empty object created on missed-validation is collected by the subsequent error path before commit). Will follow up if the framework gains a cleaner missing-key Reader override.

main/GarnetServer/Extensions/RoaringBitmap/RoaringBitmapCommands.cs
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Buffers.Text;
+using System.Diagnostics;
+using Garnet.common;
```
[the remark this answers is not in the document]
crprashant: Unused `using Garnet.common;` removed (verified by clean build with TreatWarningsAsErrors).

modules/RoaringBitmap/RoaringBitmapObject.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        /// <summary>Estimated overhead of the object shell itself, separate from container memory.</summary>
+        private const long ObjectOverhead = 32;
+
+        private readonly RoaringBitmap bitmap;
+
+        public RoaringBitmapObject(byte type)
+            : base(type, expiration: 0, size: ObjectOverhead)
+        {
+            bitmap = new RoaringBitmap();
```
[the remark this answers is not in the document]
crprashant: `Size` initialization is now consistent — both ctors set `Size = ObjectOverhead + bitmap.ByteSize`, and `bitmap.ByteSize` reports a small constant baseline for an empty bitmap so deltas accumulate correctly.

modules/RoaringBitmap/GarnetRoaringBitmap.csproj
crprashant: Done — `Microsoft.VisualStudio.Threading.Analyzers` added to `GarnetRoaringBitmap.csproj`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Outdated comment block above the chunks field removed; the `(added in a follow-up commit)` parenthetical on the class XML is also dropped.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `RoaringBitmap.Add(uint)` now returns `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `Remove(uint)` returns `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `SetBit(uint, bool)` returns the previous value as `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `GetBit(uint)` returns `bool`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `BitPos` now takes `bool bit` and the bridge to RESP's int happens in `RoaringBitmapObject`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Removed — no longer needed once `bit` is a `bool`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Acknowledged — applied a `var` (IDE0007) sweep across all modified files (`ArrayContainer.cs`, `BitmapContainer.cs`, `RoaringBitmap.cs`, `RoaringBitmapCommands.cs`, `RoaringBitmapObject.cs`, both test files).

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Refactored — `chunks` is now two parallel arrays (`ushort[] highKeys` + `IContainer[] containers`) with `MemoryExtensions.BinarySearch` for chunk lookup. Both `BitPos` loops now skip to the first chunk with `highKey >= fromHi` in `O(log N)`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Same refactor — see thread above. Both loops are now logarithmic.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `RoaringBitmap` now implements `IEnumerable<uint>`. `Enumerate()` retained as a thin wrapper for backward compatibility with existing tests.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `Serialize` now uses `ArgumentNullException.ThrowIfNull(writer);`.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — `Deserialize` now uses `ArgumentNullException.ThrowIfNull(reader);`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — replaced with `ArgumentNullException.ThrowIfNull(sortedUnique);`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — custom `BinarySearch` removed; all callers use `MemoryExtensions.BinarySearch` over `values.AsSpan(0, count)`.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — split onto separate lines.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — split onto separate lines.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — `SerializeBody` now writes `MemoryMarshal.AsBytes<ushort>(values.AsSpan(0, count))` in a single call.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Done — `DeserializeBody` uses `reader.BaseStream.ReadExactly(MemoryMarshal.AsBytes(arr.AsSpan()))` and validates strict-ascending monotonicity in a separate pass.

modules/RoaringBitmap/Containers/ArrayContainer.cs
crprashant: Adopted your suggestion verbatim — `EnsureCapacity` now grows `1,2,4,...,4096` via `BitOperations.RoundUpToPowerOf2` capped at `ArrayThreshold`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `ArgumentNullException.ThrowIfNull(preExistingWords);`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — split onto separate lines.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — split onto separate lines.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: You're right — `BitmapContainer` is only created at cardinality > `ArrayThreshold` (4096). The unreachable single-bit branch in `Remove` is removed; demotion to `ArrayContainer` is the only exit path below threshold.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Confirmed impossible — empty `BitmapContainer` cannot occur (demotion happens before cardinality reaches 0). Guard removed.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `First()` now uses `Span<ulong>.IndexOfAnyExcept(0UL) + BitOperations.TrailingZeroCount`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Same — `Last()` is gated on the same caller invariant (cardinality > 0). Empty-case guard removed.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `Last()` now uses `Span<ulong>.LastIndexOfAnyExcept(0UL) + BitOperations.LeadingZeroCount`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: You're right — `from` is always passed in `[0, 65535]` from `RoaringBitmap.BitPos` via `(ushort)(from & 0xFFFF)`. The illegal-value guards are removed.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Same — guards removed; callers respect the documented `[0, 65535]` precondition.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `NextUnsetBit` rewritten with `IndexOfAnyExcept` over a complemented span.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `Clone()` is now `new BitmapContainer(words.ToArray(), cardinality)`.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `SerializeBody` writes `MemoryMarshal.AsBytes<ulong>(words.AsSpan())` in a single call.

modules/RoaringBitmap/Containers/BitmapContainer.cs
crprashant: Done — `DeserializeBody` reads via `reader.BaseStream.ReadExactly(MemoryMarshal.AsBytes(w.AsSpan()))` and pop-counts in a separate pass for validation.

modules/RoaringBitmap/Containers/IContainer.cs
crprashant: Done — removed `Run = 3` from the `ContainerKind` enum (planned for a follow-up if/when run-length encoding is added).

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Refactored — `chunks` is now `ushort[] highKeys` + `IContainer[] containers` (two parallel arrays). Lookup uses `MemoryExtensions.BinarySearch`, insertion shifts the tail via `Array.Copy`, and growth doubles in geometric steps. Drops the SortedDictionary tree overhead and gives logarithmic point lookups.

modules/RoaringBitmap/RoaringBitmap.cs
crprashant: Done — default ctor now chains to a primary ctor that takes the arrays + counts.

test/standalone/Garnet.test/RespRoaringBitmapTests.cs
crprashant: Added `ConcurrentReadsWhileWrites_StayConsistent` in `RespRoaringBitmapTests.cs` — 4 reader tasks running 5000 `R.GETBIT`/`R.BITCOUNT` ops concurrently with a writer toggling bits 1000..1999. All ops complete without exceptions and final state is verified.

test/Garnet.test/RoaringBitmapDataTests.cs
crprashant: Done — every `global::GarnetRoaringBitmap.` prefix stripped (the namespace import resolves the type).

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: Left as-is for now and added an explicit architectural note in code (above each affected command) flagging the question for @badrishc. The Reader path is also implemented as an existing-key fast-path, but `NeedInitialUpdate` is currently the only hook that can emit Redis-compatible missing-key responses (0 / -1) without `nil` semantics. Happy to switch if @badrishc confirms a cleaner Reader+miss override is planned.

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: Same note as above — `Reader` is now implemented on `RGetBit`, `RBitCount`, `RBitPos` as the existing-key fast-path; `NeedInitialUpdate` handles the typed missing-key response. Architectural comment added in-file pointing at @badrishc.

website/docs/commands/roaring-bitmap.md
crprashant: Fixed — `website/docs/commands/roaring-bitmap.md` now says the module lives in `modules/RoaringBitmap/` and must be loaded manually via `MODULE LOADCS`.

crprashant:
## Updated — all review comments addressed

Force-pushed `47d62b29b` (rebased onto current `dev` from `dd3e2dabb`).

**Highlights:**
- `RoaringBitmap` chunk store refactored from `SortedDictionary<ushort, IContainer>` to two parallel arrays (`ushort[] highKeys` + `IContainer[] containers`) with `MemoryExtensions.BinarySearch` — `BitPos` now skips to the first relevant chunk in O(log N).
- `Add` / `Remove` / `SetBit` / `GetBit` / `BitPos.bit` migrated from `int` to `bool` (RESP int↔bool bridging stays in `RoaringBitmapObject`).
- `RoaringBitmap` now implements `IEnumerable<uint>`.
- `ArrayContainer` and `BitmapContainer` rewritten using `Span<T>.IndexOfAnyExcept(0)` / `LastIndexOfAnyExcept(0)`, `MemoryExtensions.BinarySearch`, `MemoryMarshal.AsBytes` for bulk serialize, and `Stream.ReadExactly` for bulk deserialize. `EnsureCapacity` adopts the geometric `1,2,4,…,4096` growth pattern via `BitOperations.RoundUpToPowerOf2`.
- All unreachable / illegal-state guards removed; impossible `BitmapContainer.cardinality == 0` / `cardinality == 1` branches gone.
- `Reader` overrides added on `RGetBit` / `RBitCount` / `RBitPos` as the existing-key fast path; `NeedInitialUpdate` retained for typed missing-key responses (with an inline architectural note flagging the question for @badrishc).
- `var` (IDE0007) sweep across all changed files; `ArgumentNullException.ThrowIfNull` swapped in everywhere; style nits resolved.
- `ContainerKind.Run` removed (unused).
- `Microsoft.VisualStudio.Threading.Analyzers` added to the module csproj.
- Docs corrected — module is loaded manually via `MODULE LOADCS`, not wired into the default host.
- New `ConcurrentReadsWhileWrites_StayConsistent` integration test (4 reader tasks running 5000 `R.GETBIT`/`R.BITCOUNT` ops concurrently with a bit-toggling writer).

**Build:** 0 warnings, 0 errors.
**Tests:** 44/44 Roaring tests passing on both `net8.0` and `net10.0`.

Two threads contain open questions for @badrishc about whether the read commands can be expressed as pure `Reader` ops (avoiding the `ReadModifyWrite` registration that `NeedInitialUpdate` requires for missing-key responses). I've replied to each thread inline and the code carries a matching architectural comment — happy to refactor if you confirm a cleaner pattern.

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: @badrishc — flagging this for your input.

We have R.GETBIT / R.BITCOUNT / R.BITPOS, all conceptually read-only. To return the Redis-compatible missing-key response (`0` for GETBIT/BITCOUNT, `-1` for BITPOS) without materializing an empty object, we currently register them as `CommandType.ReadModifyWrite` and emit the response from `NeedInitialUpdate`. The `Reader` path is only invoked when the key exists.

Is there an idiomatic way to express "read-only with typed missing-key response" in `CustomObjectFunctions` today, or is the `NeedInitialUpdate` hack the recommended pattern? If a cleaner `Reader`+miss override is on the roadmap, we'd happily refactor these three commands. As of `47d62b29b` the `Reader` override is already in place as a fast path for the existing-key case.

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: @badrishc — same question for R.BITPOS specifically (missing key must return `-1` per Redis semantics, not nil). Tagging you so this isn't lost; full context in the sibling thread above. Happy to refactor if `CustomObjectFunctions` gains a cleaner Reader-with-miss hook.

crprashant:
@badrishc — two open architectural questions on this PR I'd love your read on. Both relate to how a read-only custom object command should emit a Redis-compatible missing-key response (`0` / `-1`) without materializing an empty object:

- R.GETBIT / R.BITCOUNT — https://github.com/microsoft/garnet/pull/1741#discussion_r3174777004
- R.BITPOS — https://github.com/microsoft/garnet/pull/1741#discussion_r3174780061

TL;DR: the three read commands are currently registered as `ReadModifyWrite` because `NeedInitialUpdate` is the only hook that lets us return a typed missing-key response. `Reader` is wired in as the existing-key fast path. Threads are left unresolved.

kevin-montrose:
On a second review of this, we have concerns about the write commands being registered as RMW.  That implies reads will proceed sequentially rather than in parallel.

Unfortunately there's no existing way to customize the RESP response of a custom read command today, [you always get a null](https://github.com/microsoft/garnet/blob/42051c1f26006711f8752ff9155b379fbad04162/libs/server/Custom/CustomRespCommands.cs#L338-L344).

I'm going to take a look at fixing that in general (and moving existing extensions to use it if necessary).  Once that's merged you can update this PR to use the new infra and catch up to `main` (now that `dev` has been merged in we no longer need to target `dev`).  I'll try and get that done tomorrow, or shortly after the holiday weekend.

crprashant:
Thank you. Sounds good @kevin-montrose will look for updates on this from you following week.

kevin-montrose:
Infra changes for this are pending review here: #1820

kevin-montrose:
Note that linked PR has merged.

crprashant:
Thanks @kevin-montrose will push new changes this week.

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: Addressed in the latest push. These read commands are now phrased as Reader ops (with CommandType.Read): existing keys flow through Reader, and missing keys are handled by the new NotFound callback (from #1820) so no key is ever created on a read. The NeedInitialUpdate/Updater workaround is gone for R.GETBIT/R.BITCOUNT/R.BITPOS.

modules/RoaringBitmap/RoaringBitmapCommands.cs
crprashant: For the read commands (R.GETBIT/R.BITCOUNT/R.BITPOS) this validation-in-`NeedInitialUpdate` is removed — they are now `Reader` ops and validate inside `Reader`/`NotFound`. Only R.SETBIT (a genuine write) keeps validation in `NeedInitialUpdate`, intentionally: it rejects a malformed offset/value before the framework allocates an empty object, so a bad R.SETBIT never leaves an empty key behind.

crprashant:
Pushed an update (`12a9ce2`) addressing the open review feedback and re-syncing with `main`:

**Addressed @kevin-montrose's review (both threads resolved):**
- R.GETBIT / R.BITCOUNT / R.BITPOS no longer use the `NeedInitialUpdate`/`Updater` (RMW-on-read) workaround. They're now `CommandType.Read` and flow through `Reader` for existing keys and the new `NotFound` callback (#1820) for missing keys, so a read never creates a key.
- R.SETBIT (a genuine write) intentionally keeps validation in `NeedInitialUpdate` so a malformed offset/value is rejected before an empty object is allocated.

**Re-synced with main:**
- Merged latest `main` (resolved the test-project relocation `test/Garnet.test` -> `test/standalone/Garnet.test`). PR is mergeable again.
- Dropped the Allure test attributes (present on `dev` but not `main`) so the tests build against `main`.

**Verification:**
- All 44 Roaring Bitmap tests pass (net10.0).
- Full solution builds clean (0 warnings / 0 errors).
- `dotnet format` clean.
- Diff vs `main` is purely additive (new `modules/RoaringBitmap` + tests; only single-line solution/test-project wiring touched), no changes to `libs/` or core, so no breaking changes.

kevin-montrose (approved):

PR #1758, opened 2026-04-30
Fix VectorSet index keys not deleted after MIGRATE KEYS

tiagonapoli:
## Summary

After transmitting VectorSet indexes in the KEYS migration path, mark the VectorSet keys in the sketch as transmitted (\Item2 = true\) so that the existing \DeleteKeys()\ flow deletes them during the DELETING phase.

Previously, \TransmitKeys()\ skipped VectorSet keys (they are in the ignore set for out-of-band transmission), leaving \Item2 = false\, which caused \DeleteKeys()\ to skip them as well. This resulted in VectorSet keys remaining on the source node after migration.

### Changes

- **Fix MIGRATE KEYS deletion**: Mark VectorSet keys as transmitted in the sketch after out-of-band transmission, so \DeleteKeys()\ properly removes them.
- **Add DEBUG EXISTS subcommand**: Raw store existence check bypassing cluster routing/EPSM, used to verify key deletion in tests.
- **Add eviction leak test**: Documents a known limitation where DEL after eviction below HeadAddress leaks the DiskANN native index (blind tombstone path skips \TryDeleteVectorSet\).
- **Add VectorManager.InUseContextCount**: Internal accessor for test introspection of context metadata bitmap.

### DIFF
--- a/libs/cluster/Server/Migration/MigrateSessionKeys.cs
+++ b/libs/cluster/Server/Migration/MigrateSessionKeys.cs
@@ -104,6 +104,20 @@ namespace Garnet.cluster
                         logger?.LogCritical("Final flush after Vector Set migration failed");
                         return false;
                     }
+
+                    // Mark VectorSet keys for deletion so DeleteKeys() removes them during the DELETING phase
+                    var sketchKeys = migrateTask.sketch.Keys;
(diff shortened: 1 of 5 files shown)


tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/20

PR #1767, opened 2026-05-04
Fix MultiDatabaseSaveInProgressTest

badrishc:
The general BGSAVE path returned 'Background saving started' before its async helper had pause-locked any per-DB checkpoint locks. A subsequent BGSAVE <dbId> could win the race against TryPauseCheckpoints(dbId) and succeed, instead of failing with 'ERR checkpoint already in progress'.

This caused MultiDatabaseTests.MultiDatabaseSaveInProgressTest to flake in CI. SingleDatabaseManager.TakeCheckpointAsync already does the right thing by pausing synchronously before returning, which is why the single-DB equivalent test (SeSaveInProgressTest) does not flake.

Fix:
- Restructure MultiDatabaseManager.TakeCheckpointAsync(bool, ILogger, CancellationToken) so the synchronous portion now acquires databasesContentLock, multiDbCheckpointingLock (when multi-db), AND calls TryPauseCheckpoints(dbId) for every active DB before returning. Compactly store only successfully paused DB IDs in dbIdsToCheckpoint. Roll back any partially-acquired state on exception in the sync phase.
- Add alreadyPaused parameter to TakeDatabasesCheckpointAsync so the new caller can skip the inline TryPauseCheckpoints. Add a catch fallback that resumes pre-paused DB IDs not yet handed off to per-DB helpers, preventing stranded locks.
- Existing AOF-size-driven caller (TaskCheckpointBasedOnAofSizeLimitAsync) is unaffected; it uses the default alreadyPaused=false.

Regression test: MultiDatabaseGeneralSaveBlocksGeneralSaveTest verifies that a second general BGSAVE while one is in flight (multi-db) reliably returns 'ERR checkpoint already in progress' (multiDbCheckpointingLock is now held synchronously).

### DIFF
--- a/libs/server/Databases/MultiDatabaseManager.cs
+++ b/libs/server/Databases/MultiDatabaseManager.cs
@@ -147,41 +147,67 @@ namespace Garnet.server
         }
 
         /// <inheritdoc/>
-        public override async Task<bool> TakeCheckpointAsync(bool background, ILogger logger = null, CancellationToken token = default)
+        public override Task<bool> TakeCheckpointAsync(bool background, ILogger logger = null, CancellationToken token = default)
         {
             var lockAcquired = TryGetDatabasesContentReadLock(token);
-            if (!lockAcquired) return false;
+            if (!lockAcquired) return Task.FromResult(false);
 
-            var checkpointTask = TakeCheckpointHelperAsync();
+            var checkpointLockTaken = false;
+            var pausedCount = 0;
 
-            if (background)
-                return true;
-
-            return await checkpointTask.ConfigureAwait(false);
-
-            async Task<bool> TakeCheckpointHelperAsync()
+            try
             {
-                // Force async 
-                await Task.Yield();
+                var activeDbIdsMapSize = activeDbIds.ActualSize;
 
-                var checkpointLockTaken = false;
-
-                try
+                if (activeDbIdsMapSize > 1)
                 {
-                    var activeDbIdsMapSize = activeDbIds.ActualSize;
-
-                    if (activeDbIdsMapSize > 1)
+                    if (!multiDbCheckpointingLock.TryWriteLock())
                     {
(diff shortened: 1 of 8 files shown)


libs/server/Databases/MultiDatabaseManager.cs
```diff
@@ -147,141 +140,157 @@ public override void RecoverCheckpoint(bool replicaRecover = false, bool recover
         }
 
         /// <inheritdoc/>
-        public override async Task<bool> TakeCheckpointAsync(bool background, ILogger logger = null, CancellationToken token = default)
+        public override Task<bool> TakeCheckpointAsync(bool background, ILogger logger = null, CancellationToken token = default)
```
vazois: nit; Should we rename this since it not longer async? Also, since you are there, I feel the logger parameter need to appear last

vazois (approved):

libs/server/Databases/MultiDatabaseManager.cs
badrishc: Reordered the logger parameter to come after CancellationToken in eb3b952ba6 — agree it should match the codebase convention (CommitToAofAsync, WaitForCommitToAofAsync, etc.). Also reordered the IDatabaseManager / DatabaseManagerBase / SingleDatabaseManager signatures, the StoreWrapper wrappers, and the new private helpers for consistency. Had to drop the `dbId = -1` default on StoreWrapper's per-DB dispatcher overload to avoid resolution ambiguity (both overloads now have CancellationToken as the second-positional defaultable param).

For the rename: I kept the `Async` suffix. Per .NET convention, any method that returns Task / ValueTask gets the `Async` suffix regardless of whether the implementation uses the `async` keyword internally — `async` is an implementation detail. Renaming would also force changing IDatabaseManager and all implementations across SingleDatabaseManager etc., which feels out of scope for this PR. Happy to revisit if you'd prefer otherwise.

vazois (approved):

PR #1773, opened 2026-05-06
Refactor disk-based replication checkpoint shipping and safe hlog segment truncation

vazois:
## Summary

Refactors the checkpoint shipping pipeline for disk-based replication, introducing clean abstractions for reading and transmitting Tsavorite checkpoint data. Also adds safe hybrid log segment truncation to prevent deletion of segments actively being read by syncing replicas.

### Key changes

**Checkpoint shipping abstractions (send side):**
- ISnapshotReader / ISnapshotTransmitSource / ISnapshotDataSource interfaces
- TsavoriteSnapshotReader / TsavoriteCheckpointReader for reading checkpoint files
- FileTransmitSource / TsavoriteMetadataTransmitSource for transmitting data
- SnapshotTransmissionDriver orchestrating the send pipeline
- ReplicaSyncSession refactored to use the new abstractions

**Checkpoint shipping abstractions (receive side):**
- ISnapshotDataSink interface
- FileDataSink / MetadataDataSink implementations
- Unified ReceiveCheckpointHandler with ProcessSnapshotData entry point
- Unified CLUSTER SNAPSHOT_DATA command (previous per-type commands are deprecated but not removed)

**Safe hlog segment truncation (PerformInternalCleanup):**
- Added PerformInternalCleanup property to ICheckpointManager interface
- GarnetClusterCheckpointManager overrides as alse -- Tsavorite skips internal cleanup
- CheckpointStore.DeleteOutdatedCheckpoints() calls ShiftBeginAddress with the oldest active checkpoint begin address as the safe truncation boundary
- Ensures hlog segments are not deleted while replicas are actively reading them

**RangeIndexManager refactoring:**
- Moved AOF replication methods to RangeIndexManager.Replication.cs partial file

### Testing
- Added ClusterReplicationHlogSegmentCleanupTest validating hlog segment truncation during concurrent replica sync (25+ stable runs)
- All existing replication tests pass


libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/TsavoriteCheckpointReader.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            return device;
+        }
+
+        private IDevice GetStoreHLogDevice(bool isObj)
+        {
+            var opts = clusterProvider.serverOptions;
+            if (opts.EnableStorageTier)
+            {
+                var LogDir = !string.IsNullOrEmpty(opts.LogDir) ? opts.LogDir : Directory.GetCurrentDirectory();
```
tiagonapoli, line 172: Should we create a common helper? @badrishc is also adding a default on this PR https://github.com/microsoft/garnet/pull/1780, maybe we should do in a common place to avoid inconsistency

Image

vazois: Yes, we should if possible. However, this is just a refactoring targeted for Tsavorite snapshot management.
It does not do anything for RangeIndex yet.
The idea is that RangeIndexManager should provide these paths.
For now, Tsavorite does not have a uniform way to do it, so I am just using the existing options instance.

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileDataSource.cs
```diff
@@ -0,0 +1,152 @@
...
+            long numBytesToRead = size;
+            numBytesToRead = (numBytesToRead + (device.SectorSize - 1)) & ~(device.SectorSize - 1);
+
+            var buffer = bufferPool.Get((int)numBytesToRead);
+            unsafe
+            {
+                device.ReadAsync(address, (IntPtr)buffer.aligned_pointer, (uint)numBytesToRead, IOCallback, null);
+            }
+
+            if (!await signalCompletion.WaitAsync(timeout, cancellationToken).ConfigureAwait(false))
```
tiagonapoli, line 129: If the semaphore times out before the IO completes, and the IO actually completes at some point, it will write to `buffer.aligned_pointer` which was a returned pointer, so may be already unsafe memory or someone else's buffer, then we corrupt the memory

Looks like that signalCompletion.WaitAsync can also throw because of CancellationToken - In that case I think we don't do a buffer.Return - It's tricky though because if we catch and return the buffer, CancellationToken will cancel the semaphore but it will not cancel the IO, which requires that buffer. So then again we have unsafe behavior here where the IO will write to a buffer that was returned

libs/client/ClientSession/GarnetClientSessionReplicationExtensions.cs
```diff
@@ -258,6 +258,83 @@ public Task<string> ExecuteClusterSendCheckpointFileSegment(Memory<byte> fileTok
             return tcs.Task;
         }
 
+        static ReadOnlySpan<byte> snapshot_data => "SNAPSHOT_DATA"u8;
```
tiagonapoli: nit: maybe use `private static` for explicit access modifiers

libs/client/ClientSession/GarnetClientSessionReplicationExtensions.cs
```diff
@@ -258,6 +258,83 @@ public Task<string> ExecuteClusterSendCheckpointFileSegment(Memory<byte> fileTok
...
+            //5
+            while (!RespWriteUtils.TryWriteArrayItem(startAddress, ref curr, end))
+            {
+                Flush();
+                curr = offset;
+            }
+            offset = curr;
+
+            //6
+            while (!RespWriteUtils.TryWriteBulkString(data, ref curr, end))
```
tiagonapoli, line 331: nit: Do we have an enforcement somewhere that the network buffer will always be bigger than data?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileTransmitSource.cs
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    var resp = await gcs.ExecuteClusterSnapshotData(
+                        fileTokenBytes,
+                        (int)DataSource.Type,
+                        startAddress: result.ChunkStartAddress,
+                        result.Buffer.GetSlice(result.BytesRead)).WaitAsync(timeout, cancellationToken).ConfigureAwait(false);
+
+                    if (!resp.Equals("OK"))
+                        ExceptionUtils.ThrowException(new GarnetException(
+                            $"Primary error at SendFileSegments {DataSource.Type} {resp} [{DataSource.StartOffset},{DataSource.CurrentOffset},{DataSource.EndOffset}]"));
```
tiagonapoli: nit: Should the log mention `ExecuteClusterSnapshotData`? Or `TransmitAsync`?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileTransmitSource.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            // Send empty package to indicate end of transmission
+            var endResp = await gcs.ExecuteClusterSnapshotData(
+                fileTokenBytes, (int)DataSource.Type, DataSource.CurrentOffset, [])
+                .WaitAsync(timeout, cancellationToken).ConfigureAwait(false);
+
+            if (!endResp.Equals("OK"))
+                ExceptionUtils.ThrowException(new GarnetException(
+                    $"Primary error at SendFileSegments Completion {DataSource.Type} {endResp}"));
```
tiagonapoli: nit: Should the log mention `ExecuteClusterSnapshotData`? Or `TransmitAsync`?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/SnapshotTransmissionDriver.cs
```diff
@@ -0,0 +1,63 @@
...
+        public SnapshotTransmissionDriver(List<ISnapshotReader> checkpointReaders, GarnetClientSession gcs, TimeSpan timeout, ILogger logger = null)
+        {
+            this.checkpointReaders = checkpointReaders;
+            this.gcs = gcs;
+            this.timeout = timeout;
+            this.logger = logger;
+        }
+
+        public void Dispose()
+        {
```
tiagonapoli, line 37: nit: should we move the ownership of snapshotReaders to the driver and then dispose them here instead of in `ReplicaSyncSession.SendCheckpointAsync`?

libs/client/ClientSession/GarnetClientSessionReplicationExtensions.cs
vazois: It is restricted by the segment size at the caller.
I am not sure how this works with bftree though.
I will leave it for the later PR because I believe the safest approach is keep flushing chunks instead of trying to fit everything in one go.

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileDataSource.cs
```diff
@@ -0,0 +1,230 @@
...
+                // Callback hasn't fired yet; it will return the buffer when the IO eventually completes.
+                logger?.LogWarning("Timed out reading {type} checkpoint file at address {address}; buffer ownership transferred to IO callback", Type, address);
+                ExceptionUtils.ThrowException(new GarnetException(
+                    $"Timed out reading {Type} checkpoint file at address {address} (requested {numBytesToRead} bytes)"));
+                return default; // Unreachable
+            }
+
+            // Callback completed between WaitAsync returning false and our CompareExchange.
+            // The IO finished — use the result instead of discarding it.
+            return HandleIOError(buffer, address, numBytesToRead);
```
tiagonapoli, line 136: I think there might be a subtle issue here - On this case the callback completed but since we had a race we might have one extra count in the shared semaphore, and since on this case we're not throwing and continuing the operation, the next `WaitAsync` on the shared semaphore will finish immediately before the IO actually finishing, so we'll access invalid data on the buffer.
Might be safer to release the buffer and throw here on this case - it should be a pretty rare scenario anyways

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/FileDataSink.cs
```diff
@@ -0,0 +1,126 @@
...
+            this.device = device;
+            this.bufferPool = bufferPool;
+            this.writeSemaphore = writeSemaphore;
+            this.timeout = timeout;
+            this.cancellationToken = cancellationToken;
+            this.logger = logger;
+        }
+
+        /// <inheritdoc/>
+        public unsafe void WriteChunk(long startAddress, ReadOnlySpan<byte> data)
```
tiagonapoli, line 63: For my learning - do we need `startAddress` to be sector aligned as well? Should throw if not valid?
(I suppose startAddresses are sector aligned in the source replica - but what if for some reason in a Managed Garnet cluster we have two instances that have different underlying Disks, with different sector sizes - I don't know if we full control over the actual disk SKU)

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/FileDataSink.cs
```diff
@@ -0,0 +1,126 @@
...
+            long numBytesToWrite = data.Length;
+            numBytesToWrite = (numBytesToWrite + (device.SectorSize - 1)) & ~(device.SectorSize - 1);
+
+            var pbuffer = bufferPool.Get((int)numBytesToWrite);
+            try
+            {
+                fixed (byte* bufferRaw = data)
+                    Buffer.MemoryCopy(bufferRaw, pbuffer.aligned_pointer, data.Length, data.Length);
+
+                device.WriteAsync((IntPtr)pbuffer.aligned_pointer, (ulong)startAddress, (uint)numBytesToWrite, IOCallback, null);
```
tiagonapoli: We might have a similar race issue here where we might free the buffer because `.Wait` times out and later the WriteAsync IO callback writes to a freed buffer, and when we have a use-after-free

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/ReceiveCheckpointHandler.cs
```diff
@@ -0,0 +1,167 @@
...
+            if (data.Length == 0)
+            {
+                activeSink?.Complete();
+                activeSink = null;
+                return;
+            }
+
+            if (activeSink == null)
+            {
+                var device = clusterProvider.replicationManager.CreateCheckpointDevice(token, type);
```
tiagonapoli, line 66: nit: If we're in retry scenario where a previous attempt was made to send the checkpoint file, but failed partially, `CreateCheckpointDevice` will not create a new file, it will open existing one without truncating it (`FileMode.OpenOrCreate`), and then we'll overwrite sectors - if the file on disk were bigger, then we could end up with invalid file. But on this case since we have a checkpoint GUID I guess this is not a problem since file will be the same. Is it worth calling out in a comment that an incomplete file might already exist on disk but that's fine?

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/ReceiveCheckpointHandler.cs
```diff
@@ -0,0 +1,167 @@
...
+            if (activeSink == null)
+            {
+                switch (type)
+                {
+                    case CheckpointFileType.STORE_HLOG:
+                    case CheckpointFileType.STORE_HLOG_OBJ:
+                    case CheckpointFileType.STORE_SNAPSHOT:
+                    case CheckpointFileType.STORE_SNAPSHOT_OBJ:
+                    case CheckpointFileType.STORE_INDEX:
+                        var device = clusterProvider.replicationManager.CreateCheckpointDevice(token, type);
```
tiagonapoli, line 154: nit: same from here https://github.com/microsoft/garnet/pull/1773/changes#r3239372743

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/FileDataSink.cs
vazois: I don't think we are unaligned here. The addresses come from the primary who reading aligned chunks.
The case we are not unaligned will throw an error at the callback.
I can add a small defensive check there just for sanity but is not really necessary.

tiagonapoli: Yes, I think would be super rare - but the sanity check would be worth it in the rare case where we have a Primary with a different Disk hardware from a replica (e.g., primary has sector size of 512b while replica has sector size of 4096b). I'm not sure if in Managed Garnet we'll be able to have exact same disk SKU and same sector size required (other than selecting Premium SSD + Disk size for example, we don't have an option on the exact hardware we're getting)

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/FileDataSink.cs
tiagonapoli: Thoughts on using TaskCompletionSource for simplicity?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileDataSource.cs
vazois: Simplified to:

     - IOCallbackContext — just a SectorAlignedMemory buffer field that roots the pinned byte[] for GC safety while IO is in-flight
     - Callback — unconditionally releases the semaphore (with ObjectDisposedException guard)
     - Caller — single WaitAsync call; on timeout throws, on cancellation propagates naturally; buffer abandoned in both cases (GC collects after IO completes)
     - IO error — just throws; buffer abandoned same as other error paths

    Rationale: Timeout/cancellation/IO-error are all catastrophic for the replication session. No subsequent reads will use the shared semaphore, so stale counts are harmless. Abandoned buffers stay alive via the IOCallbackContext reference chain until the callback fires, then GC collects them.

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/FileDataSource.cs
vazois: I am treating the timeout/cancellation as a catastrophic event. I am not going to attempt to capture the IO output after any of these has occurred to simplify the workflow.
In that case, release should not matter because we terminate operations.

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/FileDataSink.cs
vazois: I think I am going to use the existing pattern that we had.
It just needs to be augmented to handle timeout and cancellation.

It is similar to FileDataSource workflow

    Simplified to:

     - IOCallbackContext — just a SectorAlignedMemory buffer field that roots the pinned byte[] for GC safety while IO is in-flight
     - Callback — unconditionally releases the semaphore (with ObjectDisposedException guard)
     - Caller — single WaitAsync call; on timeout throws, on cancellation propagates naturally; buffer abandoned in both cases (GC collects after IO completes)
     - IO error — just throws; buffer abandoned same as other error paths

    Rationale: Timeout/cancellation/IO-error are all catastrophic for the replication session. No subsequent reads will use the shared semaphore, so stale counts are harmless. Abandoned buffers stay alive via the IOCallbackContext reference chain until the callback fires, then GC collects them.

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/ReceiveCheckpointHandler.cs
vazois: There is no problem here as you mentioned because the chunks are streamed from scratch and it is the same file. So, data will be overwritten.
We could potentially optimize this by retrieving information about the file state.
However, this might be complex, and I am trying to retain the original functionality before making big changes.

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/ReceiveCheckpointHandler.cs
vazois: non-issue given that we overwrite the file.

tiagonapoli (approved):

PR #1774, opened 2026-05-06
Fix LockSublogs mutual exclusion and GC hole in AofAddress.Span

vazois:
## Summary

Two targeted fixes for the AOF multi-log replay infrastructure:

### 1. Fix LockSublogs mutual exclusion

`ShardedLog.LockSublogs()` was not enforcing mutual exclusion - it ORed bits without checking if they were already held. This could allow concurrent multi-sublog enqueues sharing a physical sublog to reorder TxnCommit markers, causing barrier-level deadlocks during replay.

**Fix**: Spin-wait with CAS until none of the requested bits are set, then atomically set them. This serializes multi-sublog enqueues that share any physical sublog.

### 2. Fix GC hole in AofAddress.Span property

The previous implementation used a `fixed` pointer that escaped scope, creating a dangling `Span`. The new implementation uses `MemoryMarshal.CreateSpan` with `Unsafe.As` and `[UnscopedRef]` to return a GC-tracked managed reference that remains valid as long as the struct is alive.

## Testing

- Solution builds with 0 warnings, 0 errors
- Existing AOF replay tests pass


libs/server/AOF/ShardedLog.cs
```diff
@@ -55,13 +55,25 @@ public ShardedLog(int physicalSublogCount, TsavoriteLogSettings[] logSettings, I
...
+                    spinWait.SpinOnce();
+                    continue;
+                }
+
+                // Atomically set our bits
                 var newLockMap = currentLockMap | logAccessBitmap;
                 if (Interlocked.CompareExchange(ref lockMap, newLockMap, currentLockMap) == currentLockMap)
                     break;
+
+                // CAS failed due to racing update — spin before retrying
```
TedHartMS: You could simplify this by doing CAS if there are no set bits and break if that succeeds else fall through to a single call to SpinOnce() due to bits being set by caller or CAS failure.

vazois: Done — restructured to attempt CAS only when no bits overlap, with a single \SpinOnce()\ at the bottom for both failure paths.

TedHartMS (approved):

PR #1775, opened 2026-05-06
Pause VectorManager cleanup before cluster/other resets of log

badrishc:
1. Cover reset with a boolean Initializing check that is epoch-protected, so that concurrent iterators end gracefully.
2. Add ability to pause vector cleanup, used before we reset as a second line of defense

### DIFF
--- a/test/Garnet.test/VectorCleanupVsResetRaceTests.cs
+++ b/test/Garnet.test/VectorCleanupVsResetRaceTests.cs
@@ -3,9 +3,10 @@
 
 using System;
 using System.Buffers.Binary;
-using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using Allure.NUnit;
+using Garnet.server;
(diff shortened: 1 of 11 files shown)


test/Garnet.test/VectorCleanupVsResetRaceTests.cs
```diff
@@ -51,6 +57,16 @@ private static object GetStoreWrapper(global::Garnet.GarnetServer s)
             return field?.GetValue(s);
         }
 
+        private static VectorManager GetVectorManager(object storeWrapper)
```
kevin-montrose: Consider using [`[UnsafeAccessor]`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-10.0) instead of reflection.  It's the preferred idiom these days.

kevin-montrose (approved):

PR #1780, opened 2026-05-08
Update lifecycle for compaction, truncate, improve bftree file management

badrishc:


libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -204,86 +280,141 @@ private static uint RoundUpToPowerOf2(uint v)
...
+                }
+                // Already activated by a racing thread — caller should dispose the duplicate.
+                return existing;
+            });
+            // If we lost the race (existing.Tree non-null), dispose our duplicate.
+            if (liveIndexes.TryGetValue(keyId, out var winner) && !ReferenceEquals(winner.Tree, bfTree))
+            {
+                try { bfTree.Dispose(); }
+                catch (Exception ex) { logger?.LogWarning(ex, "Failed to dispose duplicate BfTree on register race"); }
+            }
```
tiagonapoli: minor: According to .NET documentation `AddOrUpdate`:
> updateValueFactory delegates may be executed multiple times to verify the value was added or updated as expected. However, they are called outside the locks to avoid the problems that can arise from executing unknown code under a lock. Therefore, [AddOrUpdate](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=netframework-4.8.1) is not atomic with regards to all other operations on the [ConcurrentDictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=netframework-4.8.1) class

So might not be safe to do mutation in the `AddOrUpdate` callback - theoretically the `(_, existing) =>` callback which changes the existing data in place can have a race, so we could have a scenario like:
- ThreadA: Sees that `existing.Tree is null` and will execute the `if` body, but for some reason is scheduled out
- ThreadB: enters `AddOrUpdate` and updates `existing.Tree`
- ThreadB: Does a TryGetValue and sees that it won the race, so it won't release its handle
- ThreadA: Sets `existing.Tree = bftree` and goes out of `AddOrUpdate`, does a TryGetValue and sees that it won the race, so it won't release its handle

But now ThreadB's handle leaked

I think it would be super rare (maybe not even possible) since I believe we won't have threads racing calling `RegisterIndex` for same index (?), but maybe worth fixing since it could be a concurrent code smell

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -96,42 +126,88 @@ internal sealed class TreeEntry
...
+            this.riLogRoot = riLogRoot;
+            this.cprDir = cprDir;
             this.removeOutdatedCheckpoints = removeOutdatedCheckpoints;
             this.logger = logger;
             rangeIndexLocks = new ReadOptimizedLock(Environment.ProcessorCount);
+
+            if (enabled && !string.IsNullOrEmpty(riLogRoot))
+            {
+                try { Directory.CreateDirectory(riLogRoot); }
+                catch (Exception ex) { logger?.LogWarning(ex, "Failed to create RI log root: {Path}", riLogRoot); }
```
tiagonapoli: nit: If we're not able to create `riLogRoot` directory everything else might also break, right? (e.g., snapshot creation, etc)

Should we log and re-throw the exception so Garnet initialization fails? (at least for managed garnet it would be preferrable to catch these issues early on, sometimes they are related to process permissions for example)

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -148,7 +224,7 @@ internal BfTreeService CreateBfTree(
             string filePath = null;
             if (storageBackend == StorageBackendType.Disk)
             {
-                filePath = DeriveWorkingPath(keyBytes);
+                filePath = LogDataPathFor(keyBytes);
                 Directory.CreateDirectory(Path.GetDirectoryName(filePath)!);
```
tiagonapoli: super nit: directory should already exist since it was created on `RangeIndexConstructor` right? (`LogDataPathFor` will be `riLogRoot/...`)

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -204,86 +280,141 @@ private static uint RoundUpToPowerOf2(uint v)
...
+        /// A crash mid-copy is self-healing: post-recovery either <c>OnRecoverySnapshotRead</c>
+        /// (above-FUA stub) or the next RIPROMOTE-cold (IsFlushed=true stub) re-pre-stages and
+        /// overwrites any partial file before <c>RestoreTree</c> can observe it.</para>
         /// </summary>
-        internal void SnapshotTreeForFlush(ReadOnlySpan<byte> key, ReadOnlySpan<byte> valueSpan)
+        internal void PreStageAndRegisterPending(ReadOnlySpan<byte> keyBytes, long srcFlushAddress)
         {
-            ref readonly var stub = ref ReadIndex(valueSpan);
-            if (stub.TreeHandle == nint.Zero)
+            if (string.IsNullOrEmpty(riLogRoot))
```
tiagonapoli, line 469: Should we enforce that `riLogRoot` is not null nor empty on initialization instead? Looks like having it null or empty could cause some headaches

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -204,86 +280,141 @@ private static uint RoundUpToPowerOf2(uint v)
...
+            // Acquire the per-key exclusive lock for the duration of the file copy AND the
+            // pending-entry registration. This blocks any concurrent RestoreTree (which also
+            // takes the exclusive lock) from observing a partial data.bftree.
+            rangeIndexLocks.AcquireExclusiveLock(keyHash, out var lockToken);
             try
             {
-                var flushPath = DeriveFlushPath(key);
-                Directory.CreateDirectory(Path.GetDirectoryName(flushPath)!);
-                entry.Tree.SnapshotToFile(flushPath);
+                Directory.CreateDirectory(riLogRoot);
```
tiagonapoli: nit: maybe not needed since it was created in the constructor?

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -292,55 +423,175 @@ internal void SnapshotTreeForFlush(ReadOnlySpan<byte> key, ReadOnlySpan<byte> va
...
+        /// <para>Recovery is single-threaded with no concurrent readers, and a crash mid-copy is
+        /// self-healing — the next recovery attempt re-fires <c>OnRecoverySnapshotRead</c> for
+        /// the same stub and fully overwrites any partial file before any <c>RestoreTree</c>
+        /// can observe it.</para>
         /// </summary>
-        internal string DeriveWorkingPath(ReadOnlySpan<byte> keyBytes)
-            => Path.Combine(dataDir ?? string.Empty, "rangeindex", HashKeyToDirectoryName(keyBytes), "data.bftree");
+        internal void RebuildFromSnapshotIfPending(ReadOnlySpan<byte> keyBytes)
+        {
+            if (string.IsNullOrEmpty(riLogRoot) || string.IsNullOrEmpty(cprDir) || recoveredCheckpointToken == Guid.Empty)
```
tiagonapoli: nit: Should we throw instead? If someone is calling `RebuildFromSnapshotIfPending` and we don't have these then this is probably some bug or error condition"?

tiagonapoli (approved):

tiagonapoli (approved):

PR #1784, opened 2026-05-09
Fix Allure wiring: remove duplicate attributes and fix CI check for transitive dependencies

vazois:
## Summary

Fixes two CI issues with Allure test wiring in the newly split cluster replication test projects.

### Issue 1 - Runtime Allure error

Derived test classes (TLS, AsyncReplay, MultiLog) had `[AllureNUnit]` applied directly **and** inherited it from their base classes. NUnit discovers `ITestAction` attributes from both, causing Allure's `BeforeTest` to run twice and crash with:

> `System.InvalidOperationException: Unable to change the container context because the test context is active.`

**Fix:** Removed the duplicate `[AllureNUnit]` from the 4 derived classes - the inherited attribute from the base class is sufficient.

### Issue 2 - CI Allure wiring check failure

The CI `Verify Allure wiring` step used `Assembly.Load()` to find `AllureTestBase` in referenced assemblies. This failed because:

1. `Assembly.Load()` does not probe the build output directory (only works for GAC/default paths)
2. `AllureTestBase` lives in `Garnet.test.cluster`, which is a **transitive** dependency for some projects (e.g., TLS references it through `Garnet.test.cluster.replication`, not directly)

**Fix:** Updated the CI script to search `AppDomain.GetAssemblies()` - all transitive dependencies are already loaded by the runtime during `GetTypes()`.

### Changes

- `ClusterReplicationTLS.cs` - remove duplicate `[AllureNUnit]`
- `ClusterReplicationAsyncReplay.cs` - remove duplicate `[AllureNUnit]`
- `ClusterReplicationShardedLog.cs` - remove duplicate `[AllureNUnit]`
- `ClusterReplicationDisklessSyncShardedLog.cs` - remove duplicate `[AllureNUnit]`
- `.github/workflows/ci.yml` - fix `AllureTestBase` resolution for transitive dependencies


libs/storage/Tsavorite/cs/test/test.hlog/BasicStorageTests.cs
badrishc, line 1: all tests should be under a test folder else the organization gets complicated under cs folder

badrishc (approved):

badrishc (approved):

PR #1787, opened 2026-05-11
Fix GarnetClientSession leak and diskless replication dedup failure

nattress:
Two related bugs in AofTaskStore caused unbounded accumulation of AofSyncTaskInfo tasks on clusters using diskless replication.

1. TryAddReplicationTasks (the diskless path) compared existing tasks against rss.replicaNodeId for dedup. ReplicaSyncSession has two node ID fields: replicaNodeId (set by the disk-based constructor, null for diskless) and replicaSyncMetadata.originNodeId (set by the diskless constructor). The AofSyncTaskInfo was created with originNodeId, but dedup compared against the null replicaNodeId — so it never matched and every call added a new task. Over time numTasks grew unboundedly, inflating the RoleInfo[] from INFO REPLICATION until the response exceeded the network output buffer.

   Fix: use rss.replicaSyncMetadata.originNodeId in the dedup comparison. The singular TryAddReplicationTask (disk-based and CLUSTER AOFSYNC) is unaffected.

2. AofSyncTaskInfo.Dispose() did not dispose its owned GarnetClientSession. When ReplicaSyncTaskAsync is running, CTS cancellation causes it to exit and the finally block cleans up. But when ReplicaSyncTaskAsync has not yet started (e.g. the task fails to be added), Dispose() is the only cleanup path and the session was leaked.

   Fix: add garnetClient?.Dispose() to AofSyncTaskInfo.Dispose() and remove the redundant call from ReplicaSyncTaskAsync's finally block, giving a single disposal site.


test/Garnet.test.cluster/ReplicationTests/AofSyncTaskInfoTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using Garnet.client;
+using Garnet.cluster;
+using Garnet.common;
+using NUnit.Framework;
+
+namespace Garnet.test.cluster
+{
+    [TestFixture]
+    [NonParallelizable]
+    public class AofSyncTaskInfoTests
```
vazois: Is it possible to make this part of an existing test category? Maybe Negative or Management Tests?

nattress: Added \[Category(\REPLICATION\)]\ to the test method, plus \[AllureNUnit]\ and \AllureTestBase\ inheritance on the class to match the repo test conventions.

libs/cluster/Server/Replication/PrimaryOps/AofSyncTaskInfo.cs
```diff
@@ -123,7 +126,6 @@ await iter.BulkConsumeAllAsync(
             }
             finally
             {
-                garnetClient.Dispose();
                 var (address, port) = clusterProvider.clusterManager.CurrentConfig.GetWorkerAddressFromNodeId(remoteNodeId);
                 logger?.LogWarning("AofSync task terminated; client disposed {remoteNodeId} {address} {port} {currentAddress}", remoteNodeId, address, port, previousAddress);
```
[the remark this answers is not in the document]
nattress: Fixed in commit 3f41ad51 - the log message now says 'AofSync task terminated for remote node' instead of 'client disposed'.

Mathos1432:
I recreated your PR after the v2 fork

release/v1: https://github.com/microsoft/garnet/pull/1791
main: https://github.com/microsoft/garnet/pull/1794

PR #1800, opened 2026-05-13
Rework Vector Set deletion and eviction

kevin-montrose:
Rework Vector Set deletion and eviction to use `GarnetRecordTriggers`.

This matches how Range Indexes work, removes the `BADSTATE` possibility, gets a `GUID` comparison out of the hot path, and lays some groundwork for lots of Vector Sets.

Concretely:
 - Switches Vector Set deletion detection to `GarnetRecordTriggers.OnDispose`, getting rid of all the cancelled Tsavorite op stuff
 - _Blocks_ on marking a context needs cleanup, removing all the partial deletion tracking logic
 - Removes the `GUID` stuff for detecting index recreation, instead using the `OnEvict` and `OnDiskRead` callbacks in the same manner as Ranged Indexes
 - Fixes `RecordType` handling during copy updates, fixing expiration

TODO:
 - [x] I'm not convince this approach cleanups up element data if we crash before marking - need to validate that
 - [x] Tests for expiration
 - [x] Update vector-sets.md with final delete design


libs/server/Storage/Functions/GarnetRecordTriggers.cs
```diff
@@ -59,12 +66,19 @@ public GarnetRecordTriggers(CacheSizeTracker cacheSizeTracker, RangeIndexManager
...
         {
-            // Free BfTree and delete data files on key deletion.
-            if (!logRecord.Info.ValueIsObject
-                && reason == DisposeReason.Deleted
-                && logRecord.RecordDataHeader.RecordType == RangeIndexManager.RangeIndexRecordType)
+            if (!logRecord.Info.ValueIsObject)
             {
-                rangeIndexManager?.DisposeTreeUnderLock(logRecord.Key, logRecord.ValueSpan, deleteFiles: true);
+                // Free BfTree and delete data files on key deletion.
+                if (reason == DisposeReason.Deleted && logRecord.RecordDataHeader.RecordType == RangeIndexManager.RangeIndexRecordType)
```
badrishc: nit: add expired here as well to be consistent with vector.

badrishc (approved):

badrishc (approved):

PR #1807, opened 2026-05-18
Add Support for RangeIndex Diskbased Replication

vazois:
## Summary

Adds support for shipping RangeIndex (BfTree) files during disk-based cluster replication, so replicas can fully recreate range index state after checkpoint recovery.

## Work Items

- [x] Validate flush cleanup safety
- [x] Replica-side flush.bftree cleanup on resync
- [x] Add CheckpointFileType entries (STORE_RANGEINDEX_FLUSH, STORE_RANGEINDEX_SNAPSHOT)
- [x] Implement RangeIndex data source (RangeIndexFileDataSource)
- [x] Implement RangeIndex transmit source (RangeIndexFileTransmitSource)
- [x] Implement RangeIndexSnapshotReader
- [x] Implement RangeIndex data sink (RangeIndexFileDataSink)
- [x] Integrate into ReceiveCheckpointHandler
- [x] Integrate into ReplicaSyncSession
- [x] Implement replication tests

### Primary Side (Sender)
- **File enumeration**: `RangeIndexManager.EnumerateFilesForReplication()` yields all flush and snapshot .bftree files for a given checkpoint
- **Data source**: `RangeIndexFileDataSource` — FileStream-based chunked reader with `GetMetadata()` for binary serialization of keyHash + address
- **Transmit source**: `RangeIndexFileTransmitSource` — sends metadata header (startAddress=-1) then streams file content in 128KB chunks
- **Snapshot reader**: `RangeIndexSnapshotReader` — ISnapshotReader that enumerates RI files and yields transmit sources
- **Session integration**: `ReplicaSyncSession.SendCheckpointAsync` adds `RangeIndexSnapshotReader` to the transmission driver when `EnableRangeIndexPreview` is set

### Replica Side (Receiver)
- **Data sink**: `RangeIndexFileDataSink` with `FromMetadata()` factory — zero-copy deserialization from metadata span, derives target path via `RangeIndexManager` helpers, writes chunks via FileStream
- **Handler integration**: `ReceiveCheckpointHandler.ProcessSnapshotData` handles RI metadata headers and file segments

### Shared Infrastructure
- **CheckpointFileType enum**: Added `STORE_RANGEINDEX_FLUSH=7`, `STORE_RANGEINDEX_SNAPSHOT=8`
- **Visibility**: Made `LogFlushPath`, `CheckpointSnapshotPath`, `rangeIndexManager` public for cross-assembly access

## Wire Protocol

Each RI file is transmitted as:
1. **Header** (startAddress=-1): binary metadata — keyHash (32B ASCII) + address (8B LE for flush)
2. **Content chunks** (startAddress=offset): raw file data in 128KB chunks
3. **End marker** (empty data): signals completion


libs/server/StoreWrapper.cs
```diff
@@ -121,7 +121,7 @@ public sealed class StoreWrapper
         /// <summary>
         /// RangeIndex (BfTree) manager shared across sessions
         /// </summary>
-        internal readonly RangeIndexManager rangeIndexManager;
+        public readonly RangeIndexManager rangeIndexManager;
```
badrishc: why does this need to become public?

vazois: Need to pass it down to ClusterProvider through the storeWrapper instance so it can be used during shipping snapshot during full sync.
I can possible pass the RangeIndexManager instance directly in the ClusterProvider.
Other fields in StoreWrapper are public (i.e. appendOnlyFile, store etc.), though they are accessed as a property.

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -55,9 +56,20 @@ public sealed partial class RangeIndexManager : IDisposable
         /// <summary>Size of the RangeIndex stub in bytes.</summary>
         internal const int IndexSizeBytes = RangeIndexStub.Size;
 
+        // Flush filename layout: <hash:32>.<addr:16>.flush.bftree
+        private const int HashPrefixLength = 32;
+        private const int AddrHexLength = 16;
+        private const string FlushSuffix = ".flush.bftree";
```
tiagonapoli, line 62: nit: looks unused

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -891,5 +889,128 @@ void TryDelete(string p)
...
+        /// Enumerates all <c>*.flush.bftree</c> files under <c>riLogRoot</c>, parsing and
+        /// yielding the embedded logical address from each valid filename.
+        /// </summary>
+        /// <returns>Tuples of (fullPath, fileName, logicalAddress). Skips files that don't
+        /// match the expected naming pattern. Returns empty if riLogRoot is not configured
+        /// or does not exist.</returns>
+        private IEnumerable<(string Path, string Name, long Address)> EnumerateFlushFiles()
+        {
+            if (string.IsNullOrEmpty(riLogRoot) || !Directory.Exists(riLogRoot))
+                yield break;
```
tiagonapoli, line 909: nit: In the constructor we throw if riLogRoot is null or empty or if directory creation fails - maybe here would be worth to log an unexpected error and throw as well to avoid masking some issue (e.g., directory got unexpectedly deleted at runtime. What do you think?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/RangeIndexFileDataSource.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <inheritdoc/>
+        public async Task<DataSourceReadResult> ReadNextChunkAsync(CancellationToken cancellationToken = default)
+        {
+            stream ??= new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read, bufferSize: chunkSize, useAsync: true);
+
+            var remaining = EndOffset - CurrentOffset;
+            var bytesToRead = (int)Math.Min(remaining, chunkSize);
+            var buffer = new byte[bytesToRead];
```
tiagonapoli: nit: Since BFTree files can get big, maybe for follow up PR we could optimize these allocations?
We could use `ArrayPool<byte>.Shared`, but the problem is `DataSourceReadResult` currently doesn't have a way to custome `Buffer.Return` - Maybe we would need to implement a `IDataSourceReadResult` and make a `ReturnBuffer` interface method

libs/cluster/Server/Replication/ReplicaOps/DiskbasedReplication/ReceiveCheckpointHandler.cs
```diff
@@ -112,6 +112,13 @@ public void ProcessSnapshotData(Guid token, CheckpointFileType type, long startA
             {
                 switch (type)
                 {
+                    case CheckpointFileType.STORE_RANGEINDEX_FLUSH:
+                    case CheckpointFileType.STORE_RANGEINDEX_SNAPSHOT:
+                        // Create sink immediately from metadata; data chunks will follow
+                        if (!clusterProvider.serverOptions.EnableRangeIndexPreview)
+                            ExceptionUtils.ThrowException(new GarnetException("RangeIndex not enabled but received RI checkpoint data"));
+                        activeSink = RangeIndexFileDataSink.FromMetadata(type, token, data, clusterProvider.rangeIndexManager, logger);
```
tiagonapoli, line 122: nit: just as sanity check, should we check if activeSink is null before assigning it? (and throw and log unexpected if not)

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
```diff
@@ -0,0 +1,328 @@
...
+                var result = context.clusterTestUtils.Execute(endpoint, "RI.GET",
+                    [indexName, $"field{i:D4}"],
+                    logger: context.logger);
+                ClassicAssert.AreEqual($"value{i:D4}", (string)result,
+                    $"Mismatch at field{i:D4} on node {nodeIndex}");
+            }
+        }
+
+        /// <summary>
+        /// Create RI keys on primary, take checkpoint, attach replica via checkpoint-based sync.
```
tiagonapoli: For my understanding - looks like this test is using AOF sync, not checkpoint-based sync - am I missing something? Or comment is not up to date?

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
```diff
@@ -0,0 +1,328 @@
...
+            ValidateRangeIndex(replicaIndex, "idx1", 20);
+        }
+
+        /// <summary>
+        /// Checkpoint sync → shutdown replica → more RI.SET on primary → restart replica.
+        /// Verify replica catches up.
+        /// </summary>
+        [Test, Order(5)]
+        [Category("REPLICATION")]
+        public void ClusterRangeIndexReplicationRestartReplica()
```
tiagonapoli, line 246: nit: Do you think it's worth adding validations around the files on disk to ensure we have exercised the intended replication paths? (e.g., snapshot and/or flush)

libs/server/Resp/RangeIndex/RangeIndexManager.cs
vazois: Good idea, though in this case this method is used for discovery which means if we throw, we break the whole sync process.  The directory might legitimately not exist Maybe add a logging message here?

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/RangeIndexFileDataSource.cs
vazois: Good idea. Since the read is sequential right now, we can re-use the buffer and keep chunk size under 80kb limit to avoid fragmentation.
I made it part of the dataSource not the readResult which should be just a lightweight wrapper around the read result, in my opinion.

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
vazois: This is probably a test for the RangeIndex itself. These tests are exercising replication which means that validation is implicit. We write something on the primary and then wait to observe it on the replica.

We should have test for this but it should be part of the standalone range index tests, right?

libs/server/Resp/RangeIndex/RangeIndexManager.cs
tiagonapoli: Yup, agree!

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
tiagonapoli: > We should have test for this but it should be part of the standalone range index tests, right?

Yes, makes sense, I suppose standalone range index already have tests around AOF replication to replicas (?) - Maybe worth checking, if not we can move these ones there

libs/cluster/Server/Replication/PrimaryOps/DiskbasedReplication/RangeIndexFileDataSource.cs
```diff
@@ -0,0 +1,161 @@
...
+            stream ??= new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read, bufferSize: chunkSize, useAsync: true);
+
+            var remaining = EndOffset - CurrentOffset;
+            var bytesToRead = (int)Math.Min(remaining, chunkSize);
+
+            var bytesRead = await stream.ReadAsync(buffer, 0, bytesToRead, cancellationToken).ConfigureAwait(false);
+
+            if (bytesRead == 0)
+            {
+                logger?.LogWarning("RangeIndexFileDataSource: unexpected EOF at offset {currentOffset}, expected {endOffset} for {filePath}", CurrentOffset, EndOffset, filePath);
```
tiagonapoli: nit: should we throw instead to avoid sending corrupt/incomplete data?

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
```diff
@@ -0,0 +1,328 @@
...
+            var nodesCount = 2;
+
+            context.CreateInstances(nodesCount, disableObjects: true, enableAOF: true,
+                lowMemory: true, enableRangeIndexPreview: true);
+            context.CreateConnection();
+            var (_, _) = context.clusterTestUtils.SimpleSetupCluster(1, 1, logger: context.logger);
+
+            // Populate range indexes — low memory should trigger eviction
+            PopulateRangeIndex(primaryIndex, "idx1", 10);
+            PopulateRangeIndex(primaryIndex, "idx2", 10);
```
tiagonapoli, line 161: nit: should we assert the expected files are on disk to ensure we entered the eviction to disk scenario?

tiagonapoli (approved):

badrishc (approved):

test/cluster/Garnet.test.cluster.replication.rangeindex/ClusterRangeIndexReplicationTests.cs
vazois: there is an existing test that I added validation to ([17cbde7](https://github.com/microsoft/garnet/pull/1807/commits/17cbde7b9e5109c446927c820b66bd258896dd87)), so that it ensures that flush.bftree files are created.

tiagonapoli (approved):

PR #1822, opened 2026-05-22
Add per-name ACL support for custom (extension) commands

### LINKED ISSUE #1821: ACL: support per-name allow/deny rules for custom commands
### Feature request type

enhancement

Garnet's ACL system tracks per-command permissions in a fixed-size bitmap indexed by `RespCommand`. Built-in commands fit this model cleanly, but custom (extension) commands have `RespCommand` IDs assigned dynamically at module load and live outside the bitmap's range.

The practical effect is that ACL rules cannot reference individual custom commands. The only knob available today is the all-or-nothing `+@custom` category, which either grants access to every loaded custom command or to none of them. There is no way to express a rule like `+json.set` or `-json.get` for a specific extension command.

This makes it difficult to grant scoped access to custom commands in any deployment that has more than one custom command and more than one role. For example, a read-only role cannot be granted `JSON.GET` while being denied `JSON.SET` without also affecting every other extension command on the server.

Introduce per-name allow/deny entries on the user's permission set, stored alongside the existing command bitmap. Teach the ACL parser to accept custom command names as first-class tokens, and have dispatch for custom commands consult the per-name entries before falling back to the bitmap.

Expected behavior:
- `ACL SETUSER alice +json.get -json.set` should grant `alice` access to `JSON.GET` while denying `JSON.SET`, independent of any category-level rule.
- An explicit per-name deny (`-name`) should beat a category-level allow (`+@custom`, `+@all`), matching the precedence model that Redis uses for built-in commands.
- ACL file load should remain robust to module-load order: a file that references a custom command whose module has not yet loaded should not break startup. An opt-in strict mode can be offered for operators who want startup-time typo protection.
- `ACL SETUSER` issued at runtime should be strict, since modules are already loaded by then.

Happy to put up a PR for this if the direction seems reasonable.

- Keep using `+@custom` and rely on module-level separation. This forces an all-or-nothing trust boundary per server and doesn't compose with multi-role deployments.
- Mint static `RespCommand` IDs for known modules so they fit the existing bitmap. This couples the ACL system to a fixed module set and breaks the dynamic-loading model.

### Additional context

_No response_

Ankith-Kandala:
Closes #1821

## Problem

Garnet's ACL system tracks per-command permissions in a fixed-size bitmap indexed by `RespCommand`. That works well for built-in commands but breaks down for custom/extension commands: their `RespCommand` IDs are assigned dynamically at module load and live outside the bitmap's range. The only way to grant access to an extension command today is the all-or-nothing `+@custom` category. There is no way to write `+json.set` or `-json.get` for a specific extension command, which blocks deployments that need fine-grained per-command RBAC.

## What this change does

Adds first-class per-name allow/deny entries for custom commands on the user's permission set, stored alongside the existing command bitmap. ACL rules now accept custom command names directly, and dispatch consults the per-name lists before falling back to the bitmap.

### Highlights

- **`CommandPermissionSet`** carries two `FrozenSet<string>` instances (`_customAllowed`, `_customDenied`) next to the existing bitmap. Lookup is `OrdinalIgnoreCase` to match `CustomCommandManager`'s normalization. Custom-command dispatch checks deny first, then allow, then falls back to the bitmap.
- **`ACLParser`** routes unknown but syntactically valid tokens (ASCII alphanumerics plus `.`, `_`, `-`, `|`, up to 64 chars, must start with an alphanumeric) to the custom-name path instead of throwing.
- **`User`** exposes `AddCustomCommand`/`RemoveCustomCommand` with the same CAS pattern used for `AddCommand`, plus a per-user cap (`MaxCustomCommandsPerUser = 512`) to bound description growth and per-user memory.
- **ACL file load** is lenient by default: a name whose module hasn't loaded yet is accepted, so file loads aren't brittle to module-load order. A new opt-in config flag `acl-strict-custom-commands` makes startup fail if any name is still unresolved after modules finish loading.
- **`ACL SETUSER` at runtime** is always strict: a new `+name`/`-name` op must resolve against `CustomCommandManager`, otherwise it throws. Names already on the user (e.g. loose-loaded from an ACL file) can be toggled allow/deny without re-checking registration.
- **`GarnetServer`** runs a `ValidateCustomCommandACLs` pass after modules load to surface the strict/lenient outcome to the operator.
- **Docs:** `website/docs/commands/acl.md` documents the new per-name SETUSER syntax and the `acl-strict-custom-commands` flag; `website/docs/getting-started/configuration.md` lists the new flag in the config table.

### Design notes

- The on-disk ACL file format is unchanged; only the set of accepted tokens is widened.
- The built-in command bitmap layout is unchanged. Dispatch for non-custom commands takes the exact same path as before.
- `IsEquivalentTo` was tightened to also compare per-name sets so description rationalization can't silently drop a `-json.set` entry.

### New test suite

`test/standalone/Garnet.test.acl/Resp/ACL/CustomCommandACLTests.cs` adds 36 tests covering:

- Parser fallback and name validation: whitespace, non-ASCII, leading punctuation, length-limit boundary, last-write-wins, deny precedence on the same name, case-insensitive normalization.
- Persistence round-trip via `DescribeUser`.
- Per-user cap: enforcement, idempotent re-add, allow/deny swap at cap, the same on the Remove path.
- Public API null safety.
- Case-insensitive per-name store.
- Permission-set semantics: deny-over-bitmap, allow-over-missing-bitmap, `All` sentinel behavior including `+@all -name`, default fail-closed, `IsEquivalentTo` considers custom sets.
- SETUSER strict validation: unknown name rejected, registered name accepted, invalid syntax rejected, command registered without `RespCommandsInfo` accepted, plus a regression test that allow/deny swap of loose-loaded names is still accepted after the strict check.
- Dispatch end-to-end across all four custom `RespCommand` variants (`CustomTxn`, `CustomRawStringCmd`, `CustomObjCmd`, `CustomProcedure`).
- `ACL LIST` round-trip.

### Existing test runs

- `Garnet.test.acl` (excluding tests that require the native vector/range-index libraries): 451/451 on net8.0 and net10.0.
- `Garnet.test` cross-feature filter (ACL/Auth/Lua/Parse/Command/Permission/User): 789/789 on net10.0 (3 Azure-only tests skipped).
- `Garnet.test` built-in custom-command tests: 7/7.
- `Garnet.test.cluster` ACL/Auth/Custom subset: 9/9.
- Full solution build: 0 warnings, 0 errors.

### Pre-existing test updated

`AclConfigurationFileTests.BadInputMalformedStatement` had sub-cases asserting that `+none` and `+invalid` would throw on ACL file load. Both tokens are syntactically valid custom names under the new parser fallthrough and now load successfully under lenient mode. The sub-cases were updated to assert clean startup with the tokens present and to document that operators wanting typo protection can enable `acl-strict-custom-commands`. The remaining `badinput`-in-user-name-position sub-case still throws and is unchanged.

### DIFF
--- a/libs/host/Configuration/Options.cs
+++ b/libs/host/Configuration/Options.cs
@@ -174,7 +174,7 @@ namespace Garnet
         public string AclFile { get; set; }
 
         [OptionValidation]
-        [Option("acl-strict-custom-commands", Required = false, HelpText = "If true, the server refuses to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings.")]
+        [Option("acl-strict-custom-commands", Required = false, HelpText = "If true (default), the server refuses to start when an ACL rule references a custom (extension) command name that no loaded module has registered. Set to false to load unresolved names as-is and log warnings.")]
         public bool? AclStrictCustomCommands { get; set; }
 
         [Option("aad-authority", Required = false, HelpText = "The authority of AAD authentication.")]
@@ -855,7 +855,7 @@ namespace Garnet
                 ParallelMigrateTaskCount = ParallelMigrateTaskCount,
                 FastMigrate = FastMigrate.GetValueOrDefault(),
                 AuthSettings = GetAuthenticationSettings(logger),
-                AclStrictCustomCommands = AclStrictCustomCommands.GetValueOrDefault(),
+                AclStrictCustomCommands = AclStrictCustomCommands.GetValueOrDefault(true),
                 EnableAOF = EnableAOF.GetValueOrDefault(),
                 EnableLua = EnableLua.GetValueOrDefault(),
                 LuaTransactionMode = LuaTransactionMode.GetValueOrDefault(),
--- a/libs/host/Configuration/Options.cs
+++ b/libs/host/Configuration/Options.cs
@@ -173,6 +173,10 @@ namespace Garnet
         [Option("acl-file", Required = false, HelpText = "External ACL user file.")]
         public string AclFile { get; set; }
 
+        [OptionValidation]
+        [Option("acl-strict-custom-commands", Required = false, HelpText = "If true, the server refuses to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings.")]
+        public bool? AclStrictCustomCommands { get; set; }
+
         [Option("aad-authority", Required = false, HelpText = "The authority of AAD authentication.")]
         public string AadAuthority { get; set; }
 
@@ -851,6 +855,7 @@ namespace Garnet
                 ParallelMigrateTaskCount = ParallelMigrateTaskCount,
                 FastMigrate = FastMigrate.GetValueOrDefault(),
                 AuthSettings = GetAuthenticationSettings(logger),
+                AclStrictCustomCommands = AclStrictCustomCommands.GetValueOrDefault(),
                 EnableAOF = EnableAOF.GetValueOrDefault(),
                 EnableLua = EnableLua.GetValueOrDefault(),
                 LuaTransactionMode = LuaTransactionMode.GetValueOrDefault(),
--- a/libs/host/GarnetServer.cs
+++ b/libs/host/GarnetServer.cs
@@ -356,7 +356,8 @@ namespace Garnet
             {
                 // Strict mode: fail closed so operators can't accidentally ship an ACL with typos
                 // that would silently match no command (and therefore deny by default at dispatch).
-                throw new GarnetException($"ACL strict mode: {unresolved.Count} unresolved (user, custom-command) entries in ACL rules. Disable acl-strict-custom-commands or load the appropriate module(s).");
+                var entries = string.Join(", ", unresolved.Select(t => $"({t.user},{t.name})"));
+                throw new GarnetException($"ACL strict mode: {unresolved.Count} unresolved (user, custom-command) entries in ACL rules: {entries}. Disable acl-strict-custom-commands or load the appropriate module(s).");
             }
         }
 
--- a/libs/host/GarnetServer.cs
+++ b/libs/host/GarnetServer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -302,6 +303,61 @@ namespace Garnet
                 servers[i].Register(WireFormat.ASCII, Provider);
 
             LoadModules(customCommandManager);
+
+            // ACL rules are parsed before modules load, so per-name custom-command entries land
+            // on users unresolved. Sweep them now against the registered modules.
+            ValidateCustomCommandACLs(customCommandManager);
+        }
+
+        private void ValidateCustomCommandACLs(CustomCommandManager customCommandManager)
+        {
+            if (opts.AuthSettings is not AclAuthenticationSettings)
+            {
+                return;
+            }
+
+            var acl = storeWrapper.accessControlList;
+            if (acl == null)
+            {
+                return;
+            }
+
+            var unresolved = new List<(string user, string name)>();
+            foreach (var kv in acl.GetUserHandles())
+            {
+                var u = kv.Value.User;
+                foreach (var name in u.CustomCommandsAllowed)
+                {
+                    if (!customCommandManager.IsCustomCommandRegistered(name))
+                    {
+                        unresolved.Add((kv.Key, name));
+                    }
+                }
+                foreach (var name in u.CustomCommandsDenied)
+                {
+                    if (!customCommandManager.IsCustomCommandRegistered(name))
+                    {
+                        unresolved.Add((kv.Key, name));
+                    }
+                }
+            }
+
+            if (unresolved.Count == 0)
+            {
+                return;
+            }
+
+            foreach (var (user, name) in unresolved)
+            {
+                logger?.LogWarning("ACL rule references custom command '{name}' for user '{user}' which is not registered with any loaded module", name, user);
+            }
+
+            if (opts.AclStrictCustomCommands)
+            {
+                // Strict mode: fail closed so operators can't accidentally ship an ACL with typos
+                // that would silently match no command (and therefore deny by default at dispatch).
+                throw new GarnetException($"ACL strict mode: {unresolved.Count} unresolved (user, custom-command) entries in ACL rules. Disable acl-strict-custom-commands or load the appropriate module(s).");
+            }
         }
 
         private GarnetDatabase CreateDatabase(int dbId, GarnetServerOptions serverOptions, ClusterFactory clusterFactory,
--- a/libs/host/defaults.conf
+++ b/libs/host/defaults.conf
@@ -108,8 +108,8 @@
 	/* External ACL user file. */
 	"AclFile" : null,
 
-	/* Refuse to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings. */
-	"AclStrictCustomCommands" : false,
+	/* Refuse to start when an ACL rule references a custom (extension) command name that no loaded module has registered. Default true (strict). Set to false to load unresolved names as-is and log warnings. */
+	"AclStrictCustomCommands" : true,
 
 	/* The authority of AAD authentication. */
 	"AadAuthority" : "https://login.microsoftonline.com",
--- a/libs/host/defaults.conf
+++ b/libs/host/defaults.conf
@@ -108,6 +108,9 @@
 	/* External ACL user file. */
 	"AclFile" : null,
 
+	/* Refuse to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings. */
+	"AclStrictCustomCommands" : false,
+
 	/* The authority of AAD authentication. */
 	"AadAuthority" : "https://login.microsoftonline.com",
 
--- a/libs/server/ACL/ACLParser.cs
+++ b/libs/server/ACL/ACLParser.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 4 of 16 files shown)


libs/host/Configuration/Options.cs
```diff
@@ -173,6 +173,10 @@ internal sealed class Options : ICloneable
         [Option("acl-file", Required = false, HelpText = "External ACL user file.")]
         public string AclFile { get; set; }
 
+        [OptionValidation]
+        [Option("acl-strict-custom-commands", Required = false, HelpText = "If true, the server refuses to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings.")]
+        public bool? AclStrictCustomCommands { get; set; }
```
kevin-montrose, line 178: I would like to see tests for this option - look at `GarnetServerConfigTests.cs` for examples (things like `EnableVectorSetPreview` are the minimum).  We've had problems in the past of configs "working" because the defaults are reasonable, but not actually being usable in practice.

libs/host/GarnetServer.cs
```diff
@@ -302,6 +303,61 @@ private void InitializeServer()
...
+            {
+                logger?.LogWarning("ACL rule references custom command '{name}' for user '{user}' which is not registered with any loaded module", name, user);
+            }
+
+            if (opts.AclStrictCustomCommands)
+            {
+                // Strict mode: fail closed so operators can't accidentally ship an ACL with typos
+                // that would silently match no command (and therefore deny by default at dispatch).
+                throw new GarnetException($"ACL strict mode: {unresolved.Count} unresolved (user, custom-command) entries in ACL rules. Disable acl-strict-custom-commands or load the appropriate module(s).");
+            }
```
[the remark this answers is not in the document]
kevin-montrose: I agree with this comment - add a test to cover this being thrown when expected.

libs/server/ACL/ACLParser.cs
```diff
@@ -304,6 +319,49 @@ static bool IsInvalidCommandToAcl(RespCommand command)
             => command == RespCommand.INVALID || command == RespCommand.NONE || command.NormalizeForACLs() != command;
         }
 
+        /// <summary>
+        /// Maximum length (in chars) of a custom command name accepted in an ACL rule.
+        /// </summary>
+        internal const int MaxCustomCommandNameLength = 64;
```
kevin-montrose: nit: pull this up to the top of the type

libs/server/ACL/ACLParser.cs
```diff
@@ -304,6 +319,49 @@ static bool IsInvalidCommandToAcl(RespCommand command)
...
+        /// first character must be alphanumeric.
+        /// </summary>
+        internal static bool IsValidCustomCommandName(string name)
+        {
+            if (string.IsNullOrEmpty(name) || name.Length > MaxCustomCommandNameLength)
+            {
+                return false;
+            }
+
+            char first = name[0];
```
kevin-montrose: Loop like this is suboptimal - we can use some `SearchValues` to hopefully vectorize it.

Something like:

```csharp
private static SearchValues<char> LegalFirstChars { get; } = SearchValues.Create("ABCDEFGHIJKLMNOPQRSTUVWXYZabdefghijklmnopqrstuvwxyz0123456789");
private static SearchValues<char> LegalRestChars { get; } = SearchValues.Create("ABCDEFGHIJKLMNOPQRSTUVWXYZabdefghijklmnopqrstuvwxyz0123456789._-|");

static bool IsValidCustomCommandName(string name)
{
	if (string.IsNullOrEmpty(name) || name.Length > MaxCustomCommandNameLength)
    {
        return false;
    }
	
	var nameSpan = name.AsSpan();
	
	return LegalFirstChars.Contains(nameSpan[0]) && !nameSpan[1..].ContainsAnyExcept(LegalRestChars); 
}
```

libs/server/ACL/User.cs
```diff
@@ -79,6 +79,31 @@ public User(User user)
...
+        /// <param name="genericCmd">One of CustomRawStringCmd/CustomObjCmd/CustomTxn/CustomProcedure.</param>
+        /// <param name="customName">The custom command name (case-insensitive match).</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool CanAccessCustomCommand(RespCommand genericCmd, string customName)
+        => this._enabledCommands.CanRunCustomCommand(genericCmd, customName);
+
+        /// <summary>
+        /// Maximum number of per-name custom command entries (allow + deny combined) per user.
+        /// </summary>
+        internal const int MaxCustomCommandsPerUser = 512;
```
kevin-montrose: What's the justification for this limit?  The upper bound is the number of registered custom  commands.

libs/host/defaults.conf
```diff
@@ -108,6 +108,9 @@
 	/* External ACL user file. */
 	"AclFile" : null,
 
+	/* Refuse to start when an ACL rule references a custom (extension) command name that no loaded module has registered. If false (default), unresolved names are loaded as-is and logged as warnings. */
+	"AclStrictCustomCommands" : false,
```
kevin-montrose: I'd argue this should be strict by default, preserving older behavior and also defaulting to safer behavior.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -46,6 +46,13 @@ public class GarnetServerOptions : ServerOptions
         /// </summary>
         public IAuthenticationSettings AuthSettings = null;
 
+        /// <summary>
+        /// If true, the server refuses to start when an ACL rule references a custom command name
+        /// that no loaded module has registered. If false (default), unresolved names are logged
+        /// as warnings.
+        /// </summary>
+        public bool AclStrictCustomCommands = false;
```
kevin-montrose: Per prior comment, I think default `true` is more correct here.

kevin-montrose (changes requested):

libs/server/ACL/User.cs
```diff
@@ -338,6 +363,97 @@ public void RemoveCommand(RespCommand command)
...
+        /// </summary>
+        /// <param name="customName">Custom command name. Normalized to uppercase; matching is case-insensitive.</param>
+        public void AddCustomCommand(string customName)
+        {
+            ArgumentNullException.ThrowIfNull(customName);
+
+            string normalized = customName.ToUpperInvariant();
+
+            CommandPermissionSet prev = this._enabledCommands;
+            string descUpdate = $"+{customName.ToLowerInvariant()}";
```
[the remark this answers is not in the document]
Ankith-Kandala: Addressed in a1d08ccd. Both methods now validate via `ACLParser.IsValidCustomCommandName` and the persisted `descUpdate` is built from the normalized token.

website/docs/commands/acl.md
```diff
@@ -141,6 +141,12 @@ Create an ACL user with the specified rules or modify the rules of an existing u
 
 Manipulate Garnet ACL users interactively. If the username does not exist, the command creates the username without any privilege. It then reads from left to right all the rules provided as successive arguments, setting the user ACL rules as specified. If the user already exists, the provided ACL rules are simply applied in addition to the rules already set.
 
+#### Custom (extension) command rules
+
+In addition to built-in commands and `@category` rules, individual custom command names can be used directly as rules. For example, `+json.set` grants access to `JSON.SET` and `-json.get` denies access to `JSON.GET`. An explicit per-name deny takes precedence over a category-level allow, so `+@custom -json.set` allows every custom command except `JSON.SET`. Name matching is case-insensitive.
+
+`ACL SETUSER` is strict at runtime: a `+name`/`-name` rule whose name is not registered with any loaded module is rejected. ACL file load is lenient by default so files are not brittle to module load order; set `acl-strict-custom-commands` to `true` if you want startup to fail when an ACL file references an unresolved custom command name.
```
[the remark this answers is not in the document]
Ankith-Kandala: Addressed in a1d08ccd: added a sentence noting that names already on the user (e.g. loose-loaded from an ACL file before modules registered) can be toggled allow/deny without re-validation.

libs/host/GarnetServer.cs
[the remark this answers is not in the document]
Ankith-Kandala: Added in a1d08ccd: `StrictMode_StartupFails_WhenAclFileReferencesUnregisteredCustomCommand` and `LenientMode_StartupSucceeds_WhenAclFileReferencesUnregisteredCustomCommand` in `CustomCommandACLTests.cs`.

libs/host/Configuration/Options.cs
Ankith-Kandala: Added `AclStrictCustomCommands()` to `GarnetServerConfigTests.cs` in a1d08ccd, modeled on `EnableVectorSetPreview`. Covers CLI default, CLI explicit true/false, JSON empty/true/false, and JSON invalid-value rejection.

libs/server/ACL/ACLParser.cs
Ankith-Kandala: Addressed in a1d08ccd.

libs/server/ACL/ACLParser.cs
Ankith-Kandala: Addressed in a1d08ccd using your suggested `SearchValues` pattern.

libs/server/ACL/User.cs
Ankith-Kandala: Dropped in a1d08ccd. The number was arbitrary, and the description-injection vector it loosely guarded against is now covered by the input validation in `User.AddCustomCommand`/`RemoveCustomCommand`.

libs/host/defaults.conf
Ankith-Kandala: Agreed and addressed in a1d08ccd. Flipped to `true` across `defaults.conf`, `GarnetServerOptions`, and `Options.GetValueOrDefault`. Docs and help text updated to match.

libs/server/Servers/GarnetServerOptions.cs
Ankith-Kandala: Addressed in a1d08ccd.

libs/server/ACL/ACLParser.cs
```diff
@@ -11,6 +12,19 @@ class ACLParser
     {
         private static readonly char[] WhitespaceChars = [' ', '\t', '\r', '\n'];
 
+        /// <summary>
+        /// Maximum length (in chars) of a custom command name accepted in an ACL rule.
+        /// </summary>
+        internal const int MaxCustomCommandNameLength = 64;
```
kevin-montrose: Thinking on this a little more - custom command registration (in `CustomCommandManager.cs`) doesn't impose this limit, and glancing at Redis code neither do they.

Imposing this limit means it'd be possible to register a command that couldn't be ACL'd, which is unexpected.  Let's just drop it.

test/standalone/Garnet.test/TestUtils.cs
```diff
@@ -272,6 +272,7 @@ public static GarnetServer CreateGarnetServer(
             string defaultPassword = null,
             bool useAcl = false, // NOTE: Temporary until ACL is enforced as default
             string aclFile = null,
+            bool aclStrictCustomCommands = false,
```
kevin-montrose: Default this true to match .conf defaults.

kevin-montrose (changes requested):
Two small changes, but quite close to ready IMO.

libs/server/ACL/ACLParser.cs
Ankith-Kandala: Nice find. Removed `MaxCustomCommandNameLength` and the related test.

test/standalone/Garnet.test/TestUtils.cs
Ankith-Kandala: Done. Flipped the default to `true` so the test utility matches the production default. Existing tests that rely on the relaxed behavior pass `false` explicitly.

kevin-montrose (approved):

PR #1827, opened 2026-05-26
Fix missing cluster slot verification for CustomRawStringCmd and CustomObjCmd

### LINKED ISSUE #1826: Cluster: missing slot verification for custom raw-string and custom object commands
### Describe the bug

In cluster mode, single-key custom raw-string commands and custom object commands (registered via `server.Register.NewCommand(...)`) are dispatched on the local node without going through cluster slot verification. The node accepts and serves the command regardless of which slot the key belongs to.

In a multi-node cluster this has two effects:

1. A client sending a custom command to the wrong node never receives a `-MOVED <slot> <ip:port>` redirection, so it has no way to route correctly.
2. If the same key is written through two different nodes, both nodes silently accept the write and end up with divergent values for the same key.

Built-in single-key commands like `SET`/`GET` against the same key on the same node behave correctly (the non-owning node returns `MOVED`).

Custom transactions (registered via `NewTransactionProc`) are not affected. They already go through per-key verification via `TxnKeyManager` since PR #712. The gap is only on the raw-string (`CustomRawStringCmd`) and object (`CustomObjCmd`) dispatch paths.

The stock `main/GarnetServer` build already registers `MYDICTSET`/`MYDICTGET` (custom object) and `SETIFPM`/`SETWPIFPGT`/`DELIFM` (custom raw-string), so no module loading or extra registration is needed.

1. Start two `GarnetServer` instances with `--cluster`.
2. Form a cluster via `CLUSTER MEET` and assign slots so node A owns slot 0 and node B owns the rest.
3. From a RESP client connected to node B (which does NOT own slot 0), send a custom command against a key that hashes to slot 0:
   - `MYDICTSET keyOwnedByNodeA field1 value1` -> node B replies `+OK`.
   - The raw-string path behaves the same way: e.g. `SETIFPM keyOwnedByNodeA value prefix` is also accepted locally on node B.
4. Compare with a built-in command against the same key on the same node:
   - `SET keyOwnedByNodeA value` -> node B replies `-MOVED 0 <nodeA-ip:port>`, as expected.

Node B silently accepted the custom-command write for a slot it does not own. If a different client follows the correct `MOVED` redirection to node A and reads the same key, the two nodes hold different values for it.

### Expected behavior

Single-key custom raw-string and custom object commands should go through the same verification path as built-in single-key commands:

- Return `-MOVED <slot> <ip:port>` when the key's slot is owned by another node.
- Return `-ASK <slot> <ip:port>` when the slot is migrating out from this node.
- Honor the same replica-read rules (e.g. `READONLY` semantics).

Both dispatch contracts already place the user-supplied key at `parseState[0]` as a single key, and `CommandType` (`Read` vs. `ReadModifyWrite`/`Write`) is tracked per-session, so a single-key spec routed through the existing `NetworkMultiKeySlotVerify` path is enough to close the gap.

### Screenshots

_No response_

### Release version

_No response_

### IDE

_No response_

### OS version

_No response_

### Additional context

- Same class of correctness gap as PR #712, which introduced the iterative slot verification interface and wired custom transactions through it. This issue is the remaining gap on the other two extension command paths: `CustomRawStringCmd` and `CustomObjCmd`.
- `CustomProcedure` (non-transactional procedures registered via `NewProcedure`) is out of scope here. Procedure authors validate keys themselves and the key contract is more open-ended; may be worth a separate follow-up if a centralized path is desirable there too.

Ankith-Kandala:
Closes #1826

## Problem

`RespServerSession.CanServeSlot` decides whether a command needs cluster slot verification by checking `RespCommand.IsDataCommand()`. That check returns `false` for the four custom-command enum values (`CustomRawStringCmd`, `CustomObjCmd`, `CustomTxn`, `CustomProcedure`) because they sit past `LastDataCommand` (EVALSHA) in the enum range.

As a result, `CustomRawStringCmd` and `CustomObjCmd` skipped slot verification entirely in cluster mode, so a client could hit a node that doesn't own the key and never get a `MOVED` redirect back. `CustomTxn` was unaffected because `CustomTransactionProcedure.RegisterKey` already runs every key through `txnManager.VerifyKeyOwnership`. `CustomProcedure` is left to the procedure author since procedures can touch arbitrary keys. The gap this PR closes is the two remaining types, both of which have a fixed contract: the key is always at `parseState[0]`, enforced by `TryCustomRawStringCommand` and `TryCustomObjectCommand`.

## What this change does

Adds an explicit switch arm at the top of `CanServeSlot` that handles `CustomRawStringCmd` and `CustomObjCmd` by driving `NetworkMultiKeySlotVerify` with a synthetic single-key spec.

- A `static readonly` `CustomCommandSingleKeySpec` (`SimpleRespKeySpec[]`) describes the contract: `BeginSearch(index: 1)`, `FindKeys(keyStep: 1, lastKeyOrLimit: 0, isLimit: false)`. Allocated once at type load, so no per-request allocation is added.
- `csvi.readOnly` is derived from the parsed command's type (`currentCustomRawStringCommand.type` or `currentCustomObjectCommand.type == CommandType.Read`). `cmd.IsReadOnly()` cannot be used here because both custom enum values fall above `LastReadCommand`.
- The new arm runs before the existing `IsDataCommand()` bypass, so no other command's slot check is affected.

## Testing

Added `test/cluster/Garnet.test.cluster.migrate/RedirectTests/TestClusterCustomCommands.cs`, which registers four minimal custom commands so the existing slot-verification matrix covers both branches of the new code path:

- `CLUSTERTESTRAWCMD` and `CLUSTERTESTOBJCMD` exercise `readOnly=false` (`CommandType.ReadModifyWrite`).
- `CLUSTERTESTRAWREADCMD` and `CLUSTERTESTOBJREADCMD` exercise `readOnly=true` (`CommandType.Read`). The read-only stub throws from every write-path method so an accidental write-path dispatch becomes a hard test failure.

`ClusterSlotVerificationTests.OneTimeSetUp` registers all four via `Register.NewCommand` on each node and adds them to the `TestCommands` array, so the existing `CLUSTERDOWN`/`OK`/`MOVED`/`ASK`/`CROSSSLOT`/`TRYAGAIN` matrix automatically runs against the new commands.

Test runs on `net10.0` Debug:

- `Garnet.test.cluster.migrate`: 54/54
- `Garnet.test.cluster`: 137/137
- `Garnet.test.cluster.multilog`: 144/144
- `Garnet.test.cluster.replication`: 107/107
- `Garnet.test.cluster.replication.asyncreplay`: 105/105
- `Garnet.test.cluster.replication.disklesssync`: 29/29
- `Garnet.test.cluster.replication.tls`: 105/105
- `Garnet.test.cluster.vectorsets`: 16/16
- `Garnet.test`: 789/792 (3 pre-existing skips unrelated to this change)

`dotnet format Garnet.slnx --verify-no-changes` and `dotnet format Tsavorite.slnx --verify-no-changes` both clean. Full solution build: 0 warnings, 0 errors.

### DIFF
--- a/libs/server/Resp/RespServerSessionSlotVerify.cs
+++ b/libs/server/Resp/RespServerSessionSlotVerify.cs
@@ -10,7 +10,7 @@ namespace Garnet.server
     /// </summary>
     internal sealed unsafe partial class RespServerSession : ServerSessionBase
     {
-        // Single-key spec for custom raw-string/object commands (key at parseState[0])
+        // Key spec for CustomRawStringCmd and CustomObjCmd: a single key at arg index 1.
         private static readonly SimpleRespKeySpec[] CustomCommandSingleKeySpec =
         [
             new SimpleRespKeySpec
@@ -21,7 +21,7 @@ namespace Garnet.server
         ];
 
         /// <summary>
-        /// Validate if this command can be served based on the current slot assignment
+        /// Validates whether a command can be served based on the current slot assignment
         /// </summary>
         /// <param name="cmd"></param>
         /// <returns></returns>
@@ -29,27 +29,14 @@ namespace Garnet.server
         {
             Debug.Assert(clusterSession != null);
 
-            // CustomRawStringCmd and CustomObjCmd fall outside IsDataCommand()'s range and
-            // would otherwise skip slot verification, so route them through a synthetic
-            // single-key spec (both dispatch via parseState[0] as the only key)
-            if (cmd is RespCommand.CustomRawStringCmd or RespCommand.CustomObjCmd)
-            {
-                // We can't use cmd.IsReadOnly() since both custom enum values fall above LastReadCommand
-                var isReadOnly = cmd == RespCommand.CustomRawStringCmd
-                    ? currentCustomRawStringCommand.type == CommandType.Read
-                    : currentCustomObjectCommand.type == CommandType.Read;
-
-                csvi.keySpecs = CustomCommandSingleKeySpec;
-                csvi.isSubCommand = false;
-                csvi.readOnly = isReadOnly;
-                csvi.sessionAsking = SessionAsking;
-                csvi.waitForStableSlot = false;
-                return !clusterSession.NetworkMultiKeySlotVerify(ref parseState, ref csvi, ref dcurr, ref dend);
-            }
-
-            // Verify slot for command if it falls into data command category
             if (!cmd.IsDataCommand())
+            {
+                // Custom commands sit outside IsDataCommand but still touch user keys.
+                if (cmd is RespCommand.CustomRawStringCmd or RespCommand.CustomObjCmd)
+                    return CanServeSlotForCustomCommand(cmd);
+
                 return true;
+            }
 
             cmd = cmd.NormalizeForACLs();
             if (!RespCommandsInfo.TryGetSimpleRespCommandInfo(cmd, out var cmdInfo))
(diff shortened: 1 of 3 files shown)


libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -19,6 +29,24 @@ bool CanServeSlot(RespCommand cmd)
         {
             Debug.Assert(clusterSession != null);
 
+            // CustomRawStringCmd and CustomObjCmd fall outside IsDataCommand()'s range and
+            // would otherwise skip slot verification, so route them through a synthetic
+            // single-key spec (both dispatch via parseState[0] as the only key)
+            if (cmd is RespCommand.CustomRawStringCmd or RespCommand.CustomObjCmd)
```
vazois: maybe try putting this in a local method and call if cmd.IsDataCommand() fails.
This way you avoid the branch evaluation for the common case commands.

vazois (approved):

libs/server/Resp/RespServerSessionSlotVerify.cs
Ankith-Kandala: Moved the custom-command path into a separate `CanServeSlotForCustomCommand` that's only called from the `!cmd.IsDataCommand()` branch, so the hot path is unchanged for built-in data commands. Thanks!

libs/server/Resp/RespServerSessionSlotVerify.cs
```diff
@@ -10,6 +10,16 @@ namespace Garnet.server
...
     {
+        // Single-key spec for custom raw-string/object commands (key at parseState[0])
+        private static readonly SimpleRespKeySpec[] CustomCommandSingleKeySpec =
+        [
+            new SimpleRespKeySpec
+            {
+                BeginSearch = new SimpleRespKeySpecBeginSearch(index: 1),
+                FindKeys = new SimpleRespKeySpecFindKeys(keyStep: 1, lastKeyOrLimit: 0, isLimit: false),
+            }
+        ];
```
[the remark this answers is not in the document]
Ankith-Kandala: - Spec is correct - `index: 1` and `parseState[0]` are the same key (different counting: `index: 1` counts the command name as 0; `parseState[0]` is after the parser strips the command). `lastKeyOrLimit: 0` = single key. MOVED/ASK/CLUSTERDOWN tests prove it.
- Comment wording was inconsistent, latest push fixes it to "arg index 1".

vazois (approved):

PR #1841, opened 2026-06-01
Add What's New section to README

badrishc:
Add new apis and paper link


badrishc:
Folded into PR #1839 so the README's Vector Sets link can point at the polished /docs/commands/vector-sets page that lands in the same PR.

PR #1854, opened 2026-06-04
More IO optimization

badrishc:


libs/storage/Tsavorite/cs/src/core/Device/LocalMemoryDevice.cs
```diff
@@ -2,191 +2,352 @@
...
-
-            for (int i = 0; i < num_segments; i++)
+            // capacity <= 0 (including CAPACITY_UNSPECIFIED == -1): default to DefaultMaxSegments;
+            // segments are lazily allocated so the only cost of a generous max is the byte[][]/byte*[]
+            // pointer arrays.
+            if (capacity <= 0)
             {
-                orig_ram_segments[i] = GC.AllocateArray<byte>((int)sz_segment, true);
-                ram_segment_ptrs[i] = (byte*)Unsafe.AsPointer(ref orig_ram_segments[i][0]);
+                capacity = checked(sz_segment * DefaultMaxSegments);
```
TedHartMS: The base ctor has already been passed this and it's now in base.Capacity, so instead of duplicating that you can skip this assignment, move the comment into the param description, check the "multiple of sz_segment" here, and use base.Capacity in the WriteLine below

libs/storage/Tsavorite/cs/test/LowMemoryTests.cs
```diff
@@ -23,7 +23,7 @@ public class LowMemoryTests : TestBase
         public void Setup()
         {
             DeleteDirectory(MethodTestDir, wait: true);
-            log = new LocalMemoryDevice(1L << 28, 1L << 25, 1, latencyMs: 20, fileName: Path.Join(MethodTestDir, "test.log"));
+            log = new LocalMemoryDevice(1L << 28, 1L << 25, 1, latencyUs: 20_000, fileName: Path.Join(MethodTestDir, "test.log"));
```
TedHartMS: This can use DefaultLocalMemoryDeviceLatencyUs

test/standalone/Garnet.test/ScratchBufferAllocatorTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        // Rewinding back to the very first (default) state reclaims to zero but keeps the grown buffer.
+        [Test]
+        public void FirstAllocationFromDefaultRewindsToZeroAndRetainsBuffer()
+        {
+            var sba = new ScratchBufferAllocator();
+
+            var savedOffset = sba.ScratchBufferOffset; // 0
+
+            _ = sba.CreateArgSlice(64);
+            ClassicAssert.Greater(sba.ScratchBufferOffset, 0);
```
TedHartMS: More precise comparison would be better. Garnet.test.extensions also has a ScratchBufferAllocatorTests.cs; see CreateArgSliceAndRewindTest. Perhaps these two tests could be combined.

libs/storage/Tsavorite/cs/src/core/Device/LocalMemoryDevice.cs
badrishc: Done in 2e7077cd5c — dropped the duplicate assignment and now derive `maxSegments` (and the `Debug.WriteLine`) from `base.Capacity`. The defaulting note moved into the `capacity` param doc, and the multiple-of-`sz_segment` check now sits with the other ctor argument validations.

libs/storage/Tsavorite/cs/test/LowMemoryTests.cs
badrishc: Done in 2e7077cd5c — now uses `DefaultLocalMemoryDeviceLatencyUs`.

test/standalone/Garnet.test/ScratchBufferAllocatorTests.cs
badrishc: Done in 2e7077cd5c — tightened the assertion to an exact `AreEqual(64, sba.ScratchBufferOffset)`, and consolidated these tests into the existing `Garnet.test.extensions/ScratchBufferAllocatorTests.cs` alongside `CreateArgSliceAndRewindTest`, removing the duplicate fixture that had been added in `Garnet.test`.

TedHartMS (approved):

PR #1855, opened 2026-06-05
Cache DateTime.UtcNow in AAD authenticator hot path

tiagonapoli:
## Problem

`GarnetAadAuthenticator.IsAuthorized()` is on the per-command hot path — `IsAuthenticated` (which forwards to `IsAuthorized()`) is checked twice per command via [`AdminCommands.cs:32`](https://github.com/microsoft/garnet/blob/main/libs/server/Resp/AdminCommands.cs#L32) and [`AdminCommands.cs:131`](https://github.com/microsoft/garnet/blob/main/libs/server/Resp/AdminCommands.cs#L131).

In a cluster CPU profile under AAD authentication, `DateTime.get_UtcNow` shows up as the single hottest method at **~62.7% exclusive CPU**.

AAD IsAuthorized CPU profile

## Fix

Three pieces:

1. **`Garnet.common.CoarseTimeProvider`** — an instance class wrapping any `TimeProvider`. Exposes `UtcNow` (`DateTime`) and `UtcNowTicks` (`long`); both read a cached value via a pure memory load. The cache refresh is scheduled via `TimeProvider.CreateTimer` so a `FakeTimeProvider` in tests drives refreshes when its clock is advanced — production and tests therefore exercise **the same hot path**. `CoarseTimeProvider.System` is the process-wide singleton backed by `TimeProvider.System` (amortizes the single background Timer across all callers).

2. **`Garnet.common.CoarseDateTime`** — kept as a thin static facade over `CoarseTimeProvider.System` for back-compat / convenience.

3. **`GarnetAadAuthenticator`**:
   - Constructor takes an optional `CoarseTimeProvider` (defaults to `CoarseTimeProvider.System`).
   - `AadAuthenticationSettings` exposes a public `TimeProvider` parameter and owns the `CoarseTimeProvider` lifecycle (reuses `.System` singleton when no override → zero per-settings Timer cost; owns and disposes a per-settings instance otherwise).
   - `Authenticate()` precomputes the token validity window in ticks once (`_validFromTicks` / `_validToTicks`).
   - `IsAuthorized()` reads `_coarseTime.UtcNowTicks` and compares `long` against `long`, skipping `DateTime`s per-call Kind-bit masking.

## Microbenchmark

`benchmark/BDN.benchmark/Auth/IsAuthorizedBenchmark.cs` on Windows 11 24H2, i7-12700K, .NET 10, x64, Server GC, `DOTNET_TieredPGO=0`:

| Method                            | Mean       | Ratio | Notes |
|---------------------------------- |-----------:|------:|-------|
| `IsAuthorized_DateTimeUtcNow`     | 19.62 ns   | 1.000 | baseline |
| `IsAuthorized_TickCount64`        | 1.35 ns    | 0.069 | ~14× — `Environment.TickCount64` (per @Tornhoof suggestion) |
| `IsAuthorized_CoarseDateTime`     | 0.32 ns    | 0.016 | ~61× — `CoarseDateTime` (`DateTime` variant) |
| `IsAuthorized_CoarseDateTimeTicks`| **0.12 ns**| **0.006** | **~169× — shipped variant** |

`Environment.TickCount64` is a real win over `DateTime.UtcNow` (~14×) and dramatically simpler, but on Windows it still reads from `KUSER_SHARED_DATA` via a method call and is ~10× slower than the L1-cached `Volatile.Read` of a static `long`. On Linux the gap compresses (both `TickCount64` and `DateTime.UtcNow` go through vDSO `clock_gettime`) but `CoarseDateTime` still wins. For a per-command hot path checked twice per call, `CoarseDateTime.UtcNowTicks` (0.12 ns ≈ a single `mov`) is the right choice.

## Correctness considerations

- `CoarseTimeProvider` granularity is 100ms; AAD token validity windows are measured in **hours**, so the slop is operationally invisible (smaller than typical inter-host clock skew). A token that just expired may be accepted for up to ~100ms past its true deadline. Under sustained ThreadPool starvation the staleness can exceed 100ms; still harmless for hour-long token windows.
- `GarnetAadAuthenticator` is constructed once per `RespServerSession` (per client connection) and `IsAuthorized()` / `Authenticate()` are only ever called on that session''s RESP processing thread, so `_validFromTicks` / `_validToTicks` need no cross-thread synchronization.
- Failed auth in the catch block resets `_validFromTicks = long.MaxValue` / `_validToTicks = long.MinValue` so a previously-cached authorized state cannot leak past a failed re-authentication.
- The process-wide `CoarseTimeProvider.System` instance is `static readonly`, so it survives for process lifetime; ~10 callbacks/sec is negligible ThreadPool load. Cached value writes/reads use `Volatile.Write` / `Volatile.Read` of a `long` field — atomic on 64-bit, no JIT reordering.

## Testing

- `dotnet build` clean.
- `dotnet test test/cluster/Garnet.test.cluster --filter "FullyQualifiedName~Aad"`:
  - **Existing** `ClusterAadAuthTests` (×2) — full server `AUTH` flow via `Authenticate()` → `IsAuthorized()` and subsequent commands re-checking via `IsAuthenticated`. Pass on net8.0 and net10.0.
  - **New** `GarnetAadAuthenticatorTests` (×6) — unit-level tests using `FakeTimeProvider` (`Microsoft.Extensions.TimeProvider.Testing`) plumbed through `CoarseTimeProvider`. Covers default fallback to `.System`, valid window, expiry after `Advance()`, staying valid across multiple `Advance()` calls, malformed token, and failed re-auth clearing previously authorized state. Because the authenticator always reads `_coarseTime.UtcNowTicks` (no test-only branch), advancing the `FakeTimeProvider` fires the same `TimeProvider.CreateTimer` refresh callback the production cache uses — **tests exercise the production hot path**.
- `dotnet format` clean on the changed files.


Tornhoof:
You could also look at Environment.Tickcount64 for your coarse time. It's likely not as fast as your Timer solution, but should be a lot simpler in code.

tiagonapoli:
I updated the PR description with benchmark for `Environment.Tickcount64` - I think would be a viable solution, but on the case of AAD the tokens are valida for many minutes or hours, then we can afford to have a delayed timer with even less impact

libs/server/Auth/GarnetAadAuthenticator.cs
```diff
@@ -47,14 +61,16 @@ public GarnetAadAuthenticator(
             IReadOnlyCollection<string> issuers,
             IssuerSigningTokenProvider signingTokenProvider,
             bool validateUsername,
-            ILogger logger)
+            ILogger logger,
+            CoarseTimeProvider coarseTime = null)
```
kevin-montrose: `TimeProvider` so we can inject something arbitrary in tests - `CoarseTimeProvider.System` can be the default we `??=` below.

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// cadence, which can slip under GC pauses or ThreadPool starvation).
+        /// </summary>
+        public static readonly TimeSpan RefreshPeriod = TimeSpan.FromSeconds(1);
+
+        /// <summary>
+        /// Process-wide shared instance backed by <see cref="TimeProvider.System"/>.
+        /// Use this — or <see cref="Create"/> when the provider is parameterized — in
+        /// production code; it amortizes the single background Timer across all callers.
+        /// </summary>
+        public static new readonly CoarseTimeProvider System = new(TimeProvider.System);
```
kevin-montrose: I think wrapping a `TimeProvider` here is a bit strange.

Best to take `TimeProvider` in `GarnetAadAuthenticator` and have `CoarseTimeProvider` touch `DateTime.UtcNow` directly rather than go through an inner `TimeProvider`.

My rationale is that `CoarseTimeProvider` inherently caches, so swapping the inner `TimeProvider` for one in tests is brittle - we have to wait for the cache to invalidate before our "test" time provider is involved.  That means one of the big reasons to use `TimeProvider` (obviating the need to sleep in tests) doesn't work.

In both approaches we're hoping the JIT devirtualizes cleverly for best performance, so I don't think we're giving anything up.

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        private readonly TimeProvider timeProvider;
+        private readonly ITimer timer;
+        private long utcTicks;
+        private bool disposed;
```
kevin-montrose: This (L42 through L74) goes away if we don't need to parameterize by `TimeProvider` here, instead doing so in `AadAuthenticationSettings`.

Spin the timer up in a parameter-less constructor.

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                RefreshPeriod,
+                RefreshPeriod);
+        }
+
+        private void Refresh() => Volatile.Write(ref utcTicks, timeProvider.GetUtcNow().UtcTicks);
+
+        /// <summary>
+        /// Coarse, cached UTC time. May lag the underlying provider by ~<see cref="RefreshPeriod"/>.
+        /// </summary>
+        public override DateTimeOffset GetUtcNow() => new(Volatile.Read(ref utcTicks), TimeSpan.Zero);
```
kevin-montrose: nit: Just to remove some of the volatile magic, use `UtcNowTicks` instead of reading the field directly.  JIT should produce equivalent code.

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// Coarse, cached UTC time as a <see cref="DateTime"/>. Convenience wrapper
+        /// around <see cref="GetUtcNow"/> for callers that want a <see cref="DateTime"/>
+        /// directly.
+        /// </summary>
+        public DateTime UtcNow => new(Volatile.Read(ref utcTicks), DateTimeKind.Utc);
```
kevin-montrose: Same nit.

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// because it skips <see cref="DateTime"/>'s Kind-bit masking on comparison and
+        /// avoids the <see cref="DateTimeOffset"/> struct construction.
+        /// </summary>
+        public long UtcNowTicks => Volatile.Read(ref utcTicks);
+
+        // The remaining TimeProvider surface delegates to the wrapped provider — only
+        // GetUtcNow is coarse. We intentionally do NOT cache monotonic time or wrap
+        // timers; callers that need either get the accurate underlying behaviour.
+        /// <inheritdoc/>
+        public override long GetTimestamp() => timeProvider.GetTimestamp();
```
kevin-montrose: All these overrides that just delegate go away, we can use the default implementation (which is just `TimeProvider.System`).

libs/common/CoarseTimeProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <inheritdoc/>
+        public override long TimestampFrequency => timeProvider.TimestampFrequency;
+        /// <inheritdoc/>
+        public override TimeZoneInfo LocalTimeZone => timeProvider.LocalTimeZone;
+        /// <inheritdoc/>
+        public override ITimer CreateTimer(TimerCallback callback, object state, TimeSpan dueTime, TimeSpan period)
+            => timeProvider.CreateTimer(callback, state, dueTime, period);
+
+        /// <summary>
```
kevin-montrose: We can also remove `IDisposable` if we go that way.

test/cluster/Garnet.test.cluster/GarnetAadAuthenticatorTests.cs
```diff
@@ -0,0 +1,168 @@
...
+            var (auth, coarseTime, token, objId) = BuildAuthenticator(timeProvider, tokenExpiry: now.AddMinutes(5).UtcDateTime);
+            using var _ = coarseTime;
+
+            ClassicAssert.IsTrue(auth.Authenticate(Encoding.UTF8.GetBytes(token), Encoding.UTF8.GetBytes(objId)));
+            ClassicAssert.IsTrue(auth.IsAuthenticated, "token should be valid immediately after authentication");
+
+            // Advancing the FakeTimeProvider fires CoarseTimeProvider's internal Timer (registered via
+            // TimeProvider.CreateTimer), refreshing the cached UTC ticks — same code path as production.
+            // JwtSecurityToken truncates ValidTo to whole seconds, so step past the boundary.
+            timeProvider.Advance(TimeSpan.FromMinutes(5) + TimeSpan.FromSeconds(2));
```
kevin-montrose, line 84: This is (technically) unreliable - `FakeTimeProvider` doesn't guarantee that all triggered `ITimers` are complete when `Advance` returns (specifically, if another thread is servicing timer events the current thread will return immediately).

I suspect in this particular case it will work, but a slight refactor on our part could break it.  Passing `FakeTimeProvider` directly is more reliable, since `IsAuthenticated` wouldn't depend on a timer firing.

test/cluster/Garnet.test.cluster/GarnetAadAuthenticatorTests.cs
```diff
@@ -0,0 +1,168 @@
...
+                signingTokenProvider: new MockIssuerSigningTokenProvider(new List<SecurityKey> { tokenGenerator.SecurityKey }),
+                validateUsername: true,
+                logger: null,
+                coarseTime: coarseTime);
+
+            return (authenticator, coarseTime, token, objId);
+        }
+
+        [Test]
+        public void DefaultCoarseTime_FallsBackToSystem()
```
kevin-montrose: This test seems kinda pointless, other Aad tests will fail too if this breaks yeah?

kevin-montrose (changes requested):
I think this gets simpler if we don't wrap (but still implement) `TimeProvider`.

kevin-montrose (approved):

PR #1858, opened 2026-06-05
Fix large string handling in CLUSTER|NODES and CLUSTER|SHARDS

kevin-montrose:
Same issue experimentally identified in #1781 and fixed for INFO in #1776.

### DIFF
--- a/libs/cluster/Session/RespClusterBasicCommands.cs
+++ b/libs/cluster/Session/RespClusterBasicCommands.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Buffers;
 using System.Diagnostics;
 using System.Text;
 using Garnet.common;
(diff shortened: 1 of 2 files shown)


libs/cluster/Session/RespClusterBasicCommands.cs
```diff
@@ -523,5 +524,81 @@ private bool NetworkClusterPublish(out bool invalidParameters)
...
+            var buffer = ArrayPool<byte>.Shared.Rent(message.Length);
+            try
+            {
+                var len = Encoding.ASCII.GetBytes(message, buffer);
+                var remaining = buffer.AsSpan()[..len];
+
+                while (!remaining.IsEmpty)
+                {
+                    var space = Math.Min((int)(dend - dcurr), remaining.Length);
+                    _ = RespWriteUtils.TryWriteDirect(remaining[..space], ref dcurr, dend);
```
tiagonapoli, line 549: super nit: maybe a dbg assert that the return value was `true`?

tiagonapoli (approved):

PR #1874, opened 2026-06-13
[DRAFT] Inline filtering for vector sets

metajack:
Based on Haiyang's original two queue work, this rebased onto the quantization branch and modifies it for the new inline filtering with adaptive L.

### DIFF
--- a/libs/server/Resp/Vector/AttributeExtractor.cs
+++ b/libs/server/Resp/Vector/AttributeExtractor.cs
@@ -35,11 +35,11 @@ namespace Garnet.server
             if (s.IsEmpty || s[0] != (byte)'{') return 0;
             s = s[1..];
 
-            var found = 0;
-            var needed = selectorRanges.Length;
+                var found = 0;
+                var needed = selectorRanges.Length;
 
-            while (true)
-            {
+                while (true)
+                {
                 s = TrimWhiteSpace(s);
                 if (s.IsEmpty) return found;
                 if (s[0] == (byte)'}') return found;
@@ -50,16 +50,16 @@ namespace Garnet.server
                 if (!SkipString(ref s)) return found;
                 var keyContent = afterOpenQuote[..(afterOpenQuote.Length - s.Length - 1)];
 
-                var matchIndex = -1;
-                for (var i = 0; i < selectorRanges.Length; i++)
-                {
+                    var matchIndex = -1;
+                    for (var i = 0; i < selectorRanges.Length; i++)
+                    {
                     if (results[i].IsNone &&
                         keyContent.SequenceEqual(filterBytes.Slice(selectorRanges[i].Start, selectorRanges[i].Length)))
-                    {
-                        matchIndex = i;
-                        break;
+                            {
+                                matchIndex = i;
+                                break;
+                            }
                     }
-                }
 
                 s = TrimWhiteSpace(s);
                 if (s.IsEmpty || s[0] != (byte)':') return found;
@@ -68,22 +68,22 @@ namespace Garnet.server
                 s = TrimWhiteSpace(s);
                 if (s.IsEmpty) return found;
 
-                if (matchIndex >= 0)
-                {
+                    if (matchIndex >= 0)
+                    {
                     results[matchIndex] = ParseValueToken(json, ref s, ref program);
-                    found++;
-                    if (found == needed) return found;
-                }
-                else
-                {
+                        found++;
+                        if (found == needed) return found;
+                    }
+                    else
+                    {
                     if (!SkipValue(ref s)) return found;
-                }
+                    }
 
                 s = TrimWhiteSpace(s);
                 if (s.IsEmpty) return found;
                 if (s[0] == (byte)',') { s = s[1..]; continue; }
                 if (s[0] == (byte)'}') return found;
-                return found;
+                    return found;
             }
         }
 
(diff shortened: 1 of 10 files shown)


libs/server/Resp/Vector/VectorManager.Filter.cs
```diff
@@ -215,5 +217,138 @@ internal static int GetSelectorRanges(
...
+        }
+
+        /// <summary>
+        /// Per-candidate filter callback invoked from Rust during DiskANN inline filtered search.
+        /// Reads the candidate's external ID and attributes from Garnet storage, then evaluates
+        /// the compiled filter expression stored in <see cref="t_inlineFilterState"/>.
+        /// </summary>
+        /// <returns>1 if the candidate passes the filter, 0 otherwise.</returns>
+        [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
+        internal static unsafe byte InlineFilterCandidateCallbackImpl(ulong context, uint internalId)
```
kevin-montrose, line 282: style nit: this should be in .Callbacks.cs, and should be named `InlineFilterCallbackUnmanaged`

libs/server/Resp/Vector/VectorManager.Filter.cs
```diff
@@ -215,5 +217,138 @@ internal static int GetSelectorRanges(
...
+        /// <summary>
+        /// Per-query filter state maintained on the C# side.
+        /// Populated before calling into Rust; the callback reads it from thread-static storage.
+        /// All Span/pointer fields reference pinned scratch-buffer memory that remains
+        /// valid for the duration of the FFI call.
+        /// </summary>
+        internal unsafe struct InlineFilterState
+        {
+            /// <summary>Base Garnet context (no term bits).</summary>
+            public ulong Context;
```
kevin-montrose, line 249: We don't actually need Context here do we?  It's passed in every callback.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -637,72 +642,148 @@ ref SpanByteAndMemory filterBitmap
...
+
+                        var instrCount = ExprCompiler.TryCompile(filter, instrBuf, tuplePoolBuf, tokensBuf, opsStackBuf, out var tupleCount, out _);
+                        if (instrCount < 0)
+                        {
+                            // Compile failed — return zero results
+                            outputDistances.Length = 0;
+                            filterBitmap.Length = 0;
+                            outputIdFormat = VectorIdFormat.I32LengthPrefixed;
+                            errorMsg = Encoding.ASCII.GetBytes($"ERR compiling filter failed");
+                            return VectorManagerResult.OK;
```
kevin-montrose, line 681: Is OK correct here?  Probably want a different result type.

libs/server/Resp/Vector/AttributeExtractor.cs
```diff
@@ -35,11 +35,11 @@ public static int ExtractFields(
             if (s.IsEmpty || s[0] != (byte)'{') return 0;
```
kevin-montrose, line 35: High level style-nit that applies to a lot this, there are a number of places that put multiple statements on one line which is kinda ok (if not my favorite) for returns but is hard to read for things like `if (s[0] == (byte)'\\') { hasEscape = true; s = s[2..]; continue; }` and `var instrBuf = span.Slice(offset, MaxInstructions); offset += MaxInstructions;`.  Let's get that cleaned up so statements are on their own lines.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -722,34 +803,109 @@ ref SpanByteAndMemory filterBitmap
...
+                    {
+                        outputDistances.Length = 0;
+                        filterBitmap.Length = 0;
+                        outputIdFormat = VectorIdFormat.I32LengthPrefixed;
+                        return VectorManagerResult.OK;
+                    }
+
+                    var selectorCount = GetSelectorRanges(instrBuf[..instrCount], instrCount, filter, selectorBuf);
+
+                    fixed (byte* filterPtr = filter)
```
kevin-montrose, line 852: Stylistically this is kinda strange.

A cleaner approach would be to declare InlineFilterState as a `ref struct` and keep a pointer to it in the thread static.

Something like:
```csharp
internal ref struct InlineFilterState
{
	// Pointers into scratch buffer (pinned for FFI duration):
	public Span<ExprToken>InstrBuf;
	public Span<ExprToken> TuplePoolBuf;
	public Span<ExprToken> RuntimePoolBuf;
	public Span<ExprToken> ExtractedFields;
	public Span<ExprToken> StackBuf;
	public Span<(int Start, int Length)> SelectorRanges;

	public Span<byte> FilterBytesPtr;
}
```

Then 
```csharp
var callFilterState = new InlineFilterState { /* Use the actual spans here */ };
unsafe
{
   t_inlineFilterState = &callFilterState;
}
```

Reduces the amount of pointer shenanigans, doesn't have the `fixed` that does nothing (these buffers are already pinned), and makes it less likely we screw up the length bounds by not passing them around at all.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -722,34 +803,109 @@ ref SpanByteAndMemory filterBitmap
...
+                            outputDistances,
+                            out continuation
+                        );
+
+                    }
+                }
+                finally
+                {
+                    ActiveThreadSession.scratchBufferBuilder.RewindScratchBuffer(ref bufferSlice);
+                }
```
kevin-montrose, line 890: Zero out `t_inlineFilterState`, we don't want to leave dangling pointers laying around.

kevin-montrose (changes requested):
Some comments, but broadly looks OK.

Needs tests around failure to parse filter and some of the implied limits around filter sizes.

Also will want to target against `main`, though understandably that'll have to wait until quantization lands there (which is coming with #1882).

PR #1881, opened 2026-06-16
[RangeIndex] Handle crashes and recovery

tiagonapoli:
## What

Collects the RangeIndex (BfTree) crash/recovery repro tests in one place, **all marked `[Explicit]`** so they don't run in CI (they crash/hang by design). They become passing regression tests once the underlying issues are fixed. Base is `tiagonapoli/bftree-migration` because the cluster repros need the migration feature.

## Cluster-migration crash repros
New partial-class file `ClusterRangeIndexMigrateTests.Crashes.cs` (split out so the main migration suite stays crash-free and green). The main test class is made `partial` to host it.

- **`WhileModifyingAsync` / `WhileReadingAndWritingAsync` / `SlowTransmitWithConcurrentOpsAsync`** — drop-vs-insert **use-after-free**. The deferred `bftree_drop` (queued by `DisposeTreeUnderLock` and run via `BumpCurrentEpoch` *outside* the per-key exclusive lock) races a concurrent `InsertByPtr` that runs under the **shared** lock but *outside* store-epoch protection. Native panic (`mini_page_op.rs` / `circular_buffer/mod.rs` assertion) aborts the process.
- **`LargeTree`** — **concurrency-independent** crash: a single-threaded migration of a large multi-page tree crashes during snapshot/drop. Distinct, arguably more serious root cause.

## Standalone repros (migration-independent)
`RangeIndexConcurrentDeleteRepro.cs` proves these are not migration-specific — a whole-key `DEL` drives the same `DisposeTreeUnderLock` → deferred `bftree_drop` path as `MigrateOperation.DeleteRangeIndex`.

- **`ConcurrentSetAndWholeKeyDeleteRepro`** — the drop-vs-insert use-after-free.
- **`ConcurrentSetStarvesWholeKeyDeleteRepro`** — exclusive-lock starvation / livelock (sustained `RI.SET` shared-lock traffic starves the whole-key `DEL`'s exclusive-lock acquisition).

## Notes
- These 4 cluster tests were removed from `bftree-migration` (the feature branch) in its own commit; that suite now runs green (19/19). They live here for tracking.
- Root cause is structural: the per-key lock and the store epoch each only half-cover the native BfTree calls, and `DEL` / eviction / migration-delete all share that path.

🤖 Generated with [GitHub Copilot CLI](https://gh.io/copilot-cli)

### DIFF
--- a/test/cluster/Garnet.test.cluster.migrate.rangeindex/ClusterRangeIndexMigrateTests.Crashes.cs
+++ b/test/cluster/Garnet.test.cluster.migrate.rangeindex/ClusterRangeIndexMigrateTests.Crashes.cs
@@ -1,0 +1,501 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+#if DEBUG
+using Garnet.common;
+#endif
+using NUnit.Framework;
+using NUnit.Framework.Legacy;
+using StackExchange.Redis;
+
+namespace Garnet.test.cluster
+{
+    /// <summary>
+    /// Cluster-migration tests that currently crash the test host via a native bf-tree panic
+    /// (mini_page_op / circular_buffer assertion). They are split into this partial-class file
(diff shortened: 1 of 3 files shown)


badrishc (dismissed):

tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/18

PR #1886, opened 2026-06-17
`VectorSessionFunctions` Cleanup

kevin-montrose:
Removes two hacks from `VectorSessionFunctions`
 - All the alignment stuff, Tsavorite guarantees value alignment now
 - Mutating `VectorInput`, `VectorOutput` now receives updated values and `VectorInput` is `readonly`


libs/server/VectorOutput.cs
```diff
@@ -18,6 +18,11 @@ public struct VectorOutput
         /// </summary>
         public SpanByteAndMemory SpanByteAndMemory;
 
+        /// <summary>
+        /// If a call needs a larger <see cref="VectorInput.ReadDesiredSize"/> than was provided, it is stored here.
+        /// </summary>
+        public int? UpdatedReadDesiredSize { get; set; }
+
         public VectorOutput() => SpanByteAndMemory = new(null);
```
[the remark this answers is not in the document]
kevin-montrose: Wildly wrong.

libs/server/Resp/Vector/VectorManager.Callbacks.cs
```diff
@@ -308,15 +310,15 @@ private static unsafe bool ReadSizeUnknown(ulong context, bool forceAlignment, R
                         return false;
                     }
 
-                    if (input.ReadDesiredSize > asSpanByte.SpanByteAndMemory.Length)
+                    if (output.UpdatedReadDesiredSize != null && output.UpdatedReadDesiredSize.Value > output.SpanByteAndMemory.Length)
```
TedHartMS: .Value is retrieved twice (here and in the block). And maybe use GetValueOrDefault(-1) so the default is < SBAM.Length?

libs/server/InputHeader.cs
```diff
@@ -609,24 +609,22 @@ public unsafe int DeserializeFrom(byte* src)
...
+        public int WriteDesiredSize { get; init; }
 
-        public int Index { get; set; }
-        public nint CallbackContext { get; set; }
-        public nint Callback { get; set; }
-
-        public bool AlignmentExpected { get; set; }
+        public int Index { get; init; }
+        public nint CallbackContext { get; init; }
+        public nint Callback { get; init; }
```
TedHartMS: Would be helpful to add comments

libs/server/Storage/Functions/VectorStore/VectorSessionFunctions.cs
```diff
@@ -41,80 +39,83 @@ public readonly bool Reader<TSourceLogRecord>(in TSourceLogRecord srcLogRecord,
             Debug.Assert(srcLogRecord.HasNamespace, "Should never write a non-namespaced value with VectorSessionFunctions");
             Debug.Assert(srcLogRecord.NamespaceBytes.Length == 1, "Variable length namespaces not supported");
 
-            var value = AlignOrPin(in srcLogRecord, ref input, out var pin);
-            try
+            var value = srcLogRecord.ValueSpan;
```
TedHartMS, line 42: I'm not clear whether we're ensuring Key size is aligned to (now removed) ValueAlignmentBytes or if when we support ExtendedNamespace we'll be guaranteeing ExtendedNS+Key lengths are aligned to ValueAlignmentBytes; if not then we may still need to align the value start

kevin-montrose: That's a good point - today we'd get away with alignment because DiskANN provided keys are (generally) all 4-bytes.  But this would make that contractual.

I'll discuss and come back with a plan.

kevin-montrose: We'll update DiskANN to guarantee keys are always of alignment friendly sizes.

I'll set this back on draft until we've got a version to bump to that does that.

kevin-montrose:
Superseded by #1992.

PR #1891, opened 2026-06-19
Bump bf-tree to 0.5.4

tiagonapoli:
## What

Bumps the native [`bf-tree`](https://crates.io/crates/bf-tree) dependency from **0.5.0** to **0.5.4** (latest), fixing a crash where dropping a CprSnapshot-recovered, multi-level tree aborts the process in `BfTree`''s `Drop` with `assertion failed: next_level.is_null()` at `mini_page_op.rs:429`.

This is the root cause of the intermittent **RangeIndex cluster-migration crashes** — the migration destination index is a CprSnapshot-recovered tree, which aborts when dropped at shutdown. (Fixed upstream in bf-tree 0.5.3; 0.5.4 is the latest.)

## Required bf-tree API changes (0.5.0 → 0.5.4)

The snapshot/recovery API was redesigned; the FFI wrapper (`src/lib.rs`) and the managed call sites were adapted accordingly:

| 0.5.0 | 0.5.4 | adaptation |
|---|---|---|
| `Config::snapshot_file_path(path)` configured the snapshot destination | **removed** | no longer set on `Config`; the destination is passed at call time |
| `BfTree::cpr_snapshot()` | `BfTree::cpr_snapshot(path)` | the path is now a call-time argument |
| `BfTree::new_from_cpr_snapshot(recovery, use_snapshot, new_snapshot_path, buf_ptr, buf_size, wal)` | `new_from_cpr_snapshot(recovery, use_snapshot, buf_ptr, buf_size, wal)` | dropped the `new_snapshot_path` argument |
| `BfTree::are_all_threads_in_next_version()` | `are_all_threads_in_next_snapshot_version()` | renamed |

Data persistence is unchanged: records still accumulate in the **data file** (`Config::file_path`, e.g. `<hash>.data.bftree`); `cpr_snapshot(path)` writes a self-contained checkpoint to the given path via a non-blocking versioned sweep. Only the *timing* of naming the snapshot path moved (Config → call-time arg).

## Commits

1. **Bump bf-tree to 0.5.4** — `Cargo.toml`/`Cargo.lock`, adapt the FFI wrapper, rebuild win-x64/linux-x64 binaries. `bftree_cpr_snapshot` now takes the destination path as a parameter; managed callers pass it (`BfTreeService.CprSnapshot()` → `_snapshotFilePath`; `CprSnapshotByPtr(handle, snapshotPath)`). `bftree_create` / `bftree_new_from_cpr_snapshot` take a `use_snapshot` flag (the snapshot destination is supplied at `cpr_snapshot` call time, so no path is configured at create/recover).
2. **Snapshot directly to destination (drop scratch+copy)** — now that the destination is a call-time argument, the `OnFlush` / checkpoint / migration paths snapshot **straight to** their per-flush / per-checkpoint / migration file instead of snapshotting to a shared scratch file and `File.Copy`-ing it. The scratch+copy existed only because 0.5.0 fixed the snapshot path in `Config` and held a persistent file descriptor for it. The per-tree `SnapshotInProgress` claim is kept (bftree silently no-ops a racing `cpr_snapshot`), so concurrent consumers serialize and each writes its own destination; `TreeEntry.WaitForSnapshot` is removed. The scratch path remains only as the `use_snapshot` enabler at tree create/recover time.
3. **Docs** — update the RangeIndex dev doc to the direct-to-destination snapshot flow.

## Verification

- Full **RangeIndex** suite: **64/64 on Windows**; recover/checkpoint subset on **Linux**.
- **Cluster migrate RangeIndex** suite: **19/19 on Windows** (exercises the changed migration snapshot path).
- A multi-level recover-then-drop (1000 records) that aborted on 0.5.0 now drops cleanly with all records intact, exercised end-to-end through the FFI.
- Rebuilt the **win-x64** (`.dll`/`.pdb`) and **linux-x64** (`.so`) prebuilt binaries.

## Note on macOS binaries

The `osx-x64`/`osx-arm64` prebuilt binaries are **not** rebuilt here (no macOS cross-compile available). Dev machines with cargo rebuild from source for all platforms, and the release pipeline regenerates the prebuilts. A maintainer on macOS (or the pipeline) should refresh those two binaries.


tiagonapoli:
### How the snapshot path is threaded (updated)

bf-tree 0.5.3+ takes the CPR snapshot destination as an argument of `BfTree::cpr_snapshot(path)`. `bftree_cpr_snapshot` now takes the destination path as a parameter and the managed callers pass it:

- `BfTreeService.CprSnapshot()` passes its `_snapshotFilePath`.
- `CprSnapshotByPtr(nint handle, string snapshotPath)` — the RangeIndex `OnFlush` / checkpoint / migration sites now snapshot **directly to their final destination** (the per-flush / per-checkpoint / migration file), removing the previous scratch-file + `File.Copy` staging (which only existed because 0.5.0 fixed the snapshot path in `Config` and held a persistent FD for it).
- The per-tree `SnapshotInProgress` claim is kept (bftree silently no-ops a racing `cpr_snapshot`); `TreeEntry.WaitForSnapshot` was removed.
- `bftree_create` / `bftree_new_from_cpr_snapshot` keep their path arguments but use them only to toggle `use_snapshot` (the scratch path now serves only as that enabler; the scratch file is no longer written).

_(This supersedes the earlier `BfTreeHandle` and scratch+copy notes — both were dropped in favor of passing the destination path explicitly.)_

badrishc (approved):

kevin-montrose (approved):

PR #1897, opened 2026-06-22
[RangeIndex] Replace busy-spin snapshot claim with SemaphoreSlim to avoid CPU spikes

tiagonapoli:
## Problem

The per-tree snapshot claim in `TreeEntry` serialized concurrent CPR snapshots — `OnFlush` (flush), `SnapshotAllTreesForCheckpoint` (checkpoint), and `SnapshotForMigration` (cluster migration) — using a CAS + busy-spin:

```csharp
while (!TryClaimSnapshot())   // Interlocked.CompareExchange on an int
    Thread.Yield();
```

However, in case of pretty big BFTree, and since we may be writing multi-gigabyte files, snapshot may take more time, so any waiter will burn CPU for the entire wait. If migration happens during a checkpoint, or flushes happen with migration/checkpoint, we might burn CPU unnecessarily.

## Fix

Replace the CAS + spin with a per-tree `SemaphoreSlim` so waiters block instead of spinning:

```csharp
snapshotLock.Wait();
try { /* injected latency hook */ CprSnapshotByPtr(...); }
finally { snapshotLock.Release(); }
```

### DIFF
--- a/libs/common/Testing/ExceptionInjectionHelper.cs
+++ b/libs/common/Testing/ExceptionInjectionHelper.cs
@@ -150,8 +150,7 @@ namespace Garnet.common
         }
 
         /// <summary>
-        /// Blocks (not busy-spinning) until <paramref name="exceptionType"/> is disabled,
-        /// consuming negligible CPU.
+        /// Blocks until <paramref name="exceptionType"/> is disabled.
         /// </summary>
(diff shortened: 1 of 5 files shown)


tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/17

PR #1911, opened 2026-07-05
Add AOF replication for RangeIndex migration

tiagonapoli:
## What

Adds **AOF replication for migrated RangeIndex (BF-Tree) keys**, so a migrated
index reaches replicas and survives crash recovery with its full data — not just
an empty tree rebuilt from the stub.

## Why

When a node receives a migrated RI key, `PublishMigratedIndex` created the store
record via a `RICREATE` RMW that auto-logged **only the stub** to the AOF. On AOF
replay, `HandleRangeIndexCreateReplay` rebuilds an **empty** tree from the stub
config, so the migrated data never reached secondaries or survived recovery
(the `TODO(RI)` at `RangeIndexManager.Migration.cs`). A BF-Tree file can also
exceed a single AOF page, so it must be chunked.

## How

- **`RISTREAM`** — a new internal, AOF-only pseudo-command (never parsed from the
  network), intercepted in `AofProcessor.StoreRMW` alongside
  `RICREATE`/`RISET`/`RIDEL`.
- **Primary/publish** (`ReplicateRangeIndexStream`): the reassembled BfTree file is
  streamed into the AOF as chunked `RISTREAM` entries (reusing
  `RangeIndexChunkedSerializer`), each entry keyed by the RI key and no larger than
  `DefaultMigrationChunkSize`; the final chunk is flagged.
- **Replay** (`HandleRangeIndexStreamReplay`): a per-key
  `RangeIndexChunkedDeserializer` reassembles the file across interleaved AOF
  entries (all chunks for one key hash to a single virtual sublog, so they arrive
  in order on one replay task), then publishes via `PublishMigratedIndex` — which
  re-streams on a chained replica. Incomplete reassembly (crash mid-stream) is
  dropped by `CleanupIncompleteStreamReassembly`.
- **RICREATE suppression**: the local `RICREATE` RMW carries the
  `StreamedPublishLogArg` sentinel so `WriteLogRMW` skips its auto-log. The
  `RISTREAM` stream is the single AOF source of truth for a migrated key; keeping a
  competing empty-tree `RICREATE` entry would cause data loss on replay in either
  ordering (it would block the publish via the `KeyExists` gate, or clobber the
  reassembled tree).

Source-side deletion of the migrated key already replicates via the normal
`DELETE` -> AOF path (out of scope here).

## Tests

- `ClusterMigrateRangeIndexReplicatesToReplicaViaAof` — a migrated DISK-backed RI
  key replicates to the destination primary's replica via AOF; the replica serves
  the full data.
- `ClusterMigrateRangeIndexRecoversFromAof` — the destination primary reconstructs
  the migrated key from its own AOF on restart.
- Full `migrate.rangeindex` suite (21) green; `RangeIndexChunkedSerializerTests`
  (57) and `RangeIndexMigrationReceiveStateTests` (2) green; full solution builds
  with 0 warnings.

## Notes

- `MIGRATE REPLACE` for RI keys remains unsupported (pre-existing `TODO`).
- Preview feature (`--enable-range-index-preview`).

> Draft: opening for review; not yet linked to a tracking issue.


libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
```diff
@@ -177,5 +180,172 @@ internal void HandleRangeIndexDelReplay(StorageSession session, ReadOnlySpan<byt
...
+        private readonly ConcurrentDictionary<byte[], RangeIndexChunkedDeserializer> streamReassembly = new(ByteArrayComparer.Instance);
+
+        /// <summary>
+        /// Stream a migrated RangeIndex's serialized BfTree file into the AOF as a sequence of
+        /// chunked <see cref="RespCommand.RISTREAM"/> entries, reusing <see cref="RangeIndexChunkedSerializer"/>
+        /// for framing. No-op when <paramref name="appendOnlyFile"/> is null (AOF disabled / non-chained
+        /// replay) or <paramref name="storedProcMode"/> is set. Each chunk is one AOF entry no larger
+        /// than <paramref name="chunkSize"/>, which must fit within a single AOF page.
+        /// </summary>
+        internal unsafe void ReplicateRangeIndexStream(ReadOnlySpan<byte> key, ReadOnlySpan<byte> stub, string filePath,
```
tiagonapoli: This is only used in migratio path - why have a `bool storedProcMode`?

libs/host/Configuration/Options.cs
```diff
@@ -973,6 +970,9 @@ endpoint is IPEndPoint listenEp && clusterAnnounceEndpoint[0] is IPEndPoint anno
                 VectorSetReplayTaskCount = VectorSetReplayTaskCount,
                 VectorSetQuantizationTaskCount = VectorSetQuantizationTaskCount,
                 EnableRangeIndexPreview = EnableRangeIndexPreview,
+                RangeIndexAofStreamChunkSize = string.IsNullOrEmpty(RangeIndexAofStreamChunkSize)
```
tiagonapoli: Default should be min(DefaultMigrationChunkSize, AofChunkSize - Overhead)
Also, should check if the RangeIndexAofStreamChunkSize  + overhead will be enough to be in AOfChunkSize

libs/server/AOF/AofEntryType.cs
```diff
@@ -95,6 +95,14 @@ public enum AofEntryType : byte
         /// Unified store delete
         /// </summary>
         UnifiedStoreDelete = 0x73,
+
+        /// <summary>
+        /// Range Index migration stream chunk. Carries one chunk of a migrated RangeIndex's
+        /// serialized BfTree file. This is an internal AOF-only transport entry — it is never
+        /// produced from a client command and has no RESP command; it is enqueued during migration
+        /// publish and replayed via <c>RangeIndexManager.HandleRangeIndexStreamReplay</c>.
```
tiagonapoli: summarize

libs/host/GarnetServer.cs
```diff
@@ -386,6 +386,7 @@ private GarnetDatabase CreateDatabase(int dbId, GarnetServerOptions serverOption
                 rangeIndexManager = new RangeIndexManager(
                     riLogRoot: riLogRoot, cprDir: cprDir,
                     storeEpoch: storeEpoch,
+                    aofStreamChunkSize: serverOptions.RangeIndexAofStreamChunkSize,
```
tiagonapoli: aofStreamChunkSize is misleading - this should be specific to RangeIndex no?

libs/server/AOF/AofProcessor.cs
```diff
@@ -437,8 +438,10 @@ private bool ReplayOp<TPreprocessKey, TStringContext, TObjectContext, TUnifiedCo
         {
             // StoreRMW can queue VADDs onto different threads
             // but everything else needs to WAIT for those to complete
-            // otherwise we might loose consistency
-            if (header.opType != AofEntryType.StoreRMW)
+            // otherwise we might loose consistency.
+            // RangeIndexStreamChunk (formerly the StoreRMW-tagged RISTREAM pseudo-command) is
```
tiagonapoli: "formerly..." - why mention that if that's in the same PR that got overwritten

libs/server/AOF/AofProcessor.cs
```diff
@@ -575,6 +581,27 @@ static void StoreDelete<TStringContext>(PreparedParameters preparedParameters, T
...
+        /// (whose payload is the chunk and whose <c>arg1</c> holds the first/last framing flags) and
+        /// feeds it to <see cref="RangeIndexManager.HandleRangeIndexStreamReplay"/> for reassembly.
+        /// </summary>
+        static void RangeIndexStreamChunk(
+            PreparedParameters preparedParameters,
+            ref SessionParseState parseState,
+            RangeIndexManager rangeIndexManager,
+            RespServerSession activeServerSession)
+        {
+            if (rangeIndexManager == null)
```
tiagonapoli: Is it possible to check the EnableRangeIndexPreview flag instead?

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -108,6 +108,14 @@ public sealed partial class RangeIndexManager : IDisposable
         /// </summary>
         private readonly string cprDir;
 
+        /// <summary>
+        /// Maximum size (bytes) of each chunked <see cref="AofEntryType.RangeIndexStreamChunk"/> AOF entry emitted
+        /// by <see cref="ReplicateRangeIndexStream"/> when replicating a migrated index. Must be
+        /// smaller than the AOF page size so each chunk fits in one page. Defaults to
+        /// <see cref="DefaultMigrationChunkSize"/>.
+        /// </summary>
+        private readonly int aofStreamChunkSize;
```
tiagonapoli: Rename to something like rangeIndexAofStreamChunkSize - also summarize the comment

libs/server/Resp/RangeIndex/RangeIndexManager.Migration.cs
```diff
@@ -185,6 +190,9 @@ public unsafe PublishMigratedIndexResult PublishMigratedIndex(ReadOnlySpan<byte>
                     parseState.InitializeWithArgument(stubSlice);
 
                     var input = new StringInput(RespCommand.RICREATE, ref parseState);
+                    // Suppress the auto-AOF-log for this RICREATE: the range index stream enqueued above is
```
tiagonapoli, line 197: newline before comment

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
```diff
@@ -177,5 +180,204 @@ internal void HandleRangeIndexDelReplay(StorageSession session, ReadOnlySpan<byt
...
             }
         }
+
+        /// <summary>
+        /// Sentinel value placed in <c>StringInput.arg1</c> on the local <see cref="RespCommand.RICREATE"/>
+        /// RMW issued by <see cref="PublishMigratedIndex"/>. It tells
+        /// <c>MainSessionFunctions.WriteLogRMW</c> to skip auto-logging that RICREATE to the AOF.
+        /// For a migrated key the chunked <see cref="AofEntryType.RangeIndexStreamChunk"/> stream is the single AOF
+        /// source of truth; a competing RICREATE entry (which carries only the stub → an empty tree on
+        /// replay) would race the stream and cause data loss regardless of replay ordering.
```
tiagonapoli: remove - summarize just saying it's used in the migration code path

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
tiagonapoli: Good catch — removed it. `storedProcMode` was dead here: migration publish is only driven by the cluster migration-receive session (passes `false`) and by AOF replay (which always passes a null `appendOnlyFile`, already short-circuiting before the flag mattered). A migrated-index publish never runs inside a `CustomTransactionProcedure`, so there was no path that could pass `true`. Dropped the param from both `ReplicateRangeIndexStream` and `PublishMigratedIndex` in 7875d17c6f.

libs/host/Configuration/Options.cs
tiagonapoli: Done in 7875d17c6f. Each chunk becomes one AOF entry whose total size is `chunk payload + framing` (TsavoriteLog record header + AofHeader + length-prefixed key + StringInput framing), and `ValidateAllocatedLength` throws if that exceeds one AOF page — so the old `chunk < AofPageSize` check was insufficient.

- Added `RangeIndexManager.AofStreamChunkEntryOverhead` (a conservative upper bound on that framing, incl. an allowance for the RI key name).
- Default now resolves to `min(DefaultMigrationChunkSize, AofPage - overhead)` (empty config becomes a sentinel resolved in `GetAofSettings`, where the effective page size is known).
- Explicit values are validated against `chunk + overhead <= AofPage` (error otherwise), and I added a guard that the page is large enough to hold `MinChunkSize + overhead`.

One nuance: the key name is variable-length, so the overhead bakes in a fixed key allowance rather than the exact key. If you'd prefer this be exact, I can add a runtime clamp in `ReplicateRangeIndexStream` using the actual key length + `appendOnlyFile.Log.UnsafeGetLogPageSizeBits()` — let me know.

libs/server/AOF/AofEntryType.cs
tiagonapoli: Summarized in 74d98bfd3b — now two lines: what it is (a chunk of a migrated Range Index's BF-Tree file) and how it's replayed.

libs/server/AOF/AofProcessor.cs
tiagonapoli: Removed the `formerly RISTREAM` breadcrumb in 74d98bfd3b. You're right — RISTREAM only ever existed within this PR (introduced then replaced), so it never shipped and there's nothing for a future reader to reconcile.

libs/host/GarnetServer.cs
tiagonapoli: Renamed the ctor arg to `rangeIndexAofStreamChunkSize` (and the RangeIndexManager parameter/field to match) in 74d98bfd3b.

libs/server/Resp/RangeIndex/RangeIndexManager.cs
tiagonapoli: Renamed field `aofStreamChunkSize` -> `rangeIndexAofStreamChunkSize` and summarized the doc in 74d98bfd3b.

libs/server/Resp/RangeIndex/RangeIndexManager.Migration.cs
tiagonapoli: Added the blank line in 74d98bfd3b.

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
tiagonapoli: Summarized in 74d98bfd3b — now just states it suppresses auto-logging the migration-publish RICREATE (the stream is the single AOF source of truth) and that it's used only in the migration code path.

libs/server/AOF/AofProcessor.cs
tiagonapoli: `rangeIndexManager` here is `db.RangeIndexManager`, which is constructed only when `EnableRangeIndexPreview` is set (GarnetServer.cs) — so `rangeIndexManager == null` is exactly the `!EnableRangeIndexPreview` condition. I kept the null-check to stay consistent with the sibling RICREATE/RISET/RIDEL replay handlers right above (they use the same pattern), and because the flag isn't threaded into these static dispatch helpers. Happy to switch all four to an explicit `EnableRangeIndexPreview` check if you'd prefer that over the null-check convention — just let me know and I'll do it consistently.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -994,6 +976,45 @@ public void GetAofSettings(int dbId, out TsavoriteLogSettings[] tsavoriteLogSett
...
+                    logger?.LogError("{msg}", msg);
+                    throw new Exception(msg);
+                }
+
+                if (RangeIndexAofStreamChunkSize <= 0)
+                {
+                    // Auto default: largest of DefaultMigrationChunkSize and the page budget, capped so
+                    // chunk + overhead fits in one AOF page.
+                    RangeIndexAofStreamChunkSize = (int)Math.Min(RangeIndexManager.DefaultMigrationChunkSize, maxChunkBytes);
+                }
```
tiagonapoli: You should do this defaulting first and then check 'RangeIndexAofStreamChunkSize > maxChunkBytes'
What if the default is bigger than allowed

libs/server/AOF/AofEntryType.cs
```diff
@@ -95,6 +95,12 @@ public enum AofEntryType : byte
         /// Unified store delete
         /// </summary>
         UnifiedStoreDelete = 0x73,
+
+        /// <summary>
+        /// One chunk of a migrated Range Index's serialized BF-Tree file — an internal AOF-only
+        /// transport entry (no RESP command), replayed via <c>RangeIndexManager.HandleRangeIndexStreamReplay</c>.
```
tiagonapoli: (no RESP command) - remove

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -216,13 +222,17 @@ public TreeEntry(BfTreeService tree, long keyHash, string hashPrefix)
         /// <c>BfTree.Dispose</c> + file deletion past any in-flight reader observing the
         /// stub's <c>TreeHandle</c>. May be null in unit-test scenarios with no concurrent
         /// readers; in that case disposal is performed synchronously.</param>
+        /// <param name="rangeIndexAofStreamChunkSize">Maximum size (bytes) of each chunked
+        /// <see cref="AofEntryType.RangeIndexStreamChunk"/> AOF entry used to replicate a migrated index. Clamped
+        /// up to <see cref="RangeIndexChunkedSerializer.MinChunkSize"/>. Defaults to
+        /// <see cref="DefaultMigrationChunkSize"/>.</param>
```
tiagonapoli: Remove defaults to metnion and MinChunksize mention

libs/server/Resp/RangeIndex/RangeIndexManager.cs
```diff
@@ -232,6 +242,9 @@ public RangeIndexManager(string riLogRoot, string cprDir = null,
             this.riLogRoot = riLogRoot;
             this.cprDir = cprDir;
             this.storeEpoch = storeEpoch;
+            this.rangeIndexAofStreamChunkSize = rangeIndexAofStreamChunkSize < RangeIndexChunkedSerializer.MinChunkSize
+                ? RangeIndexChunkedSerializer.MinChunkSize
+                : rangeIndexAofStreamChunkSize;
```
tiagonapoli: Instead of clamp, throw invalid

libs/server/Servers/GarnetServerOptions.cs
tiagonapoli: Restructured in 095ace799a. The auto default is now resolved first — `RangeIndexAofStreamChunkSize = min(DefaultMigrationChunkSize, maxChunkBytes)` — and then an unconditional `if (RangeIndexAofStreamChunkSize > maxChunkBytes)` check runs (no longer an `else if`). Because the default is clamped to the page budget, it can never be bigger than allowed, so only an oversized *explicit* value trips the error. That covers the `what if the default is bigger than allowed` case directly.

libs/server/AOF/AofEntryType.cs
tiagonapoli: Removed in 095ace799a.

libs/server/Resp/RangeIndex/RangeIndexManager.cs
tiagonapoli: Done in 979efd29dd — the ctor now throws `ArgumentOutOfRangeException` when `rangeIndexAofStreamChunkSize < MinChunkSize` instead of clamping. (The auto default and explicit values are already >= MinChunkSize via the GarnetServerOptions validation, so this only fires on a genuinely invalid programmatic value — fail-fast at construction.)

libs/server/Resp/RangeIndex/RangeIndexManager.cs
tiagonapoli: Removed both the `Defaults to...` and `Clamped up to MinChunkSize` mentions from the param doc in 979efd29dd; documented the new `ArgumentOutOfRangeException` instead.

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
```diff
@@ -177,5 +180,201 @@ internal void HandleRangeIndexDelReplay(StorageSession session, ReadOnlySpan<byt
...
+                            break;
+                        written += n;
+                    }
+
+                    if (written == 0)
+                        continue;
+
+                    EnqueueRangeIndexStreamChunk(appendOnlyFile, version, sessionId, key, destBuffer.AsSpan(0, written), isFirst, serializer.IsComplete);
+                    isFirst = false;
+                }
```
tiagonapoli, line 282: This looks like lots of duplicated code from migration? Possible to improve?

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
```diff
@@ -177,5 +180,201 @@ internal void HandleRangeIndexDelReplay(StorageSession session, ReadOnlySpan<byt
...
+        {
+            var keyArr = key.ToArray();
+
+            // A new stream's first chunk supersedes any incomplete reassembly for the same key.
+            if (isFirst)
+                RemoveAndDisposeStreamReassembly(keyArr);
+
+            var deserializer = streamReassembly.GetOrAdd(keyArr, _ => new RangeIndexChunkedDeserializer(DeriveTempMigrationPath(), logger));
+
+            if (!deserializer.ProcessChunk(chunk) || deserializer.HasError)
```
tiagonapoli, line 342: Look at the activity types that exist for RI migration and create a new activity type for replication here
It will be in streamReassembly and once fully received then it will be complete and logged
Maybe in RemoveAndDisposeStreamReassembly - but will require a reason - either success or some failure

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
```diff
@@ -177,5 +180,201 @@ internal void HandleRangeIndexDelReplay(StorageSession session, ReadOnlySpan<byt
...
+        }
+
+        /// <summary>
+        /// Dispose and drop any in-progress AOF-stream reassembly state. Called at the end of an AOF
+        /// replay / recovery pass so a stream that was truncated mid-flight (e.g. a crash between
+        /// chunks) does not leak its temp file or deserializer.
+        /// </summary>
+        internal void CleanupIncompleteStreamReassembly()
+        {
+            foreach (var kvp in streamReassembly)
```
tiagonapoli: foreach in concurrent collection can be a problem here?

libs/host/GarnetServer.cs
```diff
@@ -386,6 +386,7 @@ private GarnetDatabase CreateDatabase(int dbId, GarnetServerOptions serverOption
                 rangeIndexManager = new RangeIndexManager(
                     riLogRoot: riLogRoot, cprDir: cprDir,
                     storeEpoch: storeEpoch,
+                    rangeIndexAofStreamChunkSize: serverOptions.RangeIndexAofStreamChunkSize ?? RangeIndexManager.DefaultMigrationChunkSize,
```
tiagonapoli: serverOptions should already be responsible for `??`

libs/host/Configuration/Options.cs
```diff
@@ -973,6 +970,9 @@ endpoint is IPEndPoint listenEp && clusterAnnounceEndpoint[0] is IPEndPoint anno
...
+                    ? null // Unset: resolved to min(DefaultMigrationChunkSize, AofPage - overhead) once the AOF page size is known (GarnetServerOptions.GetAofSettings).
+                    : (int?)ServerOptions.ParseSize(RangeIndexAofStreamChunkSize, out _),
```
tiagonapoli:
```suggestion
                RangeIndexAofStreamChunkSize = string.IsNullOrEmpty(RangeIndexAofStreamChunkSize) ? null : (int?)ServerOptions.ParseSize(RangeIndexAofStreamChunkSize, out _),
```

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
tiagonapoli: Deduped in 85d9f8f17c. ReplicateRangeIndexStream now reuses RangeIndexMigrationReader instead of re-implementing the chunk loop. I added a synchronous `ReadNextChunk` to the reader that shares the serializer-framing helpers (`ValidateDestination` / `SupplyFileDataOrThrow`) with `ReadNextChunkAsync` — the only difference is sync vs async file read. (I first tried reusing the async method directly, but VSTHRD002 correctly rejects sync-over-async, hence the sync twin.) `tempFilePath:null` so the reader disposes the FileStream but does not delete the snapshot, which PublishMigratedIndex moves into place after streaming.

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
tiagonapoli: Added in 85d9f8f17c. New `RangeIndexReplicationActivity` (mirrors the source-side `RangeIndexMigrationActivities`): each per-key reassembly in `streamReassembly` now carries an activity that records chunk count, bytes, and publish result. It's logged via `RemoveAndDisposeStreamReassembly(key, reason)` with the reason — `completed`, `publish failed`, `superseded by a new stream`, `chunk processing error`, `final chunk but stream incomplete`, or `incomplete at end of replay`.

libs/server/Resp/RangeIndex/RangeIndexManager.Replication.cs
tiagonapoli: Not a problem in practice, but cleaned it up in 85d9f8f17c. `CleanupIncompleteStreamReassembly` runs from `AofProcessor.Dispose` after all replay tasks finish, so there are no concurrent writers. It now iterates the `ConcurrentDictionary.Keys` snapshot (a copy) and removes via `RemoveAndDisposeStreamReassembly` — safe regardless of concurrency, since ConcurrentDictionary enumeration tolerates concurrent modification anyway.

libs/host/GarnetServer.cs
tiagonapoli: Moved it into the options in adda866611 — added `GarnetServerOptions.RangeIndexAofStreamChunkSizeOrDefault` (which does the `?? DefaultMigrationChunkSize`), and GarnetServer.cs now reads that. Note it can't just use `.Value`: RangeIndexManager is constructed (GarnetServer.cs:386) *before* `GetAofSettings` runs (:476), so the field can still be null at the call site — the default resolution has to live in the options accessor.

libs/host/Configuration/Options.cs
tiagonapoli: Applied in adda866611.

libs/server/AOF/AofProcessor.cs
```diff
@@ -575,6 +579,27 @@ static void StoreDelete<TStringContext>(PreparedParameters preparedParameters, T
...
+        /// Replay one <see cref="AofEntryType.RangeIndexStreamChunk"/> entry: a single chunk of a
+        /// migrated RangeIndex's serialized BfTree file. Deserializes the carried <see cref="StringInput"/>
+        /// (whose payload is the chunk and whose <c>arg1</c> holds the first/last framing flags) and
+        /// feeds it to <see cref="RangeIndexManager.HandleRangeIndexStreamReplay"/> for reassembly.
+        /// </summary>
+        static void RangeIndexStreamChunk(
+            PreparedParameters preparedParameters,
+            ref SessionParseState parseState,
+            RangeIndexManager rangeIndexManager,
+            RespServerSession activeServerSession)
```
tiagonapoli:
```suggestion
        static void HandleRangeIndexStreamChunk(PreparedParameters preparedParameters, ref SessionParseState parseState, RangeIndexManager rangeIndexManager, RespServerSession activeServerSession)
```

libs/server/AOF/AofProcessor.cs
tiagonapoli: Switch to EnableRangeIndexPreview

libs/server/AOF/AofProcessor.cs
tiagonapoli: Applied in f6b6ab31b9 — renamed to `HandleRangeIndexStreamChunk` (matches the sibling replay handlers and avoids colliding with `AofEntryType.RangeIndexStreamChunk`), signature on one line.

libs/server/AOF/AofProcessor.cs
```diff
@@ -575,6 +579,23 @@ static void StoreDelete<TStringContext>(PreparedParameters preparedParameters, T
...
 
+        /// <summary>
+        /// Replay one <see cref="AofEntryType.RangeIndexStreamChunk"/> entry: a single chunk of a
+        /// migrated RangeIndex's serialized BfTree file. Deserializes the carried <see cref="StringInput"/>
+        /// (whose payload is the chunk and whose <c>arg1</c> holds the first/last framing flags) and
+        /// feeds it to <see cref="RangeIndexManager.HandleRangeIndexStreamReplay"/> for reassembly.
+        /// </summary>
+        static void HandleRangeIndexStreamChunk(PreparedParameters preparedParameters, ref SessionParseState parseState, RangeIndexManager rangeIndexManager, RespServerSession activeServerSession)
+        {
+            if (rangeIndexManager == null)
```
tiagonapoli: Use EnableRangeIndexPreview flag instead

libs/server/AOF/AofProcessor.cs
```diff
@@ -575,6 +579,23 @@ static void StoreDelete<TStringContext>(PreparedParameters preparedParameters, T
             where TStringContext : ITsavoriteContext<FixedSpanByteKey, StringInput, StringOutput, long, MainSessionFunctions, StoreFunctions, StoreAllocator>
             => stringContext.Delete((FixedSpanByteKey)preparedParameters.Key);
 
+        /// <summary>
+        /// Replay one <see cref="AofEntryType.RangeIndexStreamChunk"/> entry: a single chunk of a
+        /// migrated RangeIndex's serialized BfTree file. Deserializes the carried <see cref="StringInput"/>
+        /// (whose payload is the chunk and whose <c>arg1</c> holds the first/last framing flags) and
+        /// feeds it to <see cref="RangeIndexManager.HandleRangeIndexStreamReplay"/> for reassembly.
+        /// </summary>
```
tiagonapoli: Remove this comment

libs/server/AOF/AofProcessor.cs
```diff
@@ -575,6 +579,23 @@ static void StoreDelete<TStringContext>(PreparedParameters preparedParameters, T
...
+        static void HandleRangeIndexStreamChunk(PreparedParameters preparedParameters, ref SessionParseState parseState, RangeIndexManager rangeIndexManager, RespServerSession activeServerSession)
+        {
+            if (rangeIndexManager == null)
+                ExceptionUtils.ThrowException(new GarnetException("RangeIndexPreview disabled; Replay failed"));
+
+            var stringInput = new StringInput { parseState = parseState };
+            _ = stringInput.DeserializeFrom(preparedParameters.PayloadPtr);
```
tiagonapoli, line 588: Is PayloadPtr underlying memory pinned?

libs/server/AOF/AofProcessor.cs
tiagonapoli: Done in e8d3f0735f — now guards on `activeServerSession.StoreWrapper.serverOptions.EnableRangeIndexPreview` instead of the `rangeIndexManager == null` check.

libs/server/AOF/AofProcessor.cs
tiagonapoli: Removed in e8d3f0735f.

libs/server/AOF/AofProcessor.cs
tiagonapoli: Yes. `PayloadPtr` points into the AOF scan buffer (`entryPtr` passed to `ProcessAofRecordInternal` from `ReplicaReplayTask` / `RecoverLogDriver` / `AofRecover`) — the same buffer every other replay handler deserializes from. It's valid for the duration of the replay call and the data is copied out synchronously within it: `DeserializeFrom` reads it, then `ProcessStreamChunk` does `key.ToArray()` and `deserializer.ProcessChunk(chunk)` (writing the chunk to the reassembly temp file) before the scan iterator advances or reuses the buffer.

test/standalone/Garnet.test/RangeIndexStreamReplayTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            // Attempt 2 (retry) replays the same stream. Its first chunk (isFirst) must reset the stale
+            // partial so the retry's chunks reassemble cleanly. Feed every chunk except the trailer so
+            // we stay incomplete (avoids the publish path). If the reset did NOT happen, the retry's
+            // header bytes would be fed into the stale mid-file deserializer and corrupt it (checksum
+            // failure -> the reassembly would be dropped, leaving count 0).
+            for (var i = 0; i < chunks.Count - 1; i++)
+                mgr.ProcessStreamChunk(session: null, key, chunks[i].Chunk, isFirst: i == 0, isLast: false);
+
+            ClassicAssert.AreEqual(1, mgr.PendingStreamReassemblyCount,
+                "retry stream should reassemble cleanly after the first chunk reset the stale partial");
```
tiagonapoli: no multiline

test/standalone/Garnet.test/RangeIndexStreamReplayTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            var key = Encoding.UTF8.GetBytes("k1");
+
+            // A first chunk that begins with a non-positive key length is rejected by the deserializer,
+            // and the failed reassembly must be removed (not left pending).
+            var malformed = new byte[16]; // leading 4 bytes = 0 => invalid key length
+            mgr.ProcessStreamChunk(session: null, key, malformed, isFirst: true, isLast: false);
+            ClassicAssert.AreEqual(0, mgr.PendingStreamReassemblyCount);
+
+            // count==0 alone is ambiguous; the activity proves the reassembly was created and then dropped
+            // for the right reason (chunk-processing failure) rather than silently ignored.
```
tiagonapoli: rmeove comment

test/standalone/Garnet.test/RangeIndexStreamReplayTests.cs
tiagonapoli: Collapsed the assert onto a single line in 17e738f88e.

test/standalone/Garnet.test/RangeIndexStreamReplayTests.cs
tiagonapoli: Removed the comment in 17e738f88e.

libs/server/Resp/RangeIndex/RangeIndexManager.Migration.cs
```diff
@@ -152,7 +151,16 @@ public unsafe PublishMigratedIndexResult PublishMigratedIndex(ReadOnlySpan<byte>
             {
                 var bftreeDataPath = LogDataPathFor(keyBytes);
 
-                // TODO(RI): Before publishing the migrated index, insert the chunked RI file into AOF to replicate to secondaries.
+                // TODO(RI): Acquire Checkpoint lock to prevent a concurrent checkpoint from deleting the
```
kevin-montrose, line 154: What's the plan with these TODOs?

tiagonapoli: I'll handle them in a follow up PR that I'm already tracking

libs/host/GarnetServer.cs
```diff
@@ -397,6 +397,12 @@ private GarnetDatabase CreateDatabase(int dbId, GarnetServerOptions serverOption
             var store = CreateStore(dbId, clusterFactory, customCommandManager, storeEpoch, rangeIndexManager, vectorManager, out var stateMachineDriver, out var sizeTracker, out var kvSettings);
             var aof = CreateAOF(dbId);
 
+            if (serverOptions.EnableRangeIndexPreview)
+            {
+                // Fail fast at startup if the AOF page can't hold a migrated range index stream chunk (no-op when AOF is disabled).
+                rangeIndexManager.ValidateAofPageCompatibility(aof);
```
vazois: nit; is it possible to perform this validation earlier to avoid store and AOF allocation? Maybe in Options.cs?!

vazois (approved):

vazois (approved):

kevin-montrose (approved):

PR #1913, opened 2026-07-06
Implement `RENAME(NX)` Support For Vector Sets

kevin-montrose:
Implements `RENAME(NX)` support for Vector Sets.

As part of this, adds `VectorSetFlags` to indexes.  This value is used to suppress certain cleanup operations that happen during a rename.

Rename implementation is modified to explicitly delete destination keys when required (ie. to trigger Vector Set cleanup) and hash slots in Vector Set indexes are updated if needed (note that that this can only happen in non-cluster modes).


libs/server/Resp/Vector/VectorManager.ContextMetadata.cs
```diff
@@ -717,5 +734,41 @@ public static void StoreContextInNamespace(ulong context, ref Span<byte> namespa
...
+        /// Given an old and new key and the value for the new key, update the associated hash slot in context metadata.
+        /// 
+        /// Assumes that old and new key are locked to prevent concurrent modification.
+        /// </summary>
+        public void UpdateHashSlot(ReadOnlySpan<byte> oldKey, ReadOnlySpan<byte> newKey, ReadOnlySpan<byte> value, ref VectorBasicContext vectorContext)
+        {
+            // Not a Vector Set index
+            if (value.Length != Index.Size)
+            {
+                return;
```
tiagonapoli, line 750: nit: do you think would be worth adding a log? Or maybe in the caller? (or even throw here instead and let the caller do the validations)

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -291,7 +291,7 @@ public readonly bool InitialUpdater(ref LogRecord logRecord, in RecordSizeInfo s
                     break;
                 case RespCommand.VADD:
                     {
-                        if (input.arg1 is VectorManager.VADDAppendLogArg or VectorManager.MigrateElementKeyLogArg or VectorManager.MigrateIndexKeyLogArg)
+                        if (input.arg1 is VectorManager.VADDAppendLogArg or VectorManager.MigrateElementKeyLogArg or VectorManager.MigrateIndexKeyLogArg or VectorManager.VADDSetFlagsArg)
```
tiagonapoli: nit: InitialUpdater + VADD + VADDSetFlagsArg should never happen right? Otherwise we're trying to set flags on something that doesn't exist and some invariant was already broken? Should we throw/log instead or there's some valid scenario?

test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs
```diff
@@ -3100,6 +3100,191 @@ public void VADDLowMemoryStorageTierForcesDiskSpill()
             ClassicAssert.IsNotEmpty(sim);
         }
 
+        [Test]
```
tiagonapoli, line 3103: nit: do you think would worth adding a chaos test where we do renames while there are VADDs in parallel?

tiagonapoli (approved):

test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs
kevin-montrose: I _think_ that's duplicative with general lock testing, since the exclusive locks in transactions would block the tsavorite read on the old key.

tiagonapoli (approved):

PR #1927, opened 2026-07-12
Fix `PUBLISH` failing when called from Lua scripts

lilpard-mk:
### Root Cause

`SessionScriptCache.cs` creates an internal `RespServerSession` for Lua script execution with `subscribeBroker` hardcoded to `null`. This causes `redis.call('PUBLISH', ...)` inside Lua scripts to fail with `ERR PUBLISH is disabled`, even when PubSub is enabled (the default).

The bug affects all Redisson Java users because Redisson's `RLock.unlock()` relies on `redis.call('PUBLISH', ...)` inside its Lua unlock script to notify waiters. See Discussion #1628.

### Description of Change

- Pass `storeWrapper.subscribeBroker` instead of `null` to the internal Lua processor in `SessionScriptCache.cs`, aligning its behavior with the network session created in `GarnetProvider.cs`.

### Key Technical Details

- The 6th parameter `enableScripts: false` is intentionally kept `false` to prevent nested EVAL recursion.
- `SubscribeBroker` is fully thread-safe (`ConcurrentDictionary`, `Interlocked`).
- `SUBSCRIBE` / `PSUBSCRIBE` remain blocked in Lua by the existing NoScript bitmap (Redis protocol compliant).
- When user passes `--no-pubsub`, `storeWrapper.subscribeBroker` is `null`, so Lua `PUBLISH` still correctly reports disabled — graceful degradation preserved.

### Tests Added

Positive cases (10):
- `LuaPublishSucceeds_WhenCalledViaRedisCall`
- `LuaPublishReturnsSubscriberCount`
- `LuaPublishToNoSubscribersReturnsZero`
- `LuaPublishMultipleSubscribers_AllReceive`
- `LuaPublish_PSubscribeReceives`
- `LuaPublishMixedWithSetGet`
- `LuaPublishSameChannelMultipleTimes`
- `LuaPublishUnicodePayload`
- `LuaPublishEmptyPayload`
- `LuaPublishAfterNonPubSubCommands`

NoScript enforcement (regression guard, 3):
- `LuaSubscribe_StillNoScriptBlocked`
- `LuaPSubscribe_StillNoScriptBlocked`
- `LuaUnsubscribe_StillNoScriptBlocked`

Concurrency/state (4):
- `ConcurrentLuaPublish_NoRaceCondition`
- `MultipleScriptsSameSession_AllCanPublish`
- `LuaPublishPreservesOtherSubscriberState`
- `ScriptCachePublish_WorksWithScriptFlush`

### Reproduction (before fix)

```
redis-cli PUBLISH test hi
# (integer) 0  

redis-cli EVAL "return redis.call('PUBLISH','test','hi')" 0
# (error) ERR PUBLISH is disabled, enable it with --pubsub option.  
```

### Checklist

- [x] Code follows `.editorconfig` (`dotnet format --verify-no-changes` passes)
- [x] Builds on `ubuntu-latest` and `windows-latest` (Debug + Release), 0 warnings
- [x] `Garnet.test.scripting` tests pass locally
- [x] No `Version.props` changes (handled by release automation)

### Related

Discussion #1628


lilpard-mk:

lilpard-mk:
Thanks @Copilot for the review! All 14 comments addressed in the latest push:

- Refined the SessionScriptCache.cs comment to clarify that only publish-side Pub/Sub is enabled (SUBSCRIBE/PSUBSCRIBE remain NoScript-blocked).
- Switched all ManualResetEvent instances to 'using var' or explicit Dispose() in finally blocks to prevent OS handle leaks.

Local validation: dotnet format --verify-no-changes passes, Debug+Release builds are 0-warning, all 102 tests pass.

test/standalone/Garnet.test.scripting/LuaScriptTests.cs
```diff
@@ -3384,5 +3384,706 @@ public void StressTimeouts()
...
+
+            ClassicAssert.AreEqual(
+                expected,
+                server.SubscriptionSubscriberCount(channel),
+                $"Did not observe {expected} subscriber(s) on {(string)channel}.");
+        }
+
+        // A single subscriber receives a message published from inside a Lua script via redis.call('PUBLISH', ...).
+        [Test]
+        public void LuaPublishSucceeds_WhenCalledViaRedisCall()
```
kevin-montrose, line 3457: Change itself seems fine, but these tests are kinda excessive - mostly testing channel functionality that is covered by other tests.

I think a single test would be sufficient, something like:
```csharp
[Test]
[CancelAfter(10_000)]
public async Task PublishAsync(CancellationToken cancellation)
{
    using var subRedis = ConnectionMultiplexer.Connect("foo");
    using var redis = ConnectionMultiplexer.Connect("foo");
    var sub = subRedis.GetSubscriber();

    var channel = RedisChannel.Literal("lua_pub_basic");
    const string Payload = "hello-from-lua";
    using var received = new ManualResetEvent(false);
    
    var tcs = new TaskCompletionSource<RedisValue>();

    await sub.SubscribeAsync(channel, (_, msg) => tcs.SetResult(msg)).ConfigureAwait(false);

    var count = 
        (long)await db.ScriptEvaluateAsync(
            "return redis.call('PUBLISH', KEYS[1], ARGV[1])",
            [new RedisKey("lua_pub_basic")],
            [new RedisValue(Payload)]
        ).ConfigureAwait(false);

    ClassicAssert.AreEqual(1L, count, "PUBLISH should report exactly one subscriber.");

    var lastMessage = await tcs.Task.WaitAsync(cancellation).ConfigureAwait(false);

    ClassicAssert.IsTrue(received.WaitOne(TimeSpan.FromMilliseconds(500)), "Subscriber did not receive the Lua-published message.");
    ClassicAssert.AreEqual(Payload, (string)lastMessage);

    await sub.UnsubscribeAsync(channel).ConfigureAwait(false);
}
```

Should be sufficient.

kevin-montrose (changes requested):
Simple change, but excessive (and duplicative) tests IMO.  Let's trim it down to not bloat CI time and test files as much.

kevin-montrose (approved):

kevin-montrose (approved):

PR #1934, opened 2026-07-15
Harden VectorSet stub against resurrection under concurrent VADD/VREM/DEL

tiagonapoli:
## Problem

On the primary, a single VectorSet key could spin forever in the recreate loop and peg all cores. The log showed `RECREATE ctx=0 dims=0 indexPtr=0` with **no preceding `INITIAL-CREATE`** — a fully-zeroed 56-byte index record that reads back as `OK`, so `NeedsRecreate` (which only checks `indexPtr == 0`) returned true forever and `RecreateIndex(dims: 0)` could never produce a valid pointer.

## Root cause

A `VADD` is phrased as a read once the index exists, so after a successful add `VectorManager.ReplicateVectorSetAdd` issues a **synthetic `VADDAppendLogArg` RMW** on the index key purely to get the write into the AOF (VREM does the same via `VREMAppendLogArg`). These synthetic args, plus `RecreateIndexArg` and `VADDSetFlagsArg`, only make sense **against an already-existing record**.

`NeedInitialUpdate` had **no `VADD`/`VREM` case**, so those args fell through to `default -> return true`. `VADD`/`VREM` hold only a **shared** vector lock across the add/remove + replicate, and a raw main-store `DEL`/`UNLINK` takes **no** vector lock, so a concurrent delete can tombstone the key mid-flight. The synthetic RMW then lands on an absent (or re-typed) key, `InitialUpdater`'s no-op branch leaves the pre-sized record zeroed or with garbage, and the tombstoned key is resurrected by `VADD/VREM`.

## Fix

The PR closes the resurrection hole at both RMW entry points, plus adjacent correctness fixes:

**1. `NeedInitialUpdate`**
Explicit `VADD`/`VREM` cases. `VADD` may establish a record **only** for an allowlist of genuine-create args; everything else rejects. `VREM` **always** rejects (it never creates an index). Creation is now **opt-in** via an explicit `CreateIndexArg` sentinel instead of the old implicit `arg1 == 0`, so any synthetic/stray write against a concurrently-deleted key is refused rather than fabricating a zeroed stub.

**2. `NeedCopyUpdate` — cancel for a re-typed key.**
If a synthetic `VADD`/`VREM` replication RMW reaches a record that is **no longer an index** (e.g. the key was deleted and re-created as a String via a raced `SET`), it now sets `RMWAction.CancelOperation` before any tail record is allocated — nothing is copied to the log/AOF and replicas never replay the add/remove against a non-index record.

**3. VREM/VADD `CopyUpdater` value-preservation fix.**
The VREM/VADD CopyUpdater was not copying the stub in the previous value for all branches; it now copies the old value forward to avoid corrution of the index.

**4. arg1 sentinel taxonomy + cleanup.**
Every `arg1` sentinel is documented with its two dispositions — **AOF** (logged/replicated vs node-local) and **InitialUpdater** (may create vs must reject).


libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -47,6 +47,25 @@ public readonly bool NeedInitialUpdate<TKey>(TKey key, ref StringInput input, re
...
+                case RespCommand.VADD:
+                case RespCommand.VREM:
+                    // The synthetic args below (replication append-log, migration, recreate,
+                    // set-flags) only make sense as a CopyUpdater on an already-existing index
+                    // record. If the key was concurrently deleted (e.g. a raced UNLINK, which
+                    // takes no vector lock), creating a record here via InitialUpdater leaves a
+                    // zeroed 56-byte index that NeedsRecreate flags forever, livelocking the
+                    // recreate loop. Refuse to create — the key must already exist, as with
+                    // RIPROMOTE/RIRESTORE. A genuine VADD create carries arg1 == 0 and still creates.
+                    if (input.arg1 is VectorManager.VADDAppendLogArg
```
tiagonapoli: @kevin-montrose Safe to assume any NeedInitialUpdate with these flags are invalid?
Now that I'm thinking about it, we should actually fail right? Otherwise we might return OK for a VADD that actually didn't complete because of a UNLINK that happened concurrently

kevin-montrose: The VADD failed to replicate because of a racing delete, so when the delete replicates the vector set disappears.  Dropping that write is fine IMO, there's no guarantee (due to replication delay) that a client would observe it.

test/standalone/Garnet.test.vectorset/VectorRecreateLivelockTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            var data = new byte[75];
+            new Random(2026_07_15).NextBytes(data);
+            _ = db.Execute("VADD", [Key, "XB8", (RedisValue)data, (RedisValue)elem, "XPREQ8"]);
+            ClassicAssert.IsTrue(db.KeyExists(Key), "vector set should exist after VADD");
+
+            // 2. UNLINK the key. This is a raw main-store tombstone that does NOT take the
+            //    vector lock, so in production it can land while a concurrent VADD still holds
+            //    only a shared vector lock and is about to replicate its add.
+            _ = db.KeyDelete(Key);
+            ClassicAssert.IsFalse(db.KeyExists(Key), "key should be gone after UNLINK");
```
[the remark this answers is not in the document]
tiagonapoli: Addressed in 326586e686 — the test issues a DEL via KeyDelete, which takes the same lock-free raw main-store delete path as UNLINK (both bypass the vector lock). Reworded the comments to describe it accurately rather than claiming UNLINK.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -47,6 +47,25 @@ public readonly bool NeedInitialUpdate<TKey>(TKey key, ref StringInput input, re
...
                     return false; // Key must already exist
+                case RespCommand.VADD:
+                case RespCommand.VREM:
+                    // The synthetic args below (replication append-log, migration, recreate,
+                    // set-flags) only make sense as a CopyUpdater on an already-existing index
+                    // record. If the key was concurrently deleted (e.g. a raced UNLINK, which
+                    // takes no vector lock), creating a record here via InitialUpdater leaves a
+                    // zeroed 56-byte index that NeedsRecreate flags forever, livelocking the
+                    // recreate loop. Refuse to create — the key must already exist, as with
+                    // RIPROMOTE/RIRESTORE. A genuine VADD create carries arg1 == 0 and still creates.
```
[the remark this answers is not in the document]
tiagonapoli: Addressed in 326586e686 — dropped the hard-coded 56-byte literals from the comments and describe the index record generically (the size is defined by a constant).

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -47,6 +47,16 @@ public readonly bool NeedInitialUpdate<TKey>(TKey key, ref StringInput input, re
                     return false; // Key must already exist; don't create new
                 case RespCommand.RIRESTORE:
                     return false; // Key must already exist
+                case RespCommand.VADD:
+                case RespCommand.VREM:
+                    // These synthetic args are only valid as a CopyUpdater on an existing index
+                    // record; encountering them here means the key was concurrently UNLINK'd, so we
+                    // cannot complete the operation. A genuine VADD create carries arg1 == 0.
```
tiagonapoli: These are valid as CopyUpdater or InPlaceUpdater, update comment

test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs
```diff
@@ -3637,6 +3637,168 @@ public async Task HideInternalRecordsAsync()
...
+            {
+                // Connection 1: run the operation; it modifies the index, then parks before its append-log RMW.
+                var opTask = operation == "VADD"
+                    ? Task.Run(() => dbOp.Execute("VADD", [key, "VALUES", "3", "1", "2", "3", (RedisValue)elem]))
+                    : Task.Run(() => dbOp.Execute("VREM", [key, (RedisValue)elem]));
+
+                await ExceptionInjectionHelper.WaitOnClearAsync(Pause).WaitAsync(TimeSpan.FromSeconds(30));
+
+                // Connection 2: DEL the parked key, tombstoning it while the operation holds a shared vector lock.
+                _ = dbDel.KeyDelete(key);
```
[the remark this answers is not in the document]
tiagonapoli: Addressed in 64e6a69e58 — the KeyDelete result is now asserted (`ClassicAssert.IsTrue(dbDel.KeyDelete(key), ...)`) so the test fails if the tombstone did not take.

test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs
```diff
@@ -3637,6 +3637,168 @@ public async Task HideInternalRecordsAsync()
...
+            try
+            {
+                // Connection 1: VADD creates the index, then parks before its append-log RMW.
+                var opTask = Task.Run(() => dbOp.Execute("VADD", [key, "VALUES", "3", "1", "2", "3", (RedisValue)elem]));
+
+                await ExceptionInjectionHelper.WaitOnClearAsync(Pause).WaitAsync(TimeSpan.FromSeconds(30));
+
+                // Connection 2: tombstone the vector-set key, then re-add it as a plain String.
+                _ = dbRacer.KeyDelete(key);
+                _ = dbRacer.StringSet(key, StringValue);
```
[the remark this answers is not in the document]
tiagonapoli: Addressed in 64e6a69e58 — both the KeyDelete and StringSet results are now asserted, so the test fails if the delete or the String recreation did not take.

test/cluster/Garnet.test.cluster.vectorsets/VectorSets/ClusterVectorSetTests.cs
```diff
@@ -2213,6 +2216,86 @@ public async Task FailoverStopsVectorManagerReplicationTasksAsync()
...
+                        vaddResult = opDb.Execute("VADD", Key, "VALUES", "3", "1", "2", "3", "elemA");
+                    }
+                    catch (Exception ex)
+                    {
+                        vaddError = ex;
+                    }
+                });
+
+                // Wait until the primary VADD has built its index and parked right before writing the synthetic append-log entry to the AOF.
+                await ExceptionInjectionHelper.WaitOnClearAsync(ExceptionInjectionType.VectorSet_Pause_Before_Synthetic_Replication_Rmw).ConfigureAwait(false);
```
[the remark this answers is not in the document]
tiagonapoli: Addressed in 64e6a69e58 — the pause wait is now bounded with `.WaitAsync(TimeSpan.FromSeconds(30))`, matching the standalone tests, so a missed pause surfaces as a timeout instead of hanging CI.

libs/server/Storage/Functions/MainStore/RMWMethods.cs
tiagonapoli: Addressed — the NeedInitialUpdate comment now describes the synthetic args as only valid against an already-existing index record (the CopyUpdater/InPlaceUpdater paths), and that reaching NeedInitialUpdate means the key was concurrently deleted.

libs/common/Testing/ExceptionInjectionHelper.cs
```diff
@@ -138,6 +138,19 @@ public static async Task ResetAndWaitAsync(ExceptionInjectionType exceptionType)
...
 
+        /// <summary>
+        /// Synchronous, event-driven counterpart to <see cref="ResetAndWaitAsync"/>: signals arrival by
+        /// clearing the condition, then parks on the shared <see cref="TaskCompletionSource{TResult}"/>
+        /// until it is re-enabled.
+        /// </summary>
+        /// <param name="exceptionType"></param>
+        public static void ResetAndWait(ExceptionInjectionType exceptionType)
+        {
+#pragma warning disable VSTHRD002
```
kevin-montrose: Use `AsyncUtils.BlockingWait` instead of suppressing warnings.

test/cluster/Garnet.test.cluster.vectorsets/VectorSets/ClusterVectorSetTests.cs
```diff
@@ -2213,6 +2216,86 @@ public async Task FailoverStopsVectorManagerReplicationTasksAsync()
             ClassicAssert.IsTrue(vsimRes.Length > 0);
         }
 
+#if DEBUG
```
kevin-montrose: Rather than `#if DEBUG`'ing the whole test, put a `ClassicAssert.Ignore` in here for non-DEBUG builds (see `ReplicaSyncTaskFaultsRecoverAsync` for example).  We're kinda inconsistent about this overall, but that's the pattern we have in Vector Set tests.  Keeps the test count the same between configurations and reduces the size of the `#if` block, which is nice.

test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs
```diff
@@ -3637,6 +3637,108 @@ public async Task HideInternalRecordsAsync()
             }
         }
 
+#if DEBUG
```
kevin-montrose: Same note as above.

kevin-montrose (changes requested):
Couple nits, and needs a clean test run.

test/cluster/Garnet.test.cluster.vectorsets/VectorSets/ClusterVectorSetTests.cs
tiagonapoli: Good point, much better

kevin-montrose (approved):

libs/server/Storage/Functions/MainStore/RMWMethods.cs
```diff
@@ -291,15 +298,14 @@ public readonly bool InitialUpdater(ref LogRecord logRecord, in RecordSizeInfo s
...
-                        if (input.arg1 is VectorManager.VADDAppendLogArg or VectorManager.MigrateElementKeyLogArg or VectorManager.MigrateIndexKeyLogArg)
-                        {
-                            // Synthetic op, do nothing
-                            break;
-                        }
+                        // NeedInitialUpdate ensures that only these args reach here
+                        Debug.Assert(input.arg1 is VectorManager.CreateIndexArg or VectorManager.MigrateElementKeyLogArg or VectorManager.MigrateIndexKeyLogArg, "InitialUpdater VADD arg1 called with invalid argument");
 
-                        if (input.arg1 == VectorManager.VADDSetFlagsArg)
+                        // TODO: InitialUpdater will create a new vector index recorded initialized with garbage data
```
tiagonapoli, line 304: @kevin-montrose thoughts?

tiagonapoli: MigrateElementKeyLogArg and MigrateIndexKeyLogArg are triggered against empty keys? Or they can also create one index?

kevin-montrose: They're only triggered against the empty key, and shouldn't create an index.

kevin-montrose (approved):

PR #1935, opened 2026-07-15
Add tests (and fix bugs) for commands WRONGTYPE'ing appropriately against Vector Sets

kevin-montrose:
Most commands, when used on Vector Sets, should `WRONGTYPE`.  Similarly, Vector Set commands when used on non-Vector Set keys should `WRONGTYPE`

We have a bunch of commands, and we've been pretty laxidasical about checking `GarnetStatus` so this is a pretty big PR, though mostly mechanical.

Changes:
 - [x] Add tests for all Vector Set commands operating on non-Vector Set types
 - [x] Implement `VCARD`
 - [x] Implement `VISMEMBER`
 - [x] Sketch out `VLINKS`, `VRANDMEMBER`, and `VSETATTR` - far enough to `WRONGTYPE`, but not fully implemented
 - [x] Add tests for (most) non-Vector Set commands on Vector Set keys
 - [x] Add tests to validate the commands that _overwrite_ Vector Sets (`SET`, `MSET`, etc.) correctly free up their resources
 
---

Overwriting Vector Sets is a more complicated operation - if the value is still in memory, a IPU or CU needs a preceeding delete (because the `RecordType` is changing); while if the value is on disk the blind upsert is fine, but we need to cleanup the Vector Set at some later point.

My solution is to _fail_ IPU/CU's with a `WRONG_TYPE` result and then retry in a transaction, and to implement `ICompactionFunctions.IsDeleted` on `GarnetRecordType` to cleanup Vector Sets which are unreachable but not explicitly tombstone'd.

This approach is taken for:
 - `MSET`
 - `PSETEX`
 - `SET`
 - `SETEX`
 - `SETIFGREATER`
 - `SETIFMATCH`
 - `SETWITHETAG`
 
This impacted [one RI test](https://github.com/microsoft/garnet/pull/1935/changes#diff-93c47d0502fd24ebd8c42a7e73fb098e39f943d04b75100e4aae60b2dee3c9c7R318) ; I left this commented out with a TODO for review by someone with more RI knowledge.

---

Follow up work from discussion:
 - rename `ICompactionFunctions.IsDeleted` -> `IsDeletedBeforeCompaction(in TSourceLogRecord logRecord)`
 - combine `ICompactionFunctions` and `IRecordTriggers`


test/standalone/Garnet.test.vectorset/VectorSetWrongTypeTests.cs
```diff
@@ -0,0 +1,1906 @@
...
+                        var res = await db.StringSetAsync(name, "0").ConfigureAwait(false);
+                        ClassicAssert.IsTrue(res);
+                    }
+                    break;
+                case KeyType.Geo:
+                    {
+                        var res = await db.GeoAddAsync(name, 12.3, 45.6, "fizzbuzz").ConfigureAwait(false);
+                        ClassicAssert.IsTrue(res);
+                    }
+                    break;
```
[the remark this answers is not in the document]
kevin-montrose: Wildy incorrect.

libs/server/Resp/Vector/RespServerSessionVectors.cs
```diff
@@ -1606,42 +1625,182 @@ private bool NetworkVISMEMBER<TGarnetApi>(ref TGarnetApi storageApi)
...
+                            WriteNull();
+                        }
+                        break;
+                    case GarnetStatus.WRONGTYPE:
+                        WriteError(CmdStrings.RESP_ERR_WRONG_TYPE);
+                        break;
+
+                    case GarnetStatus.OK:
+                        {
+                            // TODO: implement!
```
vazois, line 1792: nit; are we waiting for something additional to be implemented here?

libs/server/Resp/Vector/VectorManager.Cleanup.cs
```diff
@@ -425,5 +435,115 @@ public void WaitForDiskANNIndexDrop(ReadOnlySpan<byte> key)
                 _ = Thread.Yield();
             }
         }
+
+        /// <summary>
+        /// For testing purposes, block until all cleanup requests are processed.
+        /// </summary>
+        internal void WaitForCleanupRequests()
```
vazois, line 442: if this is for testing only, can we wrap it in DEBUG?

libs/server/Resp/Vector/VectorManager.Cleanup.cs
```diff
@@ -425,5 +435,115 @@ public void WaitForDiskANNIndexDrop(ReadOnlySpan<byte> key)
...
+            potentiallyDeleted[context] = key.ToArray();
+        }
+
+        /// <summary>
+        /// Called when a checkpoint completes, signalling that Vector Sets passed to <see cref="VectorSetPotentiallyDeleted(ReadOnlySpan{byte}, ReadOnlySpan{byte})"/> should be processed.
+        /// </summary>
+        public unsafe void CheckpointCompleted()
+        {
+            _ = Interlocked.Increment(ref postCheckpointTasksRunning);
+            _ = Task.Run(() => QueueCleanups(this));
```
vazois, line 480: Should we associate this with some type of token?
I am wondering also if such background tasks should be associated with the TaskManager.cs, even if they are transient.

libs/server/Resp/BasicCommands.cs
```diff
@@ -354,11 +357,30 @@ private bool NetworkSET<TGarnetApi>(ref TGarnetApi storageApi)
...
             var status = storageApi.SET(key, value);
 
+            // Type already stored in key which requires an explicit delete
             if (status == GarnetStatus.WRONGTYPE)
             {
-                while (!RespWriteUtils.TryWriteError(CmdStrings.RESP_ERR_WRONG_TYPE, ref dcurr, dend))
-                    SendAndReset();
-                return true;
+                var createTransaction = false;
+                if (txnManager.state != TxnState.Running)
```
vazois: nit; MSET uses the same piece of code, maybe use a helper method?

vazois (approved):

libs/server/Resp/Vector/RespServerSessionVectors.cs
kevin-montrose: Yeah, diskann-garnet needs to expose this information.  So I've implemented just enough to get `WRONGTYPE` tests working, but I can't do the whole thing yet.  Same with `VRANDMEMBER` and `VSETATTR`.

libs/server/Resp/Vector/VectorManager.Cleanup.cs
kevin-montrose: These tests also run in RELEASE builds, so we can't make this a conditionally compiled method.

libs/server/Resp/Vector/VectorManager.Cleanup.cs
kevin-montrose: Broadly I think it'd make sense to consolidate all task creation & management, but there are a bunch of places that use the `_ = Task.Run(...)`-pattern already - so I think it'd make more sense to do that in a separate change.

libs/server/Resp/BasicCommands.cs
kevin-montrose: Pushed a helper and converted my changes to use it.

vazois (approved):

PR #1948, opened 2026-07-20
Fix SynchronizationLockException on pub/sub self-publish

### LINKED ISSUE #1615: Pub/sub: network lock error, and allows more than it should (Garnet 2.0.0 64 bit)
### Describe the bug

First, minor (ish): it looks like in RESP2 mode, even in pub/sub mode, the server is allowing arbitrary operations like `set`, `get` etc; this is invalid as it is not reliably possible to distinguish request/response responses from out-of-band message (which would be "push" types in RESP3). Redis responds with an error indicating that only a few operations are valid in that state - basically: `[S|P][UB]SUBSCRIBE`, `PING` and `QUIT`, noting that `PING` has different semantics in pub/sub mode. To repro:

> subscribe foo
> 
> get bar <==== this should be a -ERR, and isn't (in RESP2 mode)

Second, this (equally invalidly) allows `publish`, which does weird things to the server:

> subscribe foo
>
> publish foo bar <==== this should also be a -ERR, and isn't (in RESP2 mode)

now switch to the server console:

> 03::45::44 fail: GarnetServer[0] An error occurred at RecvEventArg_Completed System.Threading.SynchronizationLockException: The calling thread does not hold the lock.    at System.Threading.SpinLock.ExitSlowPath(Boolean useMemoryBarrier)    at Garnet.common.GarnetTcpNetworkSender.ExitAndReturnResponseObject() in /_/libs/common/Networking/GarnetTcpNetworkSender.cs:line 150    at Garnet.server.RespServerSession.TryConsumeMessages(Byte* reqBuffer, Int32 bytesReceived) in /_/libs/server/Resp/RespServerSession.cs:line 500    at Garnet.networking.NetworkHandler`2.TryProcessRequest() in /_/libs/common/Networking/NetworkHandler.cs:line 484    at Garnet.networking.NetworkHandler`2.Process() in /_/libs/common/Networking/NetworkHandler.cs:line 340    at Garnet.networking.NetworkHandler`2.OnNetworkReceiveWithoutTLS(Int32 bytesTransferred) in /_/libs/common/Networking/NetworkHandler.cs:line 274    at Garnet.common.TcpNetworkHandlerBase`2.HandleReceiveWithoutTLS(Object sender, SocketAsyncEventArgs e) in /_/libs/common/Networking/TcpNetworkHandlerBase.cs:line 208

(above)

### Expected behavior

_No response_

### Screenshots

_No response_

### Release version

_No response_

### IDE

_No response_

### OS version

_No response_

### Additional context

_No response_

hexonal:
Fixes #1615 (the network-lock-error part of that issue; the RESP2 command-restriction part — Garnet not yet rejecting non-`(P)SUBSCRIBE`/`PING`/`QUIT`/`RESET` commands while in subscribe mode — is a separate, larger concern and out of scope for this PR).

## Root cause

`GarnetTcpNetworkSender`'s `spinLock` field is a `SpinLock` struct initialized with `new()` and no user-declared parameterless constructor. That zero-inits `_owner`, which per `SpinLock` semantics enables thread-owner tracking (equivalent to `new SpinLock(true)`), so a same-thread reentrant `Enter()` throws.

A session can trigger a reentrant call into its own network sender: `PUBLISH` on a connection that is itself subscribed to the target channel causes `SubscribeBroker.PublishNow`/`Broadcast` to call back into that same connection's `Publish()`/`PatternPublish()` synchronously, while `TryConsumeMessages` for the `PUBLISH` command is still on the stack and still holding the sender's `SpinLock`. Garnet does not currently enforce RESP2's "only a few commands allowed while subscribed" restriction, so this is reachable without RESP3.

`EnterAndGetResponseObject` on that reentrant call throws before touching `lockTaken`/`responseObject`, but the old code's `finally` block unconditionally called `ExitAndReturnResponseObject()`. Since `SpinLock.Exit()`'s ownership check is by thread ID (not call-frame/recursion depth), the erroneous reentrant `Exit()` succeeded and silently released the *outer* frame's lock early. The outer `TryConsumeMessages` `finally` then called `Exit()` again on an already-released lock, throwing an uncaught `SynchronizationLockException` and tearing down the connection.

As a secondary effect, the same unconditional `finally` also returned `responseObject` to the free buffer pool on the reentrant frame — but `responseObject` there is still the *outer* frame's actively-in-use buffer (never reset, since the failed `Enter` never got that far), so the old code was also returning a buffer to the pool while it was still being written to.

## Fix

In `PubSubCommands.cs`, `Publish()` and `PatternPublish()` now track whether `EnterAndGetResponseObject` actually succeeded (`entered`, following the same `lockTaken`-guard pattern already used in `EnterAndGetResponseObject`/`GarnetTcpNetworkSender.cs`) and only call `ExitAndReturnResponseObject()` in `finally` if it did. This also incidentally fixes the buffer-pool double-use described above, since the erroneous `Exit`/return no longer happens.

The `Debug.Assert(isSubscriptionSession == false)` in `NetworkPUBLISH` is removed: in Debug builds without a debugger attached, `Debug.Assert` failure calls `Environment.FailFast` and kills the whole process, so this exact scenario would crash the test process outright regardless of the lock fix. The assert was asserting a false invariant anyway — self-publish-while-subscribed is a real, currently-unrestricted scenario.

## Behavior note / known tradeoff

With this fix, a message a session publishes to a channel it is itself (pattern-)subscribed to is **not delivered back to that same connection** — the reentrant delivery attempt is now a no-op instead of corrupting the connection, so the self-message is silently dropped rather than crashing. `PUBLISH` itself still completes normally and still reports the correct subscriber count. This is a deviation from full Redis pub/sub semantics (a subscriber normally does receive its own publish) and is a deliberate, minimal tradeoff to fix the crash without restructuring the network-sender locking or deferring self-delivery to after the outer lock is released. Delivering the self-message correctly (e.g. by queuing it for delivery once the outer `TryConsumeMessages` lock is released) would be a larger change and is left as potential follow-up; happy to discuss if maintainers would prefer that approach instead.

## Testing performed

- `dotnet build test/standalone/Garnet.test/Garnet.test.csproj` in both Debug and Release: both succeed, 0 warnings, 0 errors.
- `dotnet test ... --filter "FullyQualifiedName~RespPubSubTests"` on net10.0, Debug and Release: **8/8 pass** (7 pre-existing + 2 new regression tests, one for `SUBSCRIBE`/`Publish` and one for `PSUBSCRIBE`/`PatternPublish`).
- Reverted the `entered`-guard fix in `Publish()` only (kept the `Debug.Assert` removal) and reran the new `SelfPublishOnSubscribedChannelDoesNotCorruptConnection` test alone: it fails, with the connection dying mid-`PUBLISH` as expected. Restored the fix: passes again, confirmed stable across repeated runs.
- `dotnet format --verify-no-changes` on the changed files: clean.
- `git diff --check`: no whitespace/line-ending issues.
- Manual review of the diff against existing lock-guard conventions elsewhere in the same files (`GarnetTcpNetworkSender.cs`, `EnterAndGetResponseObject`).
- Not run: the full Garnet test suite (only the pub/sub-focused subset above was targeted) and any cluster/RESP3-specific pub/sub paths beyond what these tests cover.

Diff touches `libs/server/Resp/PubSubCommands.cs` and `test/standalone/Garnet.test/RespPubSubTests.cs` only.

### DIFF
--- a/libs/server/Resp/PubSubCommands.cs
+++ b/libs/server/Resp/PubSubCommands.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using Garnet.common;
 using Tsavorite.core;
 
@@ -21,9 +20,18 @@ namespace Garnet.server
         /// <inheritdoc />
         public override unsafe void Publish(PinnedSpanByte key, PinnedSpanByte value)
         {
+            // Tracks whether EnterAndGetResponseObject succeeded, so we only release the lock we actually
+            // acquired. A session can receive its own published message while it is itself the publisher
+            // (e.g. RESP3 PUBLISH to a channel it is subscribed to), in which case this is a reentrant call
+            // on a thread that is already holding the same non-reentrant SpinLock via the outer
+            // TryConsumeMessages/EnterAndGetResponseObject call. Without this guard, the failed re-entrant
+            // Enter would still be unconditionally Exit-ed here, releasing the outer lock early and causing
+            // a SynchronizationLockException when the outer call later tries to exit it itself.
+            var entered = false;
             try
             {
                 networkSender.EnterAndGetResponseObject(out dcurr, out dend);
+                entered = true;
 
                 WritePushLength(3);
 
@@ -44,16 +52,21 @@ namespace Garnet.server
             }
             finally
             {
-                networkSender.ExitAndReturnResponseObject();
+                if (entered)
+                    networkSender.ExitAndReturnResponseObject();
             }
         }
 
         /// <inheritdoc />
         public override unsafe void PatternPublish(PinnedSpanByte pattern, PinnedSpanByte key, PinnedSpanByte value)
         {
+            // See the comment in Publish() above: guard against releasing a lock we never acquired when this
+            // is a reentrant call on a thread that already holds the sender's SpinLock.
+            var entered = false;
             try
             {
                 networkSender.EnterAndGetResponseObject(out dcurr, out dend);
+                entered = true;
 
                 WritePushLength(4);
 
@@ -74,7 +87,8 @@ namespace Garnet.server
             }
             finally
             {
-                networkSender.ExitAndReturnResponseObject();
+                if (entered)
+                    networkSender.ExitAndReturnResponseObject();
             }
         }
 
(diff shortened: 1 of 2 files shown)


badrishc:
Thank you @hexonal for your contributions! For this particular PR, note that there was already an open PR https://github.com/microsoft/garnet/pull/1669 that handles the lock issue as well as RESP command restrictions. We have incorporated aspects such as the test and RCA from here into that one.

PR #1952, opened 2026-07-20
Fix spurious NOTFOUND from ContinuePendingRead after a successful disk read

### LINKED ISSUE #1950: Data loss on recovery: SeSaveRecoverMultipleKeysTest("25k","18k") intermittently loses a key (~50% repro rate)
### Describe the bug

`RespAdminCommandsTests.SeSaveRecoverMultipleKeysTest("25k","18k")` (`test/standalone/Garnet.test/RespAdminCommandsTests.cs`) intermittently fails after a server restart with `tryRecover: true`: one specific key, `SeSaveRecoverTestKey0693`, comes back as an empty string instead of its expected value after recovery, even though it was written and verified present before the restart. All 1999 other keys in the same test recover correctly.

This looks like a real (if narrow) data-durability bug in the recovery path, not simple test flakiness — see reproduction data below.

```
dotnet test test/standalone/Garnet.test/Garnet.test.csproj -c Release --framework net10.0 --filter "FullyQualifiedName~SeSaveRecoverMultipleKeysTest"
```

Run it repeatedly (single run isn't enough — this is timing-dependent). Over 18 local runs (macOS, arm64, .NET 10.0.10, Release config) it failed 9 times (**50%**) — every single failure was the exact same test case, `("25k","18k")`, and the exact same key, `SeSaveRecoverTestKey0693`. It's not a random/different key each time.

We also independently hit this same test (different parameterizations of the same signature — an empty-string assertion failure) twice in CI on unrelated PRs (#1945, #1946), and once saw a full 45-minute CI job timeout/hang after this test's `("16k","16k")` case completed with no further log output at all (PR #1949's `ubuntu-latest, net10.0, Debug` job) — possibly a related but rarer failure mode in the same area, not independently reproduced locally yet.

### Expected behavior

`SeSaveRecoverTestKey0693` (and every other key) should recover with its original value (`"SeSaveRecoverTestValue"`) after the server restart, matching the assertion already in the test.

### What we found so far

The test writes 1000 keys, does a background `SAVE`, writes 1000 more keys, does `COMMITAOF`, disposes the server, then restarts it with `tryRecover: true` at a *smaller* `memorySize`/`recoveryMemorySize` than the original run — forcing recovery to happen under tighter memory pressure than the data was originally written under, which we believe is what makes this reproducible at all (the other, larger-memory test cases in the same parameterized test did not fail in our runs).

We traced (without being able to pin an exact line within the time we spent) that the likely interaction is between:
- The synchronous eviction path added in #1932 (`AllocatorBase.IssueShiftAddress`, invoked whenever the `LogSizeTracker` resizer `IsRunning` is false — which is true during recovery), and
- The multi-phase async snapshot-checkpoint recovery logic in `libs/storage/Tsavorite/cs/src/core/Index/Recovery/Recovery.cs` (`RecoverHybridLogAsync` / `RecoverHybridLogFromSnapshotFileAsync`), which already has to reason about "evictions happening mid-flight" per its own comments while adjusting `HeadAddress`.

Our hypothesis (not confirmed at the exact line level): a record landing near a page/segment boundary during this eviction-while-recovering interleaving can end up with its index entry pointing below the newly-adjusted `HeadAddress` without the on-disk read fallback correctly kicking in for it, making it read back as absent/empty. #1932's own added comments (`LogSizeTracker.cs`, the `resizerTask.IsCompleted` "defense in depth... even if `OnStopped()` somehow did not run" check) suggest the original fix wasn't fully confident it eliminated every timing case in this area, which is consistent with what we're seeing.

We don't have deep expertise in Tsavorite's recovery state machine to pin this down further ourselves with confidence, but wanted to file this promptly given how reliably it reproduces (a maintainer familiar with this code, with the repro rate/key above, should be able to nail it quickly with targeted tracing/logging around key 0693's record address during that specific test case).

Happy to help test a fix, or dig further ourselves if useful — let us know.

hexonal:
Addresses part of #1950 (see that issue for the full reproduction and analysis — this PR fixes one of at least two distinct causes of the flakiness described there).

## Root cause

In `ContinuePendingRead` (`libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/ContinuePending.cs`), after a pending (disk) read's `Reader()` call succeeds, `status` is only explicitly set to `SUCCESS` inside the two branches gated by `operationState.readCopyOptions.CopyTo` (`MainLog` or `ReadCache`). When `CopyFrom != ReadCopyFrom.None` but `CopyTo == ReadCopyTo.None` — a valid configuration reachable via `ReadCopyFrom.Inherit` resolution — neither branch runs, and `status` is left holding whatever `FindTagAndTryEphemeralSLock`'s out-parameter set earlier in the call, which is only meaningful on a failed lock/find and effectively behaves like `NOTFOUND` here. A fully successful disk read, with the correct value already retrieved into `output`, was therefore reported back to the caller as `NOTFOUND`, discarding the correct value.

This is reachable whenever a plain read is forced to disk (record evicted below `HeadAddress`) under that specific `CopyFrom`/`CopyTo` combination — uncommon in normal steady-state operation, but reliably triggered right after recovering into a tighter memory budget than the data was originally written under (exactly what `RespAdminCommandsTests.SeSaveRecoverMultipleKeysTest("25k","18k")` does, which is how this was found).

## Fix

One line: explicitly set `status = OperationStatus.SUCCESS;` immediately after a successful `Reader()` call, before the `CopyTo`-gated branching — matching the already-correct pattern in the synchronous read path (`InternalRead.cs`, where `status = OperationStatus.SUCCESS` is set unconditionally right after `Reader()` succeeds, before any `CopyFrom`-gated logic runs). The async/pending path was simply missing the equivalent line.

## Testing

- Reproduced the underlying symptom via `SeSaveRecoverMultipleKeysTest("25k","18k")`: on unpatched `main`, 6/20 (~30%) local runs failed. With this fix, 6/150 (~4%) failed — a ~7x reduction, confirming this addresses a real, distinct cause.
- Ran the broader relevant test surface to check for regressions: `RespAdminCommandsTests` (56 tests), `RespConfigTests` + `AofShardedTxnRecoveryTests` (18), `Tsavorite.test.recovery` (204), `CompletePendingTests`/`LowMemoryTests`/`StateMachineDriverTests` (33), `ReadAddressTests`/`ReadCacheTests` (171). All pass, no regressions observed.

## This does not fully close #1950

The residual ~4% failure rate is a second, distinct issue — in a captured failure, `ContinuePendingRead`'s `request.logicalAddress` was `0` instead of the correct address at completion time. I was not able to pin that down with confidence within a reasonable investigation budget and did not want to guess at a fix for it in this sensitive async-completion path. Full details and a working diagnostic-instrumentation approach are in the issue; happy to keep digging into the second cause separately.

### DIFF
--- a/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/ContinuePending.cs
+++ b/libs/storage/Tsavorite/cs/src/core/Index/Tsavorite/Implementation/ContinuePending.cs
@@ -174,6 +174,13 @@ namespace Tsavorite.core
                     // See if we are copying to read cache or tail of log. If we already found the record in memory, we're done.
                     if (operationState.readCopyOptions.CopyFrom != ReadCopyFrom.None && !memoryRecord.IsSet)
                     {
+                        // The Reader() call above already succeeded, so the operation is a SUCCESS regardless of whether a copy is
+                        // actually performed below (e.g. CopyTo is None, which is a valid "track CopyFrom for stats but don't copy"
+                        // configuration, or the ReadCache copy's guard conditions are not met). Without this, 'status' retains
+                        // whatever FindTagAndTryEphemeralSLock set via its out-parameter above, which is only meaningful when that
+                        // call returns false (i.e. it is NOT a valid "found" status here) - previously causing a spurious NOTFOUND to
+                        // be returned for a completed, correctly-read pending Read() whenever neither branch below ran.
+                        status = OperationStatus.SUCCESS;
                         if (operationState.readCopyOptions.CopyTo == ReadCopyTo.MainLog)
                         {
                             // Plumb source logical address so PostCopyToTail can name per-flush snapshot files.


hexonal:
CI finished: 209/211 checks green. Two cluster jobs failed:
- `Garnet Cluster (ubuntu-latest, net10.0, Debug, Garnet.test.cluster.multilog)` — `ClusterDivergentReplicasMMTest`, failing with `TaskCanceledException` from `GarnetServerNode.GossipAsync`
- `Garnet Cluster (windows-latest, net10.0, Release, Garnet.test.cluster.vectorsets)` — `ClusterVectorSetTests.VectorSetMigrateManyBySlot`, also preceded by a `GossipAsync` `TaskCanceledException`, then an `IndexOutOfRangeException` in the test's own array indexing

I looked into whether these could be caused by this PR's change before assuming otherwise, since it touches core read-path status handling. My assessment, with the caveats noted:

- Both failures share the same underlying trigger (a gossip round-trip timing out), not a data-correctness symptom — this PR only changes when `ContinuePendingRead` reports `SUCCESS` vs a stale status after an already-successful disk read; it doesn't touch cluster gossip, AOF streaming, or migration slot-assignment logic.
- I don't see `ReadCopyOptions`/`ContinuePendingRead` referenced anywhere in `libs/cluster/Server/Migration` or `libs/cluster/Server/Replication`.
- I don't have permission to re-run just the failed jobs on this commit to check for pure flakiness (not a repo admin), and my local attempt to reproduce these specific cluster tests hit an unrelated environment limitation (fails at cluster bootstrap itself, a different symptom than the CI failure) — so I can't claim a clean before/after confirmation, just the reasoning above.

Flagging this transparently rather than asserting certainty, given this PR touches core Tsavorite read logic. Let me know if you'd like me to dig further, or if this pattern (gossip-timeout flakiness in cluster CI jobs) is already familiar.

hexonal:
Follow-up on the two cluster failures above — dug further rather than leaving it at "probably unrelated":

**`ClusterDivergentReplicasMMTest`**: pulled the raw job log directly. The actual failure chain is `AddDelSlotsRange error ERR Slot 0 is already busy` → a later `IncrBy` fails because slot ownership ends up inconsistent → `Assert.Fail()`. This is a cluster slot-assignment/gossip-convergence race in `ClusterManager` — no code path through `ContinuePendingRead`. The `GossipAsync TaskCanceledException` lines are teardown-time noise; the same pattern shows up throughout both passing and failing tests across the whole job log. Also ran `ClusterDivergentReplicasMMTest` 160 times (20 iterations × 8 parametrized cases) locally on this PR's branch and, as a control, the same 160 runs on the exact `upstream/main` base commit: **0/160 failures on both** — no measurable difference.

**`VectorSetMigrateManyBySlot`**: couldn't get a clean local A/B for this one (blocked by an unrelated local/macOS-arm64 issue — the native `diskann_garnet.dylib` isn't available in this environment, so every `VADD` fails locally regardless of branch). Traced the code instead: vector-set migration's `WriteOrSendRecordAsync` does call into `VectorBasicContext.Read()`/`CompletePending`, which *can* reach `ContinuePendingRead` if a migrated element was evicted to disk — so there is a real, if narrow, path from this PR's change to this test. But the test only adds 4 tiny elements per primary with no low-memory configuration, making disk eviction implausible here. The failure itself (`IndexOutOfRangeException` inside the test body) is preceded by the same benign gossip-teardown noise as the other failure.

Net: high confidence the first failure is unrelated (direct root-cause trace + 320-run local parity), moderate-high confidence the second is too (same noise signature, and the one theoretically-reachable path is impractical given the test's tiny data size — the local block on confirming that statistically is an unrelated native-library gap, not something this PR affects). Standing by the PR as-is.

TedHartMS:
Thanks for the report and investigation. #1957 supersedes this; the change here is not needed, as comments in that PR at this point indicate:

// status is already OperationStatus.SUCCESS here: FindTagAndTryEphemeralSLock (at the top of the loop) returns true
// only after TryEphemeralSLock has set status to SUCCESS, and nothing below changes it before this point.

hexonal:
Thanks for the correction, and confirming here rather than letting it stand unverified: you're right, and I traced through why.

`FindTagAndTryEphemeralSLock` (`Helpers.cs`) only returns `true` via the path where `TryEphemeralSLock` (`TransientLocking.cs`) already set `status = OperationStatus.SUCCESS`; the failure path returns `false` before that point, and nothing between there and where I added the line reassigns `status`. So the line I added really is a structural no-op at that spot, not a hunch that happened to be wrong — the original root-cause claim in this PR (that `status` was left at some stale non-SUCCESS value) doesn't hold up.

I also went back and re-tested the "30% → 4%" improvement I cited as evidence, since a no-op code change obviously can't have caused a real effect, and the numbers deserved a harder look rather than being left dangling. Built `main`, this PR's branch, and #1957's branch from fresh clones and ran the same repro loop against all three, sequentially and uncontended this time (the original two sessions weren't run under comparable conditions - system load looks like the actual confound, unrelated tests started flaking too under contention, including a multi-minute hang on `main` that's probably the same thing as the "45-min CI hang" from the original report). Clean re-run, 30 iterations each on `("25k","18k")`: `main` 1/30, this PR 1/30, #1957 2/30 - statistically indistinguishable. The original "before/after" numbers were almost certainly a session-to-session environment artifact, not a real signal. Closing this out.

One thing worth flagging while I have the harness set up: `SeSaveRecoverMultipleKeysTest("25k","18k")` still failed occasionally against your #1957 branch in that same clean re-test (2/30, and 4/30 under load) - same signature both times (`Key SeSaveRecoverTestKey0693`, expected 22-char string, got empty). Small sample, so not conclusive, but it's the same key every time, which matches the pattern from before. Possibly relevant: the residual `logicalAddress == 0` case I mentioned in this PR's original description might be the same root cause your fix already addresses, just not fully - your own new `SpanByteRecoverySnapshotEvictionTests` (direct Tsavorite-level) passed every time for me, so if there's a gap it's plausibly specific to the RESP+AOF-layered recovery path that test doesn't exercise. Flagging in case it's useful before merge; happy to share the raw logs if wanted.

TedHartMS:
Yes--turns out there was another issue. Try the current 1957 (it's draft now as I'm investigating something related to the earlier fix). Thanks!

PR #1957, opened 2026-07-21
Fix intermittent key loss in SeSaveRecoverMultipleKeysTest("25k","18k")

TedHartMS:
This pull request makes significant improvements to the hybrid log recovery process, particularly around snapshot recovery, memory management, and durability guarantees. The changes ensure that pages read from a snapshot are properly flushed to the main log before eviction, supporting correct recovery even when recovering into a smaller memory budget than was checkpointed. Additionally, the object loading phase is now more robust and flexible, selecting the appropriate object log device per page.

The most important changes are:

**Snapshot Recovery and Durability Improvements:**

* Introduced `snapshotScanFromAddress` in `RecoveryStatus` to track the hybrid-log/snapshot boundary, ensuring that pages read from the snapshot are flushed to the main log before eviction, making them durable and preventing data loss. [[1]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8R31-R35) [[2]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L872-R941)
* Added `FlushIfNeededThenEvictPageForRecovery` and `FlushSnapshotPageForRecovery` methods to encapsulate the logic for flushing snapshot-region pages before eviction, replacing the previous unconditional flush of every snapshot page with a targeted approach.

**Recovery Logic and Memory Management:**

* Updated page eviction logic in `TrimLogPages` and `TrimResidentPagesToBudget` to use the new flush-before-evict mechanism, ensuring that only necessary pages are flushed and that eviction does not compromise data integrity. [[1]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L613-L635) [[2]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L644-R711) [[3]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L663-R727)
* Improved the post-recovery state alignment to maintain correct read-only and flushed boundaries, preventing issues such as failed disk reads due to unflushed pages.

**Object Loading Enhancements:**

* Refactored `RecoveryLoadObjectsPass2` to perform a single top-down pass over the recovered range, selecting the correct object log device per page based on the snapshot boundary, and ensuring that evicted snapshot pages are flushed first. [[1]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L908-R983) [[2]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L940-R1005) [[3]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8R1024) [[4]](diffhunk://#diff-1a9d48a0a34b6d560075b3ca0ba30ded8a511e39bb152e9b09119779fe425fb8L737-R801)

**Code Safety and Cleanup:**

* Made disposal of `objectLogRecoveryDevice` null-safe to prevent potential exceptions during cleanup.


hexonal:
Possibly-related data point: I've seen the same *works-before-save, one-record-missing-after-`tryRecover`* symptom show up in an unrelated code path — `RespVectorSetTests.RecreateIndexesOnRestoreAsync` (`test/standalone/Garnet.test.vectorset/RespVectorSetTests.cs:2348`) — twice today, on two different unrelated PR branches, both freshly rebased onto current `main`:

- https://github.com/microsoft/garnet/actions/runs/29979858823/job/89123471985 (`Garnet Standalone (ubuntu-latest, net8.0, Release, Garnet.test.vectorset)`)
- https://github.com/microsoft/garnet/actions/runs/29980825950/job/89122569138 (same job name)

Both fail identically: after `VADD`ing 2 elements to a vector set, a `VSIM ... ELE ...` query correctly returns 2 matches, then a `ForegroundSave` + `WaitForCommitAsync` + server restart with `tryRecover: true`, and the same query only returns 1 match (`Expected: 2, But was: 1`). Deterministic across both runs, not a random/timing-shaped flake.

I haven't been able to bisect or reproduce this locally (no native `diskann` build available on macOS/arm64 in my sandbox, so every `VADD` fails locally regardless of branch), so take this as a raw observation, not a confirmed diagnosis. But the shape matches this PR's fix description closely: VectorSet element records live in the main store as fixed-size stub records, and `RecoverSnapshotPages`'s old behavior (skip-flush for non-object-allocator stores) would apply to that store too. If a stub record's page got evicted before being flushed during recovery, the RecreateIndex-on-first-access path would rebuild the native HNSW index from whatever main-store records actually survived — silently dropping the evicted element, which matches "1 of 2 elements found" exactly.

Not asking you to scope this PR to cover it — just flagging in case it's useful corroborating evidence (or in case you want to add a VectorSet-specific case to `SpanByteRecoverySnapshotEvictionTests` once this lands). Happy to help verify once #1957 merges if that'd be useful.

badrishc (approved):

hexonal:
Following up on my hedged comment above, since I said I'd verify once this landed: **it did fix `RecreateIndexesOnRestoreAsync` too.**

That test failed deterministically on the two runs I linked (2026-07-23, both on `ubuntu-latest, net8.0, Release, Garnet.test.vectorset`, both `Expected: 2, But was: 1`). Since this merged, it passes everywhere I can find it:

- Run 30499380235 (`Garnet .NET CI`, 2026-07-29) and 30532002635 (`Garnet Nightly Tests`, 2026-07-30), both on `52aef6e6` — 16/16 in each: all 8 `Garnet.test.vectorset` legs × both `RespVectorSetTests(0)` and `(1000)` parameterizations, including the `ubuntu net8.0 Release` leg that had been failing.

So the mechanism I guessed at — VectorSet element stubs living in the main store, a page from the final resident snapshot batch getting evicted before flush, and `RecreateIndex` then rebuilding the HNSW index from only the records that survived — looks like it was the right shape after all. Thanks for the fix; the VectorSet case came along for free.

One thing you may already know, mentioning it only because it's the same bug: `release/v1` doesn't carry this fix (`fc92f53c` is not an ancestor of `release/v1`), so the recovery data loss is still live there. Not volunteering to backport it — it's your subsystem and I'd rather not guess at what v1 needs — but happy to help verify if you do decide to.

PR #1961, opened 2026-07-21
Fix VREM of disk-evicted Vector Set element spuriously reporting not-found

tiagonapoli:
## Problem
`VREM` of a Vector Set element silently reported "not removed" (`:0`, no error) whenever that element's records had been evicted to disk.

The DiskANN delete callback (`VectorManager.DeleteCallbackUnmanaged`) gated its return value on `status.Found`. Tsavorite deletes are tombstone appends and never read the old record back, so for a disk-resident record `status.Found == false` even though the key logically exists. The failing `0` then propagated:

`DeleteCallbackUnmanaged` returns `0` → native `remove` != 1 → `DiskANNService.Remove` `false` → `VectorManager.TryRemove` -> `MissingElement` → `NetworkVREM` writes `:0`.

The client could not distinguish this from "element was never there" — a data-correctness bug with no exception.

## Fix
Return success on `status.IsCompletedSuccessfully` (dropping the `&& status.Found`), matching the sibling `Write`/`RMW` callbacks which already only check `IsCompletedSuccessfully`.

## Tests
Two layered regression tests, both forcing every record to disk via `Log.FlushAndEvict` first:

- **`VREMOnDiskFlushedElementSucceeds`** — end-to-end over RESP: asserts `VREM` returns `1`, a repeat returns `0`, and a `VSIM` consistency check.
- **`VREMOnDiskFlushedElement_ServiceRemoveSucceeds`** — lower-level: drives `VectorManager.TryRemove` → `DiskANNService.Remove` directly (bypassing RESP parsing/encoding and AOF replication), asserting `VectorManagerResult.OK`.

Verified before/after by toggling the fix:

| Layer | With fix | Without fix |
|---|---|---|
| E2E (`VREM`) | `1` | `Expected: 1, But was: 0` |
| `Service.Remove` (`TryRemove`) | `OK` | `Expected: OK, But was: MissingElement` |

### DIFF
--- a/libs/server/Resp/Vector/VectorManager.Callbacks.cs
+++ b/libs/server/Resp/Vector/VectorManager.Callbacks.cs
@@ -388,7 +388,9 @@ namespace Garnet.server
             var status = ctx.Delete(keyWithNamespace);
             Debug.Assert(!status.IsPending, "Deletes should never go async");
 
-            return status.IsCompletedSuccessfully && status.Found ? (byte)1 : default;
+            // Note: we don't check status.Found here because Tsavorite deletes are just a tombstone entry.
+            // If the key was flushed to disk, status.Found will be false even though the key exists.
+            return status.IsCompletedSuccessfully ? (byte)1 : default;
         }
 
         [UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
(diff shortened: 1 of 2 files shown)


tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/16

PR #1965, opened 2026-07-21
Add runtime management of eligible server options via CONFIG GET/SET

vazois:
## What

Introduces a runtime **configuration management registry** for RESP `CONFIG GET` and `CONFIG SET`.

## Design

Configuration parameters are explicitly registered as either:

- **Read-only:** Startup or physical settings that cannot be changed without restarting the server. They are exposed through `CONFIG GET` for informational purposes and are formatted from the retained `GarnetServerOptions`.
- **Read-write:** Settings that can change without restarting the server. Changes either take effect through live reads or may require restarting a related background task.

Runtime values are managed by a shared `RuntimeServerConfig` instance. Numeric values, booleans, enums, and durations are stored in numeric slots, while read-only parameters may also use string formatters.

`RuntimeServerConfig` provides typed accessors—including `GetInt`, `GetLong`, `GetBool`, `GetEnum<T>`, `GetTimeSpan`, and duration-unit conversions—so callers can consume each value in the form they require.

For settings associated with independently restartable background tasks, `ConfigMeta.UpdateAction` applies the change by reconciling the affected task without restarting the server. This PR includes examples for AOF commit frequency, expired-object collection, and expired-key deletion scanning, but does not attempt to cover every parameter that could support this behavior.

Adding parameters to the configuration management registry is intentionally opt-in. A parameter must be registered with metadata and, for read-write settings, its consumers must read from `RuntimeServerConfig`. This PR adds a representative set of read-only and read-write parameters to demonstrate the feature and establish the extension pattern.

## Additional behavior

- Parameter names are resolved case-insensitively and may define aliases.
- `CONFIG SET` validates types, ranges, and enum members before publishing a value.
- Failed update actions restore the previous value.
- `CONFIG GET *` includes registered read-only and read-write parameters.
- Session-specific settings such as `slave-read-only` remain outside the shared registry.

## Tests

Tests cover runtime GET/SET operations, supported value types, aliases, validation, read-only rejection, `CONFIG GET *`, live slowlog updates, and background-task configuration changes.


libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+            void SetReadOnly(ServerConfigType t, string name, ConfigKind kind)
+                => m[(int)t] = new ConfigMeta(name, kind, 0, 0, null, IsRuntime: true, ReadOnly: true);
+
+            // Read-only parameters: exposed through CONFIG GET (and GET *) via this table, but CONFIG SET
+            // rejects them because they are constants or physical (require restart). Their values are
+            // computed in FormatReadOnly from the owning StoreWrapper.
+            // NOTE: slave-read-only is intentionally NOT here: it is a per-session value (READWRITE/READONLY,
+            // https://redis.io/docs/latest/commands/readwrite/) and is handled directly by the CONFIG GET
+            // handler, which has the calling session in scope.
+            SetReadOnly(ServerConfigType.TIMEOUT_SECONDS, "timeout", ConfigKind.Int32);
```
Mathos1432: Should this be ConfigKind.Seconds ?
```suggestion
            SetReadOnly(ServerConfigType.TIMEOUT_SECONDS, "timeout", ConfigKind.Seconds);
```

libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+        {
+            var seconds = Volatile.Read(ref values[(int)type]);
+            return seconds <= 0 ? Timeout.InfiniteTimeSpan : TimeSpan.FromSeconds(seconds);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public TEnum GetEnum<TEnum>(ServerConfigType type) where TEnum : unmanaged, Enum
+        {
+            var value = (int)Volatile.Read(ref values[(int)type]);
+            return Unsafe.As<int, TEnum>(ref value);
```
kevin-montrose: This assumes all config enums are backed by ints, bad things will happen if there's a `enum Foo : byte` or `enum Bar : short`.

We either need a check/assert (if we want to keep the `int` backing requirement), or to switch on ` Enum.GetUnderlyingType(...)`.

libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        /// <summary>
+        /// Validate <paramref name="value"/> and, if valid, update the slot for <paramref name="type"/>.
+        /// Returns <see langword="null"/> on success, otherwise an error message prefixed with "ERR ".
+        /// </summary>
+        public string TrySet(ServerConfigType type, string value)
```
kevin-montrose: Kind of a strange pattern to have a `TryXXX` return `string`.  `bool TrySet(..., out string error)` is more idomatic.

libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+    /// Kind of value stored in a runtime configuration slot. A slot is a raw 8-byte cell; the kind
+    /// determines how it is parsed (CONFIG SET), validated, formatted (CONFIG GET) and read back.
+    /// </summary>
+    internal enum ConfigKind : byte
+    {
+        Int32,
+        Int64,
+        Bool,
+        Enum,
+        Seconds,
```
Mathos1432: I wonder if it would be possible to instead have TimeSpan as a ConfigKind and use that instead of Seconds/Int32 for encoding/decoding?

It would avoid the need to encode the unit as a suffix which reduces the risk of incorrect storing/retrieval?

libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+                            case "0":
+                                Volatile.Write(ref values[(int)type], 0);
+                                return null;
+                            default:
+                                return $"ERR Invalid value for '{meta.Name}': expected 'yes' or 'no'.";
+                        }
+                    }
+                case ConfigKind.Enum:
+                    {
+                        if (!Enum.TryParse(meta.EnumType, value, ignoreCase: true, out var parsed) ||
```
kevin-montrose: Similar issue with this assuming `int`-backed enums.  This probably won't break for smaller types, but we'll want symmetry with any checks in `GetEnum<T>(...)`

libs/server/RuntimeServerConfig.cs
```diff
@@ -0,0 +1,293 @@
...
+            Type EnumType,
+            bool IsRuntime,
+            bool ReadOnly);
+
+        // Metadata for every ServerConfigType (index == (int)ServerConfigType). Non-runtime types keep
+        // the default (IsRuntime == false) and are served by the bespoke CONFIG code, not this table.
+        static readonly ConfigMeta[] Meta = BuildMeta();
+
+        // Wire name (and aliases) -> ServerConfigType, for CONFIG parameter parsing.
+        static readonly Dictionary<string, ServerConfigType> NameToType = BuildNameLookup();
```
kevin-montrose: A `byte[]` key probably makes more sense, and would let us use `AlternateLookup<ReadOnlySpan<byte>>` in .NET 9+ to avoid allocations.

Similarly a `FrozenDictionary` would be more efficient, since the set doesn't change.

libs/server/ServerConfigType.cs
```diff
@@ -3,15 +3,37 @@
...
+        CLUSTER_REPLICATION_REESTABLISHMENT_TIMEOUT_SECONDS,
+        COMPACTION_MAX_SEGMENTS,
+        COMPACTION_FORCE_DELETE,
+        COMPACTION_TYPE,
+        SLOWLOG_LOG_SLOWER_THAN_MICROS,
+        OBJECT_SCAN_COUNT_LIMIT,
+        SG_GET,
+
+        // Sentinel: number of ServerConfigType values, used to size the runtime value table.
+        COUNT
```
kevin-montrose: A sentinel isn't wrong, but it's a little weird.

We can get the max value with `((int[])Enum.GetValuesAsUnderlyingType<ServerConfigType>()).Max()` and then we don't have to make sure `COUNT` isn't accidentally accepted somewhere as a config name.

Mathos1432:
If the configurations are still present in GarnetServerOptions, is there a risk that some code paths go through the "static" GarnetServerOptions instead of the runtime options, causing some paths to use the updated value and some to use the boot time value?

Is it possible to remove the runtime configs from the GarnetServerOptions, or make it hard to accidentally read their values?

libs/server/StoreWrapper.cs
```diff
@@ -200,13 +206,23 @@ public StoreWrapper(
...
+            RuntimeServerConfig runtimeConfig = null)
         {
             this.version = version;
             this.redisProtocolVersion = redisProtocolVersion;
             this.servers = servers;
             this.startupTimestamp = Stopwatch.GetTimestamp();
             this.serverOptions = serverOptions;
+            // Share the runtime config table across cloned wrappers (e.g. the AOF copy ctor) so that a
+            // CONFIG SET is observed consistently; otherwise seed a fresh table from the startup options.
+            this.runtimeConfig = runtimeConfig ?? new RuntimeServerConfig();
```
kevin-montrose: Rather than this if, pass `this` and `serverOptions` to `RuntimeServerConfig` constructor.

test/standalone/Garnet.test/RespConfigTests.cs
```diff
@@ -738,5 +739,121 @@ public void ConfigSetHeapMemorySizeUtilizationTest(int smallerSize)
...
+        /// and that a rejected value does not mutate the option.
+        /// </summary>
+        [Test]
+        public void ConfigSetRuntimeOptionValidationTest()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+
+            // Out-of-range integer (min is 0).
+            _ = Assert.Throws<RedisServerException>(() => db.Execute("CONFIG", "SET", "replica-sync-delay", "-5"));
```
kevin-montrose: We should check that the error messages are what we expect.

test/standalone/Garnet.test/RespConfigTests.cs
```diff
@@ -738,5 +739,121 @@ public void ConfigSetHeapMemorySizeUtilizationTest(int smallerSize)
...
+        /// parameters and the per-session slave-read-only value.
+        /// </summary>
+        [Test]
+        public void ConfigGetAllAndReadOnlyRejectionTest()
+        {
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+
+            // CONFIG SET on read-only parameters is rejected.
+            _ = Assert.Throws<RedisServerException>(() => db.Execute("CONFIG", "SET", "appendonly", "yes"));
```
kevin-montrose: Similarly, we should check the error messages are correct.

kevin-montrose (changes requested):
Couple nits, and some potential bugs.

libs/server/RuntimeServerConfig.cs
vazois: I had to add some extensions but now should be formally validated.

libs/server/RuntimeServerConfig.cs
vazois: I added the appropriate Getter for this and also a collection of flags to indicate differents it can be retrieved as, i.e. TimeSpan, Int seconds, Int Milliseconds, etc.

badrishc (approved):

kevin-montrose (approved):

Mathos1432 (approved):

PR #1966, opened 2026-07-22
Fix BITOP failures in MULTI/EXEC transactions (#1954)

### LINKED ISSUE #1954: `BITOP` failures in `MULTI`/`EXEC` transactions
### Describe the bug

Running `BITOP` commands in transactions fails under certain circumstance.

Two different symptoms, one for `BITOP|NOT` which appears to validation based in (in `NetworkSKIP`):

```csharp
[Order(42)]
[Test]
[Category("BITFIELD")]
public async Task BitopNotTransactionAsync()
{
    const string Key = nameof(BitopNotTransactionAsync);

    var db = redis.GetDatabase();

    _ = await db.StringSetAsync(Key, new byte[] { 0 }).ConfigureAwait(false);

    // BITOP in transaction
    {
        var trans = db.CreateTransaction();
        var opRes = trans.ExecuteAsync("BITOP", "NOT", Key, Key).ConfigureAwait(false);

        _ = await trans.ExecuteAsync();

        _ = await opRes;
    }

    // Check result
    var res = (byte[])await db.StringGetAsync(Key).ConfigureAwait(false);
    ClassicAssert.IsTrue(new byte[] { 255 }.SequenceEqual(res));
}
```

This fails with `EXECABORT Transaction discarded because of previous errors.`.

For binary ops (`BITOP|AND` in this case) we hit an assert failure:

```csharp
[Order(43)]
[Test]
[Category("BITFIELD")]
public async Task BitopAndTransactionAsync()
{
    const string Key = nameof(BitopNotTransactionAsync);
    const string KeyA = nameof(BitopNotTransactionAsync) + "A";
    const string KeyB = nameof(BitopNotTransactionAsync) + "B";

    var db = redis.GetDatabase();

    _ = await db.StringSetAsync(KeyA, new byte[] { 0b1001_1001 }).ConfigureAwait(false);
    _ = await db.StringSetAsync(KeyB, new byte[] { 0b0111_1110 }).ConfigureAwait(false);

    // BITOP in transaction
    {
        var trans = db.CreateTransaction();
        var opRes = trans.ExecuteAsync("BITOP", "AND", Key, KeyA, KeyB).ConfigureAwait(false);

        _ = await trans.ExecuteAsync();

        _ = await opRes;
    }

    // Check result
    var res = (byte[])await db.StringGetAsync(Key).ConfigureAwait(false);
    ClassicAssert.IsTrue(new byte[] { 0b0001_1000 }.SequenceEqual(res));
}
```

Assert is `Attempting to use a non-XLocked key in a Transactional context (requesting XLock): XLocked False, Slocked False`.

### Expected behavior

Both tests should pass - and both do if you remove the transactional logic.

### Screenshots

_No response_

### Release version

Latest `main` (`2056547502693ad822f3531cf818849a17fcacc9`)

### IDE

_No response_

### OS version

_No response_

### Additional context

_No response_

vazois:
## Description

Fixes #1954 — `BITOP` commands fail inside `MULTI`/`EXEC` transactions.

`BITOP` is parsed as a subcommand-style command: the operation token (`AND`/`OR`/`XOR`/`NOT`/`DIFF`) is consumed by the parser, so `parseState` starts at the destination key. However, its command metadata (`Arity = -4`, key-spec indices `2`/`3`) describes the flat form where the operation token is still present, and `NormalizeForACLs` maps `BITOP_*` to the parent `BITOP` (whose `IsSubCommand` is `false`). The transaction path never accounted for the consumed token, causing two failures:

- **`BITOP NOT` → `EXECABORT`**: the queue-time arity check in `NetworkSKIP` computed a minimum of 3 args instead of 2, wrongly rejecting the valid `BITOP NOT dst src` (2 args).
- **`BITOP AND` (binary ops) → transactional-lock assert**: `LockKeys` resolved key-spec indices without the subcommand offset, so the destination key was never `XLock`ed (`Attempting to use a non-XLocked key in a Transactional context`).

## Fix

Apply the same BITOP special-case already used by cluster slot verification (`RespServerSessionSlotVerify.cs`) to the transaction path:

- `TxnRespCommands.NetworkSKIP`: treat `cmd == RespCommand.BITOP` as a subcommand when computing arity, and pass the flag through to key locking.
- `TxnKeyManager.LockKeys`: accept an explicit `isSubCommand` and forward it to `TryGetKeySearchArgsFromSimpleKeySpec`, which already applies the correct `-2` key-index offset.

## Tests

Added the two repro tests from the issue (`BitopNotTransactionAsync`, `BitopAndTransactionAsync`) to `GarnetBitmapTests`. Both fail without the fix and pass with it.


test/standalone/Garnet.test.complexstring/GarnetBitmapTests.cs
```diff
@@ -1067,6 +1068,64 @@ public void BitOp_OverflowValues_StableUnderGCCompaction(
             }
         }
 
+        // Regression tests for https://github.com/microsoft/garnet/issues/1954
```
kevin-montrose: nit: not much point in this comment, we can always just blame for history - and since there's more than 1 test it's not obvious when they _stop_ coming from that issue.

kevin-montrose (approved):

kevin-montrose (approved):

PR #1983, opened 2026-07-23
Deterministic repro: vectorset cleanup/recycle race on restore

tiagonapoli:
## What

Adds a **deterministic reproduction** of the flaky `RespVectorSetTests.RecreateIndexesOnRestoreAsync` failure (originally seen on PR #1967 / CI job 89250167325). No fix yet — the repro is committed `[Ignore]`d so CI stays green.

## Root cause

Garnet uses a **single unified store**; a vector set's index-config key, element records, and per-context metadata all recover from **one consistent checkpoint**. The bug is **not** a cross-store checkpoint skew.

The background cleanup scan (`PostDropCleanupFunctions.Reader`) deletes element records matched by **context-id namespace only, with no generation/epoch tag**, and context ids are **recycled**. In steady state a reuse barrier keeps this safe (a context isn't freed until its cleanup scan completes). **Recovery resets that barrier**: `ResumePostRecovery` re-arms cleanup for any in-use context whose index record was not recovered and pumps the cleanup task — which can then delete the records of a **newer, live generation** that recycled the same context id. Manifests as lost elements / `ERR asked quantization mismatch` / an `AccessViolationException` on recreate.

Empirically confirmed: a delay sweep showed delaying the cleanup scan makes the failure vanish; a TLA+ model reproduced the invariant violation.

## Changes

- `VectorManager.Cleanup.cs` — two test-only hooks: `TestOnlyRequeueCleanupForContext` (mimics the recovery re-arm) and `TestOnlyIsCleaningUp` (await the scan).
- `RespVectorSetTests.cs` — `RequeuedCleanupOnLiveContextLosesElementsOnRestoreAsync`: VADD 2 elements, requeue cleanup on the **live** context, save + recover, then assert the elements survive. Fails today with the exact CI signature (`ERR asked quantization mismatch with existing vector set`), 2/2 fixtures, deterministically.

## Fix direction (follow-up)

Make cleanup **generation-safe** by tagging element records with a per-context generation so a cleanup pass only deletes the exact generation it was armed for. Un-ignore the test once landed.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
Copilot-Session: 50b7a035-5389-46a7-8423-7806ef1635c3

### DIFF
--- a/libs/server/Resp/Vector/VectorManager.Cleanup.cs
+++ b/libs/server/Resp/Vector/VectorManager.Cleanup.cs
@@ -370,6 +370,42 @@ namespace Garnet.server
         }
 
         /// <summary>
+        /// Test-only hook that reproduces the recovery re-arm in <see cref="ResumePostRecovery"/>
+        /// (see the "mark for cleanup" loop at its end): mark an in-use context for cleanup and
+        /// pump the background cleanup task. Recovery does exactly this for any context still
+        /// marked in-use whose index record was not recovered; because context ids are recycled
+        /// and the cleanup scan matches records by context namespace only (no generation tag),
+        /// this can delete the records of a newer, live generation that recycled the same id.
+        /// </summary>
+        public void TestOnlyRequeueCleanupForContext(ulong context)
+        {
+            var (contextIndex, contextValue) = ContextMetadata.DecomposeContext(context & ~(ContextStep - 1));
+
+            lock (this)
+            {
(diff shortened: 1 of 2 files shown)


tiagonapoli:
Closing: reverting to an empirical approach — will add logging, reproduce the real failure ordering in Docker, and then force that exact ordering with injection points instead of synthetic hooks.

PR #1995, opened 2026-07-29

tiagonapoli:
## Problem

A Vector Set index record persists the primary's native DiskANN handle (`IndexPtr`) inside its value. Diskless sync streams log records verbatim into the replica's memory, so the replica receives the **primary's** pointer.

The only hook that zeroes that field is `GarnetRecordTriggers.OnDiskRead`, which by definition never fires for records streamed straight into memory. Since `VectorManager.NeedsRecreate` is exactly `indexPtr == 0`, a foreign pointer is indistinguishable from a healthy local one: the lazy `Service.RecreateIndex` rebuild is skipped and the raw value reaches the P/Invoke unvalidated. In production this faults the replica with SIGSEGV inside `NativeDiskANNMethods.card`.

## Fix

`NetworkClusterSync` clears the pointer as records are deserialized, so the replica rebuilds its own index on first touch.

## Tests

Adds `Garnet.test.cluster.replication.vectorsets`. The shared harness asserts two independent things, because either can hold while the other is broken:

1. **Index ownership** — a node may only dereference a DiskANN handle it allocated itself.
2. **Element-level equality** — every element VADDed on the source exists on the target with a byte-identical embedding, cardinality and dimensionality agree, and VSIM is navigable. Matching cardinality alone is far too weak: an aliased index reports the *source's* count perfectly well.

The pre-existing cluster fixture cannot catch this. It attaches replicas before writing, so each replica builds its own index from replicated VADD payloads; the defect needs an **already-populated** Vector Set to travel over a full sync.

Covers diskless full sync, failover promotion, and async replay. The diskless tests are the executable form of the TLA+ model in `tla/VectorIndexLifetime.tla`, whose counterexamples all share one 6-step trace.

## Validation

- 12/12 pass with the fix.
- **5/12 fail without it** (`VectorSetReadableOnReplicaAfterDisklessFullSync`, `VectorSetReplicatedToEveryReplicaByDisklessFullSync`, `VectorSetsStayPartitionedAcrossDisklessFullSync`, `VectorSetSurvivesFailoverAfterDisklessFullSync`, `VectorSetReplicatedUnderAsyncReplayAfterFullSync`).
- Regressions green: `Garnet.test.cluster.vectorsets` 80/80, `replication.disklesssync` 28/28, `replication` 103/103.

> First of three stacked PRs. #2 makes recovery re-entrant; #3 fixes AOF replay-channel shutdown.

### DIFF
--- a/test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetAsyncReplayTests.cs
+++ b/test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetAsyncReplayTests.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System.Linq;
 using NUnit.Framework;
 using NUnit.Framework.Legacy;
 
@@ -30,9 +29,8 @@ namespace Garnet.test.cluster
                 replayTaskCount: replayTaskCount,
                 vectorSetReplayTaskCount: vectorSetReplayTaskCount,
                 timeout: timeout);
-            context.CreateConnection();
 
-            context.clusterTestUtils.FormCluster(PrimaryIndex, [.. Enumerable.Range(1, nodeCount - 1)], context.logger);
+            FormClusterAllNodes(nodeCount);
         }
 
         /// <summary>Baseline: asynchronously replayed VADDs must land with identical embeddings.</summary>
--- a/test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetAsyncReplayTests.cs
+++ b/test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetAsyncReplayTests.cs
@@ -1,9 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Extensions.Logging;
 using NUnit.Framework;
 using NUnit.Framework.Legacy;
 
@@ -19,13 +17,6 @@ namespace Garnet.test.cluster
     {
         private const int PrimaryIndex = 0;
         private const int ReplicaIndex = 1;
-
-        protected override Dictionary<string, LogLevel> MonitorTests => new()
-        {
-            [nameof(VectorSetReplicatedUnderAsyncReplay)] = LogLevel.Trace,
-            [nameof(VectorSetReplicatedUnderAsyncReplayAfterFullSync)] = LogLevel.Trace,
-            [nameof(VectorSetReplicatedUnderParallelAsyncReplay)] = LogLevel.Trace,
-        };
 
         private void SetupAsyncReplayCluster(int nodeCount, bool disklessSync, int replayTaskCount = 1, int vectorSetReplayTaskCount = 0)
         {
@@ -47,7 +38,6 @@ namespace Garnet.test.cluster
         /// <summary>Baseline: asynchronously replayed VADDs must land with identical embeddings.</summary>
         [Test]
         [Category("REPLICATION")]
-        [CancelAfter(180_000)]
         public void VectorSetReplicatedUnderAsyncReplay()
         {
             const string Key = "{vsdisk}async";
@@ -69,7 +59,6 @@ namespace Garnet.test.cluster
         /// </summary>
         [Test]
         [Category("REPLICATION")]
-        [CancelAfter(180_000)]
         public void VectorSetReplicatedUnderAsyncReplayAfterFullSync()
         {
             const string Key = "{vsdisk}asyncfullsync";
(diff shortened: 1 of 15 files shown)


test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
```diff
@@ -0,0 +1,306 @@
...
+            AssertFullyReplicated(PrimaryIndex, ReplicaIndex, Key);
+        }
+
+        /// <summary>
+        /// Multiple sets can make a foreign handle point at the wrong live index instead of faulting.
+        /// Different sizes plus element checks expose cross-set substitution.
+        /// </summary>
+        [Test]
+        [Category("REPLICATION")]
+        [CancelAfter(180_000)]
```
tiagonapoli: too much time

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
```diff
@@ -0,0 +1,306 @@
...
+            AssertFullyReplicated(PrimaryIndex, ReplicaIndex, Key);
+        }
+
+        /// <summary>
+        /// Every replica takes its own diskless full sync from the same populated primary.
+        /// Each must rebuild its own index, not share per-primary state.
+        /// </summary>
+        [Test]
+        [Category("REPLICATION")]
+        [CancelAfter(180_000)]
```
tiagonapoli: too much time - all of these using 180s is too much

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
```diff
@@ -0,0 +1,306 @@
...
+    /// Executable form of the model in tla/VectorIndexLifetime.tla.
+    /// </summary>
+    [TestFixture]
+    [NonParallelizable]
+    public class ClusterVectorSetDisklessSyncTests : VectorSetReplicationTestBase
+    {
+        private const int PrimaryIndex = 0;
+        private const int ReplicaIndex = 1;
+
+        protected override Dictionary<string, LogLevel> MonitorTests => new()
```
tiagonapoli: Needed? Can't we have defualt for all somehow instead of listing each individual test?

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetAsyncReplayTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            MakeReadable(ReplicaIndex);
+            AssertFullyReplicated(PrimaryIndex, ReplicaIndex, Key);
+        }
+
+        /// <summary>
+        /// Multiple replay tasks apply VADDs for different keys concurrently; each key must keep exactly its own elements.
+        /// </summary>
+        [Test]
+        [Category("REPLICATION")]
+        [CancelAfter(180_000)]
```
tiagonapoli: too much time, all these tets using 3min, no

tiagonapoli: Removed here too - no more `[CancelAfter(180_000)]`, 60s base timeout instead.

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
tiagonapoli: Good call - replaced the per-test dictionary with `protected virtual LogLevel MonitorLogLevel => LogLevel.Trace` on `VectorSetReplicationTestBase`. Setup now builds the dictionary for whatever the current test is, so no fixture has to enumerate its tests. All three `MonitorTests` overrides are gone.

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
tiagonapoli: Removed all the 180s `[CancelAfter]` attributes in this fixture; the 60s base timeout applies uniformly now.

test/cluster/Garnet.test.cluster.replication.vectorsets/ReplicationTests/ClusterVectorSetDisklessSyncTests.cs
tiagonapoli: Dropped the per-test `[CancelAfter(180_000)]`. Timeout now comes from the base setup (`testTimeout` = 60s). Full suite runs in 82s total (~7s/test), so 60s is plenty of headroom.

tiagonapoli:
Closing while I complete a self-review pass in my fork (tiagonapoli/garnet#4). Will reopen a fresh upstream PR once it's review-ready.

PR #1996, opened 2026-07-29
Make Vector Set recovery re-entrant so disk-based full sync sanitizes handles

tiagonapoli:
> **Stacked on #1** — review that one first; this PR targets `tiagonapoli/vs-1-diskless`, so its diff shows only this change.

## Problem

`VectorManager` collects the contexts it sees during recovery into `recoveredIndexes` / `recoveredMetadata`, and `ResumePostRecovery` reconciles them against the recovered store. Both were **released** at the end of that method, on the assumption that a process recovers exactly once.

A replica does not. It re-recovers the store on **every disk-based full sync**, and nothing re-ran `ResumePostRecovery` on that path.

So the second recovery threw `NullReferenceException` out of `RecoveredVectorSetIndexKey`. That exception escaped `ClearBitsOnPage` — the per-record walk that fires `OnDiskRead` — so **the walk aborted after the first record on the page**. Every Vector Set record behind it kept the primary's raw DiskANN handle.

Instrumented recovery of three vector sets, before the fix:

```
CBOP page=0 startLA=0 endLA=16777216 until=156296 endOff=156296 snapFrom=64
  rec off=64   size=8224  ->  msf_            (context metadata)
  rec off=8288 size=88    ->  {vsdisk}dblarge (sanitized OK)
  EX System.NullReferenceException
     at VectorManager.RecoveredVectorSetIndexKey(LogRecord&) VectorManager.cs:384
     at GarnetRecordTriggers.OnRecoverySnapshotRead(LogRecord&)
     at TsavoriteKV.ClearBitsOnPage(...) Recovery.cs:1259
```

`dbsmall` and `dbmedium` were never visited, and came up on the replica holding `IndexPtr == primary's pointer`.

The half-finished recovery also leaked the checkpoint file handle, which is what made teardown hang ~60s deleting the directory. The disk-based fixture went **1m07s → 7s**.

**This path matters more than the diskless one in practice, since `ReplicaDisklessSync` defaults to `false`.**

## Fix

- Recovery state is **cleared** rather than released, so a second recovery starts from a clean slate instead of a null reference.
- A replica's disk-based recovery reconciles it via the new `StoreWrapper.RecoverVectorSets()`, mirroring what the startup path already did.

## Validation

- 16/16 pass with the fix.
- **All 4 disk-based tests fail without it.**
- Regressions green: `Garnet.test.cluster.vectorsets` 80/80, `replication.disklesssync` 28/28, `replication` 103/103.

### DIFF
--- a/libs/cluster/Server/Replication/ReplicaOps/ReplicaDiskbasedSync.cs
+++ b/libs/cluster/Server/Replication/ReplicaOps/ReplicaDiskbasedSync.cs
@@ -320,8 +320,9 @@ namespace Garnet.cluster
                 logger?.LogInformation("Initializing AOF");
                 storeWrapper.appendOnlyFile.Log.Initialize(beginAddress, recoveredReplicationOffset);
 
-                // Before we can use the replication offset, we must wait for queued Vector Set ops to complete
-                storeWrapper.DefaultDatabase.VectorManager?.WaitForVectorOperationsToComplete();
+                // The store was recovered again, so recovered Vector Set contexts need reconciling just
+                // as they are on the startup path.
+                storeWrapper.RecoverVectorSets();
 
                 // Before we can use the replication offset, we must wait for queued Vector Set ops to complete
                 storeWrapper.DefaultDatabase.VectorManager?.WaitForVectorOperationsToComplete();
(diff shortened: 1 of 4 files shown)


tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/14

PR #1997, opened 2026-07-29
Keep the Vector Set replay channel alive across AofProcessor disposal

tiagonapoli:
> **Stacked on #1996** (which is stacked on #1995) — review those first; this PR targets `tiagonapoli/vs-2-recovery-reentrancy`, so its diff shows only this change.

## Problem

`AofProcessor.Dispose()` called `VectorManager.ShutdownReplayTasks()`, which is a **permanent** shutdown — it completes the replication replay channel writer and parks the replay tasks for good.

But the `VectorManager` belongs to the **database**, not to the processor, and a node constructs and disposes several `AofProcessor`s over its lifetime. `SingleDatabaseManager.ReplayAOF` builds a transient one during startup AOF recovery.

So on any node started with `Recover` enabled — a normal production replica config — the channel was already closed before the node began serving. From that point `HandleVectorSetAddReplication` could no longer queue:

```csharp
var queued = replicationReplayChannel.Writer.TryWrite(...);
if (!queued) replicationBlockEvent.Decrement();   // no log, no error
```

It **discards whatever it cannot queue, silently**. The node dropped every VADD replicated to it while continuing to report healthy AOF offsets — silent data loss with no signal at all.

## Fix

`Dispose()` now only drains. `VectorManager.Dispose()` still performs the real shutdown when the database goes away.

## Tests

Adds `ClusterVectorSetRecoveryTests`, which restarts the replica, the primary, and both nodes with `tryRecover: true` — the configuration that exposed this.

Note these tests also depend on #1996: without it, an aborted recovery leaves the store corrupt enough that a subsequent read dies with `AccessViolationException` in `TsavoriteKV.TraceBackForKeyMatch`, taking down the test host. That is why this is the top of the stack.

## Validation

- 19/19 pass with the fix (full suite, top of stack).
- **All 3 recovery tests fail without it.**
- Regressions green: `Garnet.test.cluster.vectorsets` 80/80, `replication.disklesssync` 28/28, `replication` 103/103. `ShutdownReplayTasks()` is left public but now has no callers.

### DIFF
--- a/libs/server/AOF/AofProcessor.cs
+++ b/libs/server/AOF/AofProcessor.cs
@@ -152,8 +152,10 @@ namespace Garnet.server
         /// </summary>
         public void Dispose()
         {
+            // Drain but don't shut down: the VectorManager belongs to the database, and startup AOF
+            // recovery builds a transient AofProcessor. Completing the replay channel here would make
+            // HandleVectorSetAddReplication silently drop every VADD replicated afterwards.
             activeVectorManager?.WaitForVectorOperationsToComplete();
-            activeVectorManager?.ShutdownReplayTasks();
             activeRangeIndexManager?.DisposeIncompleteStreamReassembly();
             aofReplayCoordinator?.Dispose();
         }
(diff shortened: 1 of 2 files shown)


tiagonapoli:
Moving this to my fork: https://github.com/tiagonapoli/garnet/pull/13

PR #2002, opened 2026-07-30
Update Vector Set docs to match current behavior

tiagonapoli:
## What

The Vector Set command documentation had drifted from the implementation. This updates `website/docs/commands/vector-sets.md` and one line of `website/docs/commands/api-compatibility.md` to match current behavior.

No product code is changed — documentation only.

## How this was verified

Every statement below was checked against a locally built server on `main` (`96d31943f`) started with `--enable-vector-set-preview`, driven over raw RESP — not inferred from reading code.

### Quantization

`Q8` and `BIN` became functional in #1882 ("Add quantization tasks and calls into DiskANN"), but the docs still described them as rejected.

| Documented | Actual |
|---|---|
| `Q8` / `BIN` "❌ Rejected — `ERR Unsupported quantization type`" | Both supported. That error string no longer exists anywhere in the codebase. |
| Default `Q8` "currently fails" | Works. |
| Quantizer list is `NOQUANT \| XPREQ8` | Also `Q8`, `BIN`, `XNOQUANT_U8`, `XNOQUANT_I8`, `XBIN_U8`, `XBIN_I8`; `XPREQ8` is a deprecated alias for `XNOQUANT_U8`. |
| `X*` quantizers "require `XB8` input" | They also accept `FP32` / `VALUES`, subject to byte-range validation. |
| `REDUCE` "not allowed with `XPREQ8`" | Rejected for all `XNOQUANT_*` / `XBIN_*`; works with `NOQUANT` / `Q8` / `BIN`. |

Also added the `VINFO quant-type` string each token reports, and a note that `Q8`/`BIN` quantization is applied **asynchronously** (background table build, then backfill) — which is why results on small or newly created indexes are identical to `NOQUANT`.

### Commands

| Documented | Actual |
|---|---|
| `VEMB ... RAW` "parsed but not yet implemented — invoking it currently throws" | Implemented. Documented the 3-element reply (4 for `Q8`) and that the norm/range entries are placeholders. |
| `VSIM` "currently only supports RESP2" | RESP3 works. Returns a map keyed by element ID when `WITHSCORES`/`WITHATTRIBS` are set, with scores as RESP3 doubles. |
| `VDIM` returns "the dimension before any `REDUCE`" | Returns the **reduced** dimension when `REDUCE` is configured. |
| `VINFO` `reduced-dimensions` "same as `input-vector-dimensions` if no projection" | Returns `0`. Both worked examples updated. |
| `TYPE key` "currently reports `string`" | Reports `vectorset`. |
| Filter compile errors "yield an empty result set rather than a server-side error" | Returns `ERR Compiling filter failed`. |

`api-compatibility.md`: dropped the now-inaccurate "(no `RAW`)" qualifier on `VEMB`.

## Known issues documented (not fixed here)

Two real behaviors are now called out in the docs so users stop being surprised by them. Both are pre-existing and unrelated to any change in this PR — happy to split them into separate issues.

1. **`VCARD` / `VISMEMBER` / `VLINKS` / `VRANDMEMBER` / `VSETATTR` close the connection.** The docs claimed they "return `+OK` regardless of arguments". They actually validate arguments and keys correctly — `:0` / nil for a missing key, `WRONGTYPE` for another type — but when invoked against an *existing* Vector Set key they reach the `default:` arm in `libs/server/Storage/Functions/MainStore/PrivateMethods.cs` and throw `ERR Garnet Exception: Unsupported operation on input`, terminating the client connection.

2. **`VADD` on an existing element ID duplicates instead of updating.** It returns `:1` (not `:0`), `VINFO` `size` grows by one, and `VSIM` can return the same element more than once. `VEMB` reads back the most recently written vector. Reproduces on every quantizer. The docs now recommend `VREM` before re-adding until this is fixed.

## Testing

Documentation-only change; no tests added. Verified by exercising the documented commands and replies against a live server as described above.

### DIFF
--- a/website/docs/commands/vector-sets.md
+++ b/website/docs/commands/vector-sets.md
@@ -120,7 +120,9 @@ Insert a vector with the given element ID into a Vector Set, creating the index 
 
 ```bash
 VADD key [REDUCE dim] (FP32 vector | XB8 vector | VALUES n v1 ... vN) element
-         [CAS] [NOQUANT | XPREQ8] [EF n] [SETATTR attr] [M n]
+         [CAS]
+         [NOQUANT | Q8 | BIN | XNOQUANT_U8 | XNOQUANT_I8 | XBIN_U8 | XBIN_I8]
+         [EF n] [SETATTR attr] [M n]
          [XDISTANCE_METRIC L2 | COSINE | IP | XCOSINE_NORMALIZED]
 ```
 
@@ -136,9 +138,9 @@ VADD key [REDUCE dim] (FP32 vector | XB8 vector | VALUES n v1 ... vN) element
 
 | Option | Default | Description |
 |--------|---------|-------------|
-| `REDUCE dim` | _disabled_ | Project the input vector down to `dim` dimensions. `dim` must be ≤ the input dimensions. Not allowed with `XPREQ8`. Only honored on the first `VADD` (when the index is created). |
+| `REDUCE dim` | _disabled_ | Project the input vector down to `dim` dimensions. `dim` must be ≤ the input dimensions. Not supported with the `XNOQUANT_*` / `XBIN_*` quantizers. Only honored on the first `VADD` (when the index is created). |
 | `CAS` | _off_ | Accepted for parser compatibility with Redis; currently a no-op. |
-| `NOQUANT` \| `XPREQ8` | _required_ | Quantization (see [Quantization](#quantization)). `Q8` and `BIN` are parsed but rejected with `ERR Unsupported quantization type`. |
+| _quantizer token_ | `Q8` | Quantization (see [Quantization](#quantization)). All of `NOQUANT`, `Q8`, `BIN`, and the Garnet `X*` extensions are supported. |
 | `EF n` | `200` | Build-time exploration factor (DiskANN `R` candidate-list size). Must be in `[1, 1000000]`. |
 | `SETATTR attr` | _none_ | Attach an arbitrary byte string to the element (typically a JSON object). Retrieve later with `VGETATTR` or via `WITHATTRIBS` on `VSIM`. |
 | `M n` | `16` | DiskANN max out-degree per node. Must be in `[4, 4096]`. |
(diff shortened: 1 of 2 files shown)


website/docs/commands/vector-sets.md
```diff
@@ -157,18 +159,25 @@ VADD movies FP32 <16-byte float blob> dune XDISTANCE_METRIC COSINE M 32
...
 #### Resp Reply
 
-- RESP2: integer `:1` if a new element was inserted, `:0` if the element was already present.
+- RESP2: integer reply. Currently always `:1` on a successful insert (see the warning below).
 - RESP3: boolean `true` / `false`.
 - Errors include `ERR Vector dimension mismatch`, `ERR REDUCE dimension must be <= vector dimensions`,
   `ERR M must be an integer between 4 and 4096`, `ERR EF must be an integer between 1 and 1000000`,
   `ERR vector exceeds maximum of 65536 dimensions`, `ERR Vector Set key cannot be empty`.
 
+:::warning
```
tiagonapoli, line 174: remove this comment

website/docs/commands/vector-sets.md
```diff
@@ -534,7 +594,20 @@ reserved for future implementation:
 | `VRANDMEMBER key [count]` | Return random element IDs. |
 | `VSETATTR key element attr` | Update an element's attribute in place. Today the only way to set/replace an attribute is to re-run `VADD ... SETATTR`. |
 
-Treat these as no-ops in preview builds — do not rely on their return value.
+:::warning
+Argument validation and key lookup are wired up, so these commands behave sensibly when the key is missing or
```
tiagonapoli, line 598: I don't know if this is really needed

badrishc (approved):

website/docs/commands/vector-sets.md
```diff
@@ -157,18 +159,25 @@ VADD movies FP32 <16-byte float blob> dune XDISTANCE_METRIC COSINE M 32
...
   `ERR M must be an integer between 4 and 4096`, `ERR EF must be an integer between 1 and 1000000`,
   `ERR vector exceeds maximum of 65536 dimensions`, `ERR Vector Set key cannot be empty`.
 
+:::warning
+Re-adding an element ID that is already present does **not** currently replace it in place. `VADD` returns `:1`
+rather than `:0`, the index `size` grows by one, and the element can be returned more than once by `VSIM`.
+`VEMB` reads back the most recently written vector. Until this is fixed, issue `VREM key element` before
+re-adding an existing element ID.
+:::
+
```
tiagonapoli, line 180: rmeove this

website/docs/commands/vector-sets.md
```diff
@@ -157,18 +159,25 @@ VADD movies FP32 <16-byte float blob> dune XDISTANCE_METRIC COSINE M 32
 VADD movies VALUES 3 0.12 0.34 0.56 inception \
      NOQUANT SETATTR "{\"year\":2010,\"rating\":4.2}" XDISTANCE_METRIC COSINE
 
-# Insert a uint8 vector with the XPREQ8 pseudo-quantizer
-VADD photos XB8 <128-byte blob> photo:42 XPREQ8
+# Insert a uint8 vector with the XNOQUANT_U8 pseudo-quantizer
+VADD photos XB8 <128-byte blob> photo:42 XNOQUANT_U8
```
tiagonapoli, line 163: why changed?

website/docs/commands/vector-sets.md
tiagonapoli, line 180: remove these chagnes

PR #2009, opened 2026-08-03
Fix LREM with count = int.MinValue killing the session

hexonal:
`LREM key -2147483648 <element>` terminates the RESP session.

```
RPUSH L a b c
LREM L -2147483648 a    -> connection closed, no reply
PING                    -> Connection reset by peer
```

Server log:

```
crit: Session[0] ProcessMessages threw an exception: System.OverflowException:
Negating the minimum value of a twos complement number is invalid.
   at Garnet.server.ListObject.ListRemove(ObjectInput&, ObjectOutput&)
      in libs/server/Objects/List/ListObjectImpl.cs
   at Garnet.server.ListObject.Operate(...)
```

The server process stays up and other sessions are unaffected — the blast radius is the issuing connection.

### Cause

`ListRemove` normalizes a negative count with `Math.Abs(count)`, which throws for `int.MinValue` because `|int.MinValue|` does not fit in an `int`.

### Fix

Clamp rather than negate. The removal loop is `while (removedCount < count && currentNode != null)`, and `removedCount` can never exceed the list length, which cannot exceed `int.MaxValue` — so `int.MaxValue` removes exactly the same elements that a correctly negated `2147483648` would. Direction is unaffected: `fromHeadToTail` is decided from the original sign before the clamp.

### Scope

I checked the other `Math.Abs` call sites that take a client-supplied count — `HRANDFIELD` (`HashObjectImpl.cs:138`), `SRANDMEMBER` (`SetObjectImpl.cs:219`) and `ZRANDMEMBER` (`SortedSetObjectImpl.cs:651,656`) — by driving each with `-2147483648` against a live server. None of them terminates the session, so `LREM` is the only instance and the production change stays at one line.

### Testing

- New test `LREMWithIntMinValueCountRemovesAllMatches`: fails without the change with `RedisConnectionException : SocketClosed (ReadEndOfStream)`, passes with it.
- `RespListTests`: **104/104**.
- Semantics spot-checked against a live server: `LREM key -2147483648 a` on `a b a c a` returns `3` and leaves `b c`; `LREM key -2 a` still returns `2` and leaves `a b c` (tail-first order preserved); `LREM key 2147483647 a` unchanged.
- `dotnet format --verify-no-changes` clean on both files.
- macOS arm64, net10.0 Release. `net8.0` builds clean but its tests were not run here — no net8.0 runtime in this environment.


hexonal:
The red check on `Garnet Standalone (ubuntu-latest, net10.0, Debug, Garnet.test.collections)` looks unrelated to this change — happy to be corrected.

19 tests failed: 18 in `RespHashTests` plus `RespListGarnetClientTests.AddElementsToListTail_WithCallback`. All of them fail with the same connection error rather than an assertion, e.g.

```
RedisConnectionException : SocketClosed (ReadEndOfStream, last-recv: 0) on 127.0.0.1:34400/Interactive,
Idle/MarkProcessed, last: HMSET, origin: ReadFromPipe
```

with `last:` being `HMSET` / `HPEXPIRE` / `HRANDFIELD` — hash commands. This PR only touches `libs/server/Objects/List/ListObjectImpl.cs` and `RespListTests.cs`, and no `RespHashTests` failed on its own merits.

The branch sits on the current main tip (26b1882) with no other commits, and main's own runs are green. Running the full `Garnet.test.collections` project locally on this branch gives 747/747 passed, 0 failed, on net10.0 — including all 18 of those hash tests.

I do not have permission to re-run the job. Could someone re-trigger it?

hexonal:
Rebased onto current main. The branch was 2 commits behind and GitHub had it as `BEHIND`; it's now up to date. The patch is unchanged — `diff` of the pre- and post-rebase commit is byte-identical, and neither of the two new upstream commits touches `ListObjectImpl.cs`.

Good news on the `Garnet.test.collections` check I asked about earlier: the rebase re-triggered it, and **it is now green in all 8 configurations**, including the `ubuntu-latest, net10.0, Debug` one that had failed. So that was flaky and has cleared itself — no re-run needed, and thanks for not having to chase it.

One different check went red on this run, `Garnet Standalone (ubuntu-latest, net10.0, Release, Garnet.test.scripting)`. I don't think it's this PR either, and here's what I checked rather than just asserting it:

- The 4 failures are all in `LuaScriptTests`: `RedisLog`, `RedisPCall`, `RedisCallErrors`, `RedisDebugAndBreakpoint` — all assertions about `redis.log()` / `redis.debug()` / `redis.breakpoint()` argument handling. This PR's diff is a single line clamping the LREM `count` argument.
- The other **7 of 8** `Garnet.test.scripting` configurations passed on this same commit.
- Running exactly those 4 tests locally on this branch (`-c Release -f net10.0`) gives 26/26 passed.

Happy to rebase again for a fresh run if that's easier than a re-trigger.

kevin-montrose (approved):

PR #2015, opened 2026-08-04
Harden LightEpoch: make the epoch announce part of the slot-claim CAS

tiagonapoli:
**Note: this does not affect production use of Garnet. Tsavorite's CRUD paths (Read/Upsert/RMW/Delete) already issue interlocked operations and memory barriers around epoch protection, so the required ordering is established independently and the scenario below cannot occur there. This change hardens `LightEpoch` itself for non-Tsavorite-core paths that use it directly, without relying on the caller's barriers.**

## Summary

A thread entering a protected region announces its epoch with a **plain store**, which is not ordered against the reclaimer's later **load** of the same slot. A reclaimer can scan a live reader's slot, see it as free, raise `SafeToReclaimEpoch` past the reader's epoch, and free a page the reader is about to dereference.

The fix changes **which word the claim CAS operates on - localCurrentEpoch instead of threadId**. No barrier and no atomic is added — the `lock cmpxchg` was already there, it just lands on a different word.

Scoped to **x86-64**. ARM64 additionally need more fixes; that ordering work, along with herd7 models and TLA+ specs, is deliberately left out of this change.

## The bug

```csharp
ReserveEntryForThread(ref entry);                              // CAS: threadId 0 -> tid
(*(tableAligned + entry)).localCurrentEpoch = CurrentEpoch;    // plain store: the announce
```

Store-then-load-of-another-address is the one reordering TSO permits:

| Reader | Reclaimer |
| --- | --- |
| CAS `threadId` = tid | |
| store `localCurrentEpoch` = 5 *(buffered)* | |
| | scan slot → reads **0**, treats it as free |
| | `SafeToReclaimEpoch` = 5, free the page |
| load page → **use-after-free** | |

## The fix

The CAS writes `localCurrentEpoch` directly, so claiming the slot and announcing the epoch are one locked RMW

```csharp
if (Interlocked.CompareExchange(ref (tableAligned + entry)->localCurrentEpoch, epoch, 0) != 0)
    return false;

// The slot is now exclusively ours, so threadId needs no interlocked write.
(*(tableAligned + entry)).threadId = Metadata.threadId;
```

### 8-hour hardware run

Two Azure VMs, quarantine litmus, baseline vs fixed (2026-07-30, westus2, x86-64):

| impl | violations | sampled rounds |
| --- | --- | --- |
| baseline | **3,680** | 8.97 × 10⁷ |
| this fix | **0** | 5.60 × 10⁸ |

### Local repro

`--buggy` points the harness at `BuggyLightEpoch`, a frozen copy of `main`'s version, so both arms run back to back on one machine (20 logical processors, x86-64, 15 s runs):

| arm | run 1 | run 2 | run 3 | run 4 | exit |
| --- | ---: | ---: | ---: | ---: | --- |
| fixed | 0 | 0 | 0 | 0 | 0 (pass) |
| buggy | 34 | 77 | 131 | 245 | 1 (violation) |

## Benchmarks

BDN `RawStringOperations`, 3 interleaved runs per arm on a dedicated x86 VM (Xeon 8272CL, 16 vCPU): mean |Δ| **1.46%**, 5 faster / 5 slower, 0 B allocated in both arms.

`GetNotFound` had the widest spread, so it was re-run alone with 8 interleaved runs per arm:

| arm | mean | median | min | max |
| --- | ---: | ---: | ---: | ---: |
| baseline | 6.960 µs | 6.848 | 6.519 | 7.761 |
| this fix | 6.846 µs | 6.826 | 6.628 | 7.104 |

The fix is 1.65% faster, but the baseline's own run-to-run spread is far larger than that — it's host noise, not a real effect. **No measurable cost.**


tiagonapoli:
## Epoch BDN results — no regression

Ran `EpochTests` (`BDN-Tsavorite.Benchmark`) on a dedicated Azure `Standard_D16ds_v5` (Ice Lake), Ubuntu 24.04, .NET 10.0.302, `DOTNET_TieredPGO=0`, `taskset`-pinned. Baseline `main` @ `4c004e2a` vs PR `6e72265e`, 5 rounds alternating main/PR, fresh process each round.

| Benchmark | `main` mean | PR mean | Δ | baseline spread | Verdict |
|---|---:|---:|---:|---:|---|
| `ResumeSuspend` | 8.768 ms | 8.781 ms | +0.15% | 0.30% | noise |
| `ProtectAndDrain` | 1.439 ms | 1.225 ms | −14.9% | 0.07% | artifact — pinned-heap placement, reproduces  `main` |
| `BumpCurrentEpoch` | 18.837 ms | 18.892 ms | +0.29% | 0.68% | noise |

For `ResumeSuspend`/`BumpCurrentEpoch` the two arms' ranges fully overlap and the baseline alone wanders as much as the reported delta. Same 5-round protocol against the merge-base (`c6cc866b`) gave the same picture.

### Bottom line

**No regression** on any of the three epoch benchmarks. `ResumeSuspend` and `BumpCurrentEpoch` are flat within noise, and `ProtectAndDrain` is genuinely unchanged — its apparent 15% gain is an allocator-placement artifact I reproduced on unpatched `main`, so please don't read it as a win either.

badrishc (approved):

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
tiagonapoli:
```suggestion
            var epoch = Volatile.Read(CurrentEpoch);
            (*(tableAligned + entry)).localCurrentEpoch = epoch;
```

tiagonapoli: Remove the ReserveEntry method and inline it here

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -518,16 +518,8 @@ void Acquire()
             Debug.Assert(entry == kInvalidIndex,
                 "Trying to acquire protected epoch. Make sure you do not re-enter Tsavorite from callbacks or IDevice implementations. If using tasks, use TaskCreationOptions.RunContinuationsAsynchronously.");
 
-            // Reserve an entry in the epoch table for this thread
```
tiagonapoli, line 521: Readd this comment:
// Reserve an entry in the epoch table for this thread and set localCurrentEpoch

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -291,9 +291,20 @@ public bool TrySuspend()
...
+        /// <remarks>
+        /// <para><see cref="ComputeNewSafeToReclaimEpoch"/> derives <see cref="SafeToReclaimEpoch"/> from the
+        /// minimum announced epoch across occupied slots, so a thread that holds its slot without refreshing
+        /// pins that minimum at the epoch it entered on. That stalls reclamation and every pending drain
+        /// action process-wide, not just this thread's. Refreshing is how a long-lived holder lets the minimum
+        /// advance, which is why loops that hold the epoch across many operations must call this periodically.</para>
+        /// <para>Announcing a newer epoch relinquishes protection for the previous one. Callers may therefore
+        /// only refresh at a point where they hold no references acquired under the epoch they were
+        /// previously announcing.</para>
+        /// </remarks>
```
tiagonapoli: less verbose

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -302,13 +313,16 @@ public void ProtectAndDrain()
             Debug.Assert(entry > 0, "Trying to refresh unacquired epoch");
             DebugAssertEpochAcquired(entry);
 
-            // Protect CurrentEpoch by copying it to the instance-specific epoch table
-            // so that ComputeNewSafeToReclaimEpoch() will see it.
-            (*(tableAligned + entry)).localCurrentEpoch = CurrentEpoch;
+            // Announce the epoch this thread is now on, so ComputeNewSafeToReclaimEpoch() stops holding
+            // SafeToReclaimEpoch down at the epoch previously announced here. The read is volatile because a
+            // cached CurrentEpoch would re-announce a stale epoch, which is safe but defeats the purpose of
+            // the refresh.
```
tiagonapoli: 1 line comment is enough - just say refresh current epoch

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.TestHooks.cs
```diff
@@ -20,18 +22,20 @@ public sealed unsafe partial class LightEpoch
...
         /// </summary>
-        internal long TestHookAnnouncedEpochAt(int entry) => EntryAt(entry).localCurrentEpoch;
+        internal long TestHookAnnouncedEpochAt(int entry) => Volatile.Read(ref EntryAt(entry).localCurrentEpoch);
 
         /// <summary>
         /// The thread id recorded in epoch table slot <paramref name="entry"/>, or 0 if the slot is free.
+        /// Read volatile: callers observe slots owned by other threads.
         /// </summary>
-        internal int TestHookThreadIdAt(int entry) => EntryAt(entry).threadId;
+        internal int TestHookThreadIdAt(int entry) => Volatile.Read(ref EntryAt(entry).threadId);
```
tiagonapoli: tests should probably not add memory barriers

playground/LightEpochLitmus/helpers/BuggyLightEpoch.cs
```diff
@@ -908,7 +908,8 @@ struct EpochActionPair
         /// <summary>
         /// The epoch announced in epoch table slot <paramref name="entry"/>, or 0 if the slot is free.
         /// Mirrors the test hook of the same name on the fixed epoch so the harness can drive both.
+        /// Read volatile: the harness disturber threads poll slots owned by other threads.
         /// </summary>
-        internal long TestHookAnnouncedEpochAt(int entry) => (*(tableAligned + entry)).localCurrentEpoch;
+        internal long TestHookAnnouncedEpochAt(int entry) => Volatile.Read(ref (*(tableAligned + entry)).localCurrentEpoch);
```
tiagonapoli: tests should probably not add memory barriers

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -291,9 +291,13 @@ public bool TrySuspend()
         }
 
         /// <summary>
-        /// Enter the thread into the protected code region
+        /// Refresh this thread's announced epoch and run the drain actions that thereby become safe. The
+        /// thread must already hold a slot; this is a refresh, not an entry.
+        /// <para><see cref="SafeToReclaimEpoch"/> is the minimum announced epoch across occupied slots, so a
+        /// holder that never refreshes stalls reclamation and every pending drain action process-wide.</para>
+        /// <para>Refreshing relinquishes protection for the previously announced epoch, so callers must hold
+        /// no references acquired under it.</para>
```
tiagonapoli: 2 line comment is enough

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -291,24 +291,24 @@ public bool TrySuspend()
...
             Debug.Assert(entry > 0, "Trying to refresh unacquired epoch");
-            Debug.Assert((*(tableAligned + entry)).threadId > 0, "Epoch table entry missing threadId");
+            DebugAssertEpochAcquired(entry);
 
-            // Protect CurrentEpoch by copying it to the instance-specific epoch table
-            // so that ComputeNewSafeToReclaimEpoch() will see it.
-            (*(tableAligned + entry)).localCurrentEpoch = CurrentEpoch;
+            // Refresh the announced epoch to CurrentEpoch
+            var epoch = Volatile.Read(ref CurrentEpoch);
+            EntryAt(entry).localCurrentEpoch = epoch;
```
tiagonapoli:
```suggestion
            EntryAt(entry).localCurrentEpoch = Volatile.Read(ref CurrentEpoch);;
```

libs/storage/Tsavorite/cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -291,24 +291,23 @@ public bool TrySuspend()
...
-            // Protect CurrentEpoch by copying it to the instance-specific epoch table
-            // so that ComputeNewSafeToReclaimEpoch() will see it.
-            (*(tableAligned + entry)).localCurrentEpoch = CurrentEpoch;
+            // Refresh the announced epoch to CurrentEpoch
+            EntryAt(entry).localCurrentEpoch = Volatile.Read(ref CurrentEpoch);
 
             // Max epoch across all threads may have advanced, so check for pending drain actions to process
             if (drainCount > 0)
-                Drain((*(tableAligned + entry)).localCurrentEpoch);
+                Drain(EntryAt(entry).localCurrentEpoch);
```
tiagonapoli: do not change - revert to *(tableAligned + entry)...

TedHartMS (approved):

PR #2019, opened 2026-08-04
Extract Vector Set cleanup work channel and work set

tiagonapoli:
Pulls the ad-hoc `Channel` and `ConcurrentDictionary` plumbing in the Vector Set cleanup pipeline behind two small types, and uses them in `VectorManager`. Behaviour is unchanged.

### `VectorSetCleanupWorkChannel<T>`

Replaces all three queues. They differ only in how the consumer reads:

| queue | payload | read shape |
|---|---|---|
| `requestDropTaskChannel` | none - a wake-up | whole backlog per pass |
| `requestCleanupTaskChannel` | `(ulong Context, TaskCompletionSource)` | whole backlog per pass |
| `cleanupTaskChannel` | none, but one item = one scan | one item per iteration |

The first two service all outstanding work in one pass, so a redundant item costs nothing and the backlog collapses - which gives

```csharp
// Drain all wake up signals
while (requestDropTaskChannel.Reader.TryRead(out _)) { }
```

a home as `requestDropTaskChannel.DrainPending()`. The third acts on the item itself, so it takes one per iteration.

Two of the queues carry no payload, because the work they refer to lives elsewhere - in `requestedDrops` for the drop queue, and in the marked contexts for the cleanup queue. Both `DrainPending()` and the payload-free `TryPublish()` are extension methods on `VectorSetCleanupWorkChannel<object>` rather than members, so neither can be called on `requestCleanupTaskChannel`: publishing a `default` tuple there would queue a null `TaskCompletionSource`, and discarding its backlog would strand every waiter on it.

The type also folds the three-step shutdown

```csharp
requestDropTaskChannel.Writer.Complete();
AsyncUtils.BlockingWait(requestDropTaskChannel.Reader.Completion);
AsyncUtils.BlockingWait(requestDropTask);
```

into `requestDropTaskChannel.CompleteAndWaitForConsumerTask(requestDropTask)`, and does not expose the underlying `ChannelReader`/`ChannelWriter`.

### `VectorSetCleanupWorkSet<TValue>`

Wraps `requestedDrops`, moving the `NET9_0_OR_GREATER` alternate-lookup branching and the poll-until-absent loop out of `VectorManager`. `DropRequested` and `WaitForDiskANNIndexDrop` become one-liners over `Contains` / `WaitForCompletion`.

### Tests

Unit tests over both types.

### DIFF
--- a/libs/server/Resp/Vector/Cleanup/VectorSetCleanupWorkChannel.cs
+++ b/libs/server/Resp/Vector/Cleanup/VectorSetCleanupWorkChannel.cs
@@ -1,0 +1,71 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Garnet.common;
+
+namespace Garnet.server
+{
+    /// <summary>
+    /// An unbounded single-consumer queue of Vector Set cleanup work, where each item is a unit of work.
+    /// </summary>
+    internal sealed class VectorSetCleanupWorkChannel<T>
+    {
+        private readonly Channel<T> channel;
+
+        public VectorSetCleanupWorkChannel()
+        {
+            channel = Channel.CreateUnbounded<T>(new() { SingleWriter = false, SingleReader = true, AllowSynchronousContinuations = false });
+        }
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


tiagonapoli:
Reopening from the fork branch instead, to avoid keeping a duplicate copy of the branch on this repo.

PR #2042, opened 2026-08-07

tiagonapoli:
## Problem

A Vector Set index record persists the primary's native DiskANN handle (`IndexPtr`) inside its value. Diskless sync streams log records verbatim into the replica's memory, so the replica receives the **primary's** pointer.

The only hook that zeroes that field is `GarnetRecordTriggers.OnDiskRead`, which by definition never fires for records streamed straight into memory. Since `VectorManager.NeedsRecreate` is exactly `indexPtr == 0`, a foreign pointer is indistinguishable from a healthy local one: the lazy `Service.RecreateIndex` rebuild is skipped and the raw value reaches the P/Invoke unvalidated. In production this faults the replica with SIGSEGV inside `NativeDiskANNMethods.card`.

## Fix

- `NetworkClusterSync` clears the pointer as records are deserialized, so the replica rebuilds its own index on first touch.
- **Context reservation after diskless full sync:** a diskless full sync SETs the streamed index and `ContextMetadata` records straight into the store, bypassing the RMW path that maintains the in-memory context-reservation bitmap. Unlike startup recovery, nothing rebuilt it, so after promotion a fresh Vector Set could be handed a context a streamed set already owns, corrupting both. Streamed records are now fed through the recovery bookkeeping during `CLUSTER SYNC` (`SanitizeAndTrackIngestedRecordIfApplicable`) and the reservation is rebuilt via `ReconcileRecoveredState` once the stream completes.

`ReconcileRecoveredState` takes `requireNoReservedContexts` on the diskless path: the rebuild replaces `contextMetadatas` wholesale, so a context that survived the preceding flush would be dropped without its index cleaned up. Failing loudly beats silently leaking one.

## Tests

Adds `Garnet.test.cluster.replication.vectorsets`, registered in `Garnet.slnx` and in the CI and nightly matrices.

The shared harness asserts index ownership *and* element-level equality — matching cardinality alone is too weak, since an aliased index reports the source's count perfectly well. Coverage includes diskless full sync, async replay, multi-database sync, and `FreshVectorSetDoesNotReuseStreamedContextAfterDisklessFullSync` for the context-reservation regression, plus two exception-injection tests covering an aborted full sync (recovery after retry, and that the aborted attempt leaves no orphaned namespaced records).

## Deferred (separate PRs)

- Vector Set failover tests.
- RangeIndex diskless-replication test coverage.
- Disk-based full-sync of a populated Vector Set (a distinct, more severe foreign-`IndexPtr` AVE + disk-based reservation gap).


tiagonapoli:
Superseded by #2043, which is opened from the branch in my fork (\	iagonapoli:tiagonapoli/vs-1-diskless\) and squashed to a single commit. Identical tree, no content change.

PR #2043, opened 2026-08-07

tiagonapoli:
## Problem

A Vector Set index record persists the primary's native DiskANN handle (`IndexPtr`) in its value. Diskless sync streams records verbatim into the replica's memory, so the replica receives the **primary's** pointer.

The only hook that zeroes that field is `GarnetRecordTriggers.OnDiskRead`, which never fires for records streamed straight into memory. Since `NeedsRecreate` is exactly `indexPtr == 0`, a foreign pointer looks healthy: the lazy `RecreateIndex` rebuild is skipped and the raw value reaches the P/Invoke, faulting the replica with SIGSEGV inside `NativeDiskANNMethods.card`.

## Fix

- **Index pointer:** `NetworkClusterSync` clears it as records are deserialized, so the replica rebuilds its own index on first touch.
- **Context reservation:** a diskless full sync SETs streamed index and `ContextMetadata` records directly, bypassing the RMW path that maintains the in-memory reservation bitmap. Nothing rebuilt it, so after promotion a fresh Vector Set could be handed a context a streamed set already owns, corrupting both. Streamed records now go through `SanitizeAndTrackIngestedRecordIfApplicable`, and the reservation is rebuilt via `ReconcileRecoveredState` once the stream completes. Its `requireNoReservedContexts` flag fails loudly if a context survived the preceding flush, since the rebuild replaces `contextMetadatas` wholesale and would otherwise leak that index.
- **Flush (found in review):** `FlushGuard.Dispose` reset `contextMetadatas` but left `dirtyContextMetadatas` populated, so a concurrent `UpdateContextMetadata` drain could index out of range, or re-persist metadata for a context the flush just removed. Now cleared with the rest of the invalidated state.

## Tests

Adds `Garnet.test.cluster.replication.vectorsets`, registered in `Garnet.slnx` and the CI/nightly matrices.

The harness asserts index ownership *and* element-level equality — matching cardinality alone is too weak, since an aliased index reports the source's count perfectly well. Covers diskless full sync, async replay, multi-database sync, `FreshVectorSetDoesNotReuseStreamedContextAfterDisklessFullSync` for the reservation regression, and two exception-injection tests for an aborted full sync (recovery after retry, and no orphaned namespaced records left behind).

## Deferred (separate PRs)

- Vector Set failover tests.
- RangeIndex diskless-replication coverage.
- Disk-based full sync of a populated Vector Set (distinct, more severe foreign-`IndexPtr` AVE + disk-based reservation gap).


libs/server/Resp/Vector/VectorManager.ContextMetadata.cs
```diff
@@ -440,6 +440,10 @@ public readonly void Dispose()
                 // Clear out all context data
                 manager.contextMetadatas = new ContextMetadata[1];
 
+                // The flushed records are gone, so any inflight recovery state should be cleared since it's now invalid
+                manager.recoveredIndexes.Clear();
+                manager.recoveredMetadata.Clear();
```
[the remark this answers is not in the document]
tiagonapoli: Good catch — this one is real, and it's the same shape as the bug this PR is fixing. Fixed in the latest push.

`UpdateContextMetadata` drains the dirty set under `lock (this)` and indexes straight into the array:

```csharp
contextIndex = dirtyContextMetadatas.First();
_ = dirtyContextMetadatas.Remove(contextIndex);
MemoryMarshal.Cast<byte, ContextMetadata>(dataSpan)[0] = contextMetadatas[contextIndex];   // line 640
```

`FlushGuard` does block new work — it takes `vectorSetLocks.AcquireAllExclusiveLock` plus `Monitor.Enter(manager)` — but it does not drain `dirtyContextMetadatas`, and the drain loop releases `lock (this)` between iterations to issue the RMW. So:

1. A drain pops index 5, releases the lock, and issues the (possibly pending) RMW.
2. `FLUSHDB` takes the monitor, resets `contextMetadatas` to `new ContextMetadata[1]`, releases.
3. The drain loop re-enters, pops index 7, and hits `contextMetadatas[7]` on a length-1 array → `IndexOutOfRangeException`.

Multiple entries in flight is not exotic: recovery marks every recovered context dirty in a loop (`VectorManager.cs:324` and `:373`).

There's a second, quieter failure even with a single queued entry — the RMW re-persists a `ContextMetadata` record for a context the flush just deleted, resurrecting metadata in a store that is supposed to be empty. That's exactly the "stale state survives a wipe" class of bug this PR is about.

Fix is to clear it alongside the other flushed state in `FlushGuard.Dispose`:

```csharp
// Clear out all context data
manager.contextMetadatas = new ContextMetadata[1];

// Indices queued against the old array would index out of range, and would re-persist
// metadata for contexts the flush just removed
manager.dirtyContextMetadatas.Clear();

// The flushed records are gone, so any inflight recovery state should be cleared since it's now invalid
manager.recoveredIndexes.Clear();
manager.recoveredMetadata.Clear();
```

Clearing is safe: after the flush `contextMetadatas` is a fresh empty array with nothing reserved and the store holds no metadata records, so there is nothing legitimate left to persist.

libs/server/Resp/Vector/VectorManager.cs
```diff
@@ -409,6 +428,45 @@ public void RecoveredContextMetadata(ref LogRecord record)
...
+        /// </summary>
+        public void SanitizeAndTrackIngestedRecordIfApplicable<TSourceLogRecord>(ref TSourceLogRecord record) where TSourceLogRecord : ISourceLogRecord
+        {
+            if (record.Info.Tombstone)
+            {
+                return;
+            }
+
+            // ContextMetadata records are identified by their namespace, index records by their record type
+            var ns = record.Namespace;
```
kevin-montrose: This should presumably be guarded by .HasNamespace

test/cluster/Garnet.test.cluster/ClusterTestUtils.cs
```diff
@@ -2201,6 +2201,90 @@ public List<SlotItem> ClusterSlots(IPEndPoint endPoint, ILogger logger = null)
...
+                BackOff();
+            }
+
+            Assert.Fail($"node {nodeIndex} never began serving '{key}'");
+        }
+
+        /// <summary>
+        /// Writes filler keys so a re-attaching replica falls behind the primary's AOF window and cannot sync incrementally.
+        /// </summary>
+        public void PushPrimaryAhead(int primaryIndex, int keyCount = 256, int valueLength = 64)
```
tiagonapoli: That's too generic name

test/cluster/Garnet.test.cluster/ClusterTestUtils.cs
```diff
@@ -2201,6 +2201,90 @@ public List<SlotItem> ClusterSlots(IPEndPoint endPoint, ILogger logger = null)
...
+        {
+            var primary = GetEndPoint(primaryIndex);
+            for (var i = 0; i < keyCount; i++)
+            {
+                _ = Execute(primary, "SET", [$"{{padding}}key{i}", new string('x', valueLength)], skipLogging: true);
+            }
+        }
+
+        /// <summary>Attaches a replica to a primary and waits until it is in sync.</summary>
+        public void Attach(int replicaIndex, int primaryIndex, bool waitForRecovery = false, ILogger logger = null)
```
tiagonapoli: AttachReplicaToPrimary?

kevin-montrose (changes requested):
The dirty namespaces Copilot seems correct, plus I think some HasNamespace checks are missing (things might still pass in Release, Debug I would expect assert failures).

libs/cluster/Server/Replication/ReplicaOps/ReplicaDisklessSync.cs
```diff
@@ -238,6 +238,13 @@ public AofAddress TryReplicaDisklessRecovery(SyncMetadata primarySyncMetadata, o
                 // Before advertising updated replication offset, wait for Vector Set ops to finish
                 storeWrapper.DefaultDatabase.VectorManager?.WaitForVectorOperationsToComplete();
 
+                // Full sync only: a partial sync replays AOF instead of streaming raw records.
+                if (primarySyncMetadata.fullSync)
+                {
+                    // In full sync, a flush is done right before, so no reserved contexts should exist
+                    storeWrapper.DefaultDatabase.VectorManager.ReconcileRecoveredState(requireNoReservedContexts: true);
```
tiagonapoli:
```suggestion
                    storeWrapper.DefaultDatabase.VectorManager?.ReconcileRecoveredState(requireNoReservedContexts: true);
```

test/cluster/Garnet.test.cluster/ClusterTestUtils.cs
tiagonapoli: Renamed to `AdvancePrimaryPastReplicaAofWindow`, which says what it is for rather than what it does.

test/cluster/Garnet.test.cluster/ClusterTestUtils.cs
tiagonapoli: Renamed to `AttachReplicaToPrimary`.

libs/cluster/Server/Replication/ReplicaOps/ReplicaDisklessSync.cs
tiagonapoli: Applied.

libs/server/Resp/Vector/VectorManager.cs
tiagonapoli: Done in `babecbc82d` - the metadata-namespace check is now behind `record.HasNamespace`.

kevin-montrose (approved):

PR #2048, opened 2026-08-08
Add Garnet.shared project and move LightEpoch into it

tiagonapoli:
Introduces a low-level `Garnet.shared` project shared by Tsavorite and the Garnet libraries, and moves Tsavorite's `LightEpoch` into it.

### `Garnet.shared` project

- New `libs/shared/Garnet.shared.csproj` (namespace `Garnet.shared`), added to `Garnet.slnx` and `Tsavorite.slnx`
- Moved `LightEpoch.cs`, `LightEpoch.EntryTable.cs`, `LightEpoch.TestHooks.cs` and `IEpochAccessor.cs` from `Tsavorite.core/Epochs` into `libs/shared/Epochs`
- `Tsavorite.core` now takes a `ProjectReference` on `Garnet.shared`
- `Murmur3` moved to `Garnet.shared.HashUtility` so `LightEpoch` has no dependency on `Tsavorite.core`; `Tsavorite.core.Utility.Murmur3` delegates to it, preserving the public API
- Consumers import the moved types via a project-level global using (`<Using Include="Garnet.shared" />` in `Garnet.server.csproj` and `Tsavorite.core.csproj`) rather than ~45 per-file `using` directives; the few consumers outside those two projects keep a plain `using`

`Garnet.client.LightEpoch` is intentionally left untouched in this PR; deduplicating it is follow-up work.

### Tests

- New `test/standalone/Garnet.shared.test` replaces `libs/storage/Tsavorite/cs/test/test.epoch`; namespace `Tsavorite.test.epoch` -> `Garnet.shared.test.epoch`
- `InternalsVisibleTo` for the epoch test hooks moved from `Tsavorite.core` to `Garnet.shared`
- Added `Garnet.shared.test` to the standalone test matrices in `ci.yml` and `nightly.yml`. The epoch tests were previously **not** run by CI at all (`Tsavorite.test.epoch` was absent from every matrix), so this is a coverage gain

## Validation

- `dotnet build Garnet.slnx -c Release` -> 0 errors / 0 warnings (net8.0 + net10.0)
- `dotnet build Tsavorite.slnx -c Release` -> 0 errors / 0 warnings
- `dotnet test test/standalone/Garnet.shared.test` -> 15/15 passed
- `dotnet format` -> no changes


Tornhoof:
Isn't that naming wise not a bit inconsistent?, Tsavorite is the underlying storage engine, shouldn't it be then called tsavorite.shared, or something more generic, like framework, common or similar?

libs/shared/HashUtility.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Hash helpers shared across Garnet and Tsavorite.
+    /// </summary>
+    public static class HashUtility
+    {
+        /// <summary>
+        /// A 32-bit murmur3 implementation.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int Murmur3(int h)
```
Tornhoof, line 17: Isn't FMix32 a better name for that function?, it's not full Murmur3, but just that Final 32 Bit mix of it.

tiagonapoli:
Closing this in favour of the shared-source approach suggested in review.

## Why

This PR moves `LightEpoch` into a separate assembly. Garnet publishes ReadyToRun, and `LightEpoch`'s hot methods (`ProtectAndDrain`/`Resume`/`Suspend`) are `[MethodImpl(AggressiveInlining)]`. Today Tsavorite's calls to them are intra-assembly and inline into the R2R image; behind a DLL boundary R2R suppresses that inlining for version resilience.

I tried to rescue this with composite ReadyToRun and measured it end to end. Result: the regression is real, but composite is not a viable fix.

### The regression is real

Minimal two-assembly repro — an `[AggressiveInlining]` method called across an assembly boundary in a hot loop, `DOTNET_TieredCompilation=0`, median of 3 runs:

| R2R mode | `DOTNET_ReadyToRun=1` | `=0` (pure JIT) |
| --- | --- | --- |
| per-assembly | 1045 ms | 633 ms |
| composite | 627 ms | 628 ms |

Per-assembly R2R costs ~65% on the cross-assembly inlined call, while the JIT inlines it fine. So the concern raised in review is not theoretical.

### Composite ReadyToRun is not a viable fix

Composite does fully restore the inlining (row 2), but:

- **It is not supported for framework-dependent publish.** dotnet/runtime#90454 ("Allow Composite ReadyToRun to be used with framework-dependent deployment") is still open. The SDK actively force-disables it in `Microsoft.NET.CrossGen.targets`, still present in `dotnet/sdk` main:

  ```xml
  <PublishReadyToRunComposite Condition="'$(PublishReadyToRunUseCrossgen2)' != 'true' or '$(SelfContained)' != 'true'">false</PublishReadyToRunComposite>
  ```

  Because that assignment is unconditional w.r.t. a user-supplied value and evaluates after the publish profile is imported, enabling composite requires an MSBuild target that runs after SDK evaluation purely to defeat the guard. (The docs note "In .NET 6, Composite ReadyToRun is only supported for self-contained deployment" reads as version-scoped, but net7+ is no different — that is exactly the confusion behind #90454.)
- **Artifact size roughly doubles**: published `GarnetServer.exe` goes from 31 MB to 60 MB (net10.0, win-x64).
- **No startup win**: CPU time to accepting connections, median of 7 runs — 281 ms non-composite vs 297 ms composite. The larger image offsets the better code.
- Garnet does not disable tiered compilation, and composite is only recommended for apps that do. Tiering re-JITs these hot methods anyway, so the benefit is confined to warmup.
- Composite diagnostics/profiling support is still incomplete, which matters for perf investigation on this codebase.

Forcing an officially unsupported publish mode, by defeating an SDK guard, to buy warmup-only codegen that tiering recovers is not a good trade.

## What to do instead

De-duplicate via shared source rather than a shared assembly:

```xml
<Compile Include="..\..\common\LightEpoch.cs" Link="Epochs\LightEpoch.cs" />
```

This removes the duplicate `Garnet.client` copy while `LightEpoch` still compiles into each assembly: no DLL boundary, R2R inlining preserved, no `internal` -> `public` changes, and Tsavorite stays dependency-free.

Two things from this PR worth keeping when it is redone:

- `Murmur3` -> `FMix32` — it is MurmurHash3's 32-bit finalization mix, not the full hash, and it matches the existing `fmix64` naming in `libs/common/HashUtils.cs` (thanks @Tornhoof).
- Adding the epoch tests to the CI matrices — they were previously not run by CI at all.

PR #2051, opened 2026-08-09
[Storage] Add optional native (mimalloc / direct-VM) allocator for pooled memory

badrishc:
### Description of Change

Adds an **opt-in** native (off-managed-heap) allocator so Garnet's large, pooled, long-lived memory can live in native memory instead of the .NET GC heap. Default behavior is unchanged (`off` = fully managed). Controlled by a single setting:

```
--native-allocator <off | buffer-pool | full>
```

**Modes** (`NativeAllocatorSurfaces` flags enum, exposed as ordered modes):

| Mode | Surfaces moved off-heap | Backend |
|------|-------------------------|---------|
| `off` (default) | none — behavior-identical to today | managed (`GC.AllocateArray`) |
| `buffer-pool` | `SectorAlignedBufferPool` IO buffers | **mimalloc** |
| `full` | above **+** main/object log pages, `TsavoriteLog` pages, hash-index table, recovery/flush frames | mimalloc (pool) + **direct-VM** (`mmap`/`VirtualAlloc`) |

The primary win is `buffer-pool`: mimalloc's thread-local heaps + cross-thread free lists **replace the manual per-level free-list recycling** in `SectorAlignedBufferPool`, removing its single shared `ConcurrentQueue` as a scaling bottleneck. Transient/small allocations intentionally stay managed — only the pooled, hot buffers move.

**Key changes:**
- New `Tsavorite.core.Native` surface: `Mimalloc` (P/Invoke + RID resolution), `INativePinnedAllocator`, `MimallocPooledAllocator`, `DirectVirtualMemory`, `NativeMemoryTracker`, `NativePageBlockRegistry`, `NativeAllocatorInitializer`/`NativeAllocatorSurfaces`.
- `SectorAlignedBufferPool` grows an optional native backend (`NativeAllocator`); when set, `Get`/`Return` allocate/free via mimalloc instead of the managed free-list.
- `full` mode routes log pages, the hash-index table, and `BlittableFrame`s through the direct-VM allocator. These regions use **lazy physical backing** (reserve address space, fault pages on first touch) so memory is not committed on the instantiating socket — matching managed `GC.AllocateArray` NUMA behavior.
- On Linux, direct-VM regions are 2 MB-aligned and hinted `MADV_HUGEPAGE` (transparent huge pages), lowering dTLB pressure on random access.
- Native usage is registered with Garnet: `NativeMemoryTracker.Bytes` is surfaced as **`native_allocator_bytes` in `INFO memory`** alongside `gc_heap_bytes`.
- Config: `--native-allocator` option (`Options.cs`, `defaults.conf`) → `GarnetServerOptions.NativeAllocatorSurfaces`, installed once at startup via `NativeAllocatorInitializer.Initialize` before any store/pool is built.
- Docs: `website/docs/getting-started/memory.md` (modes, GC/`GCHeapHardLimit` sizing guidance, telemetry).
- Packaging/CI: prebuilt `mimalloc` binaries for `linux-x64`, `linux-arm64`, `linux-musl-x64`, `linux-musl-arm64`, `win-x64`, `win-arm64`; `build-mimalloc.sh` (POSIX sh) + `native-mimalloc-build.yml` workflow + `publish-native-prebuilts` action; docker-image validation gains a native-allocator phase.

### Key Technical Details

**Affected types/interfaces:**
- `NativeAllocatorSurfaces` — `[Flags]` capability set (`BufferPool`/`LogPages`/`HashIndex`/`Frames`/`Network`; `Full` = all but `Network`).
- `NativeAllocatorInitializer` — process-wide install/teardown; `EnabledSurfaces`, env-var baseline (`GARNET_NATIVE_ALLOCATOR`) for test harnesses.
- `INativePinnedAllocator` / `MimallocPooledAllocator` — pluggable native backend for the pool.
- `DirectVirtualMemory` / `NativePageBlockRegistry` — direct-VM regions + a registry that keeps native page blocks alive until every reader has drained.
- `NativeMemoryTracker` — process-wide byte counter (queried on demand for `INFO`, not per-op on the hot path).

**Fail-fast, not silent fallback:** `buffer-pool` and `full` require the mimalloc native library. If it cannot be loaded for the RID, **startup throws** (`TsavoriteException`) rather than quietly running on the managed pool — a missing binary is a loud config error, not a silent slowdown. The direct-VM surfaces use `mmap`/`VirtualAlloc` and are always available.

**Finalization / memory safety (`full` mode):** direct-VM log pages and frames can be evicted while a snapshot flush or scan still references them. Frees are **deferred** (to a finalizer/refcount-driven drain) so a page is only unmapped once all IO/readers have released it, preventing use-after-free/AccessViolation. Evicted direct-VM log pages are recycled through an `OverflowPool` to bound churn. This is the one place native needs extra bookkeeping the GC gave us for free (managed pages were kept alive by references; native pages must be explicitly kept mapped).

### Performance

- **Buffer-pool microbenchmark** (direct `Get`/`Return`, no device, NUMA-pinned, Mops/s): the managed single-`ConcurrentQueue` pool peaks at 1 thread (~35) and **collapses under concurrency** (~0.6 @ 64t, cache-line ping-pong on the shared queue). mimalloc scales smoothly to **~385 @ 32t (~480×)** via per-thread heaps, with no fixed shard cap.
- **In-memory KV read A/B** (`full`, 100M keys, 100% reads): **+6–17%** throughput and ~1.8× faster load, attributable to 2 MB THP backing the index/log (measured 12.5 GB `AnonHugePages` vs 0 for managed) reducing page-walk/dTLB cost.

### What NOT to Do (for future agents)

- ❌ **Don't move transient or small allocations off-heap** — only pooled/large/long-lived buffers benefit; managed is fine (and cheaper) for the rest.
- ❌ **Don't free direct-VM log pages/frames eagerly in `full` mode** — defer through the registry/finalization drain; a snapshot flush or scan may still hold the page (UAF/AccessViolation otherwise).
- ❌ **Don't add per-op native tracking on the hot path** — `native_allocator_bytes` is computed on demand for `INFO`; per-op `Interlocked` on the counter was measured to cap throughput and was removed.
- ❌ **Don't assume native fallback** — a mimalloc-backed mode fails fast if the library is missing; don't add silent managed fallback.
- ❌ **Don't eager-commit direct-VM regions** — reserve + fault-on-touch to preserve NUMA-local placement like `GC.AllocateArray`.

### Edge Cases

| Scenario | Handling |
|----------|----------|
| mimalloc binary missing for RID | Startup fails fast with a clear `TsavoriteException` (buffer-pool/full only) |
| Page evicted during snapshot flush / scan | Free deferred until IO/reader refcount drains |
| Container memory limits | Native memory is outside the GC heap — size `GCHeapHardLimit(Percent)` with native headroom; monitor `native_allocator_bytes` |
| NUMA placement | Direct-VM regions reserve address space and fault lazily, so pages land on the touching socket |
| Windows / musl / arm64 | Prebuilt binaries shipped for all supported RIDs; separate CI builds/validates them |

### Testing

34 native-allocator tests across the pool, store, hash-index, snapshot-stress, and end-to-end server surfaces (`NativeAllocatorTests`, `NativeAllocatorStoreTests`, `NativeHashIndexTests`, `NativeSnapshotStressTests`, `NativeAllocatorServerTests`), plus existing suites parametrized to run under `buffer-pool`/`full` via `GARNET_NATIVE_ALLOCATOR`, a config test, and a docker-image native phase. Memory-freeing/finalization hardened over multiple two-model (GPT-5.6 + Gemini 3 Pro) review passes; full Garnet + Tsavorite suites run clean under both native modes with no leaks.

### Issues Fixed

No tracked issue — opt-in performance feature. (Default `off` preserves current behavior.)


libs/storage/Tsavorite/cs/src/core/Native/NativeMemoryTracker.cs
```diff
@@ -0,0 +1,81 @@
...
+    public static class NativeMemoryTracker
+    {
+        // Longs per stripe (128-byte stride) so adjacent stripes never share a cache line regardless of the
+        // managed array's base alignment.
+        const int Stride = 16;
+
+        static readonly long[] slots;
+        static readonly int mask;
+
+        static NativeMemoryTracker()
```
kevin-montrose: Avoid static constructors ([CA1810](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1810); perhaps we should enable that warning?).

libs/storage/Tsavorite/cs/src/core/Native/NativeMemoryTracker.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// make deterministic assertions about direct-VM allocation/free without the noise of mimalloc's
+        /// process-global committed total, which fluctuates asynchronously as mimalloc returns memory to the OS.
+        /// </summary>
+        internal static long DirectVmBytes
+        {
+            get
+            {
+                long total = 0;
+                for (var i = 0; i < slots.Length; i += Stride)
+                    total += Interlocked.Read(ref slots[i]);
```
kevin-montrose: nit: Looks like we only have mimalloc on 64-bit architectures, on which Interlocked.Read is unnecessary.

libs/storage/Tsavorite/cs/src/core/Native/Mimalloc.cs
```diff
@@ -0,0 +1,207 @@
...
+                {
+                    available = false;
+                    logger?.LogWarning(ex, "Failed to bind mimalloc exports.");
+                }
+                initialized = true;
+            }
+            return available;
+        }
+
+        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.AggressiveInlining)]
```
kevin-montrose: nit: why fully qualify all these `[MethodImpl]`s instead of adding a `using`?

kevin-montrose (changes requested):

badrishc:
Closing - will reopen a new smaller surface PR with only direct-VM and managed buffer-pool v2.

PR #2063, opened 2026-08-13
[Tsavorite] Scalable origin-return buffer pool + direct-VM native allocator

badrishc:
Two independent performance features for Tsavorite/Garnet, both targeting GC pressure and buffer-pool scalability under many concurrent IO-completion threads.

- The **scalable buffer pool** becomes the default managed backend (the legacy per-level pool stays selectable via a flag).
- The **native allocator** is opt-in and off by default.

### New configuration switches

| Switch | Option | Default | Purpose |
|---|---|---|---|
| `--use-legacy-buffer-pool` | `UseLegacyBufferPool` | `false` | Select the legacy per-level `ConcurrentQueue` pool |
| `--buffer-pool-memory-budget` | `BufferPoolMemoryBudget` | `1g` | Per-pool managed byte budget (`0` disables caching) |
| `--use-native-allocator` | `UseNativeAllocator` | `false` | Route large regions off the managed heap via direct VM |

---

### Problem

The single per-level `ConcurrentQueue` pool collapses under concurrent cross-thread `Get`/`Return`. Because completion threads free buffers that issuing threads allocated, completion threads accumulate every buffer while issuing threads keep allocating fresh — a memory-bloat / zero-reuse asymmetry on top of the head/tail cache-line contention.

### Design

Each buffer is routed back to its **originating** thread rather than the freeing (IO-completion) thread.

- **Ownership** keyed by `(pool, thread, size-class)` via a `[ThreadStatic]` recyclable-slot scheme — no `ThreadLocal<T>`, no monotonic-index leak.
- **`Get` tiers:** owner-local stack (no atomics) → cross-thread claim → shared depot → allocate.
- **Per-pool immutable mode** and frozen `UnpinOnReturn`, so a buffer is always returned under the pin policy it was allocated under.

#### Size-class ladder

Coarse-linear-then-geometric, `O(1)` lookup in both directions. At a 512 B sector it spans **512 B … 16 MB** across 28 classes:

| Region | Classes | Sizes | Granularity |
|---|---|---|---|
| exact tiny | 2 | 512 B, 1 KB | 1 sector (zero waste) |
| coarse linear | 4 | 2, 4, 6, 8 KB | 4-sector stride |
| geometric | 22 | 12 KB … 16 MB | 2 classes/doubling (≤ 1.5× waste) |

The 16 MB ceiling pools the 4 MB object-log flush buffer and full records built on inline values up to ~16 MB. Requests above the ceiling bypass the cache (allocate-on-`Get`, free-on-`Return`).

#### Byte budget

A hard per-pool byte budget, with one permit per buffer taken at birth and released exactly once at death (standalone `BudgetState` + finalizer reclamation). It is split into **isolated small/large sub-budgets** (25% / 75%) so a burst of large record/flush buffers cannot starve caching on the hot small-buffer path.

`--buffer-pool-memory-budget` sets it at startup. Raise it for workloads with a large working set of big values paged to/from disk (fewer LOH allocations, less GC churn); lower it to cap the pool footprint. Ignored in legacy mode.

Setting it to **`0` is an explicit opt-out of buffer caching**: it routes to the pool's `Disabled` short-circuit (honored by both backends) so every `Get` allocates and every `Return` drops — trading reuse for the smallest steady-state footprint, and avoiding the shard lookup plus doomed permit reservation that a literal zero budget would pay on every `Get`. A negative value is only reachable by overflowing the size parser (e.g. `16000000p`), so it is a misconfiguration rather than an intent and is rejected with a warning.

#### Large classes are depot-only

Buffers above `LargeTierMinBytes` (256 KB) skip the per-thread local/cross-thread tiers on `Return` (both owner and foreign paths) and go straight to the shared striped depot, so any thread can reuse them under the shared `largeBudget`.

Under a wide value-size mix a thread rarely re-requests the same large class back-to-back, so per-thread parking strands big buffers and multiplies the working set across `(thread × class)` — inflating peak RSS and Gen2 GC *even while within budget*. Global sharing restores legacy-like memory behavior for big buffers while small classes keep the atomic-free per-thread fast path. Large-buffer ops are low-rate and disk workloads are bandwidth-bound, so the depot handoff is effectively free.

Measured on a 100 B–10 MB disk read mix at 64 threads, 1 GiB budget:

| Metric | Before | After |
|---|---|---|
| Large-class reuse | ~12–50% | ~54–96% |
| New allocations | baseline | ~halved |
| Gen2 collections | baseline | ~halved |
| Peak RSS | baseline | lowered |
| Throughput | baseline | unchanged |

#### Concurrency hardening

- The owner-hot local stack and producer-hot cross-thread stack sit on **distinct 64-byte cache lines** via explicit `[StructLayout(LayoutKind.Explicit)]`. A reference-type class ignores `LayoutKind.Sequential`, so padding fields alone do **not** separate reference fields.
- **Seal-aware retirement/teardown** for dead threads and pool close. `Free()` seals every registered shard; for a shard still owned by a *live foreign* thread (whose owner-private stack it must not mutate) it **atomically detaches** that stack before releasing its permits. Buffers are detached rather than dropped, so one that the owner is concurrently popping stays valid.
  - Detaching is required, not merely tidy: that thread's thread-static slot roots `shard → buckets → localHead → buffers`, so leaving the chain in place would pin a disposed pool's buffers for the remaining life of a long-lived (e.g. thread-pool) thread — and invisibly, since the permits backing them were already returned to the budget.
- Teardown fully drains only the freeing thread's own (quiescent) shard, so a concurrent owner `Get`/`Return` is never handed a buffer whose backing array was nulled underneath it.
- The per-pool close is **elected exactly once** (`PoolActive` CAS), and the shard registry compacts dead weak-references so a long-lived pool under thread churn stays bounded by the live shard count.

### Results

Direct pool microbenchmark (Release, 160-core):

| Metric | Result |
|---|---|
| Scaling (1 → 64 threads) | **22 → 1260 Mops/s** (legacy: 44 → 1.0, collapses) |
| Working set | 8M ops → 528 allocations (8.2× reuse) |
| Cross-thread latency | p99 ≤ 2 µs |

### Diagnostics

An opt-in `SectorAlignedBufferPool.Stats` surface records per-size-class new-buffer allocations vs. cache reuses, for validating allocation stabilization and reuse under load.

The recording call sites in the `Get` hot path are `[Conditional("BUFFER_POOL_STATS")]`, so a default build compiles them out entirely — byte-for-byte identical `Get` codegen, provably zero cost. Counters are compiled in only with `-p:BufferPoolStats=true` (KV.benchmark warns if `--pool-stats` is used without it; `Stats.Compiled` reports whether the build records counts).

---

## 2. Direct-VM native allocator

Routes large memory regions (log pages / hash index / recovery frames) off the managed GC heap through direct OS virtual memory (`mmap` / `VirtualAlloc`). **No native library is required.**

- New `--use-native-allocator` switch (`UseNativeAllocator`), plus `NativeAllocatorInitializer`, `DirectVirtualMemory`, `NativeMemoryTracker`, and `NativePageBlockRegistry`.
- New `INFO` field **`native_allocator_bytes`** surfaces off-heap committed bytes.
- Native index grow allocates the replacement block **before** retiring the old one, so a throwing (OOM) map leaves the live table intact.
- Transparent huge-page alignment is gated to Linux (where `MADV_HUGEPAGE` applies).

---

## Testing

**`SectorAlignedBufferPoolTests`** — 25 CI correctness tests covering the size-class ladder, cross-thread origin routing and lazy-clear, the depot, budget/permit accounting, unpin-on-return, disabled/oversize bypass, teardown racing live traffic, in-flight returns after `Free`, thread-static slot recycling across pools, and dead-thread finalizer reclamation.

Also included:

- `NativeAllocator*Tests` (hlog / recovery / server) and direct-VM Docker validation.
- Config parsing tests for all three switches.
- `[Explicit]` scaling / reuse / latency / soak stress harnesses.

**Disk-oriented KV.benchmark stress run** (variable 100 B–10 MB inline values paged to NVMe, multi-threaded reads via the pending-read pool path) validates that new allocations stabilize while cache reuse dominates across every size class and thread count, with bounded memory. New KV.benchmark options `--value-size-max`, `--pool-stats` (requires `-p:BufferPoolStats=true`), `--pool-budget`, and `--use-legacy-buffer-pool` drive, instrument, and A/B it.

**No regression on the tuned path:** fixed-length in-memory read throughput is unchanged versus the pre-instrumentation baseline (36 / 70 / 109 Mops/s at 16 / 32 / 64 threads) — the fixed-size load/read path is preserved verbatim, and in-memory reads never touch the pool `Get` path.

---

## Documentation

- **Design from first principles:** `website/docs/dev/tsavorite/buffer-pool.md`
- **Configuration reference table:** `website/docs/getting-started/configuration.md` — all three switches
- **Tuning and GC-sizing guidance:** `website/docs/getting-started/memory.md`


libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -2057,11 +2293,40 @@ void FlushRunner()
...
+                            // its callback owns releasing this page's IO unit and buffers — releasing here too would
+                            // underflow nativeSnapshotIoOutstanding and could unmap a page under an in-flight write.
+                            // If it was NOT submitted, no callback will fire, so release the unit (exactly once) and
+                            // the buffers here. Then rethrow into the outer catch for logging + SetException.
+                            flushCompletionTracker.SetException(writeEx);
+                            if (!asyncResult.snapshotDeviceWriteIssued)
+                            {
+                                if (asyncResult.TryClaimSnapshotUnitRelease())
+                                    EndNativeSnapshotFlush();
+                                _ = asyncResult.Release();
```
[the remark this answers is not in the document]
badrishc: Thanks — I looked at this closely and I don't think it can double-return buffers. `PageAsyncFlushResult.Release()` is exactly-once by construction: it acts (returns `freeBuffer1`, calls `CompleteFlush`) **only** on the `Interlocked.Decrement(ref count) == 0` transition, and `count` starts at 1.

So in the synchronous-callback-then-throws scenario:
- the callback's `Release()` takes `count` `1 → 0` and does the actual return (and nulls `freeBuffer1`);
- the `catch` block's `Release()` takes `count` `0 → -1`, `result != 0`, so it is inert — no second `Return()`, no second `CompleteFlush()` (and `freeBuffer1` is already `null`).

The `count == 0` check *is* the exactly-once guard for buffer/flush-completion release; a separate flag would be redundant. The snapshot IO unit (`nativeSnapshotIoOutstanding`) has its own one-shot guard via `TryClaimSnapshotUnitRelease()` (`Interlocked.Exchange`), so it also cannot underflow. No code change needed here.

libs/storage/Tsavorite/cs/src/core/Utilities/BufferPool.OriginReturn.cs
```diff
@@ -0,0 +1,835 @@
...
+        // the ~16 MB max inline value (header + key + up-to-~16 MB value) is still pooled rather than bypassed.
+        private const int TinyExactClasses = 2;         // exact 512 B, 1 KB
+        private const int LinearStrideSectors = 4;      // 2 KB step at 512 B sectors
+        private const int StrideClasses = 4;            // 2 / 4 / 6 / 8 KB classes
+        private const int LinearClasses = TinyExactClasses + StrideClasses;         // 6
+        private const int LinearTopSectors = LinearStrideSectors * StrideClasses;   // 16 (must be a power of two)
+        private const int Log2LinearTop = 4;            // BitOperations.Log2(LinearTopSectors)
+        private const int GeometricDoublings = 11;      // 8 KB .. 16 MB
+        private const int NumClasses = LinearClasses + 2 * GeometricDoublings;      // 28
+        private const int MaxPooledSectors = LinearTopSectors << GeometricDoublings; // 32768 (16 MB at 512 B sectors)
```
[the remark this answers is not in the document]
badrishc: Good catch on the precision. This is working as intended — the 16 MB ceiling is deliberate — but you're right that the "still pooled" claim over-stated it at the absolute maximum. A record whose value is at the `0xFFFFFE` inline cap, plus key + header, rounds just past 32,768 sectors and takes the **bypass** path (allocate-on-`Get`, free-on-`Return`), which is correct and safe; it simply isn't cached. Parking a >16 MB buffer per (thread, size-class) for these rare maximum-size records isn't worth the retained memory.

I've softened the wording rather than raising the ceiling, in the source comment (`BufferPool.OriginReturn.cs`) and the design doc (`website/docs/dev/tsavorite/buffer-pool.md`) to state precisely that values up to *nearly* 16 MB (16 MB minus header/key overhead) are pooled and the absolute-max record intentionally bypasses.

website/docs/getting-started/memory.md
```diff
@@ -74,4 +74,67 @@ Read cache helps bring in records from disk to memory in a separate read cache r
...
+policies). It is independent of `--use-native-allocator`: the buffer pool always uses managed (Pinned Object
+Heap) buffers in this release, regardless of the native-allocator setting.
+
+## Native (off-heap) allocator
+
+By default, Garnet's large, long-lived buffers — the hash index, hybrid-log pages, IO/flush buffers, and
+recovery frames — are allocated as pinned arrays on the managed .NET heap (the Pinned Object Heap). The optional
+`--use-native-allocator` setting moves these allocations *off* the managed heap, into native memory, which reduces
+GC pause times (the collector has far less to scan/compact) and removes POH fragmentation. It is **off by
+default** and opt-in.
```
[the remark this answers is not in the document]
badrishc: Fixed — removed "IO/flush buffers" from the off-heap intro list and added a parenthetical clarifying that the `SectorAlignedBufferPool` IO/flush buffers always remain on the managed Pinned Object Heap regardless of `--use-native-allocator`, consistent with the rest of the section and the implementation.

test/docker-tests/validate_docker_images.py
```diff
@@ -415,6 +415,85 @@ def test_native_device(images: list[ImageDef], results: TestResult, base_port: i
...
+        os.makedirs(vol, exist_ok=True)
+        os.chmod(vol, 0o777)
+        container_name = f"garnet-na-validate-{img.name}"
+
+        na_args = (
+            "--use-native-allocator true --logger-level Information "
+            "--storage-tier true "
+            "--logdir /data/store --checkpointdir /data/checkpoint "
+            "--aof true"
+        )
```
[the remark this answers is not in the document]
badrishc: Fixed — removed the undefined `({mode})` interpolation from all five f-strings in `test_native_allocator` (leftover from the enum→bool refactor). `python3 -m py_compile` is clean.

libs/storage/Tsavorite/cs/src/core/Utilities/BufferPool.Legacy.cs
```diff
@@ -0,0 +1,176 @@
...
+                var localPool = new ConcurrentQueue<SectorAlignedMemory>();
+                Interlocked.CompareExchange(ref queue[index], localPool, null);
+            }
+
+            if (!SectorAlignedBufferPool.Disabled && queue[index].TryDequeue(out SectorAlignedMemory page))
+            {
+#if CHECK_FREE
+                page.Free = false;
+#endif // CHECK_FREE
+                if (SectorAlignedBufferPool.UnpinOnReturn)
```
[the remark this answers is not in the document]
badrishc: Fixed — `LegacyBufferPool` now captures `unpinOnReturn` into a `readonly` field at construction (passed from `SectorAlignedBufferPool`'s per-pool captured policy) and all four Get/Return sites read the field instead of the mutable static, so a toggled static can't mix pin policies within a pool.

test/standalone/Garnet.test/NativeAllocatorServerTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        GarnetServer server;
+
+        [SetUp]
+        public void Setup() => TestUtils.DeleteDirectory(TestUtils.MethodTestDir, wait: true);
+
+        [TearDown]
+        public void TearDown()
+        {
+            server?.Dispose();
+            server = null;
```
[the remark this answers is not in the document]
badrishc: Fixed — the fixture teardown now calls `NativeAllocatorInitializer.Initialize(false)` to restore the process-global managed allocator (mirroring the Tsavorite-side native fixtures) and adds `TestUtils.OnTearDown()`, so a `true` case can no longer bleed native-backed allocation into later tests in the same process.

libs/storage/Tsavorite/cs/src/core/Utilities/BufferPool.OriginReturn.cs
```diff
@@ -0,0 +1,835 @@
...
+
+            ThreadShard shard;
+            lock (registryLock)
+            {
+                var bornSealed = poolState != PoolActive;
+                shard = new ThreadShard(this, NumClasses, sectorSize, classCaps, bornSealed);
+                if (bornSealed)
+                    shard.drainedOnce = 1;      // nothing cached; no drain, no permits held
+                else
+                    registry.Add(new WeakReference<ThreadShard>(shard));
```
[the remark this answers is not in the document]
badrishc: Fixed — `CreateShardSlow` now compacts dead (`TryGetTarget` failed) weak-references in place before appending, gated by an amortized `registryCompactThreshold` (starts at 64, resets to 2× the surviving count). A long-lived device pool under thread churn now keeps its registry bounded by the live shard count instead of growing with every historical thread.

libs/host/defaults.conf
```diff
@@ -42,6 +42,22 @@
...
+	   (per-thread) pool. The origin-return pool returns each buffer to the thread that allocated it, so it
+	   scales with concurrent IO-completion threads under a per-pool byte budget. */
+	"UseLegacyBufferPool": false,
+
+	/* Per-pool managed byte budget for the default origin-return SectorAlignedBufferPool (ignored when
+	   UseLegacyBufferPool is set). Bounds the total cacheable buffer bytes across all IO-completion threads,
+	   reserving 25% for small size-classes and 75% for large so large record/flush buffers cannot starve
+	   caching of hot small buffers. Requests above the pooled ceiling, or beyond the budget, allocate on
+	   demand and free on return (bounded memory, lower reuse). E.g. 1g, 8g. */
+	"BufferPoolMemoryBudget": "1g",
```
TedHartMS, line 61: Should have the "Set to 0 to disable buffer caching entirely..." sentence here too.

libs/server/Servers/GarnetServerOptions.cs
```diff
@@ -21,6 +21,32 @@ public class GarnetServerOptions : ServerOptions
...
+        /// is created.
+        /// </summary>
+        public bool UseLegacyBufferPool = false;
+
+        /// <summary>
+        /// Per-pool managed byte budget for the default origin-return <see cref="SectorAlignedBufferPool"/>
+        /// (ignored when <see cref="UseLegacyBufferPool"/> is set). Bounds the total cacheable buffer bytes
+        /// across all IO-completion threads, split 25% small size-classes / 75% large so large record/flush
+        /// buffers cannot starve caching of hot small buffers. Applied at startup to
+        /// <see cref="SectorAlignedBufferPool.ManagedBudgetBytes"/> before any pool is created.
```
TedHartMS: Should have the "Set to 0 to disable buffer caching entirely..." sentence here too.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -105,6 +105,56 @@ public abstract unsafe partial class AllocatorBase<TStoreFunctions, TAllocator>
...
+        /// block is freed deterministically in <c>ReturnPage</c> when the page is evicted (post-flush, epoch-drained
+        /// via OnPagesClosed — no in-flight IO), which bounds the footprint. Any still-installed blocks at Dispose
+        /// (which may have an in-flight flush/read) are handed to <see cref="nativePageRegistry"/> instead.</summary>
+        readonly DirectVmBlock[] pageBlocks;
+
+        /// <summary>Captured once at construction: whether this allocator's circular-buffer pages use the direct-VM
+        /// backend. Read on the (possibly concurrent) page-alloc/return hot paths instead of the process-global
+        /// <see cref="NativeAllocatorInitializer.Enabled"/> so an allocator never switches backend mid-life
+        /// (and never dereferences a null registry if the global flag is flipped after construction).</summary>
+        protected readonly bool useNativeLogPages;
```
TedHartMS: Same comment as for TsavoriteBase.useNativeHashIndex: There is currently no path to allow switching this mid-life; it is set only on GarnetServer.Initialize, so this part of the docstring may be obsolete.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -105,6 +105,56 @@ public abstract unsafe partial class AllocatorBase<TStoreFunctions, TAllocator>
         /// <summary>Array of pages kept to ensure the pinned pages are not garbage collected.</summary>
         protected readonly byte[][] pageArrays;
 
+        /// <summary>Per-page direct-VM backing block when direct-VM log pages are in use (<see cref="useNativeLogPages"/>);
+        /// the parallel <see cref="pageArrays"/> entry is null. Empty (default) for the managed backend. A page's
+        /// block is freed deterministically in <c>ReturnPage</c> when the page is evicted (post-flush, epoch-drained
+        /// via OnPagesClosed — no in-flight IO), which bounds the footprint. Any still-installed blocks at Dispose
+        /// (which may have an in-flight flush/read) are handed to <see cref="nativePageRegistry"/> instead.</summary>
+        readonly DirectVmBlock[] pageBlocks;
```
TedHartMS: A lot of this is duplicated in TsavoriteBase and could be consolidated into a DirectVmBlockOwner struct:
- page/tableBlocks, nativeTableRegistry, nativeIndexIoOutstanding, deferredIndexFrees and nativeIndexFreeGate, useNativeLogPages/HashIndex
More readable and easier to spot the parts that are unique to each.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -751,12 +861,131 @@ internal void VerifyRecoveryInfo(HybridLogCheckpointInfo recoveredHLCInfo, bool
...
+        /// <see cref="nativeFreeGate"/> so a block can never be enqueued into an already-drained pool (which would leak
+        /// the mapping). Cold path (page eviction, snapshot completion, teardown) — never the record read/write or
+        /// <see cref="AllocatePinnedPageArray"/> allocation path, which reads the pool lock-free via TryGet.</summary>
+        void RecycleOrFreeNativeBlock(in DirectVmBlock block)
+        {
+            lock (nativeFreeGate)
+            {
+                if (!nativeFreePoolDisposed)
+                {
+                    _ = nativeFreePagePool.TryAdd(block);
```
TedHartMS, line 903: Add a comment here (I know there's one above but that's far away) that the disposer calls DVM.Free if there's no capacity to add

libs/storage/Tsavorite/cs/benchmark/KV.benchmark/Options.cs
```diff
@@ -271,8 +298,32 @@ internal string Resolve()
             if (ValueSize > ResolvedMaxInlineValueSizeBytes)
                 return $"--value-size ({ValueSize}) exceeds --max-inline-value-size ({ResolvedMaxInlineValueSizeBytes}); values larger than the inline threshold overflow to heap and skew the benchmark.";
 
+            // Variable value sizes (--value-size-max): per-key log-uniform in [ValueSize, max].
+            ResolvedValueSizeMaxBytes = 0;
+            VariableValueSize = false;
+            if (!string.IsNullOrWhiteSpace(ValueSizeMax) && ValueSizeMax != "0")
+            {
+                var vmax = KvSize.ParseSize(ValueSizeMax);
+                if (vmax <= 0) return $"--value-size-max invalid: {ValueSizeMax}";
```
TedHartMS: crlf

libs/host/defaults.conf
badrishc: Added — `defaults.conf` now reads:

```
Requests above the pooled ceiling, or beyond the budget, allocate on
demand and free on return (bounded memory, lower reuse). Set to 0 to disable buffer caching entirely
(every IO buffer is allocated on demand and reclaimed by the GC; applies to both pool backends).
```

libs/server/Servers/GarnetServerOptions.cs
badrishc: Added, with the same wording as `defaults.conf` so the two stay in sync:

```csharp
/// buffers cannot starve caching of hot small buffers. Set to 0 to disable buffer caching entirely
/// (every IO buffer is allocated on demand and reclaimed by the GC; applies to both pool backends).
```

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
badrishc: You're right that production never flips it mid-life — it's set once in `GarnetServer.Initialize`. But the capture isn't obsolete, because **test fixtures do flip the process-global between runs in a single process** (`NativeHashIndexTests` and `NativeAllocatorServerTests` toggle `NativeAllocatorInitializer.Enabled` in `[SetUp]`/`[TearDown]`). Without capturing, a store built while the global was `true` could later observe `false` and take the managed branch against direct-VM state — or dereference a now-null owner.

Reworded both docstrings to say exactly that rather than implying a production mid-life switch:

> Production sets that global once at startup (GarnetServer.Initialize) and never changes it, but test fixtures do flip it between runs in a single process, so capturing keeps this allocator on the backend that was in effect when it was built.

Also, following your consolidation suggestion below, both flags are now *derived* rather than separately captured — `useNativeLogPages => nativeLog is not null` and `useNativeHashIndex => nativeIndex is not null`, where those owners are assigned once at construction. So there is now a single source of truth per surface.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
badrishc: Done — added `DirectVmBlockOwner` in `libs/storage/Tsavorite/cs/src/core/Native/DirectVmBlockOwner.cs`, and rewired both `AllocatorBase` and `TsavoriteBase` onto it.

**Consolidated state:** `Blocks[]` (page index / table version), `Gate`, the outstanding-IO counter, the deferred-free list, and the finalization `NativePageBlockRegistry`.

**Consolidated operations:** `BeginIo` / `EndIo` / `TryDeferFree` / `TryDrainDeferred` / `HandOffInstalledBlocks` / `HandOffDeferredBlocks` — i.e. the whole retire → park → drain → hand-off-at-teardown protocol, which was genuinely identical on both sides.

**What stayed unique** (and is now much easier to see, which was your point):
* the log recycles a quiesced block into a bounded `nativeFreePagePool` to avoid munmap+mmap+re-fault on the free-head/alloc-tail churn; the hash index unmaps a superseded table outright;
* the two diagnostic counter pairs.

I made it a `sealed class` rather than a struct: it owns a lock object, a list and a registry — all reference state — and a null owner now cleanly denotes the managed backend, which removed the separate `useNativeLogPages` / `useNativeHashIndex` fields (they're properties over it now). None of these paths are hot (checkpoint issuance, page eviction, teardown), so the indirection is free. Happy to switch it to a struct if you'd rather.

Net effect in the two core classes: **−196 / +186** lines, plus the ~120-line shared type.

Validation after the refactor: Tsavorite **1527** tests across all 7 test projects (`test.hlog` 547, `test.recovery` 197, `Tsavorite.test` 269, `test.recordops` 220, `test.session` 155, `test.session.context` 124, `test.epoch` 15) and **Garnet.test 1038** — all green; hlog/recovery/core re-run on net10.0 as well. Both `dotnet format` gates clean. I also had an independent reviewer diff the before/after specifically for lock-scope, counter short-circuit, deferred-list nullability, re-check-under-lock, teardown ordering, empty-block handling, and the field→property change; no semantic drift found. The one intentional difference: `TryDeferFree` treats a *negative* counter as "IO outstanding" and parks, where the old code's `> 0` test would have unmapped — fail-closed.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorBase.cs
badrishc: Added at the call site:

```csharp
// At capacity, TryAdd hands the block to the pool's disposer, which is DirectVirtualMemory.Free
// (see the nativeFreePagePool construction) — so an overflow block is unmapped, never leaked.
_ = nativeFreePagePool.TryAdd(block);
```

libs/storage/Tsavorite/cs/benchmark/KV.benchmark/Options.cs
badrishc: Split into two lines:

```csharp
if (vmax > ResolvedMaxInlineValueSizeBytes)
    return $"--value-size-max ({vmax}) exceeds --max-inline-value-size (...)";
```

...and likewise for the two adjacent one-liner `if (...) return $"...";` pairs, so they match the style of the surrounding validators.

One check: I read "crlf" as *"insert a line break here"* — there are no literal CR bytes in the file (verified in both the working tree and the committed blob, and `.gitattributes` has `*.cs text`). Let me know if you actually meant a line-ending problem and I'll dig further.

libs/storage/Tsavorite/cs/src/core/Utilities/BufferPool.OriginReturn.cs
```diff
@@ -0,0 +1,1036 @@
...
+            var owner = bucket.owner;
+
+            // Small class, same-thread owner (this pool's shard is in our slot, still alive)? -> owner-only local
+            // push, no atomics. On this path the bucket is reachable via page.originBucket and owned by this live
+            // thread, so no finalizer race is possible and no GC.KeepAlive is needed.
+            var arr = t_shards;
+            var slot = slotIndex;
+            if (arr is not null && slot < arr.Length && ReferenceEquals(arr[slot], owner) && owner.state == ThreadShard.Alive)
+            {
+                PushLocal(bucket, page);
```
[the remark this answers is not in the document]
badrishc: Good catch — this was a **real bug**, and it turned out to be the cause of an intermittent CI failure (`ConcurrentFreeDuringActiveTrafficIsSafe`, `Expected 0 / But was 4608`). Fixed.

**Mechanism.** As you say, the owner check isn't atomic with `PushLocal`. `PushLocal` is a non-atomic read-then-write pair:

```csharp
page.next = bucket.localHead;
bucket.localHead = page;
```

so a racing `SealAndDrainShard` can `Interlocked.Exchange(ref bucket.localHead, null)` *between* those two lines, and the owner's subsequent store resurrects the buffer into a bucket that has already been drained. The permit was then stranded permanently, because `SealAndDrainShard` had already CAS'd `drainedOnce` 0→1, so `~ThreadShard` — which already contains the exact repair walk — returned early forever.

**Fix.** `SealAndDrainShard` now re-arms the flag for shards `Free()` did not own, handing repair back to the finalizer:

```csharp
if (!ownedByCurrentThread)
    Volatile.Write(ref shard.drainedOnce, 0);
```

This is safe because permit release is CAS-exactly-once, and no finalizer can race the drain (`Free` keeps the shard reachable via its live-shard list).

**Evidence.** I reproduced it deterministically by injecting a delay between the read and the write inside `PushLocal`, then ran a control experiment (40 iterations each):

| | stranded pre-GC | stranded post-GC |
|---|---|---|
| fix disabled | 39/40 | **39/40 — permanent** |
| fix enabled | 5/40 | **0/40** |

Worth noting for the record: it was never a memory leak or a use-after-free. The stranded permit sits on a dead `BudgetState`, and the buffers unroot on the owner's next `Get` via the `ReferenceEquals(s.pool, this)` identity check. The observable symptom was only that a pool freed under concurrent traffic read its budget permanently short.

Also hardened the test that caught it (retries 20 → 40) and added `ManyShortLivedPoolsDoNotAccumulatePerThreadState`, which runs 200 sequential pool lifetimes and asserts the budget unwinds each time.

libs/storage/Tsavorite/cs/src/core/Utilities/BufferPool.OriginReturn.cs
```diff
@@ -0,0 +1,1036 @@
...
+                    else
+                    {
+                        // Detach atomically with respect to other interlocked accesses, then release the permits of
+                        // what we took. A racing owner pop may still resurrect a suffix (it writes localHead
+                        // non-atomically); that suffix is valid memory and is unrooted when the owner's next Get
+                        // replaces the shard or when the thread dies.
+                        var localChain = Interlocked.Exchange(ref bucket.localHead, null);
+                        bucket.localCount = 0;
+                        bucket.localBytes = 0;
+                        ReleaseChainPermits(localChain);
```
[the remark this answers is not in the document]
badrishc: Fixed by the same change as the sibling comment on the push side — `SealAndDrainShard` now re-arms `drainedOnce` for shards the caller doesn't own, so `~ThreadShard` performs the repair walk and releases any permits the racing owner's non-atomic update left behind.

One correction to the diagnosis, though: I initially assumed the **pop** side was the resurrection mechanism too, and could not reproduce it that way — 200 raced-`Free` iterations, then amplified with batched buffers, then with a `Thread.Sleep(1)` injected right after the detach exchange, then pinned to two cores with `taskset` — all produced **0** stranded permits. Injecting the delay inside `PushLocal` between its read and its write reproduced immediately. So the **push** side is the dominant mechanism; both are non-atomic read-then-write pairs and the one fix covers both. I updated the code comment to say that rather than describing the pop side.

libs/storage/Tsavorite/cs/src/core/Index/Recovery/IndexCheckpoint.cs
```diff
@@ -117,7 +123,17 @@ void FlushRunner()
...
+                            BeginNativeIndexCheckpointIo();
+                            try
+                            {
+                                device.WriteAsync(chunkStartBucket, numBytesWritten, chunkSize, AsyncPageFlushCallback, result);
+                            }
+                            catch
+                            {
+                                // Submission failed: no completion callback will fire, so release this chunk's unit here.
+                                EndNativeIndexCheckpointIo();
+                                throw;
```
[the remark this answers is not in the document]
badrishc: Correct, and fixed. `LocalMemoryDevice` invokes completion callbacks synchronously and propagates their exceptions out of `WriteAsync`, so the callback's `finally` could release the unit and then the issuer's `catch` would release it a second time, underflowing `nativeIndexIoOutstanding` — which could let a later checkpoint free a table still referenced by an in-flight write.

The fix needed one non-obvious step: `HashIndexPageAsyncFlushResult` is a **struct**, and it's boxed when handed to the device as the `object context`, so the issuer's copy and the callback's copy are distinct values — a plain `int` field could not be shared between them. So the guard is a reference-typed one-shot cell:

```csharp
public StrongBox<int> ioUnitReleaseGuard;

public readonly bool TryClaimIoUnitRelease()
    => Interlocked.Exchange(ref ioUnitReleaseGuard.Value, 1) == 0;
```

allocated per chunk in the issuance loop and claimed at all three release sites — both `catch` blocks and `AsyncPageFlushCallback`'s `finally`. This mirrors `PageAsyncFlushResult.TryClaimSnapshotUnitRelease()`, which works with a plain field only because that type is a class.

Added `IndexCheckpointIoUnitReleaseIsExactlyOnce` to `NativeHashIndexTests`, which asserts the guard is genuinely shared across the boxing boundary (a plain `int` field fails it) and that exactly one of 8 racing claimants wins.

Note this is native-hash-index-only: `EndNativeIndexCheckpointIo` early-returns for the managed backend, so managed builds were never affected.

TedHartMS (approved):

PR #2070, opened 2026-08-16
Stop the size-tracker resizer before clearing objectPages in ObjectAllocatorImpl.Dispose

### LINKED ISSUE #2069: ObjectAllocatorImpl.Dispose() nulls objectPages before the LogSizeTracker resizer is stopped, FailFasting the process
## Summary

`ObjectAllocatorImpl.Dispose()` nulls `objectPages` **before** the `LogSizeTracker` resizer is stopped, so a resizer already inside `ShiftAddresses` can reach `EvictRecordsInRange` and dereference the null array. Since #2056 that `NullReferenceException` reaches `Environment.FailFast`, which kills the process.

This is a store-shutdown race, not a data-path bug. It surfaces in CI as a whole test run aborting mid-suite with no failing assertion.

## Observed

`Garnet Standalone (ubuntu-latest, net8.0, Release, Garnet.test.complexstring)`, during `Garnet.test.HyperLogLogTests`:

```
Process terminated. OnPagesClosedWorker failed; allocator state is unrecoverable.
ClosedUntilAddress=114752, OngoingCloseUntilAddress=131072
   at System.Environment.FailFast(System.String, System.Exception)
   at Tsavorite.core.AllocatorBase`2.OnPagesClosedWorker()
   at Tsavorite.core.AllocatorBase`2.OnPagesClosed(Int64)
   at Tsavorite.core.LightEpoch.Drain(Int64)
   at Tsavorite.core.LightEpoch.BumpCurrentEpoch(System.Action)
   at Tsavorite.core.AllocatorBase`2.ShiftHeadAddress(Int64)
   at Tsavorite.core.AllocatorBase`2.ShiftAddressesWithWait(Int64, Int64, Boolean)
   at Tsavorite.core.LogAccessor`2.ShiftAddresses(Int64, Int64, Boolean)
   at Tsavorite.core.LogSizeTracker`2.ResizeIfNeeded(System.Threading.CancellationToken)
   at Tsavorite.core.LogSizeTracker`2+<ResizerTask>d__41.MoveNext()

System.NullReferenceException: Object reference not set to an instance of an object.
   at Tsavorite.core.ObjectAllocatorImpl`1.EvictRecordsInRange(Int64 startAddress, Int64 endAddress, EvictionSource source, Boolean isRecovery)
      in libs/storage/Tsavorite/cs/src/core/Allocator/ObjectAllocatorImpl.cs:line 399
   at Tsavorite.core.AllocatorBase`2.OnPagesClosedWorkerCore()
   at Tsavorite.core.AllocatorBase`2.OnPagesClosedWorker()
```

The whole run aborted after 383 passing tests; the remainder never executed.

## Mechanism

Line numbers below are against `8c54cc278` (current `main`).

`AllocatorBase.Dispose()` states the required ordering explicitly (`AllocatorBase.cs:501`):

```csharp
public virtual void Dispose()
{
    disposed = true;

    // Stop the size-tracker resizer and wait for it to exit BEFORE tearing down the epoch, buffer pool, flush event
    // (and, by the owner, the log device) that it uses; otherwise a still-running resizer can spin on or dereference
    // these cleared resources. ...
    logSizeTracker?.Stop(wait: true);
    ...
```

`ObjectAllocatorImpl.Dispose()` tears down its own resources first and only then calls that (`ObjectAllocatorImpl.cs:492`):

```csharp
public override void Dispose()
{
    var localValues = Interlocked.Exchange(ref objectPages, null);   // (1) objectPages is gone
    if (localValues != null)
    {
        freePagePool.Dispose();
        foreach (var value in localValues)
            value.Clear();                                            // (2) every ObjectIdMap cleared
        base.Dispose();                                               // (3) only now: disposed = true; Stop(wait: true)
    }
}
```

Between (1) and (3) the allocator is in a state the resizer is not prepared for: `objectPages == null` **and** `disposed == false`.

`LogSizeTracker` checks `runState` at `LogSizeTracker.cs:275` and then calls `ShiftAddresses`; a resizer already past that check runs to completion. It reaches `ObjectAllocatorImpl.cs:411`:

```csharp
var objectIdMap = objectPages[GetPageIndexForAddress(address)].objectIdMap;
```

which is the null dereference in the trace.

Two consequences worth noting:

* **`if (disposed) return;` in `OnPagesClosedWorkerCore` would not close this window.** `disposed` is set inside `base.Dispose()`, i.e. *after* the null, so the hazardous interval is exactly the one where `disposed` is still `false`. This is the most likely way a well-meaning fix lands and the flake survives.
* Step (2) gives a second, quieter signature of the same window: a resizer that already loaded its `objectIdMap` local walks a cleared map rather than crashing.

## Why it started showing up on 2026-08-12

The race is older than the crash. #2056 (`eeaa6ffc4`) changed the handler in `AllocatorBase.OnPagesClosedWorker` (`AllocatorBase.cs:1707`):

```diff
-                logger?.LogCritical(ex, "OnPagesClosedWorker failed, page closing will not resume. ...");
-                throw;
+                Environment.FailFast($"OnPagesClosedWorker failed; allocator state is unrecoverable. ...", ex);
```

Before that, the identical NRE unwound into `ResizerTask`'s catch-all and was invisible — the store was being torn down anyway, so the wedged `OngoingCloseUntilAddress` was harmless. From 2026-08-12 the same interleaving kills the host. The `FailFast` is not the bug; it made an existing one visible.

## Suggested direction

Hoist the background-work shutdown ahead of the derived teardown, e.g. a `protected void StopBackgroundWork()` on `AllocatorBase` (setting `disposed = true` and `logSizeTracker?.Stop(wait: true)`) that each derived `Dispose()` calls as its first statement. `Stop` CASes on `runState` (`LogSizeTracker.cs:173`), so the later call from `base.Dispose()` is a no-op.

I have not opened a PR: `ObjectAllocatorImpl.cs` has active owners and the epoch/resizer invariants here are yours, not mine. Happy to put one up if that is useful.

Two related observations, offered without a claim that they are the same defect:

* `SpanByteAllocatorImpl.Dispose()` (`SpanByteAllocatorImpl.cs:224`) has the identical shape — `Interlocked.Exchange(ref freePagePool, null)` before `base.Dispose()` — and `freePagePool` is reached from the same worker via `FreePage` → `ReturnPage`.
* `GarnetDatabase.Dispose` calls `Store?.Dispose()` before `SizeTracker?.Stop()`, and that `Stop` is `wait: false`, which widens the window from the Garnet side.

One thing I could **not** establish: the reported `ClosedUntilAddress=114752` / `OngoingCloseUntilAddress=131072` do not by themselves pin line 399 — `ClosedUntilAddress` is only advanced after `EvictRecordsInRange` returns for the whole page, so the arithmetic is equally consistent with an NRE in the record walk further down. The stack frame is what identifies the line.

nileshpatil6:
Fixes #2069.

`ObjectAllocatorImpl.Dispose()` clears `objectPages` first and calls `base.Dispose()` last:

```csharp
var localValues = Interlocked.Exchange(ref objectPages, null);
if (localValues != null)
{
    freePagePool.Dispose();
    foreach (var value in localValues)
        value.Clear();
    base.Dispose();
}
```

`AllocatorBase.Dispose()` is where the resizer is stopped, and it deliberately does that first, with the reason written out above the call:

> Stop the size-tracker resizer and wait for it to exit BEFORE tearing down the epoch, buffer pool, flush event (and, by the owner, the log device) that it uses; otherwise a still-running resizer can spin on or dereference these cleared resources.

`objectPages` is the same kind of resource, but it is cleared before that call runs, so the invariant does not hold for it. A resizer still inside `ResizeIfNeeded` -> `ShiftAddresses` reaches `EvictRecordsInRange`, which dereferences it directly:

```csharp
var objectIdMap = objectPages[GetPageIndexForAddress(address)].objectIdMap;
```

so the NullReferenceException in the issue's stack trace, which `OnPagesClosedWorker` escalates to `Environment.FailFast`, taking the whole process down.

This moves the stop to the top of the override, so the resizer is stopped and awaited before the array it uses is cleared:

```csharp
logSizeTracker?.Stop(wait: true);
```

Notes on why this is safe:

- `base.Dispose()` still calls `Stop(wait: true)` afterwards. `Stop` compare-exchanges `runState` from `Running`, so once the resizer has stopped the second call takes neither the signal path nor the wait loop, and returns immediately.
- The `Interlocked.Exchange` on `objectPages` still guards the rest of the method, so the dispose-once behaviour is unchanged. `Stop` is safe to reach on a second `Dispose()` for the same reason as above.
- `logSizeTracker` is `internal` on `AllocatorBase`, so no accessibility change was needed.

I did not add a test. The failure is a shutdown race, and the observable outcome is `Environment.FailFast`, which terminates the test host rather than failing an assertion, so a test that reproduces it would take the run down with it rather than report. What is testable is the ordering itself, and after this change that ordering is enforced by construction in the override.

Built `Tsavorite.core` clean, and ran the in-memory allocator tests (`Name~NativeInMem`, 11 passed / 3 skipped) to check for fallout on the normal dispose path.


nileshpatil6:
Good catch, that is a real problem with the first version and I have pushed a fix.

`Stop(wait: true)` spins on `!IsStopped && !(resizerTask is { IsCompleted: true })`, and the resizer only reaches `OnStopped()` after `ResizeIfNeeded` returns. If it was already parked in one of the `while (waitForEviction && !disposed && ClosedUntilAddress < newHeadAddress)` loops in `ShiftAddressesWithWait`, `disposed` was still false when my override called `Stop`, so the release condition the base method's comment relies on was not in place yet. That is exactly the invariant the comment above the base call spells out, and calling `Stop` from the override skipped the half of it that makes the wait safe.

Fixed as suggested. Both statements now live in one protected helper on `AllocatorBase`:

```csharp
protected void StopSizeTrackerForDispose()
{
    disposed = true;
    logSizeTracker?.Stop(wait: true);
}
```

`AllocatorBase.Dispose()` calls it in place of the two lines it had, so its ordering is byte-for-byte what it was, and `ObjectAllocatorImpl.Dispose()` calls the same helper before the `Interlocked.Exchange` on `objectPages`. The override now gets `disposed = true` first as well, so a parked resizer is released rather than waited on.

It is idempotent for the second call from `base.Dispose()`: `disposed` is already true, and `Stop` compare-exchanges `runState` from `Running`, so once the resizer has stopped it takes neither the signal path nor the wait loop.

`disposed` is `private volatile` on `AllocatorBase`, so a helper was needed regardless; the derived class could not set it directly.

Rebuilt `Tsavorite.core` clean and re-ran the in-memory allocator tests (`Name~NativeInMem`): 11 passed / 3 skipped on both net8.0 and net10.0.

TedHartMS (approved):

TedHartMS:
I believe you have completed addressing the Copilot comment; please Resolve if so

nileshpatil6:
Yes, that one is fully addressed. Resolved it.

Summary of where the PR stands: `AllocatorBase` now has a protected `StopSizeTrackerForDispose()` that sets `disposed` and then stops the resizer with `wait: true`. `Dispose()` calls it in place of the two statements it had, so its ordering is unchanged, and `ObjectAllocatorImpl.Dispose()` calls the same helper before the `Interlocked.Exchange` on `objectPages`. That closes the original FailFast (the resizer can no longer reach a cleared `objectPages` through `EvictRecordsInRange`) without the hang Copilot pointed out, since `disposed` is now set before the wait in both paths.

Happy to make any further changes if you would prefer a different shape.

nileshpatil6:
Thanks for the approval. Merged `main` in again to clear a conflict that appeared after it.

The conflict was in `AllocatorBase.cs`: #2081 added `TryCompleteMine()` in the same place this PR adds `StopSizeTrackerForDispose()`. Both are kept, with `TryCompleteMine()` left next to `TryComplete()` where it landed and the helper immediately above `Dispose()`. Nothing else in the resolution, and the diff against `main` is still only the two allocator files.

Rebuilt `Tsavorite.core` clean and re-ran the in-memory allocator tests on net10.0: 11 passed, 3 skipped, unchanged from before the merge.

nileshpatil6:
Correction to the comment above: the PR that added `TryCompleteMine()` was #2018 ("[Storage] Optimize IOPS for RAID-0 NVMe disks"), not #2081. The resolution itself is unchanged.

TedHartMS (approved):

nileshpatil6:
@TedHartMS thanks for the merge from `main` and the approval.

The one remaining red check, `Garnet Cluster (windows-latest, net8.0, Debug, Garnet.test.cluster.vectorsets)`, looks environmental rather than related to this change. 83 of 84 tests in that job passed; the failure is `ClusterVectorSetTests.MultipleReplicasWithVectorSetsAsync` asserting on a primary/replica count mismatch (`[127.0.0.1:7600]: 1,544088 != [127.0.0.1:7605]: 1,542456`) after a 3 minute run, with `AofSyncTask` logging `TaskCanceledException` during teardown. That is replica sync timing, not allocator disposal.

The same suites are failing independently of this PR right now:

| run | branch | failing jobs |
| --- | --- | --- |
| 32304693275 | `main` | cluster.replication, cluster.vectorsets, cluster.multilog |
| 32387481110 | `main` | collections, cluster.multilog |
| 32379659413 | `users/kmontrose/vectorSetRenameAOFFixes` | cluster.vectorsets on three platforms, cluster.replication |
| 32390622962 | `tedhar-fix-checkpoint-cleanup-crash` | vectorset |

This change only reorders disposal inside `ObjectAllocatorImpl.Dispose`, setting `disposed = true` before `logSizeTracker.Stop(wait: true)` so a resizer parked in `ShiftAddressesWithWait` cannot hang, and it does not touch replication or the AOF sync path.

Could you re-run that single job when convenient?

PR #2095, opened 2026-08-31
Fix live-object mutation and scan liveness defects (#2092)

### LINKED ISSUE #2092: List key corrupts to a garbage String value (WRONGTYPE) after a connection drop
## Summary
A List key that is only ever written via `RPUSH` occasionally comes back as a String
holding a 4-byte payload of `0xFF 0xFF 0xFF 0xFF` (the bit pattern for `-1` / `UINT32_MAX`)
after `DUMP`. Every write to that key from then on fails with `WRONGTYPE`. The corruption
is always preceded by a `ConnectionError: Connection closed by server.` on a write in the
same key family.

The corruption is confined to a `log:*` key family — `log:scans`, `log:errors`,
`log:details` — and has not been observed on any other key type or naming pattern in the
same database.

## Environment
- Garnet 2.1.5
- Config: `EnableAOF: true`, `Recover: true`, `MemorySize: 16mb`, `ObjectMemorySize: 16mb`
- Client: Python `redis.asyncio` (redis-py 8.0.0), `BlockingConnectionPool(max_connections=10, timeout=1)`
- OS: Windows 10 IoT Enterprise LTSC 2021 with all current updates
- CPU: Intel Celeron Dual Core
- RAM: 8 GB

## Background on the `log:*` key family and what precedes the corruption
Every `log:*` key is written the same way: a plain `RPUSH log:<name> <csv line>`. The
client's error-logging path treats a failure on one `log:*` key as a reason to immediately
write to a *different* `log:*` key, on the same shared connection pool:

1. `RPUSH log:scans ...` fails with `ConnectionError: Connection closed by server.`
   (the command was already sent; the connection died before the reply came back)
2. handling that exception calls `logger.error(...)`, which — via a logging handler —
   immediately fires `RPUSH log:errors ...` on the same pool, in reaction to the very
   failure being logged
3. seconds later, the next `RPUSH log:scans ...` (or `log:errors`, `log:details`, ...)
   comes back `WRONGTYPE`

This sequence — a write failing mid-flight, immediately followed by another write to a
sibling key on the same pool — has occurred every time this corruption has been observed.
Every time, all three `log:*` keys have come back corrupted with the exact same 16-byte
`DUMP` payload, including `log:details`, which nothing in that code path had written to
during the failing operation.

Attempts to reproduce this on demand have been unsuccessful. Everything above is drawn
from production occurrences. Frequency in production is once or twice a day total across
28 systems processing approximately 30,000 to 50,000 log records each.

## Client-side concurrency and commands used against these keys
The client runs on a single asyncio event loop. Any log call arriving from a different OS
thread hands off to it via `asyncio.run_coroutine_threadsafe`. The full set of commands
issued against `log:*` keys is `RPUSH` (real-time writes), `LRANGE` (a periodic save
reading entries before persisting them to a log file), and then a pipeline of one
`LMOVE log:<name> events:<name> LEFT RIGHT` per entry read, clearing the key by moving
each entry off it.

## Client-side workaround (doesn't address root cause)
Every `log:*` write now detects `WRONGTYPE`, confirms server-side that the key really is
the wrong type (never blind-deletes), deletes it, and retries the write once:

```python
async def _add_log_entry_async(self, log: str, entry: str):
    try:
        await self.database.add_log_entry_async(log, entry)

    except redis_exceptions.ResponseError as exception:
        if not await self._recover_from_wrong_type_async(log, exception):
            raise

        await self.database.add_log_entry_async(log, entry)

async def _recover_from_wrong_type_async(self, log: str, exception: Exception) -> bool:
    # A log:* key is only ever written via RPUSH, so WRONGTYPE here means the
    # key has been corrupted into some other type by something outside our
    # control and should be deleted.
    if "WRONGTYPE" not in str(exception):
        return False

    key = f"log:{log}"
    type_name = await self.database.type_async(key)
    if type_name == "list":
        return False

    await self.database.delete_async(key)
    return True
```

This keeps production from blocking on the failure. But it's a workaround, not a fix.

Random, but triggered as noted above.

### Expected behavior

No WRONGTYPE corruption.

### Screenshots

_No response_

### Release version

V2.1.1 - v2.1.5

### IDE

Python Redis 8.0.0

### OS version

Windows 10 IoT Enterprise LTSC 2021

### Additional context

I'm happy to try to assist in testing or troubleshooting.

badrishc:
Fixes #2092

## The reported symptom

List keys intermittently and **permanently** became Strings holding a 4-byte `0xFF FF FF FF` payload. Every subsequent write returned `WRONGTYPE`, and the condition survived restart.

## Root cause

`StorageSession.ListMove` called `GET` to obtain the `IGarnetObject` and then mutated the returned heap object in place:

```csharp
srcListObject.LnkList.RemoveLast();
```

That is only legal when the record is in the mutable region (`>= ReadOnlyAddress`). Otherwise it races the object log writer and throws `InvalidOperationException: Collection was modified` inside `ObjectAllocatorImpl.WriteAsync`. The page write is never issued, `FlushedUntilAddress` stalls, and the log wedges.

The `WRONGTYPE` is the *aftermath*, not the bug. On restart with recovery, the newest persisted record for the key is an older CopyUpdate source that was legitimately cleared to an inline `ObjectIdMap.InvalidObjectId` — four `0xFF` bytes, which read back as a perfectly valid **String**.

Repro (standalone server + concurrent `LMOVE` stress):

| | ops completed | net errors | flush exceptions |
|---|---|---|---|
| before | 13,141 | 285 | yes |
| after | 423,467 | **0** | **0** |

## Audit for the same anti-pattern

Per review feedback the audit covered every object operation. The pattern — `GET` an object, then mutate its collection directly — appeared in **exactly three** places:

| Site | Commands | Fix |
|---|---|---|
| `StorageSession.ListMove` | `LMOVE`, `RLMOVE` | route pop/push through the existing `ListPop`/`ListPush` RMW helpers |
| `StorageSession.SetMove` | `SMOVE` | route through RMW |
| `CollectionItemBroker` | all blocking collection commands | replaced 3 live-mutating helpers with `TryGetNextListResult` / `TryGetNextSortedSetResult` |

Four further defects surfaced while fixing those:

- **`SMOVE` destroyed the member** when the destination held a wrong type. Now returns `WRONGTYPE`.
- **Same-key singleton `LMOVE`/`BLMOVE` discarded the key TTL.** The no-op fast path now also covers `Count == 1`.
- **`BLMOVE src dst` never woke a client blocked on `dst`.** Reproduced on unmodified `main`, so this is pre-existing. The broker event is now enqueued after `txnManager.Commit(true)`, which also avoids re-entering `keysToObserversLock`.
- **A collection update served only one blocked waiter.** `TryAssignItemFromKey` returned immediately after the first `HandleSetResult`, so one event woke one observer regardless of how many items were available. Pre-existing on `main`. With three clients blocked, `LPUSH k a b c` woke **1 of 3** and stranded two items, and a no-op `BLMOVE k k LEFT LEFT` woke **1 of 3** with the item still present; Redis serves all three in both cases. The observer queue is now drained until the collection can no longer serve the observer at the head. That observer is left queued rather than skipped, which keeps the queue FIFO and removes a latent infinite loop in the old `if (currCount > 0) continue;` path, which re-peeked the same undequeued observer forever.
- **`SortedSetObject` / `HashObject` mutated `expirationTimes` during `DoSerialize`.** Serialization must be a pure read: the CAS state machine in `HeapObjectBase.Serialize` coordinates serializer-vs-serializer and serializer-vs-CopyUpdate only — plain readers never touch `SerializationPhase`, and `ObjectLogWriter.DoSerialize` takes no record lock. Confirmed empirically by instrumenting `DoSerialize` under `ZEXPIRE` load on a tiered-storage server. `DoSerialize` is now **two-pass and non-mutating**. Both passes share one `now` and one snapshot of the `expirationTimes` reference, because the element count is written *before* the entries and the writer is forward-only, so it cannot be back-patched. **Emitted bytes are byte-identical.**

## Separate defect: `KEYS` listed deleted keys

`ScanLookup`'s dedup step (`ConditionalScanPush`) is a **recency** test — *"is a newer version of this key reachable from the tag chain below `maxAddress`?"* — not a **liveness** test.

Deletion normally hides a key via two mechanisms, and elision defeats both:

```
SET key v0     -> A                             chain: HB -> A
EXPIRE key ttl -> RCU writes B, elides A        chain: HB -> B
                  A = SealAndInvalidate  (Invalid, NOT a tombstone)
DEL key        -> B.SetTombstone(); B elided    chain: HB -> (empty)
```

The iterator walks **raw log bytes**, not the chain, so both records are still visited. `B` is caught by the existing tombstone filter. `A` is not a tombstone, so it reaches `ConditionalScanPush`, `FindTag` hits an empty chain, nothing newer is found — and the stale record is pushed. The tombstone that was supposed to suppress `A` is exactly what elision removed.

### The fix, and why this specific predicate

`CanElide` requires the record to be the **sole member** of its tag chain:

```csharp
stackCtx.hei.Address == stackCtx.recSrc.LogicalAddress
    && srcRecordInfo.PreviousAddress < hlogBase.BeginAddress
```

and `TryElide` then CASes the bucket slot to `kInvalidAddress`. So an elided chain is **exactly** an empty slot, and `FindTag` failing is an exact test for *"this key is unreachable."* `OperationStatus.NOTFOUND` is set **only** on that path (`TryFindRecordInMainLogForPendingOperation` only ever sets `SUCCESS` or `RETRY_LATER`), so it is an unambiguous discriminator.

The whole change is 3 lines in one file:

```csharp
if (internalStatus == OperationStatus.NOTFOUND)
    return Status.CreateFound();
```

**This keys off reachability of the _key_, not of the _record_** — which is what makes it safe for point-in-time consumers. If the key was re-inserted above `maxAddress`, as during slot migration, the chain is non-empty and the record is still pushed.

> **Rejected alternative — filtering on `RecordInfo.Invalid`.** That is a *per-record* test. `Invalid` marks the elided record while the key may remain live at a higher address, so it **silently drops keys during slot migration**. `ClusterMigrateCustomProcDelRMW` catches this: `DELRMW` elides record A and writes B above the captured scan tail, the key never enters the migration sketch, and it is lost. Because it keys off the key instead, the predicate here needs no opt-in flag and fixes the same defect for the other bounded-`maxAddress` callers (`DeleteSlotKeys`, `HasKeysInSlots`, `VectorManager.Cleanup`, streaming snapshot).

## Tests

Every test below was verified to **fail without** its corresponding fix.

- `RespListTests` / `RespSetTest` — object-mutation and TTL regressions
- `RespBlockingCollectionTests` — broker correctness and the `BLMOVE` destination-notification gap
- `GarnetObjectTests` — non-mutating serialization
- `RespScanCommandsTests` — 3 `KEYS` regressions, asserting `KEYS` + `SCAN` + `DBSIZE` + `EXISTS` + `GET` under `Assert.Multiple` so no single view masks another
- `SpanByteIterationTests.SpanByteIterateLookupSnapshotEmitsKeyRcudAboveSnapshotTail` — locks the guarantee that an RCU landing above the captured tail does not lose the key

Four test assertions that had encoded the old `EXPIRE(key, TimeSpan.Zero)`-on-empty behavior (which left a phantom record) were updated.

## Validation

| Suite | Result |
|---|---|
| `Garnet.test` | 1049 / 0 |
| `Garnet.test.collections` | 775 / 0 |
| `Garnet.test.cluster` | 156 / 0 |
| `Garnet.test.cluster.migrate` | 56 / 0 |
| Tsavorite: base / hlog / recovery / recordops / session / session.context / epoch | 290 / 560 / 204 / 220 / 155 / 124 / 15 — all 0 failures |
| `dotnet build Garnet.slnx` | 0 errors, 0 warnings |
| `dotnet format` (Garnet + Tsavorite) | clean |

Run on `net10.0` Debug.

## Notes for reviewers

- The Tsavorite change is 12 lines in `AllocatorScan.cs` (9 of them comment) with **no public API change**.
- `ConditionalCopyToTail` (compaction) is deliberately untouched; only `ConditionalScanPush` is affected.
- Latent issues found but **not** fixed here, to keep the diff scoped — happy to file follow-ups: `DEL` on an already-expired key returns `1` where Redis returns `0`; `SMOVE key key member` returns `0` where Redis returns `1` (currently encoded in our own tests); `SortedSetIntersection` returns the live `Dictionary` via `out` for the single-key/no-weights case; the public mutable collection properties (`ListObject.LnkList`, `SetObject.Set`, `SortedSetObject.Dictionary`) would be better as read-only views; and eight redundant `EXPIRE(key, TimeSpan.Zero)` delete-on-empty hacks remain.


libs/server/Objects/ItemBroker/CollectionItemBroker.cs
```diff
@@ -383,125 +383,179 @@ private bool TryAssignItemFromKey(byte[] key)
...
+                        // list unchanged. Returning without the pop/push also avoids emptying the key, which would
+                        // drop its TTL.
+                        if (asKey.ReadOnlySpan.SequenceEqual(dstKey.ReadOnlySpan) &&
+                            (srcDirection == dstDirection || currCount == 1))
+                        {
+                            var unchanged = srcDirection == OperationDirection.Right
+                                ? srcList.LnkList.Last.Value
+                                : srcList.LnkList.First.Value;
+                            result = new CollectionItemResult(key, unchanged.AsSpan().ToArray());
+                            return true;
```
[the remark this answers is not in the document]
badrishc: Good catch, and the underlying defect is broader than this fast path. Fixed in b380605.

The starvation does not originate here — it is in `TryAssignItemFromKey`, which is unchanged from `main` in this PR. It `return true`s immediately after `HandleSetResult`, so a collection-update event serves **at most one** observer no matter how many items the collection can still supply. This fast path just happens to make it visible.

Confirmed against the pre-fix build, three clients blocked on each key:

| scenario | trigger | served |
|---|---|---|
| `BLPOP k` ×3 | `LPUSH k a b c` (one command, 3 items) | **1 / 3** — 2 items left stranded in the list |
| `BLMOVE k k LEFT LEFT` ×3 | `LPUSH k v1` | **1 / 3** — item still present |

So the first row has nothing to do with this fast path, and enqueueing another event from here would have fixed only the second. Redis serves every client the key can satisfy in both cases.

Fixed at the root instead: `TryAssignItemFromKey` now keeps draining the observer queue after a successful assignment, and stops when the collection can no longer serve the observer at the head. Both scenarios now report 3/3.

Two details worth noting:

- The old `if (currCount > 0) continue;` failure path `continue`d **without dequeuing**, so `TryPeek` would return the same observer forever. It is now a `break`: the head observer stays queued so the queue remains FIFO, and every iteration either dequeues an observer or breaks, which makes termination obvious. The only reachable `currCount > 0` failures are an unknown `BLMOVE` direction (rejected at parse time) and an unexpected `ListPop` failure; source type mismatch already returned with `currCount == 0`, so both old and new code stop there.
- `break` also lets the empty-queue cleanup below run, which the old `return true` / `return false` skipped.

Regression tests added in `RespBlockingCollectionTests`: `BlockingPopServesAllWaitersFromSingleMultiItemPush` (BLPOP and BRPOP) and `BlockingListMoveSameKeyServesAllWaiters`. All three fail on the pre-fix build with the assertion messages quoted above and pass with the fix.

Validation: `Garnet.test.collections` 778/0 (775 baseline + 3 new), `Garnet.test` 1049/0, `Garnet.test.cluster` 156/0, `Garnet.slnx` builds clean, `dotnet format` clean.

badrishc:
Addressed the review comment, plus a test-quality fix it surfaced.

**1. `CollectionItemBroker` — one waiter served per update event** (`b380605`)

Root-caused past the reported symptom. `TryAssignItemFromKey` is unchanged from `main` in this PR and `return true`s after the first `HandleSetResult`, so one collection-update event served **one** observer regardless of how many items were available. Measured on the pre-fix build with three clients blocked per key:

| scenario | trigger | before | after |
|---|---|---|---|
| `BLPOP k` ×3 | `LPUSH k a b c` | 1 / 3, two items stranded | 3 / 3 |
| `BLMOVE k k LEFT LEFT` ×3 | `LPUSH k v1` | 1 / 3, item still present | 3 / 3 |

The first row does not involve the flagged fast path at all, so notifying from that path would have fixed only half the defect. The observer queue is now drained until the collection can no longer serve the observer at the head; that observer is left queued rather than skipped, keeping the queue FIFO and removing a latent infinite loop in the old `if (currCount > 0) continue;` path, which re-peeked the same undequeued observer forever.

New tests, all three verified failing on the pre-fix build: `BlockingPopServesAllWaitersFromSingleMultiItemPush` (BLPOP, BRPOP) and `BlockingListMoveSameKeyServesAllWaiters`.

**2. Flaky serialization tests** (`2000817e`)

`SerializeDoesNotMutateSortedSetWithExpiredMembers` failed on `windows-latest/net8.0/Debug`. My bug, and a real test defect rather than a bad flake: the member got a 200 ms TTL when its bytes were built, but the deserializing constructor drops members that are already expired (`SortedSetObject.cs:185`), so a stalled runner produced an object with nothing expirable. The object is now built inside a retry that widens the window until the member survives construction, then waits on the member's own expiration timestamp instead of a fixed sleep. Verified by forcing attempt 0 to miss, which reproduces the CI failure exactly and now recovers. `HashObject.HasExpirableItems` became internal so the hash test uses the same signal, which also lets it assert that serialization leaves the expiration structures intact.

**Status:** 226 checks passing, 0 failing.

One unrelated job, `Garnet.test.cluster.replication.tls` on `ubuntu/net10.0/Release`, timed out in `ValidateNodeObjects` -> `BackOff` waiting for replica convergence, with a different test each occurrence; it passed on re-run. Confirmed not caused by this PR: the full TLS suite passes locally in that exact configuration (103/0), and every change made since the last all-green run is unreachable from it — `TryAssignItemFromKey` requires a registered observer and `cluster.replication.tls` issues no blocking commands, and the `HasExpirableItems` visibility change is not behavioral. For the same reason the earlier `DoSerialize` change is not implicated: its extra pass is skipped when `expirationTimes is null`, which is the case for the objects these tests build, and `ConditionalScanPush` is reachable only from `ScanCursor`, not from checkpointing or replication.

libs/server/Objects/Hash/HashObject.cs
TedHartMS, line 75: All *Object that have expirationTime/Queue need comments explaining their purpose and interaction

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorScan.cs
```diff
@@ -305,6 +305,18 @@ internal Status ConditionalScanPush<TInput, TOutput, TContext, TSessionFunctions
...
+            // NOTFOUND is set only when FindTag found no hash-index entry for this key's tag, which means no record for
+            // this key is reachable by any operation: the key's whole tag chain was elided (by a Delete, or by an RCU in
+            // InternalRMW/InternalUpsert, both of which SealAndInvalidate the source when CanElide holds) and it was not
+            // re-inserted. The record is still physically in the log, and the search above only asks "is there a *newer*
+            // version of this key?", so without this check it would correctly find nothing newer and push a dead record.
+            // Note this keys off the reachability of the *key*, not of this record: if the key was re-inserted -- possibly
+            // above maxAddress, as during slot migration -- the chain is non-empty, so we still push and the key is
+            // reported. That distinction is what makes this safe for point-in-time consumers (migration, streaming
+            // snapshot), which must never lose a key that is live outside the scanned range.
+            if (internalStatus == OperationStatus.NOTFOUND)
```
TedHartMS, line 317: I believe this can only happen when the scan has been run with includeClosedRecords == true and the record being returned is Invalid; otherwise, the HashBucket would not be empty. Suggest adding inside the if block a Debug.Assert(srcLogRecord.Invalid, "Expected an invalid record if its key's HashBucket is empty");

libs/server/Objects/Hash/HashObject.cs
badrishc: Done in a7316af6c4. `HashObject` and `SortedSetObject` are the only two `*Object` types with these structures, so both are now commented.

The comments cover purpose and interaction:

- The two structures are allocated together by `InitializeExpirationStructures`, are null until the first TTL is set, and are torn back down together by `CleanupExpirationStructuresIfEmpty`. `HasExpirableItems` tests `expirationTimes` and guards every access to either.
- `expirationTimes` is the source of truth for whether a field/member is expired.
- `expirationQueue` orders the same expirations soonest-first so lazy cleanup only inspects the head. Since `PriorityQueue` has no update operation, resetting a TTL enqueues a second entry and leaves the stale one behind, and removing a field leaves its entry behind entirely, so entries are only hints -- `DeleteExpiredItemsWorker` re-checks each dequeued entry against `expirationTimes` and discards it if the field is gone or now carries a different expiration.
- `expirationTimeSpanLookup` (hash only) is a `ReadOnlySpan<byte>`-keyed view of `expirationTimes` that avoids allocating a `byte[]` per lookup, and follows its lifetime.

libs/storage/Tsavorite/cs/src/core/Allocator/AllocatorScan.cs
badrishc: Added in a7316af6c4 (as `srcLogRecord.Info.Invalid`, since `ISourceLogRecord` exposes `Info` rather than `Invalid` directly).

Your reasoning checks out on both halves:

- **Closed records only reach here under `includeClosedRecords`.** `ScanCursor` builds the iterator with `includeClosedRecords: maxAddress < long.MaxValue` (`SpanByteAllocatorImpl.cs:304`, `ObjectAllocatorImpl.cs:1608`), and otherwise `var skipOnScan = !includeClosedRecords && recordInfo.SkipOnScan` filters them out (`SpanByteScanIterator.cs:229`, `ObjectScanIterator.cs:230`), where `SkipOnScan => IsClosedWord(word)` covers Invalid **or** Sealed.
- **Empty bucket implies elided implies Invalid.** `CanElide` requires the record to be the sole member of its chain, and `TryElide` then CASes the slot to `kInvalidAddress` after `SealAndInvalidate`, so an empty bucket and an Invalid record go together.

I made sure the assert isn't passing vacuously: temporarily flipping it to `Debug.Assert(false)` failed **exactly** the three new KEYS regression tests and nothing else, which confirms this branch is the one they exercise. With the real assert, Debug runs are clean -- `Garnet.test` 1049/0, `Garnet.test.collections` 778/0, `Garnet.test.cluster.migrate` 56/0, `Tsavorite.test.hlog` 560/0.

One caveat worth your call, since it's your invariant: the scan iterator copies the record into bufferPool memory before returning it, so `srcLogRecord.Info` is a snapshot. A record that was Valid and chain-reachable when snapshotted, then elided by a concurrent delete before `FindTag` runs here, would reach this branch with a stale Valid `Info` and trip the assert. I stress-tested that window (4 writers looping SET/EXPIRE/DEL against 3 concurrent `KEYS *` for 25s) without tripping it, and it's Debug-only, so I'm happy to leave it as-is -- just flagging it in case you'd rather it were weakened.

badrishc:
CI status note for a7316af6c4: the run shows red, but **every `Run tests` step passes**. The remaining failures are all the `Upload test results` step returning `Failed to FinalizeArtifact: (403) Forbidden` from GitHub's artifact service -- an infrastructure error, not a test result. It hit unrelated suites too (including Release-only Tsavorite epoch tests).

The first attempt did have one genuine test failure, `ClusterManagementTests.PrimaryUnavailableRecoveryAsync(Aof_Sync_Task_Consume,True)`, which passed on rerun. It's a pre-existing flake, not related to this PR:

- The same test fails identically on `main` without these changes -- run 33837610174, job `Garnet Cluster (windows-latest, net10.0, Debug, Garnet.test.cluster)`, same `Cancellation Requested` error and same `ClusterTestUtils.BackOffAsync:774` stack, differing only in the parameterization.
- It times out in `SimpleSetupCluster` -> `WaitForSyncAsync`, i.e. during gossip convergence at cluster formation, before any data operation runs -- nothing this PR touches.
- All 6 parameterizations pass locally in Debug.

Local Debug validation for this commit: `Garnet.test` 1049/0, `Garnet.test.collections` 778/0, `Garnet.test.cluster` 156/0, `Garnet.test.cluster.migrate` 56/0, `Tsavorite.test.hlog` 560/0, plus `dotnet format --verify-no-changes` clean on both solutions.

libs/server/Objects/Hash/HashObject.cs
TedHartMS: Out of scope for this PR but only HashObject has expirationTimeSpanLookup; it looks like the same could be applied to SortedSetObject.

TedHartMS (approved):

badrishc:
Followed up on the `Tsavorite.test.hlog` flake I earlier attributed to pre-existing noise, and root-caused it instead. Fixed in 113d7e3d74.

**`CommitRecordBoundedGrowthTest` -> `TsavoriteException: Error reading page 0 from device`**

The test builds its `TsavoriteLog` without setting `TryRecoverLatest`. That defaults to `true`, and `TsavoriteLog(logSettings, logger)` forwards it as `syncRecover`, so the constructor runs a synchronous recovery -- even though this test always starts from an empty log and never intends to recover.

`TestUtils.MethodTestDir` is keyed on `TestContext.CurrentContext.Test.MethodName` only, with no parameters, so all parameterizations of this test share one directory. A commit written late by the previous parameterization can land there after `BaseSetup`'s `DeleteDirectory` has cleaned it. Recovery then finds real commit metadata (`info.UntilAddress > 0`, so it does not take the "Unable to recover using any available commit" early return) and replays it onto this run's fresh, empty `LocalMemory` device, which fails reading page 0.

The fix sets `TryRecoverLatest = false`, matching the fixture's other two tests, which already disable recovery. This is a deterministic elimination rather than a timing tweak: with `syncRecover` false the constructor never calls `RecoverAsync`, so the entire failing stack is unreachable.

Evidence: over 60 runs of the test's parameter set, **2 failures before the change and 0 after** (control isolated to just this line), plus 5 consecutive clean full-suite runs (580 passed each).

Separately, while chasing this I found my earlier post-rebase hlog numbers were run against a stale binary -- `test.hlog` is not part of `Garnet.slnx`, so `dotnet build Garnet.slnx` doesn't rebuild it and my `--no-build` runs used pre-rebase test code. Re-validated with freshly built binaries: `Tsavorite.test.hlog` 580/0, `Tsavorite.test.recovery` 204/0, `Tsavorite.test` 290/0.

badrishc:
### Diagnosis of the `GarnetObjectStoreDisabledError` CI failure

The failing job (`Garnet Standalone (ubuntu-latest, net8.0, Release, Garnet.test)`) had **two** failures that are inverses of each other:

| Test | Symptom |
|---|---|
| `GarnetObjectStoreDisabledError` | expected `"Object store is disabled"` from `LLEN`, got a **successful reply** |
| `IncrDecrChangeDigitsWithExpiry` | a plain `INCR` **received** `"Object store is disabled"` |

That pairing is the whole clue: no string or unified code path can reach `ThrowObjectStoreUninitializedException` (the only call sites are in `ObjectStore/{List,Set,SortedSet,Common}Ops.cs`), and I confirmed on a live `--no-obj` server that `SET key 9 EX 1000` + `INCR` returns `:10` rather than erroring. So neither test was getting an answer from the server it thought it was talking to.

**Root cause.** `.NET` implements `SocketOptionName.ReuseAddress` on Unix by setting **both** `SO_REUSEADDR` and `SO_REUSEPORT`. `SO_REUSEPORT` lets a second server bind an `address:port` that is already being listened on, and the kernel then load-balances incoming connections across both listeners. Starting a Garnet server on a port already in use therefore **succeeded silently** and served a fraction of client traffic from a different store.

Reproduced in-process, two servers on the fixture's fixed port:

```
PROBE: first server (objects enabled) started
PROBE: SECOND server bound the SAME port with NO error
PROBE: LLEN answered by objects-ENABLED server 16/40 times
```

`GarnetObjectStoreDisabledError` is the one test that swaps the shared fixture server for a differently-configured one (`disableObjects: true`) on the same fixed port mid-test, which is why it is the test that trips.

**Fix** (`libs/server/Servers/GarnetServerTcp.cs`): clear `SO_REUSEPORT` after enabling `ReuseAddress`, so binding an occupied port fails with `AddressAlreadyInUse` instead of silently splitting traffic. `SO_REUSEADDR` is left set, so a listener can still rebind while old connections linger in `TIME_WAIT`. After the fix the same probe reports `SocketException: Address already in use`.

This is a production-relevant hazard beyond the test: restarting Garnet without stopping the previous instance would silently split client traffic between two stores, which looks like random data loss.

**Regression tests** (`GarnetServerPortBindingTests`): `SecondServerCannotBindPortAlreadyInUse` and `PortIsReusableAfterServerIsDisposed`. Reverting the one-line call makes the first fail and leaves the second passing.

**Validation:** `Garnet.test` net8.0 Release 1050/0 and net10.0 Debug 1055/0 (full suites, hang-blame enabled), cluster 156/0, collections 778/0, cluster.migrate 56/0, `dotnet build Garnet.slnx` 0 errors, `dotnet format` clean.

Windows is left unchanged: exclusivity there is governed by `SO_EXCLUSIVEADDRUSE`, so the new exclusivity test is excluded on Windows and that hardening is worth a separate change.

<sub>One unrelated observation while validating: a single full-suite run hung in `RespAofTornTailTests.TearDown()` with two sessions parked in `StoreWrapper.WaitForCommitAsync` → `TsavoriteLog.WaitForCommitAsync` blocking `GarnetServer.InternalDispose()`. It did not recur in subsequent full runs and is in an AOF commit path this PR does not touch, so I am noting it rather than folding it in.</sub>

badrishc:
**Correction to my previous comment — I overstated this as the root cause.**

@badrishc rightly pushed back: tests in this assembly run **serially** (no `[assembly: Parallelizable]`, CI passes no `.runsettings`, one test project per job), so I owe a precise account of how a second listener could exist at all.

Checking that: `GarnetServer.InternalDispose()` is explicitly staged — **Phase 1 `Close()`s every listening socket to free the port**, *then* Phase 2 drains active handlers, then Phase 3 disposes the provider. So a slow or even hung dispose does **not** leave a port bound. And I found no test in `Garnet.test` that creates a server without guaranteed disposal.

So the honest status is:

| Claim | Status |
|---|---|
| Two Garnet servers can silently bind and share one port | proven (reproduced both in-process and across two processes) |
| With two servers sharing a port, `LLEN` wrongly succeeds ~40% of the time | proven (16/40) — matches failure #1 exactly |
| This is what actually happened in CI run 34287012831 | **not established** |
| Failure #2 (`INCR` receiving "Object store is disabled") | **still unexplained** |

Failure #2 is the part that does not fit: I verified that `SET key 9 EX 1000` + `INCR` succeeds on a `disableObjects: true` server *and* on a normal one, so no choice of server explains a plain `INCR` receiving that error. Since no string or unified path reaches `ThrowObjectStoreUninitializedException`, that response had to originate from another session — which suggests response/stream cross-talk, a different defect from the one fixed here.

I also cannot claim the green CI run as evidence: I never reproduced the original failure locally (5 full `Garnet.test` suite runs plus 6 `RespTests` fixture runs, net8.0 Release and net10.0 Debug).

What this commit does legitimately stand on: silently binding an already-bound port is a real production hazard — restarting Garnet without stopping the previous instance splits client traffic between two stores and looks like random data loss — and the fix turns that into a loud `AddressAlreadyInUse`, which would also make any future occurrence of this CI symptom diagnosable instead of mysterious. It is hardening, not a proven fix for that specific failure, and I've left the underlying `GarnetObjectStoreDisabledError` question open.

badrishc:
Reverted the port-binding change (`a12c501`); the branch is back to `113d7e3` (force-pushed).

Rationale: it was hardening, not a root-cause fix. As noted in my previous comment, I proved the *mechanism* (on Unix, .NET maps `SocketOptionName.ReuseAddress` to SO_REUSEADDR **+ SO_REUSEPORT`, so two servers can silently bind the same port and the kernel load-balances connections between them) but never established a *trigger* for it in CI, where tests run serially and `GarnetServer.InternalDispose()` closes listeners in phase 1 specifically so a port cannot outlive its test. Nor did it ever explain the second half of the failure pair, where `StringIncrement` *received* "Object store is disabled" — `INCR` cannot generate that error on any server configuration.

Shipping it here would have wallpapered over an unexplained failure while adding unrelated socket behavior to a PR about live-object mutation and scan liveness. `GarnetObjectStoreDisabledError` / `IncrDecrChangeDigitsWithExpiry` remain **unexplained**; the strongest surviving lead is intra-process session/response cross-talk.

The SO_REUSEPORT double-bind hazard is real and reproducible on its own (restarting Garnet without stopping the old instance silently splits client traffic across two stores), so I'll raise it as a separate issue rather than smuggle it in here.

TedHartMS (approved):

PR #2108, opened 2026-09-08
Support separate client and peer cluster endpoints

### LINKED ISSUE #2107: Support separate client and peer cluster endpoints
### Feature request type
Enhancement

### Problem
A node behind per-node port translation may need to advertise a client hostname and port that differ from its directly reachable peer endpoint. Using the client endpoint for gossip or replication can break cluster communication.

### Proposed solution
Add optional client address, hostname, and port advertisements for cluster discovery and redirects while retaining separate peer endpoints for cluster traffic. Preserve existing defaults and version 1 cluster configuration compatibility. Document the requirement to upgrade all nodes before relying on translated client endpoints.

Mathos1432:
## Summary
Support per-node client port translation while retaining direct peer endpoints. Adds optional client IP address, port, and hostname settings for `CLUSTER NODES`, `CLUSTER SLOTS`, `CLUSTER SHARDS`, `MOVED`, and `ASK`.

Peer gossip, replication, migration, and failover keep using internal endpoints. Unset options preserve existing behavior. Client metadata uses an optional extension after the unchanged version 1 configuration prefix. Upgrade all nodes before relying on translated endpoints; older nodes ignore the extension and advertise peer endpoints.

Fixes #2107.

## Validation
- [Garnet .NET CI](https://github.com/microsoft/garnet/actions/runs/34273263237) was triggered automatically for this draft.
- `dotnet build Garnet.slnx -c Debug`: passed for both target frameworks, no warnings.
- `dotnet format Garnet.slnx --no-restore --verify-no-changes`: passed.
- Full Debug `net10.0` suites: Garnet.test 1,073 passed and 3 skipped; Garnet.test.cluster 175 passed; Garnet.test.cluster.replication 107 passed; Garnet.test.cluster.migrate 56 passed.
- Coverage includes command-line and embedded validation, client discovery and redirects, legacy configuration compatibility, restart, metadata refresh, replication, failover, and migration with separate endpoints.

Assisted-By: GitHub Copilot


libs/cluster/Server/ClusterConfig.cs
```diff
@@ -41,8 +41,8 @@ internal sealed partial class ClusterConfig
         public const int MAX_HASH_SLOT_VALUE = 16384;
 
         /// <summary>
-        /// Version of the cluster config serialization format.
-        /// Increment when the binary layout of <see cref="ToByteArray"/>/<see cref="FromByteArray"/> changes.
```
Mathos1432: I really feel like you **shouldn't** remove this comment.

Should we not also increment it?

PR #2112, opened 2026-09-09
Avoid int overflow in sorted-set numkeys argument-count checks

foobar:
Two sites let a numkeys near `int.MaxValue` slip bounds checks and walk the parse state far out of bounds, crashing the server:

1. The additive argument-count checks in ZMPOP, ZUNION, ZUNIONSTORE, ZINTER, and ZINTERCARD (`parseState.Count < numKeys + k`) wrap negative and pass; the handlers then index `GetArgSliceByRef` with a wild index. ZINTERSTORE was fixed in #2031; the same subtraction form is applied to the five siblings.

2. `TryGetKeySearchArgsFromSimpleKeySpec` computed `lastKeyIdx = firstKeyIdx + ((numkeys - 1) * keyStep)` bounded only by `Debug.Assert`, so `COMMAND GETKEYS ZUNION 2147483647 x` (and cluster slot verification) walked `GetArgSliceByRef` to attacker-controlled indexes before the handlers ran. The asserts are replaced with runtime guards and `lastKeyIdx` is clamped to the parse state.

Repro on a default deployment: `COMMAND GETKEYS ZUNION 2147483647 x`.

Adds regression coverage for numkeys = 2147483646/2147483647 across the five commands and the COMMAND GETKEYS boundary, each followed by `PING` to prove the connection survives.

### DIFF
--- a/libs/server/Resp/Objects/SortedSetCommands.cs
+++ b/libs/server/Resp/Objects/SortedSetCommands.cs
@@ -431,7 +431,7 @@ namespace Garnet.server
             }
 
             // Validate we have enough arguments (no of keys + (MIN or MAX))
-            if (parseState.Count < numKeys + 2)
+            if (parseState.Count - 2 < numKeys)
             {
                 return AbortWithErrorMessage(CmdStrings.RESP_SYNTAX_ERROR);
             }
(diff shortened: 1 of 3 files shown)


foobar:

vazois (approved):

PR #2128, opened 2026-09-11
Read version 2 cluster endpoints before enabling writers

Mathos1432:
## Why

Support deployments where clients reach cluster nodes through a load balancer using different addresses or ports from those used for node-to-node communication. Each advertised client endpoint must route to its corresponding node, while gossip, replication, and migration continue using directly reachable peer endpoints.

This change prepares readers for that separation so existing clusters can upgrade before a follow-up starts writing and using the new endpoint fields.

## What

- Add version 2 cluster configuration support with `ClusterAddress` and `ClusterPort` alongside the existing address and port fields.
- Accept both versions during recovery and cluster communication. Version 1 configurations initialize the new fields from the existing endpoint.
- Continue writing version 1 and retain existing routing behavior. This change does not enable separate endpoints or configure a load balancer.

## Validation

- 81 cluster configuration and management tests passed on .NET 10.
- Cross-version serialization, persisted configuration recovery, and mixed-version cluster communication were exercised against Garnet v2.1.7 and upstream `main`.
- [Garnet .NET CI](https://github.com/microsoft/garnet/actions/runs/34644233024) completed with two Windows vector-set test failures: `VectorSetMigrateManyBySlot` and `MigrateVectorStressAsync`.

Assisted-By: GitHub Copilot

### DIFF
--- a/libs/cluster/Server/ClusterConfig.cs
+++ b/libs/cluster/Server/ClusterConfig.cs
@@ -42,9 +42,14 @@ namespace Garnet.cluster
 
         /// <summary>
         /// Version of the cluster config serialization format.
-        /// Increment when the binary layout of <see cref="ToByteArray"/>/<see cref="FromByteArray"/> changes.
+        /// Increment when the binary layout of <see cref="ToByteArray(byte)"/>/<see cref="FromByteArray"/> changes.
         /// </summary>
-        public const byte ClusterConfigVersion = 1;
+        public const byte ClusterConfigVersion = 2;
+
+        /// <summary>
+        /// Production write version, retained until all nodes can read version 2.
+        /// </summary>
+        public const byte DefaultClusterConfigVersion = 1;
 
         /// <summary>
         /// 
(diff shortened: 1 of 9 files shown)


libs/cluster/Server/ClusterConfig.cs
```diff
@@ -42,9 +42,14 @@ internal sealed partial class ClusterConfig
 
         /// <summary>
         /// Version of the cluster config serialization format.
-        /// Increment when the binary layout of <see cref="ToByteArray"/>/<see cref="FromByteArray"/> changes.
+        /// Increment when the binary layout of <see cref="ToByteArray(byte)"/>/<see cref="FromByteArray"/> changes.
```
Mathos1432: Revert changes to the comment

libs/cluster/Server/ClusterConfig.cs
```diff
@@ -142,6 +149,11 @@ public ClusterConfig InitializeLocalWorker(
             newWorkers[LOCAL_WORKER_ID].ReplicaOfNodeId = replicaOfNodeId;
             newWorkers[LOCAL_WORKER_ID].ReplicationOffset = 0;
             newWorkers[LOCAL_WORKER_ID].hostname = hostname;
+            if (workers[LOCAL_WORKER_ID].Nodeid == null)
```
Mathos1432: Why is the assignment done only when nodeId == null ?

libs/cluster/Server/ClusterConfigSerializer.cs
```diff
@@ -25,16 +25,31 @@ public static bool TryPeekVersion(ReadOnlySpan<byte> data, out byte version)
             return true;
         }
 
+        /// <summary>
+        /// Check whether a cluster config serialization version can be read.
+        /// </summary>
+        /// <param name="version">Serialization version.</param>
+        public static bool IsSupportedVersion(byte version) => version is 1 or 2;
```
Mathos1432: Shouldn't we use the constants defined in ClusterConfig ?

PR #2130, opened 2026-09-13
[RESP] Parse SET expiration values as 64-bit integers

### LINKED ISSUE #2129: SET, SETEX, and PSETEX parse expiration values as 32-bit integers
### Describe the bug

Garnet parses expiration values for `SET ... EX|PX`, `SETEX`, and `PSETEX` as signed 32-bit integers. Redis parses these values as signed 64-bit integers, so valid expiration values greater than `Int32.MaxValue` are rejected by Garnet.

This creates a RESP compatibility difference for clients that use large but valid TTLs. For example, `PX 4294967296` represents about 49.7 days and is accepted by Redis, but Garnet returns:

```text
ERR value is not an integer or out of range
```

The affected parsing is in `libs/server/Resp/BasicCommands.cs`: `NetworkSETEX` and `NetworkSETEXNX` use `TryGetInt`, and the shared SET expiration helpers accept `int`.

Run the following commands against Garnet:

```text
SET set-px value PX 4294967296
SET set-ex value EX 3000000000
SETEX setex 3000000000 value
PSETEX psetex 4294967296 value
```

Garnet rejects each command with `ERR value is not an integer or out of range`.

Run the same commands against Redis. Redis accepts them with `OK`, and `PTTL` reports a value greater than `2147483647`.

### Expected behavior

Garnet should parse these expiration arguments as signed 64-bit integers, matching Redis RESP behavior. Positive values that can be safely converted to Garnet's absolute expiration ticks should be accepted.

Values that would overflow the absolute expiration representation should be rejected with:

```text
ERR invalid expire time in 'set' command
```

The error must not modify the key or terminate the client session.

### Screenshots

_No response_

### Release version

v2.1.5 / v2.1.7 tested

### IDE

dotnet CLI + Cursor | VSCode, StackExchange.Redis 3.2.0

### OS version

Windows 11 25H2,
Windows 2019

### Additional context

Redis uses a signed 64-bit value for these expiration arguments in `getExpireMillisecondsOrReply` in `src/t_string.c`.

The proposed fix is intentionally limited to the Redis-compatible `SET`, `SETEX`, and `PSETEX` command paths. `RESTORE` and Garnet-specific ETag SET commands also contain 32-bit expiration parsing, but are outside the scope of this compatibility fix.

yuseok-kim-edushare:
### Root Cause

The RESP handlers for `SET ... EX|PX`, `SETEX`, and `PSETEX` parsed expiration values with `TryGetInt` and passed them through helpers that accepted `int`. Redis parses these values as signed 64-bit integers, so Garnet rejected valid TTLs greater than `Int32.MaxValue`.

Simply changing the parser type would expose a second problem: converting very large values with `TimeSpan.FromSeconds` or `TimeSpan.FromMilliseconds` can throw, while unchecked absolute-tick arithmetic can overflow. Either outcome could terminate the session or store an invalid expiration.

### Description of Change

- Parse `SET`, `SETEX`, and `PSETEX` expiration values with `TryGetLong`.
- Pass expiration values through the SET helpers as `long`.
- Convert relative expiration values to absolute ticks with an explicit overflow check.
- Return `ERR invalid expire time in 'set' command` when the expiration cannot be represented.
- Add regression coverage for 64-bit TTLs, conditional SET paths, overflow errors, unchanged key values, and session health.

### Key Technical Details

**Affected code:**
- `NetworkSETEX` handles `SETEX` and `PSETEX`.
- `NetworkSETEXNX` parses `SET` options including `EX`, `PX`, `NX`, `XX`, and `GET`.
- `NetworkSET_EX` and `NetworkSET_Conditional` apply the parsed expiration.
- `TryGetAbsoluteExpiryTicks` validates the multiplication and addition before computing the absolute expiration.

The conversion avoids exception-based validation and performs no allocation on the RESP command path.

### Tests

- `SetExpiryAcceptsInt64Value`
  - Covers `SET PX`, `SET EX`, `SETEX`, `PSETEX`, `SET PX NX`, and `SET PX XX GET`.
  - Verifies each resulting `PTTL` remains greater than `Int32.MaxValue`.
- `SetExpiryOutOfRangeIsRejectedWithoutKillingSession`
  - Covers unrepresentable second and millisecond expirations.
  - Verifies the expected RESP error, unchanged key value, and a healthy connection.

Validation completed:

```text
dotnet build Garnet.slnx -c Debug
dotnet test test/standalone/Garnet.test/Garnet.test.csproj -c Debug --no-build -f net8.0 --filter "FullyQualifiedName~SetExpiry"
dotnet test test/standalone/Garnet.test/Garnet.test.csproj -c Debug --no-build -f net10.0 --filter "FullyQualifiedName~SetExpiryAcceptsInt64Value|FullyQualifiedName~SetExpiryOutOfRangeIsRejectedWithoutKillingSession"
```

### Scope

This change is limited to the Redis-compatible `SET`, `SETEX`, and `PSETEX` paths. It does not change `RESTORE` or Garnet-specific ETag commands.

### Issues Fixed

Fixes #2129


libs/server/Resp/BasicCommands.cs
```diff
@@ -769,15 +769,14 @@ private unsafe bool NetworkSET_EX<TGarnetApi>(RespCommand cmd, ExpirationOption
...
             where TGarnetApi : IGarnetApi
         {
-            var inputArg = expiry == 0
-                ? 0
-                : DateTimeOffset.UtcNow.Ticks +
-                  (highPrecision
-                      ? TimeSpan.FromMilliseconds(expiry).Ticks
-                      : TimeSpan.FromSeconds(expiry).Ticks);
+            long inputArg = 0;
+            if (expiry != 0 && !TryGetAbsoluteExpiryTicks(expiry, highPrecision, out inputArg))
```
vazois, line 776: nit; can we push the expiry != 0 check inside TryGetAbsolute? Should it be uniform across all calls?

test/standalone/Garnet.test/RespTests.cs
```diff
@@ -749,6 +749,83 @@ public void SetExpiryHighPrecision()
...
+        [Test]
+        public void SetExpiryAcceptsInt64Value()
+        {
+            const long LargeMilliseconds = 4_294_967_296;
+            const long LargeSeconds = 3_000_000_000;
+
+            using var redis = ConnectionMultiplexer.Connect(TestUtils.GetConfig());
+            var db = redis.GetDatabase(0);
+
+            void AssertLargeTtl(string key, long maximumTtlMilliseconds)
```
vazois: tighten the bound to ensure that wrong conversion from seconds to milliseconds does not pass the assertion
e.g.

const long LargeSeconds = 3_000_000_000;
AssertLargeTtl("set-ex", LargeSeconds * 1000);

the expected TTL is:

3,000,000,000,000 ms

But suppose the implementation accidentally treated the seconds value as milliseconds. It could return:

3,000,000,000 ms

The existing assertions would still pass:

3,000,000,000 > 2,147,483,647                 // passes
3,000,000,000 <= 3,000,000,000,000            // passes

vazois (changes requested):

test/standalone/Garnet.test/RespTests.cs
yuseok-kim-edushare: I fix  test from complare with Int32 max to comapre with expectedTTL value

libs/server/Resp/BasicCommands.cs
yuseok-kim-edushare: I think I need to think about more how can we reach uniformly useable helper.

The `expiry != 0` guard shouldn’t move into `TryGetAbsoluteExpiryTicks`, and the call sites shouldn’t be made uniform.

`TryGetAbsoluteExpiryTicks` converts a real relative EX/PX value to absolute ticks and rejects overflow. `expiry == 0` is not a relative expiry — it means “no EX option was provided.” Storage already treats `arg1 == 0` as no expiration.

Only `NetworkSET_Conditional` uses that sentinel. `GETSET`, `KEEPTTL`, and `SET NX/XX/GET` without EX all pass `expiry == 0` and must keep `arg1 = 0`. That is why the skip lived at that call site.

The other callers are different:

- `NetworkSETEX` / `SET EX|PX`: `0` is invalid and already rejected with `expiry <= 0`. If the helper treated `0` as success (“no expiry”), those command-level checks would become the only thing preventing `SETEX key 0 val` from succeeding as a SET with no TTL.
- `NetworkSET_EX`: historically always did `now + expiry`, with no `!= 0` skip. The option parser also routes `ExpirationOption.None` into this method for a blind SET. Putting the skip inside the helper changes `expiry == 0` from “absolute now” to “no expiry” on that path. Plain `SET key value` goes through `NetworkSET` instead, so this is likely unused — but it is still a contract change, not a mechanical cleanup.

vazois (approved):

